PR #240, opened 2015-10-23
Suggest medium color update based on icon

rpbaltazar:
Based on the favicon used on the website, I've updated the background color for the medium icon.

medium_icon

Also took the chance and fixed the indentation of the index.html


danleech:
@rpbaltazar Thank you!

Can you make the colour change on `src/simple-icons.json` though please? Otherwise my build script will overwrite your change with the wrong colour again next time I build!

rpbaltazar:
ok. Updated.
Once again, my editor fixed the indentation.
Let me know if that is a problem for you and i'll get it working on my machine for this project.

danleech:
Sorry, I pulled in another request before this one that caused hundreds of conflicts as a result of the indentation changes. I’ve updated the Medium colour in a 32eb193.

For future reference I’m trying to use soft tabs (four spaces long) throughout. It’s keeps the code simpler, and makes no difference to the served file size because I’m using CloudFlare to minify the HTML. There are still some tabs output by the build script because I was lazy and used `\t` a lot.

PR #258, opened 2015-10-30
Added Conekta Icon

richpeniche:


richpeniche:
Please, do not merge "Commented cname url". The "Added Conekta icon" commits are correct.

PR #260, opened 2015-11-03
Add sass

athimannil:


danleech:
Thanks! I’ve switched out the full logo for the icon for Sass.

For future reference the `index.html` and `404.html` files are generated by the `src/build.js` Node script. This compiles the page content, figures out the order and sets the icon count at the top. Neat idea you had to do it on the client-side though!

I should make the contribution instructions clearer; the easiest thing to do would be to have people contribute an SVG file and an entry in `simple-icons.json`, and I’ll run the build script when I pull.

PR #269, opened 2015-12-07
Added dynamic icon count

augbog:
I noticed the past couple of commits you simply iterate the icon count every time. I thought it'd be good to make it dynamic so you don't have to. The only issue is SEO the Google robot might not pick it up immediately but I don't think the number is necessarily important if it keeps updating.

Feel free to deny this PR if this isn't ideal.

### DIFF
--- a/index.html
+++ b/index.html
@@ -209,7 +209,7 @@
         </form>
     </header>
     <main role="main">
-        <p class="hero">184 SVG icons for popular brands <a href="https://github.com/danleech/simple-icons">Download them from GitHub</a></p>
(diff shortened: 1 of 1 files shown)


danleech:
Thanks for the commit, but the icon count is already automatically calculated when the page is built in `src/build.js` (this builds the page, embeds the SVGs, does the sorting etc). I elected to make the page build a pre-deployment task rather than a client-side task because the sorting is fairly complex.

augbog:
Ah my bad I should have checked that. Great icons btw :)

danleech:
Thanks :smiley:

PR #299, opened 2016-02-23
Usability fixes

donroyco:
Made some usability improvements:
- Tile icons are now fully clickable.
- Hover state full tiles is more obvious (underline name).
- Larger area full tiles clickable (name too).
- 'Hex code' is now better visible on hover.
- Small code improvements.


danleech:
Wow! Fantastic PR, thank you.

PR #321, opened 2016-08-06
Remove email

FletchXYZ:
I don't see the point of having email included, as there's no official logo or color to represent it as far as I know.


danleech:
While I agree there’s no official icon, and it’s not a “brand”, in real world use a lot of designers include an email link in their list of social icons (either for emailing or for subscribing to newsletters).

I’ll remove the icon for now, but if requests start coming in to include it again I might put it back.

Thanks for the PR!

FletchXYZ:
Fair enough, I understand the reasoning. Thanks! :)

PR #323, opened 2016-08-08
Add Product Hunt [#325], update Windows logo and color

FletchXYZ:
Windows: New source is Wikimedia Commons. The previous source was from 2012, this might be slightly more (or less?) accurate.

Product Hunt: Added in relation to #325.


FletchXYZ:
In relation to Windows: I see only the updated version I've submitted being used in the official OS. I'd recommend updating.

PR #335, opened 2016-12-27
Add ACM-ICPC, Codeforces and HackerRank

MathiasSM:
Some popular sites for Competitive PRogramming. One can link to Codeforces or HackerRank profile, plus ACM-ICPC "ID" (Which is a page with records for every official contest one has participated on).

There are no official svgs for any of these organizations, so I made them myself. I hope it's ok. You can see the logos in the sites: https://icpc.baylor.edu/ http://codeforces.com/ https://www.hackerrank.com/
Although Codeforces has put a temporal christmas logo...


danleech:
Thanks for the contribution! I’ll get those added. The ACM has an official SVG of their brand, which I’ll use in place of the AMC-ICPC artwork.

MathiasSM:
Awesome! They look really good. The ACM has a logo, indeed, but the
ACM-ICPC is an annual event they organize. One can have an ACM account and
have no relation to the ICPC and viceversa. If it's not too much to ask,
could you add the ICPC too? (should I leave it as an issue?)

danleech:
Oh right, I understand. Leave it as an issue. I’ll see if I can get hold of someone on the ICPC team who can send me an official vector image. I always prefer to use official artwork where possible. Thanks again!

PR #337, opened 2017-01-10
Updated Evernote color

augbog:
Wanted to update the Evernote primary color! We recently changed some colors and haven't updated it everywhere but wanted to pay it forward :) See our marketing color palette 

https://evernote.com/brand/color-palette/

Oh and let me know if the Evernote logo needs to be reordered to be somewhere else since the green has a little more blue now but I'm actually not sure how to approximate where in the list it should be. Might be worth considering a tool that calculates what the order is based on hex value?


danleech:
@augbog Thanks very much! It’s great to have the Evernote colour up to date.

The web page is built using a [Node.js script](https://github.com/danleech/simple-icons/blob/gh-pages/src/build.js) (which I’ve just run and pushed). The script takes the icon definitions from the JSON file you edited, calculates hues from the hex values, sorts the list, then outputs the web page and Sass/LESS files.

augbog:
oh sweet thanks :)

PR #342, opened 2017-02-24
Add custom properties CSS file

lowmess:
Creates a `.css` file with the brand color values from `colour-variables.scss` as CSS custom properties.


danleech:
@lowmess Thanks for this PR. CSS variables have been off my radar for a while, so I was surprised to see how much [support](http://caniuse.com/#search=css%20variables) they have! I’ve added some code to the build script to make sure this file stays up to date with future icon additions and changes.

lowmess:
Oh awesome! Should have guessed those files were built programmatically. Thanks!

PR #348, opened 2017-03-04
Add Plurk

jmlntw:
[Plurk](https://www.plurk.com/) is a free social networking and micro-blogging service. Logo and color are taken from <https://www.plurk.com/brandInfo>.


danleech:
Thanks!

PR #349, opened 2017-03-06
Improve text legibility

tomasdev:
Not sure if you really want this, but this would put text in black when the color is brighter than 50%, improving the readability of the companies' names.


danleech:
I really like this idea. Can you help me understand some of the magic numbers in this PR please?

tomasdev:
They come from the "brightness formula" (https://en.wikipedia.org/wiki/Relative_luminance) which is the way the average human eye perceives colors

tomasdev:
Now that I think about it, numbers don't 100% match (since Csrgb is variable) lol - I took the internet failed me!

danleech:
I still like the idea. I already calculate the luminance in the build script, to help with sorting (albeit without factoring in human perception). I could pick a threshold over which tiles get darker text. I’ll probably go with something like `rgb(0,0,0,0.5)` so it’s not *too* dark.

PR #354, opened 2017-03-13
Add bower, composer, php, symfony npm, grunt, css3, angular and gitkraken logos

jrmessias:


CNAME
```diff
@@ -1 +0,0 @@
-simpleicons.org
```
Haroenv: this shouldn’t be removed in this pull request

PR #371, opened 2017-04-01
Fix Product Hunt icon

hacknug:
As you can see in the attached screenshot, the Product Hunt's icon has a faulty inner shape. This PR aims to fix this.

captura de pantalla 2017-04-01 a las 17 18 02


danleech:
Thanks!

PR #385, opened 2017-04-13
added package.json to install repo via npm

Schubidu:
I like to install the repo via npm/yarn. This is the PR to make it compatible with them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,6 @@
   "version": "1.0.0",
   "description": "SVG icons for popular brands https://simpleicons.org",
   "main": "index.js",
-  "scripts": {
-    "test": "echo \"no tests available\""
-  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/danleech/simple-icons.git"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -0,0 +1,31 @@
...
+    "brand-assets",
+    "icons",
+    "logos",
+    "svg",
+    "svg-icons",
+    "design",
+    "design-assets"
+  ],
+  "author": "Dan Leech <dan@danleech.com> (https://danleech.com)",
+  "license": "SEE LICENSE IN LICENSE.md",
```
ericcornelissen: Since the license type used for this package is pretty common, the value for `license` can be set to `	CC0-1.0`. For more info read [this](https://docs.npmjs.com/files/package.json#license) 🙂

package.json
```diff
@@ -0,0 +1,31 @@
...
+  "bugs": {
```
ericcornelissen: `bugs` can be changed into a just a string of the URL in this case, as specified [here](https://docs.npmjs.com/files/package.json#bugs)

package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "simple-icons",
+  "version": "1.0.0",
+  "description": "SVG icons for popular brands https://simpleicons.org",
+  "main": "index.js",
+  "scripts": {
```
ericcornelissen: `scripts` can be omitted when it is not used

PR #459, opened 2017-05-25
Implement JS search

nietonfir:
The search is based on the icon titles and their index in the list. A 250ms debounced input event listener hides/shows elements according to the current input value.

The layout of the search box is very simple, so feel free to update it.

Imho a notification if nothing matches would be nice, but I didn't want to change too much in `.grid`.

The implementation should work on all current browsers as well as >=IE9.

### DIFF
--- a/index.html
+++ b/index.html
@@ -245,16 +245,63 @@
             }
         }
 
+        .search {
+            margin: 1rem 1.5rem;
+            text-align: center;
+        }
+        @media (min-width: 45rem) {
+            .search {
+                margin: 1rem 3rem;
+                text-align: left;
+            }
+        }
+
+        .search .search__wrapper {
+            display: inline-block;
+            position: relative;
+        }
+        .search .search__close {
+            display: none;
+            font-size: 1rem;
+            padding: 6px 12px;
+            position: absolute;
+            right: 0;
+            top: 1px;
+        }
+        .search .search__close span {
+            cursor: pointer;
+        }
+
+        .search input {
+            color: #333;
+            font-size: 1rem;
+            padding: 6px 12px;
+        }
+        @media (min-width: 45rem) {
+            .search input {
+                width: 250px;
+            }
+        }
+
+        .search--active .search__close {
+            display: block;
+        }
+        .search--active input {
+            padding-right: 32px;
+        }
+
         .grid {
             display: flex;
             flex-wrap: wrap;
             list-style: none;
             margin: 1.5rem;
+            margin-top: 0;
             padding: 0;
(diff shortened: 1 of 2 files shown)


XwhyyZ:
I wish to unsubscribe please

nietonfir:
There's a button to the right of the PR (shy over the participants badges) that reads "unsubscribe".

nietonfir:
Rebased branch to current develop [HEAD](https://github.com/danleech/simple-icons/commit/236ed8d2440febe93c630e5422030da8cef59cc3).

nietonfir:
@danleech Is there something I need to modify to get this merged?

danleech:
@nietonfir Hi! No, I’m just not organised and haven’t looked at this PR yet. I *will* get round to it though! Thanks for your patience! 😄

nietonfir:
No problem at all, take your time! ;-) Thanks for the response!

index.html
```diff
@@ -410,5 +467,90 @@ <h2 class="grid-item__title">{{ iconArray[5] }}</h2>
...
+
+            $search.addEventListener('input', debounce(function(e) {
+                e.preventDefault();
+
+                var value = e.target.value;
+                if (value) {
+                    $search.classList.add('search--active');
+                    search(value);
+                } else {
+                    $search.classList.remove('search--active');
```
birjolaxew: Should probably add `search('')` to properly clear the search when everything is deleted.

index.html
```diff
@@ -410,5 +467,90 @@ <h2 class="grid-item__title">{{ iconArray[5] }}</h2>
...
+                        result = func.apply(context, args);
+                        context = args = null;
+                    }
+
+                    return result;
+                };
+            }
+
+            function search(value) {
+                value = value.toLowerCase();
```
birjolaxew: Add `.trim()` in case user hits spacebar after searching.

nietonfir:
Implemented all of the feedback except the shortcut and the normalization.

Regarding the shortcut: I'm not a fan of overriding default browser shortcuts. And after conducting a quick search I couldn't find a site where this is done, so I wouldn't recommend it either.

Regarding the normalization: I don't see any issue with that as long as the search terms keep being the brand names. I'll look into that…

birjolaxew:
> Regarding the shortcut: I'm not a fan of overriding default browser shortcuts. And after conducting a quick search I couldn't find a site where this is done, so I wouldn't recommend it either.

Fair enough, I can see where you're coming from :)

birjolaxew:
This is now live :)

Issue #480, opened 2017-07-09
Use dark color on light background


birjolaxew:
On the website, some icons/titles are hard to read. These could be improved by using a dark color instead of white on some icons. Examples:

- Snapchat
- Gerrit
- Pingdom

Should possibly check using the color's [contrast ratio](http://leaverou.github.io/contrast-ratio/)

birjolaxew:
Since Liquid doesn't support a Math.pow equivalent, calculating the [relative luminance](https://www.w3.org/TR/WCAG20/#relativeluminancedef) of a color is impossible. This means that calculating the [contrast ratio](https://www.w3.org/TR/WCAG20/#contrast-ratiodef) is impossible. Gonna have to shelf this idea until we implement deployment that supports custom filters (or someone else finds a sneaky way).

ericcornelissen:
I'm not very familiar with Liquid, but from [their wiki](https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers#create-your-own-filters) I see you can create your own filters. Just write a power filter in Ruby as follows:
```ruby
module PowerFilter
  def pow(input, power) # will be available as the "pow" filter
    input**power
  end
end
```
And use it as so (simplified, you need to obtain the red/green/blue values from the hex):
```html
<!-- Based on if RsRGB <= 0.03928 then R = RsRGB/12.92 else R = ((RsRGB+0.055)/1.055) ^ 2.4 -->
{% if icon.hex <= 0.03928 %}
{{ icon.hex | divided_by: 12.92 }}
{% else %}
{{ icon.hex | plus: 0.055 | divided_by: 1.055 | pow: 2.4 }}
{% endif %}
```

I hope that helps 😄

birjolaxew:
@ericcornelissen Unfortunately we currently build our site using GitHub Pages, which doesn't allow custom filters. The easiest solution would probably be to do it using a bit of JS.

ericcornelissen:
That makes sense 😅 And what about opening a PR in the [Liquid](https://github.com/Shopify/liquid) repository, I'd be interested in doing that 🙂

birjolaxew:
@ericcornelissen That's definitely a worthwhile idea 🙂 Unfortunately I don't think it'll help our specific scenario much (there are many Liquid tags/filters that Jekyll don't support, so a new filter will likely take some time to reach Jekyll).

[...]

PR #488, opened 2017-08-02
Use dark color icons/text on light backgrounds

ericcornelissen:
Implements the feature specified in #480

It is using [this formula](https://www.w3.org/TR/WCAG20/#relativeluminancedef) for relative luminance combined with [this approximation](http://chilliant.blogspot.nl/2012/08/srgb-approximations-for-hlsl.html) to find the contrast ratio between the background (which is the brand color) and white (the default color of the icon/text). If this value, which is between 0 and 1, is to high it will make sure the brand icon, brand name and hex-value of the color are displayed in black (currently `#000`) instead of the current default white.
The color change is achieved by adding a class to the `li.grid-item`, either `.dark` or `.light`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -78,9 +78,9 @@
     {% endif %}
     {% assign luminance = lRed | plus: lGreen | plus: lBlue %}
     {% if luminance >= 0.83 %}
-        {% assign class = "dark" %}
+        {% assign class = "grid-item--dark" %}
     {% else %}
-        {% assign class = "light" %}
+        {% assign class = "grid-item--light" %}
     {% endif %}
 
     {% assign hslLuminance = rgbMax | plus: rgbMin | times: 50.0 %}
@@ -177,25 +177,9 @@
         }
 
         svg {
+            fill: currentColor;
             height: 1.5rem;
             width: 1.5rem;
-        }
-
-        .grid-item.light path,
-        .grid-item.light rect,
-        .grid-item.light circle,
-        .grid-item.light h2,
-        .grid-item.light p {
-            color: #FFF;
-            fill: #FFF;
-        }
-        .grid-item.dark path,
-        .grid-item.dark rect,
-        .grid-item.dark circle,
-        .grid-item.dark h2,
-        .grid-item.dark p {
-            color: #000;
-            fill: #000;
         }
 
         #carbonads {
@@ -376,6 +360,15 @@
             background-color: #757575;
             text-align: center;
         }
+        .grid-item--light {
+            color: #FFF;
+        }
+        .grid-item--dark {
+            color: #000;
+        }
+        .grid-item__link {
+            color: inherit;
+        }
         @supports not (display: grid) {
             .grid-item {
                 border: 0.1875rem solid #FFF;
@@ -402,7 +395,6 @@
         }
 
         .grid-item__link {
-            color: #FFFFFF;
             display: block;
             padding: 1rem 1rem 0;
             text-align: center;
@@ -425,7 +417,6 @@
         }
 
         .grid-item__subtitle {
-            color: #FFFFFF;
             font-size: 0.75rem;
             line-height: 1rem;
             margin: 0;
--- a/index.html
+++ b/index.html
@@ -60,6 +60,32 @@
     {% if rgbBlue < rgbMin %}
         {% assign rgbMin = rgbBlue %}
     {% endif %}
+
+
+    {% if rgbRed <= 0.03928 %}
+        {% assign lRed = rgbRed | divided_by: 12.92 | times: 0.2126 %}
+    {% else %}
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -60,6 +60,29 @@
...
+    {% else %}
+        {% assign lGreen = rgbGreen | times: 0.3053 | plus: 0.6822 | times: rgbGreen | plus: 0.0125 | times: rgbGreen | times: 0.7152 %}
+    {% endif %}
+    {% if rgbBlue <= 0.03928 %}
+        {% assign lBlue = rgbBlue | divided_by: 12.92 | times: 0.0722 %}
+    {% else %}
+        {% assign lBlue = rgbBlue | times: 0.3053 | plus: 0.6822 | times: rgbBlue | plus: 0.0125 | times: rgbBlue | times: 0.0722 %}
+    {% endif %}
+    {% assign luminance = lRed | plus: lGreen | plus: lBlue %}
+    {% if luminance >= 0.83 %}
```
birjolaxew: I find that a cutoff of `0.5` covers most characters that are hard to read. This is obviously up for debate.

index.html
```diff
@@ -158,9 +181,22 @@
             width: 1.5rem;
         }
 
-        path, rect, circle {
+        .grid-item.light path,
+        .grid-item.light rect,
+        .grid-item.light circle,
+        .grid-item.light h2,
+        .grid-item.light p {
```
birjolaxew: Simplify selector to `.grid-item.light`, and use inheritance for all the rest. IIRC this means adding `.grid-item__link { color: inherit }` and removing `.grid-item__subtitle { color: #FFFFFF }`.

index.html
```diff
@@ -158,9 +181,22 @@
...
+            color: #FFF;
             fill: #FFF;
```
birjolaxew: Add `svg { fill: currentColor; }` instead of setting `fill` here.

index.html
```diff
@@ -158,9 +181,22 @@
...
         }
+        .grid-item.dark path,
+        .grid-item.dark rect,
+        .grid-item.dark circle,
+        .grid-item.dark h2,
+        .grid-item.dark p {
+            color: #000;
```
birjolaxew: Use `#222` instead - it's a nicer color

ericcornelissen:
@birjolaxew about the numbers, you're correct that you can't find all numbers in the [original link](http://chilliant.blogspot.nl/2012/08/srgb-approximations-for-hlsl.html) I provided, I also used the [relative luminance](https://www.w3.org/TR/WCAG20/#relativeluminancedef) link you provided (I updated the PR description 👍)

Just for good measure, I'll try to explain how I applied the functions (using red as example):
This come straight from the if-statement provided in the [relative luminance](https://www.w3.org/TR/WCAG20/#relativeluminancedef) link.
```liquid
{% if rgbRed <= 0.03928 %}
```

This is the then-branch for the if, the `times: 0.2126` comes from the `L = 0.2126 * R + 0.7152 * G + 0.0722 * B` formula also found on that page.
```liquid
{% assign lRed = rgbRed | divided_by: 12.92 | times: 0.2126 %}
```

This is the else-branch for the if, but instead of the formula you see on the relative luminance link I used the approximation provided in the other [link](http://chilliant.blogspot.nl/2012/08/srgb-approximations-for-hlsl.html) (see *C_lin_3*). The `times: 0.2126` is from this the `L = 0.2126 * R + 0.7152 * G + 0.0722 * B` formula again.
```liquid
{% assign lRed = rgbRed | times: 0.3053 | plus: 0.6822 | times: rgbRed | plus: 0.0125 | times: rgbRed | times: 0.2126 %}
```

Then the luminance is calculating by adding the terms of the luminance formula `L = 0.2126 * R + 0.7152 * G + 0.0722 * B`:
```liquid
{% assign luminance = lRed | plus: lGreen | plus: lBlue %}
```

index.html
ericcornelissen: `0.5` seems like a logical cutoff, but I'm not sure if its the best. I made a screenshot so you can see it:
screenshot

I think we can identify two problems from this screenshot:
- First, imo there are a few icons that don't need to be darkened. This includes most yellow icons except for Pingdom, Snapchat and Javascript, and Upwork, Kickstarter and maybe even Gumtree from the green icons.
- Second, it seems a bit odd to me that there are some white icons in between dark icons? Not sure if the icons are not properly ordered or if it is due to the formula that I'm using 🤔 

One last thing, if you want to play around with this more effectively I recommend you check out the [develop-see-luminance](https://github.com/ericcornelissen/simple-icons/tree/develop-see-luminance) branch I created in my fork where instead of the hex-value the calculated luminance is shown so you get a better idea of what icons will be made dark if you change the value (also the cutoff is set to `0.5` on that branch).

ericcornelissen:
Also, note that some icons are colored incorrectly due to the `fill` defined on the SVGs themselves. I fixed this in another PR, see #489 👍

Issue #496, opened 2017-08-25
Request: Adobe (Reader) PDF


daybugging:
Possible starting points:
https://www.adobe.com/legal/permissions/icons-web-logos.html
https://de.wikipedia.org/wiki/Datei:Adobe_PDF.svg

Much appreciated, great work!

birjolaxew:
[Official SVG](https://acrobat.adobe.com/etc/clientlibs/beagle/ace/reimagine/source/img/doccloud/product-icon-acrobat.svg), taken from header icon on [this page](https://acrobat.adobe.com/us/en/acrobat/pdf-reader.html).

ericcornelissen:
Satisfied with #497 & #498 @daybugging? 😄

birjolaxew:
Closed with #497 and #498

PR #497, opened 2017-08-30
Add Adobe Acrobat Reader icon

ericcornelissen:
Added an icon for Adobe Acrobat Reader as requested in #496. The hex value comes from the original SVG found in the source.


birjolaxew:
:+1:

PR #498, opened 2017-08-30
Add Adobe icon

ericcornelissen:
Added an icon for Adobe. The hex value comes from the original SVG found in the source.


birjolaxew:
:+1:

PR #521, opened 2017-09-11
Fix JSON typos

birjolaxew:
Fixes two missing `,` in the JSON, causing the build to fail.

I should be more thorough when reviewing PRs in the future to catch those before merge.


ericcornelissen:
Perhaps we can use something like [jsonlint](https://www.npmjs.com/package/jsonlint) to verify the validity of the `_data/simple-icons.json` file automatically for each PR?

Issue #508, opened 2017-09-07
Static Generators and CMS


xdesro:
Thoughts about these?

- [ ]  Contentful
- [x]  Netlify
- [x]  Hexo
- [ ]  Hugo
- [x]  Prismic
- [x]  Gatsby

davidklebanoff:
Due to a lack of replies, I'm gunna close this issue.

Thoughts @xdesro?

xdesro:
Works for me. Contentful might be tricky to translate anyway.

PR #522, opened 2017-09-11
Adds Prismic.IO, Hexo, and Netlify icons

xdesro:
RE: #508. Let me know if any issues!


ericcornelissen:
Good work, I've got two suggestions:
- Maybe change the link of Hexo to just `https://hexo.io/`?
- It seems you picked the color for Prismic from their landing page (`#1DE9B6`), right? I'm not sure if that is the best color to use since that's the only place where I could find that particular color... Instead maybe use `#13C095`, a similar color which they use as an accenting color throughout their website?

xdesro:
@ericcornelissen Noted and fixed! Thanks for the guidance  🙏

birjolaxew:
Prismic seems to user 3 logo color schemes: [white](https://prismic.io/), [blue](https://prismic.io/blog) or [multicolored](https://github.com/prismicio).

<p align="center">
  
  
  
</p>

Since we'll be using single-colored, I'd actually suggest #484A7A (from their blue logo)

xdesro:
That makes sense. New commit reflects that thought.

birjolaxew:
:+1: Merged and published

PR #524, opened 2017-09-12
Update Bitbucket icon

ericcornelissen:
Updated the Bitbucket icon according to the recent redesign, you can see it at https://bitbucket.org/


birjolaxew:
Am I correct in that the second `<path>` is redundant and can be deleted?

ericcornelissen:
You are indeed correct, fixed it 👍

birjolaxew:
:+1: Thanks for the update!

PR #526, opened 2017-09-13
syntax error in matternet.svg

paulpflug:
added missing `>`


birjolaxew:
:+1: Thanks for noticing! Merged

PR #527, opened 2017-09-13
Add Json validator

ericcornelissen:
In response to #521, I added a JSON validator to validate the validity of `_data/simple-icons.json` that could/should be used with Continuous Integration (currently configured for [Travis CI](https://travis-ci.org/), but we could change to e.g. [CircleCI](https://circleci.com/) or [AppVeyor](https://www.appveyor.com/)). You can see it effects [here](https://github.com/ericcornelissen/simple-icons/commits/json-validator) or in the screenshot below (I intentionally made a commit with an incorrect .json file)
preview

I used [jsonlint](https://www.npmjs.com/package/jsonlint) for this, which I added as a `devDependency` to the project in `package.json`. To make it easier to use I also added a npm script called jsonlint which automatically lints `_data/simple-icons.json` (without this script the command would have to look something like `node_modules/.bin/jsonlint _data/simple-icons.json -q` in case jsonlint is not available globally).

In the future we could even make use of JSON Schemes (which jsonlint supports :+1:) to enforce e.g. certain data types on certain fields. You can find out more [here](http://json-schema.org/) and [here](https://spacetelescope.github.io/understanding-json-schema/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ Temporary Items
 *.css.map
 
 # End of https://www.gitignore.io/api/sass,macos,jekyll
+
+
+# NPM dependencies
+node_modules/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: node_js
+node_js:
+  - "6"
+
+script:
+  - npm run jsonlint
+
+notifications:
+  email:
+    on_success: never
+    on_failure: change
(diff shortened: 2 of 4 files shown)


birjolaxew:
:+1: Thanks for implementing the linting! I've had this planned for a bit, but it's great that someone else got it done before me :) Really appreciated!

On a related note, [svglint](https://github.com/birjolaxew/svglint) is being worked on - at this point, the only things missing is implementing the rules (and fix any problems/redesigns that arise from that)

ericcornelissen:
Great @birjolaxew 👍 

Btw, do you want me to write up an issue for the JSON schemes, or are you not interested in using that?

birjolaxew:
That would be neat @ericcornelissen!

PR #540, opened 2017-09-17
fixes the relative path

jenil:
hey @birjolaxew 

I was testing the npm package and there is an issue in reading the icons SVG file due to relative path, I just fixed it. Could you merge and republish it?

Thanks!


birjolaxew:
Thanks! Should now be published as version 1.0.2.

jenil:
Awesome! Implemented it on [icongram.jgog.in](https://icongram.jgog.in) and it works perfectly 💯

PR #542, opened 2017-09-17
Push footer to bottom of the page

ericcornelissen:
Now the `<footer>` is pushed to the bottom of the page when the icons don't fill up the screen.

(background color changed only in screenshots for clarity)

**Before:** 
before

**After:**
after

### DIFF
--- a/index.html
+++ b/index.html
@@ -160,6 +160,7 @@
 
         :root {
             font-size: 100%;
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Thanks for noticing and fixing!

Issue #502, opened 2017-09-04
Update existing icons to adhere to the rules in the CONTRIBUTING.md


ericcornelissen:
A reminder to update some of the existing icons to adhere to the rules specified in [CONTRIBUTING.md](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md). Most icons have been updated with #500, but a few where skipped because they depend on on either `fill-rule`, `clip-rule`, `stroke-linejoin` or `stroke-miterlimit`. I listed them here:

- [x] ACM
- [x] American Express
- [x] Automatic
- [x] Basecamp
- [x] Dailymotion
- [x] Ello
- [x] Empire Kred
- [x] Gauges
- [x] Geeklist
- [x] Google Hangouts
- [x] IMDb
- [x] InVision
- [x] Klout
- [x] MailChimp
- [x] Node.js
- [x] Odnoklassniki
- [x] Player.me
- [x] ProtonMail
- [x] Runkeeper
- [x] Simple Icons
- [x] Slides
- [x] Speaker Deck
- [x] Spotlight
- [x] ~~StyleShare~~ (icon outdated, see #687)
- [x] Subversion
- [x] Toggl
- [x] Travis CI
- [x] Trello
- [x] Vue.js
- [ ] Wix
- [x] Youtube (Maybe just update the icon given the rebrand?

xdesro:
@ericcornelissen YouTube updated. I'll start working on this list to get these icons up-to-date.

ericcornelissen:
Reopening this since Wix still has to be updated (see #688 for details)

PR #558, opened 2017-09-20
Update ACM per Contributing.md

xdesro:
Wanted to push more of these updates per #502 but totally got caught up in work. I'll push more later!


ericcornelissen:
If I look at the version you added I would say it is a bit rugged/jagged 🤔

xdesro:
You're totally right. Will fix and push again.

Issue #560, opened 2017-09-21
Request: Sony PS3, Sony PS4, and Nintendo GameCube


davidklebanoff:
Sony PS3, Sony PS4, and Nintendo GameCube icons.

birjolaxew:
Official logo SVG can be found in the header of [their official site](https://www.playstation.com/).

I'm not sure if PS3 and PS4 have separate logo marks, or if the PlayStation brand is the one you're talking about?

davidklebanoff:
@birjolaxew, Simple Icons already contains that logo (The generic Playstation logo). This request is for the PS4 and PS3 logos specifically that read "PS4" and "PS3" respectively.

See:
https://commons.wikimedia.org/wiki/File:PlayStation_4_logo_and_wordmark.svg
https://commons.wikimedia.org/wiki/File:PS3.svg

davidklebanoff:
I'll go ahead and knock them out. I'll leave it up to you guys if you wish to include them.

birjolaxew:
Added by @davidklebanoff in #564

PR #564, opened 2017-09-22
Added PS3, PS4, and Gamecube icons.

davidklebanoff:
I updated the playstation color as well, as it was slightly off.

The reference was old brand reference was dead, but I was able to confirm the new color using the official website corroborated with some an official style guide hosted on another domain.

Issue #560 and #565.


birjolaxew:
Where on https://www.playstation.com and https://www.nintendo.com did you find the logos? I could only find the Playstation brand as an SVG.

davidklebanoff:
The PS3 logo had come from a filter from the German retailer Saturn, now broken link:
http://www.saturn.at/flyer/10/A_SA_V_2808HTE1_01.pdf

The PS4 logo had come from a Japanese Sony Interactive Entertainment press release, now broken link:
http://www.sie.com/corporate/release/pdf/130221a_e.pdf

The gamecube logo is an SVG version of the logo seen here, the console is older, so I could not locate an official SVG version:
https://www.nintendo.com/consumer/systems/nintendogamecube/index.jsp

birjolaxew:
The `source` URL should refer to where the logos were taken from (or where the image that was traced was taken from, if no SVG is available) - that way it's easier to check for updates, and verify how correct the SVG is.

The PS4 logo source should be changed to e.g. [Wikimedia](https://commons.wikimedia.org/wiki/File:PlayStation_4_logo_and_wordmark.svg) (in lack of better/more official source - Wikimedia is preferred to most other third party sources since it includes a source of its own), while a more official/well-sourced source for the PS3 logo should be found.

Otherwise this is looking good :+1:

davidklebanoff:
@birjolaxew I updated the sources to your guidance. The logos used were the most well-sourced logos that I could find avaliable.

birjolaxew:
@davidklebanoff Thanks for the updates! Merged and published!

Issue #487, opened 2017-07-30
Programming libraries and tools


DSIW:
My suggestions:

- [x] [Ruby On Rails](http://rubyonrails.org/images/rails-logo.svg)
- [x] [Django](https://worldvectorlogo.com/logo/django)
- [x] [VIM](https://worldvectorlogo.com/logo/vim)
- [x] [MongoDB](https://worldvectorlogo.com/logo/mongodb)
- [x] [Redis](https://worldvectorlogo.com/logo/redis)
- [x] [Neo4J](https://worldvectorlogo.com/logo/neo4j)
- [x] [Arch Linux](http://cinderwick.ca/files/archlinux/artwork-official/symbol.svg)
- [x] [Ubuntu](https://worldvectorlogo.com/logo/ubuntu-4)
- [x] [Fedora](https://worldvectorlogo.com/logo/fedora-1)

I'm not sure how some of these icons are usable for a mono version.

davidklebanoff:
All the icons have been added. Special thanks to @ericcornelissen for knocking most of them out!

PR #570, opened 2017-09-25
Add Django

xdesro:
Proposed in #487.


birjolaxew:

PR #581, opened 2017-09-26
Clean up Hatena Bookmark.

davidklebanoff:
Scaled and centered the icon, it was small and off-center before.


birjolaxew:
:+1: Merged. Discussion on padding ongoing at #582

Issue #585, opened 2017-10-04
Update: Dropbox


ericcornelissen:
Dropbox updated their logo, see [link](https://www.dropbox.com/branding)

birjolaxew:
Fixed in #586

PR #586, opened 2017-10-09
Updated dropbox to new branding.

davidklebanoff:
Addressed issue #585.

Not sure if the new branding has a new color, or if the old one was just wrong. Either way, I updated it to the current branding guidelines as well.


birjolaxew:

Issue #582, opened 2017-09-26
Update: Clarify Icon Padding Guidelines


davidklebanoff:
We should discuss the guidelines for the padding on SVGs. What standards should be followed (perhaps they should be stated more explicitly as well), and how we want to clean up the current icons so they meet those standards.

The current contribution guidelines make reference to [Google’s Material Design guidelines for system icons](https://material.io/guidelines/style/icons.html#icons-system-icons), but it seems that many of the icons in Simple-Icons follow no such standard with padding spanning from 0 to quite a bit.

So perhaps:

1. Clarify guidelines.
2. Publish guidelines clearly.
3. Update existing icons to meet guidelines.
4. Discuss procedures for enforcing guidelines on future contributions.

birjolaxew:
The design spec is currently a bit of a mess.

We recently switched to 24x24 viewbox, following the Material Design guidelines. This was mostly to allow us to use their proportional sizing guides ([see this](https://github.com/simple-icons/simple-icons/issues/472#issuecomment-309722443) and [this discussion](https://github.com/simple-icons/simple-icons/issues/477)).

This means that icons should technically be fit to one of the defined guides (20x16 rectangle, 18x18 square, or 20 circle) where possible. Unfortunately this is a bit cumbersome, both for contributors and maintainers. It also isn't easily lint-able (for when I eventually get SVGLint finished).

An alternative is to say that all icons should be their maximum size inside the 24x24 viewbox. This would disregard proportional sizing, but would be far easier to not only contribute, but also maintain and lint.

A third option is to fit the viewbox to the icon (with the longest side being 24), so that no extra whitespace exists. This would allow people to size it themselves easily (e.g. decide if they want all icons to have equal width or equal height), but would also completely ignore proportional sizing.

I'm not sure which option is the best; input is appreciated.

davidklebanoff:
@birjolaxew, Thanks for providing some background on the discussion. While the spec may need improvement in parts, it’s nice to see the project has been moving in a positive direction and I think the past decisions on things like the viewbox size is a good step forward.

With so many icons, I think automation should play an important role moving forward - and like you, I think having clean and simple input to the automated processes should be prioritized. So with all that said, it seems a 24x24 viewbox, with the icon max-scaled to fit such that it touches two to four sides would be a good baseline. Then automation processes can manipulate that icon as we see fit to whatever final version it should take.

[...]

PR #590, opened 2017-10-09
Remove padding from icons and resize to 24x24

birjolaxew:
See discussion in #582.

Icons that need manual fixing:

- compropago (check if correct size)
- nodejs (failed in script)
- garmin (check if correct size)
- squarespace (check if correct size)


davidklebanoff:
I noticed icons/angular.svg had its role and aria-labelledby removed. Not sure if any other icons were affected by a similar issue as well.

birjolaxew:
@davidklebanoff That seems likely - I am running it through SVGO with default settings, which might be doing the removal.

I'll have a look later

Issue #591, opened 2017-10-09
Fix: compropago, nodejs, garmin and squarespace


birjolaxew:
These icons failed during the resizing script from #590. They should be manually resized to 24x24, and have any potential padding removed.

PR #593, opened 2017-10-09
Resize icons and remove padding

birjolaxew:
Second go at this (thanks @davidklebanoff for pointing out a mistake last time :smile:)

Still requires manual handling of the files mentioned in #591


birjolaxew:
@davidklebanoff The main incentive behind having a newline at the end is that unix files are expected to end with one. In fact, the newlines in this PR are from the `echo` commandline utility.

In the future when more work will be done by scripts, it seems reasonable to have newlines be the default. I am open for removing it if that is the general consensus though.

davidklebanoff:
Fair enough. I bounce back in forth on whether putting them in another commit will make it easier to identify important changes, or just further clutter the commit histories of each file. Your call, there is no right or wrong answer.

birjolaxew:
@davidklebanoff I've removed the newlines from the files (lowered it from ~350 to ~70, much easier to work with :smile:)

Had some git fuckery, but not enough time to look at it right now. Should be good tho.

davidklebanoff:
@birjolaxew, much more manageable. Looks good to me! Great job!

davidklebanoff:
@birjolaxew This is a big one! Thanks for getting this done!

PR #626, opened 2017-10-14
Add MongoDB icon

ericcornelissen:
As requested in #487 

The color was a bit tricky since I couldn't find any guidelines on it over [here](https://www.mongodb.com/pressroom) and the logo has multiple different shades of green... Let me know if any if you think it should be changed 🙂


davidklebanoff:
I see the icon has a `clip-rule="evenodd"`. Gotta get rid of that.

ericcornelissen:
Good catch, luckily it did nothing so it was easy to update 👍

davidklebanoff:
Hmm, so I don't know if it's the compression, but the SVG icon version looks a lot rougher than the versions I found in the media kit on this page https://www.mongodb.com/brand-resources. The edges of the leaf and the stem have wavy edges and cut-ins when it appears like they should be straight curves.

[MongoDB_Logos.zip](https://github.com/simple-icons/simple-icons/files/1384866/MongoDB_Logos.zip)

ericcornelissen:
Hmm 🤔 I agree with you that it doesn't look all that good... I'll see what I can do to improve it 👍

ericcornelissen:
@davidklebanoff this should look better, let me know if you see any more improvements. It was not the compression, it was due to the way I transformed the  `.jpg` to a vector image

davidklebanoff:
@ericcornelissen Good to know. In the future, if you can't find an SVG, but an EPS is available (like for MongoDB), you should try transforming the EPS file because it's already vector format and there won't be data loss. Going from JPG to vector will always results in a lower quality transformation.

There are lots of online tools in addition to offline tools for transforming EPS to SVG, but I'm not terribly familiar with them and don't know if some are better than others.

ericcornelissen:
I didn't know EPS was also a vector format 😅 Good to know! I will update the `.svg` based on the `.eps` in [MongoDB_Logos.zip](https://github.com/simple-icons/simple-icons/files/1384866/MongoDB_Logos.zip)

davidklebanoff:
Looks good to me! Thanks for the hard work. ;)

PR #627, opened 2017-10-14
Update Patreon icon

fritz-c:


fritz-c:
I have one reservation about this one, in that I resized the svg by hand in an svg editor in order to fit the view box of `0 0 24 24`. Is there a better way?

ericcornelissen:
I'm not sure if there is a better way to do it, but it is not a problem that you did

The new logo looks good, only the icon is not properly centered vertically, if you could change that it can be merged 😃

ericcornelissen:
I'm not sure if there is a better way to do it, but it is not a problem that you did

The new logo looks good. I have two minor things though, if you could change them it can be merged:
- The icon is not properly centered vertically
- Patreon also changed their colors to  #F96854 (Coral) and #052D49 (Navy), I would suggest using Coral for the monochrome icon

fritz-c:
@ericcornelissen I made a little discovery in that I could wrap the svg with `<g transform="scale(0.24, 0.24)">`, and it would be optimized away by svgomg, resulting in nicely modified path values. The same thing worked for `translate`. It's not the ideal method perhaps, but it's good in a pinch.

ericcornelissen:
@fritz-c that's a nice find! Perhaps that we can include it in the contributing guideline as a tip 👍 

One comment on your change, as you can read [here](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#json-data-for-simpleiconsorg) the color shouldn't be set as a `fill` on the `.svg`, but instead in the JSON data f0r SimpleIcons. In this case you only have to change the `"color"` value for the `"title": "Patreon"` to `F96854` (note no `#`) and remove the `fill` from the `.svg`.

fritz-c:
Fixed.

davidklebanoff:
Merged. Thanks so much for the contribution! Your first, Congrats! 🎉

PR #629, opened 2017-10-14
Add Vim icon

ericcornelissen:
As requested in #487

Added a monochrome version of the Vim icon. For the color I used the green of the big rectangle in the middle of the logo.


davidklebanoff:
This is a good one!

Quick question. Did you look into the idea of having the black line around the dotted `i`? I'm thinking it might make the `i` more readable.

Something like this, where the grey is transparent and the green is black:
https://commons.wikimedia.org/wiki/File:Icon-Vim.svg

Thoughts?

ericcornelissen:
Now that you point it out, I see that it makes more sense to include the line for the dot of the `i`. I will update it 👍

davidklebanoff:
Much better! Merged. Thank you!

Issue #387, opened 2017-04-14
Bower package


Soullivaneuh:
Same request as #353 for npm, could you provide and register a package definition for bower.io?

Your library looks great, it would be greater if we could install it with ease. :wink:

ericcornelissen:
Just going to reference #164 here, I personally see no harm in supporting Bower and as opposed to back them there now seem to be some people that are interested.

It seems @birjolaxew thinks the same (given he added some labels) but I still want to ask him explicitly given #164 before I start working on it 🙂

birjolaxew:
@ericcornelissen Go ahead! The more support there is, the better.

On a sidenote, this is as much my project as it is yours or @davidklebanoff 's :smile:

ericcornelissen:
Okay, thanks for the sidenote, wasn't entirely sure about that 😛

ericcornelissen:
Closing this as Bower itself is recommending users to use other systems instead (thanks to @styfle for pointing that out!), from their website:

> psst! While Bower is maintained, we recommend using Yarn and Webpack for front-end projects read how to migrate!

If anyone still feels like we should support Bower, reopen this issue with a compelling reason to do so 🙂

styfle:
> Your library looks great, it would be greater if we could install it with ease. 

Also note you can use it without installing with a CDN.
See the [README CDN Usage](https://github.com/simple-icons/simple-icons/blob/develop/README.md#cdn-usage) section that I added.

PR #639, opened 2017-10-15
Add bower.json to allow publishing on Bower.io

ericcornelissen:
As requested in #387, when merged to `master` we can publish to [bower.io](https://bower.io/)

Added a `bower.json` file based on the existing `package.json` and `.npmignore` (see the ignore section). Also changed the order of the keys in `package.json`

/cc @timgluz

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,14 @@ Icons can be downloaded as SVGs directly from [our website](https://simpleicons.
 
 ### Node Usage
 
-The icons are also available through our npm package. To install, simply run:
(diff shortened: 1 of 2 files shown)


styfle:
Does it make sense to publish new packages to bower when it seems as if they are encouraging [migrating away from bower](https://bower.io/blog/2017/how-to-migrate-away-from-bower/)?

ericcornelissen:
Thanks for the information @styfle, I will close this PR and related issues based on it 🙂

PR #640, opened 2017-10-16
Updated Contributing Guidelines

davidklebanoff:
Basically a full rewrite of the contributing guidelines. It includes guidance on how to make new icon requests, how to add icons, and how to label issues. This should hopefully make everyone's lives a bit easier, new contributors and maintainers.

I'd love to have more in depth documentation on how to prepare an icon using various different tools, but that's for another time and probably a separate set of files. Either way this should be a huge step in the right direction.

It's worth noting I took out some previous guidelines and added some new ones which I believe were unspoken but closely followed. This should eliminate a lot of confusion for new contributors as well.

Just like not all of our current icons meet our contribution guidelines, not all of them will meet these new guidelines. This is addressed by issue #502.

Issue #630. Also touches on #472 and #582, which I'll re-evaluate after this is merged.

PR #613, opened 2017-10-13
Add Redis icon

ericcornelissen:
As requested in #487 

Since there is no monochrome version defined by Redis (as far as I know) I choose to go with a version where the middle parts are transparent, as well as the figures in the top part of the logo. For the color I used to lighter red used in the logo.

In case you're interested I created 3 more monochrome versions of the logo, you can download them all [here (.zip)](https://github.com/simple-icons/simple-icons/files/1384164/Redis.zip), let me know if another one is preferred 😎

PR #644, opened 2017-10-17
Make docs more "casual"

ericcornelissen:
Small update to the changes #640 and d9f180d to make both the Contributing Guidelines and Issue Template a bit more casual/friendly/inviting as suggested by @birjolaxew ([ref](https://github.com/simple-icons/simple-icons/pull/640#pullrequestreview-69704075)):

> ... There could also be a point in thinking about how the rules are worded (e.g. instead of please make sure the issue includes ..., a more casual it'd help if your issue includes ... might be more welcoming)

I welcome any feedback to these changes, as they can probably be improved/are not complete!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,7 +107,6 @@ The final icon should:
   * This means the icon should be touching at least two sides of the viewbox.
 * Be vertically and horizontally centered.
 * Be minified to a single line with no formatting.
-* Contain only one single path.
 * Not contain extraneous attributes.
   * This includes: `width`, `height`, `fill`, `stroke`, `clip`, `font`, etc.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,19 +11,18 @@ Simple Icons welcomes contributions and corrections. Before contributing, please
 
 ## Requesting An Icon
 
-We welcome icon requests. However, before you submit a new issue please make sure the icon:
+We welcome icon requests, before you submit a new issue please make sure the icon:
 
 * Is of a _popular_ brand.
 * Has not already been requested.
-* Can be represented in a monochromatic format.
 
-When submitting a request for a new or updated icon, please submit the issue with the following information:
+When submitting a request for a new or updated icon include helpful information such as:
 
-* **Issue Title:** Include the brand name. For example:
+* **Issue Title:** The brand name. For example:
   * New Icons: `Request: GitHub Icon`
   * Icon Updates: `Update: GitHub Color` or `Update: GitHub Icon`
 
-* **Issue Body:** Include links to official sources for the brand's icon and official colors (e.g. media kits, brand guidelines, SVG files etc.)
+* **Issue Body:** Links to official sources for the brand's icon and colors (e.g. media kits, brand guidelines, SVG files etc.)
 
 ## Adding or Updating An Icon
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -11,19 +11,18 @@ Simple Icons welcomes contributions and corrections. Before contributing, please
 
 ## Requesting An Icon
 
-We welcome icon requests. However, before you submit a new issue please make sure the icon:
+We welcome icon requests, before you submit a new issue please make sure the icon:
```
davidklebanoff: These are two separate statements. Therefore, a period (or semicolon) would be more appropriate punctuation than a comma.

CONTRIBUTING.md
```diff
@@ -11,19 +11,18 @@ Simple Icons welcomes contributions and corrections. Before contributing, please
...
 
 * Is of a _popular_ brand.
 * Has not already been requested.
-* Can be represented in a monochromatic format.
```
davidklebanoff: We've had issues with this in the past, see Sendgrid. But I understand that this may not be the most clear to a new contributor.

CONTRIBUTING.md
```diff
@@ -11,19 +11,18 @@ Simple Icons welcomes contributions and corrections. Before contributing, please
...
 
-When submitting a request for a new or updated icon, please submit the issue with the following information:
+When submitting a request for a new or updated icon include helpful information such as:
```
davidklebanoff: `When submitting a request for a new or updated icon, please include helpful information such as:`
Sounds a little more casual (and less demanding) to me. :)

davidklebanoff:
I provided a little feedback for the CONTRIBUTING.md, the parts I didn't comment on look good to me!

As for the ISSUE_TEMPLATE.md, I don't feel like it's an improvement. It removes the "please" which makes it sound a little more demanding, but removes simple requests like title formatting, which just means more work for maintainers... But maybe I'm just biased.

CONTRIBUTING.md
ericcornelissen: I agree that it can cause issues, but I think it is better to have the maintainers judge if the icon is monochromatic. Take e.g. [Redis](https://github.com/simple-icons/simple-icons/pull/613): it is not clear if it is representable in a monochromatic format, but with some effort it is definitely possible

CONTRIBUTING.md
ericcornelissen: It is interesting how we have a different view on this 😅 I personally feel like including "please" is kinda demeaning. I think we should get a third opinion (@birjolaxew) on this and just go with that

CONTRIBUTING.md
ericcornelissen: Got it, will update this when there is more to change 👍

ericcornelissen:
@davidklebanoff I see your concern with regards to the ISSUE_TEMPLATE.md, but I think that in most cases the title people will choose when requesting a new/update to an icon will be clear enough and the exact format (at least to me) doesn't matter as much.

My version still asks people to include the brand name in the title of the issue, perhaps it is an idea to change `The requested icon’s brand name, preferably in the title.` to `The requested icon’s brand name in the title.`?

davidklebanoff:
@ericcornelissen Sounds good to me!

CONTRIBUTING.md
ericcornelissen: I guess [that](https://github.com/simple-icons/simple-icons/pull/644#pullrequestreview-70020565) settles it 🙂

Issue #616, opened 2017-10-13
Full Icon Review


davidklebanoff:
Sometimes brand logos change and the icons in SimpleIcons become old/incorrect. Occasionally people submit issues when they find a discrepancy between SimpleIcons and official sources, but sometimes brand updates go unnoticed. As SimpleIcons approaches 5 years in age, I think it may be time to do a quick review of all the icons to make sure they match their brand's current icon/logomark.

So, below is a list of all icons currently within simple icons. Please help us review all the icons to make sure they are up to date. It's simple:

1. Find an unchecked icon below.
1. Goto the brand's official website.
1. Check if the icon matches the official brand's current logo/icon. Good places to look are the website's header, favicon, and branding guidelines. Look closely!
1. If the SimpleIcon's icon matches the official record, check the box. If it does not, please open a new issue and put the issue # next to the icon name in the list below. A color check would be nice too.

Any help is appreciated, thanks!

- [x] 500px
- [x] aboutme (#617)
- [x] acm
- [x] addthis
- [x] adobe
- [x] adobeacrobatreader
- [x] adobeaftereffects
- [x] adobedreamweaver
- [x] adobeillustrator
- [x] adobeindesign
- [x] adobelightroom
- [x] adobephotoshop
- [x] aim (#809)
- [x] air (#621)
- [x] airbnb
- [x] amazon
- [x] americanexpress
- [x] android
- [x] angellist
- [x] angular
- [x] apple
- [x] applemusic
- [x] appnet (#618)
- [x] asana
- [x] atlassian
- [x] atom
- [x] auth0
- [x] automatic
- [x] baidu
- [x] bandcamp
- [x] basecamp  (#636)
- [x] bathasu
- [x] bebo (#634)
- [x] behance
- [x] bigcartel
- [x] bing
- [x] bitbucket (#642)
- [x] bitcoin
- [x] bitly
- [x] blogger (#643)
- [x] bootstrap
- [x] brandai
- [x] buffer
- [x] buzzfeed
- [x] cakephp
- [x] campaignmonitor (#668)
- [x] circleci
- [x] codecademy
- [x] codeforces
- [x] codeigniter
- [x] codepen
- [x] coderwall
- [x] codeschool
- [x] codio
- [x] compropago
- [ ] conekta
- [x] coop
- [x] coursera
- [x] css3
- [ ] csswizardry
- [ ] dailymotion
- [x] deezer
- [ ] delicious
- [ ] designernews
- [ ] deviantart
[...]

Issue #625, opened 2017-10-14
A list of all popular technologies/brands/companies that are missing from SimpleIcons


DoronBrayer:
It's going to be overwhelming, it's a lot...
Here is a list of all popular technologies/brands/companies that are missing from SimpleIcons:

- [ ] .NET
- [ ] Adobe PhoneGap
- [ ] Adobe XD (Experience Design)
- [ ] AFNetworking
- [ ] Akismet
- [ ] Alamofire
- [ ] Alibaba Dubbo
- [ ] Alibaba Weex
- [ ] Amazon Aurora
- [ ] Amazon EC2
- [ ] Amazon S3
- [ ] Amazon SES
- [x] AMD
- [ ] Animate.css
- [ ] Ansible
- [ ] Apache
- [ ] Apache Cassandra
- [ ] Apache Cordova [formerly PhoneGap]
- [ ] Apache CouchDB
- [ ] Apache Hive
- [ ] Apache Solr
- [ ] Apache Sprak
- [ ] Apple iCloud
- [ ] ASP.NET
- [ ] Aurelia
- [ ] Autodesk
- [ ] AWS (Amazon Web Services)
- [ ] Babel
- [ ] Backbone.js
- [ ] Beego
- [x] Bower
- [ ] Brackets
- [ ] Bulma
- [ ] ~C~ (duplicate of #469)
- [ ] C# (C Sharp)
- [ ] ~C++~ (duplicate of #469)
- [ ] Caddy
- [ ] certbot
- [ ] Ceylon
- [ ] Chart.js
- [x] Chrome (#695)
- [ ] Cloud Foundry
- [ ] CockroachDB
- [ ] ~CoffeeScript~ (duplicate of #469)
- [ ] CryEngine
- [ ] D3.js
- [ ] Debian
- [x] ElasticSearch
- [ ] Electron
- [ ] Element toolkit
- [ ] Elixir
- [ ] Elm
- [ ] Ember
- [ ] Erlang
- [ ] Express
- [ ] Fastlane
- [ ] Firebase
- [x] Firefox
- [ ] Flask
- [ ] Flux
- [ ] GIMP
- [ ] Gin-Gonic
- [x] ~GitLab~ (already exists)
- [ ] Go (Golang)
- [ ] GoDaddy
- [ ] Gogs
- [ ] Google Cloud Platform
- [ ] Grafana
- [ ] Grunt
- [ ] Handlebars
- [ ] Hugo
- [ ] Impress.js
[...]

PR #645, opened 2017-10-17
Update Swarm icon

ericcornelissen:
Related to #616

- ~~Reduce the SVG to a single <path> element.~~
- Change the color to a primary web color (`#FFA633`). I couldn't [find](https://github.com/simple-icons/simple-icons/pulls?utf8=%E2%9C%93&q=swarm%20) anything about earlier choices related to this icon nor could I found a reliable source for the color.


davidklebanoff:
Oh, this one is interesting.

So did you decide to redo this one because it has two paths? Because your outputted icon has two paths! Which of course is very interesting. It seems running the icon through SVGO outputs two paths for this icon, which challenges our assumption that the icons would all be reduced to one path! Perhaps our use of the tools is wrong,  or maybe the guidelines need to be updated to relax the rules a bit? I think a bit more investigation into what's going on here is needed.

It's probably worth looking into how many of our icons have 1 path vs more than one and understanding why that is. It's possible that we wrongly assumed only one was needed?

ericcornelissen:
> So did you decide to redo this one because it has two paths?

Initially I did it because of the color, and then I saw the icon had two paths as well so I, indeed, tried to change that and didn't notice that SVGO changed it back 😅 


> Perhaps our use of the tools is wrong, or maybe the guidelines need to be updated to relax the rules a bit?

Given that SVGO is (more or less) the last step in the process of creating an icon, and it does output icons with multiple paths I think it is a good idea to relax that rule. Perhaps even removing it and trusting that SVGO gives a good enough icon.


> It's probably worth looking into how many of our icons have 1 path vs more than one and understanding why that is. It's possible that we wrongly assumed only one was needed?

After quickly going through the iconbase I found that there are certainly a number of icons with multiple `<path>`s, most of them have white space in them (e.g. Swarm, Event Store, Redis)

davidklebanoff:
So, with some quick playing around with SVGOMG, it seems that it doesn't always optimize icons down to a single path. Even when tweaking various settings like multi-pass.

I'm curious as to why, but that's going to require some investigation that I don't have time for at the moment. It seems rather low priority anyway.

So with that said, this looks good to me!

ericcornelissen:
> I'm curious as to why, but that's going to require some investigation that I don't have time for at the moment. It seems rather low priority anyway.

Same here 😅 I guess for the time being we could remove it from the Contributing Guidelines given the problems in #625, which can be done in #644?

davidklebanoff:
@ericcornelissen Yeah, go for it.

PR #647, opened 2017-10-17
Add CDN Usage for JSDelivr and Unpkg

styfle:
I think it would be nice to show users how to use a CDN since this solution requires no downloading or installing to get started.


davidklebanoff:
Great addition! Thanks!

Issue #648, opened 2017-10-17
Allow for search query in URL


ericcornelissen:
Not hugely important, but might be nice to have 😎 

What I mean by this is to allow for URLs such as `https://simpleicons.org/?q=angular` (or a similar format), which would result in the browser initializing the page as follows:

![preview](https://user-images.githubusercontent.com/3742559/31683397-34ed12ac-b37d-11e7-98e9-ecf776b8fd03.png)

**DISCLAIMER:** Given the nature of Jekyll, as far as I know it is not possible to implement this using HTTP get request. Hence, it must be implemented using JavaScript on top of the existing [search feature](https://github.com/simple-icons/simple-icons/pull/459). This will require [manipulating the browser history](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to update the URL in the users browser, which is only supported by modern browsers (but I expect this is not a problem given our target audience).

Another issue is related to #649 where user A sharing a query-link with user B that has JavaScript disabled might cause some confusing between these two users.

Thoughts?

birjolaxew:
We might be able to do some fancy hacking with `:target` in CSS (and `#` instead of `?q=` in the URL) so that this doesn't rely on JS.

ericcornelissen:
After some fiddling I wasn't able to get it working properly, the closest I got was the snippet below but this doesn't hide icons before the selected icon:

```css
.grid-item:target {
  display: block;
}
.grid-item:target ~ .grid-item {
  display: none;
}
```

Another problem with this approach would be that you cannot share a query that gives back multiple results, although (if someone can get your suggestion working) we could use it as a fallback method, but that can still be a point of confusion...

birjolaxew:
@ericcornelissen See the edit of my previous comment. While it is possible, it brings way too many cons along with it :disappointed:

ericcornelissen:
> Although it is possible, it doesn't work with `history.replaceState`. This means that the browser's back button is completely broken when using this technique...

Can you elaborate on that @birjolaxew? As far as I know `history.replaceState()` can be used (but isn't in your example because of the limitations in Codepen, I think) but **that** would break the back-button, see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_replaceState()_method). The method used in the pen, `window.location.hash = "#"+val;`, works with the back button (at least when I test it locally).

[...]

PR #649, opened 2017-10-17
Small usability improvement for users with JavaScript disabled (low priority)

ericcornelissen:
By default the search input is now disabled and when hovered displays a message informing the user search is only available when JavaScript is enabled. When JavaScript is enabled, the input is programmatically enabled and the title is changed as well.

Feedback is welcome 🙂

### DIFF
--- a/index.html
+++ b/index.html
@@ -488,7 +488,7 @@
         <div class="search">
             <div class="search__wrapper">
                 <div class="search__close"><span>&times;</span></div>
-                <input type="text" placeholder="Search by brand …" />
+                <input type="text" placeholder="Search by brand …" title="Search not available when JavaScript is disabled" disabled />
             </div>
         </div>
         <ul class="grid">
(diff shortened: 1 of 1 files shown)


davidklebanoff:
Sounds good to me, but if when you implement #648, it seems to me the fallback solution would be to have a search button that submits the search form with an HTTP get with the query parameter, reloading the page. So this PR would be effectively removed when that's complete.

ericcornelissen:
You bring up an interesting point 😅 But I will have to look into how that will be implemented as I'm not sure if HTTP get request work with Jekyll [[ref](https://github.com/jekyll/jekyll/issues/1334)], if not that feature will/can be implemented using JavaScript...

birjolaxew:
:+1: Great idea! An alternative solution would've been to use `<noscript>`, but this seems to be fine as-is. Merged!

Issue #651, opened 2017-10-22
Cut off footer


mudlabs-monster:
Noticed a little issue with the footer, most notably the "Share this on Twitter" button. **(See screenshot)**

As tested this seems to be a problem on:
- Chrome v61
- Safari v11
- Opera v48

![screen shot 2017-10-22 at 3 29 56 pm](https://user-images.githubusercontent.com/32623552/31858219-24ffe864-b73e-11e7-9d30-159e6af6e5cc.png)

birjolaxew:
Closed in #652 thanks to @mudlabs-monster

PR #652, opened 2017-10-22
Fix for issue #651 "Cut off footer"

mudlabs-monster:
It looks like there was an issue with `<main>` taking up too much space when filled out by all the icons.
 
When I used the search/filter input to reduce the visible icons the issue eventually resolves itself when there's enough room.

I've added `align-items: center` to the `.footer` rule and it's all good now, on the tested browsers.
- Chrome v61
- Safari v11
- Opera v48
- Firefox v56

### DIFF
--- a/index.html
+++ b/index.html
@@ -232,6 +232,7 @@
         @media (min-width: 45rem) {
             .footer {
                 display: flex;
+                align-items: center;
                 justify-content: space-between;
                 padding: 3rem;
           }


birjolaxew:
:+1: Great catch! Apparently the flexbox on the footer sets the height of the children to 0 if `align-items` isn't set for some reason.

Closes #651

PR #658, opened 2017-11-01
Add support for search query in URL

ericcornelissen:
As discussed in #648, implemented JavaScript based support for a search query in the URL. This feature does two things:
1. If you go to `www.simpleicons.org/?q=adobe`, the page will load with a search query for the string `'adobe'`.
2. If you start searching on `www.simpleicons.org`, the url will be updated with each keystroke to `www.simpleicons.org/?q=hellowor`.

Since `window.history.replaceState` is being used, this changing of the URL does not mess with the browser history. I.e. if the user came from www.google.com, started searching and press the back button they will return to www.google.com immediately.

### DIFF
--- a/index.html
+++ b/index.html
@@ -536,12 +536,13 @@
     {% endfor %}
     <script type="text/javascript">
         (function(document) {
-            var icons        = [{{ allIconNames }}].map(normalizeSearchTerm),
-                $grid        = document.querySelector('.grid'),
-                $icons       = $grid.querySelectorAll('.grid-item:not(.grid-item--ad)'),
-                $search      = document.querySelector('.search'),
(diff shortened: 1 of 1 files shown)


birjolaxew:
`URLSearchParams` isn't supported by IE, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#Browser_compatibility). Perhaps we could parse the URL parameters ourselves?

ericcornelissen:
Good point, I updated it using a fallback I found [here](https://davidwalsh.name/query-string-javascript) 👍

birjolaxew:
:+1: Looking good! Merged

PR #659, opened 2017-11-01
Add fuzzy search

ericcornelissen:
Updated the search method from a simple partially-equals method to a fuzzy search (also Approximate string matching) implementation. 

**Pro**: this method is more forgiving, allowing users to make typos.
**Con**: search is more expensive in terms of CPU.

Let me know what you think and definitely try it out if you're unfamiliar with fuzzy search!

* * *

My implementation is based on this implementation: https://github.com/wouter2203/fuzzy-search

### DIFF
--- a/index.html
+++ b/index.html
@@ -585,16 +585,51 @@
 
             function search(value) {
                 var hiddenCounter = 0,
-                    value = normalizeSearchTerm(value);
+                    query = normalizeSearchTerm(value);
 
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Optimally we'd want a fuzzy algorithm that also handles substitution, but I see no need to force that onto this PR. Great idea and good work!

PR #655, opened 2017-10-24
Add Staticman icon

iriiiina:
https://staticman.net

PR #667, opened 2017-11-12
Staticman

davidklebanoff:
Adds Staticman icon that meets icon guidelines. Addresses #655.


birjolaxew:
:+1: Thanks for the work!

PR #673, opened 2017-11-22
Add Humble Bundle icon

ericcornelissen:
I decided to use the H logomark over the "bindle" because that isn't present on their homepage anymore and I couldn't find any preference on their support site.


birjolaxew:
:+1: Looking good! Merged and published.

PR #675, opened 2017-11-26
Update TripAdvisor teal

ThomasJClark:
#616 
The TripAdvisor logo is now teal instead of green


birjolaxew:
:+1: Thanks for noticing! Merged and published

PR #683, opened 2017-11-30
Fix grid item overflow due to text

### LINKED ISSUE #682: Grid item overflow/overlap in Firefox 57.0
Seems like if the title of a grid item is too long, it overflows the box. Maybe the text should instead wrap and the entire row should increase in height slightly to accommodate?

Here are three examples of the issue:

screenshot_20171128_193934

Doesn't seem to be a problem in Google Chrome for me.

ericcornelissen:
Closes #682 

Solved the grid overflow bug by allowing for wrapping in grid item titles, and centering the contents of grid-items using flexbox. Also slightly increased the size of grid-items.
Note that this solution only works properly as long as titles don't get longer than two lines, which I don't expect to happen 🙂 

@rosslh can you confirm if this solves the problem on Ubuntu as well?

### DIFF
--- a/index.html
+++ b/index.html
@@ -349,7 +349,7 @@
             .grid {
                 display: grid;
                 grid-template-columns: repeat(auto-fill, minmax(9rem, 1fr));
-                grid-auto-rows: 6.25rem;
+                grid-auto-rows: min-content;
                 grid-column-gap: 0.375rem;
                 grid-row-gap: 0.375rem;
                 grid-auto-flow: dense;
@@ -423,8 +423,6 @@
             font-weight: 400;
             line-height: 1rem;
             margin: 0;
-            overflow: hidden;
-            text-overflow: ellipsis;
         }
 
         .grid-item__subtitle {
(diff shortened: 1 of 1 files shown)


rosslh:
Just tested locally and it looks great! Thanks 👍

index.html
```diff
@@ -349,7 +349,7 @@
             .grid {
                 display: grid;
                 grid-template-columns: repeat(auto-fill, minmax(9rem, 1fr));
-                grid-auto-rows: 6rem;
+                grid-auto-rows: 6.25rem;
```
birjolaxew: I think `grid-auto-rows: min-content` would fit better (since it automatically adjusts to the content height)

index.html
```diff
@@ -422,7 +425,6 @@
             margin: 0;
             overflow: hidden;
```
birjolaxew: We can also remove `overflow: hidden` and `text-overflow: ellipsis` now that there no longer is any overflow

birjolaxew:
:+1: Thanks for fixing this!

PR #690, opened 2017-12-02
Update MailChimp, ProtonMail and VueJS color

ericcornelissen:
As per #616 

- The old color of ProtonMail was slightly off, I picked the color from [the purple logo](https://protonmail.com/images/media/logos/protonmail-logo-dark.svg) on the media kit page
- ~~The color of VueJS was also slightly off as well, I extracted the new color from [the png version of the logo](https://github.com/vuejs/art/blob/master/logo.png)~~
- Same for MailChimp, I picked the color most similar to the old color on their [Brand Assets page](https://mailchimp.com/about/brand-assets/)

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1272,7 +1272,7 @@
         },
         {
             "title": "ProtonMail",
-            "hex": "9596C6",
+            "hex": "8B89CC",
             "source": "https://protonmail.com/media-kit"
         },
         {
@@ -1792,7 +1792,7 @@
         },
         {
             "title": "Vue.js",
-            "hex": "41B883",
+            "hex": "4FC08D",
             "source": "https://github.com/vuejs/art"
         },
         {
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -922,7 +922,7 @@
         },
         {
             "title": "MailChimp",
-            "hex": "2C9AB7",
+            "hex": "239AB9",
             "source": "http://mailchimp.com/about/brand-assets"
         },
         {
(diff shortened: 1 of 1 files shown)


davidklebanoff:
I believe the previous Vue color may have been extracted from the css on the page, which looks like it was meant to match the official color of the logo. Examples of the logo next to css colors can be seen here: https://vuejs.org/ - It's not to say one is better than another, but be careful extracting colors from PNGs, GIFs, and JPGs, as they can be changed when saved due to compression or embedded color profiles.

davidklebanoff:
For ProtonMail, if you goto the [media page](https://protonmail.com/media-kit) and download the SVG version of the logo (https://protonmail.com/images/media/logos/protonmail-logo-purple.svg), the color can be found in the source as `#8b89cc`.

ericcornelissen:
> I believe the previous Vue color may have been extracted from the css on the page, which looks like it was meant to match the official color of the logo.

I couldn't find anything about it in the history of the repo, the only thing I could find was [this commit](https://github.com/simple-icons/simple-icons/commit/430a73413696b051f72f9cdfe3a4d2d90745c2de). It does seem the old color was extracted from the CSS as the "Get started" button uses `#4fc08d`, but the button "Github" next to it uses `#42b983` for its text and so do the headers just below that... that's not to say my choice is better, just pointing out that the CSS is also not perfect.


> be careful extracting colors from PNGs, GIFs, and JPGs, as they can be changed when saved due to compression or embedded color profiles.

But that's a very good point so I guess I will revert that change, especially since the difference is so minor 👍 


> For ProtonMail, if you goto the media page and download the SVG version of the logo, the color can be found in the source as `#8b89cc`.

That is even better, I'll update it 😄

davidklebanoff:
Awesome. Thanks for digging through all that @ericcornelissen.

PR #691, opened 2017-12-03
Yelp: wrong color

louistsaitszho:
According to https://www.yelp.com/brand, the primary color is the Five-star red, and it should be #D32323

Part of #616


ericcornelissen:
Thank you, much appreciated!

Issue #469, opened 2017-06-06
Request: Programming languages


birjj:
Related to #468 

- [x] JavaScript
- [x] Ruby
- [x] [Java](https://en.wikipedia.org/wiki/File:Java_programming_language_logo.svg)
- [x] [Python](https://en.wikipedia.org/wiki/File:Python_logo_and_wordmark.svg)
- [x] [PHP](https://en.wikipedia.org/wiki/File:PHP-logo.svg)
- [x] [Lua](https://en.wikipedia.org/wiki/File:Lua-logo-nolabel.svg)
- [x] [Go](https://golang.org/doc/gopher/gopherbw.png)
- [x] Java

[Edit] More, as noted by @greybax:

- [x] [Standard C++](https://github.com/isocpp/logos)
- [x] [Swift](https://commons.wikimedia.org/wiki/File:Swift_logo_with_text.svg)
- [x] [Rust](https://commons.wikimedia.org/wiki/File:Rust_programming_language_black_logo.svg)
- [x] [CoffeeScript](http://coffeescript.org/)
- [x] [Kotlin](https://commons.wikimedia.org/wiki/File:Kotlin-logo.png)
- [x] [Scala](https://en.wikipedia.org/wiki/File:Scala_logo.png)
- [x] [PowerShell](https://commons.wikimedia.org/wiki/File:PowerShell_Core_6.0_icon.png)
- [x] [Clojure](https://commons.wikimedia.org/wiki/File:Clojure_logo.svg)

greybax:
Take some more popular languages from https://github.com/showcases/programming-languages:

- [ ] C++ [this](https://commons.wikimedia.org/wiki/File:The_C%2B%2B_logo.svg) or [this one](https://worldvectorlogo.com/ru/logo/c)
- [ ] [swift](https://camo.githubusercontent.com/de32b354687f1cd9b05a89e4aa03c7f2d311f294/68747470733a2f2f73776966742e6f72672f6173736574732f696d616765732f73776966742e737667)
- [ ] [Rust](https://www.rust-lang.org/logos/rust-logo-blk.svg)
- [ ] [coffeescript](https://worldvectorlogo.com/ru/logo/coffeescript)
- [ ] [kotlyn](https://worldvectorlogo.com/ru/logo/kotlin)
- [ ] [scala](https://github.com/kaeawc/scala-logo/blob/master/img/logo_shadow.svg)
[...]

PR #700, opened 2017-12-05
add Rust logo

rosslh:
<s>Circle-S icon retrieved from Soylent's [Twitter](https://twitter.com/soylent) and [Facebook](https://facebook.com/Soylent) pages.</s>


davidklebanoff:
Hi @rosslh, thanks for the contribution!

Unfortunately the circle looks quite rough, how did you go about extracting the icon? Did you use a PNG to SVG converter?

At quick glance I did not see any official SVGs of their icon logo. Just the main logo on their website: https://www.soylent.com/ - I suppose someone could re-create the icon logo fairly easily by extracting the s from the official logo and re-drawing the circle around it to match... but that's no small task.

rosslh:
@davidklebanoff ok I'll look into fixing that, in the meantime I'll use this pull request to add the Rust logo, related to #469

davidklebanoff:
Ok. Happy to have you helping out! In the future I recommend making a branch per icon on your forked repos so you can easily have multiple pull requests at once without conflicts. :) 

As for the rust icon, it looks like you're missing the `role="img"` attribute we include between the `aria-labelledby="simpleicons-rust-icon"`  and `viewBox="0 0 24 24"` attributes.

Additionally, the quality looks lower to me than we would like. For instance, the circles on the gear aren't very smooth, an artifact of compression I would assume. Can you walk me through how you prepared/compressed the file? If you followed our [contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and used SVGO with the `-p 3` flag, we may have to reconsider our recommended precision level.

rosslh:
Here's the process I went through:
1. download logo from https://upload.wikimedia.org/wikipedia/commons/d/d5/Rust_programming_language_black_logo.svg
2. resize to 24x24 in vectr.com
3. run through svgomg default settings, compression level 3 (you're right, this is where the quality decreases)
4. refactor manually in text editor to specifications laid out in contributing.md (I forgot about role="img" in this case)

Here is the logo at each stage of this process: [rust.zip](https://github.com/simple-icons/simple-icons/files/1529584/rust.zip)

davidklebanoff:
Thanks for the detailed write up! It appears that step 2 is actually the culprit. The issues I'm seeing, the "low polygon" curves on the circles are visible in rust2.svg that you included in the .zip. Additionally, uploading rust1.svg to [SVGOMG](file:///Users/dave/Downloads/rust/rust1.svg) and playing with the precision shows that lowering the precision to something like 1 introduces artifacts, but not like the ones we're seeing here.

So to sum up, I would avoid using vectr.com as it's doing some weird compression that's trashing the icon quality. I personally use [icomoon](https://icomoon.io/app/) to do quick online resizing.

rosslh:
Thanks for the help @davidklebanoff! I've resized using icomoon and the icon looks a lot better.

davidklebanoff:
Awesome, looks much better. Thanks for the contribution.

In the future, please try to use official websites to source the SVGs. This time you used wikimedia, which works out in this case because they pulled it from the official website (you can see in the attribution), but just know official sources are much preferred when available.

Issue #693, opened 2017-12-03
Not sure about the color of Xing


louistsaitszho:
1) The [official logo rules](https://dev.xing.com/logo_rules) does not specify what color it should be
2) there are multiple versions of grey being used through out their website
- [This icon](https://corporate.xing.com/build/images/navigation/logo.svg) extracted from their [corporate website](https://corporate.xing.com/) uses #006567
- If you sign in to [your Xing account](https://www.xing.com/app/startpage), it uses #007575

Neither of them matches the current color we have, which is #005A5F

davidklebanoff:
Yeah, sometimes even the brands themselves aren't internally consistent. Makes things a bit of a challenge. I would use the "icon extracted from their corporate website" color, and just make sure it's referenced in the source.

Thanks for taking the time to look into it all so thoroughly!

ericcornelissen:
Unfortunately the commit that set the color, be1d16b, doesn't help either 😅 I agree that we should use the color from the icon you references, perhaps you can include this in #692 👍

louistsaitszho:
@ericcornelissen got it. thx

PR #702, opened 2017-12-05
Fix sources of 3 icons and color for XING

louistsaitszho:
1. fix sources for ubuntu, xing, and zendesk (see #616)
2. fix color of XING (see #693)

Part of #616 of course


ericcornelissen:
Looking good, thanks for the changes! 😃

PR #710, opened 2017-12-11
Added PHP icon

William-Owen:


birjolaxew:
This is one of the cases where the `fill-rule` is actually needed for the icon to render properly ;) You should either cut out the "PHP" from the circle using a vector program, or leave in the `fill-rule="evenodd"` on the path. As it is now, it's just a black circle.

William-Owen:
Hi @birjolaxew Thanks for the feedback. Shouldn't have rushed this one while doing other things.

birjolaxew:
:+1: Thanks for the contribution! Relevant to #625 and #469

Issue #701, opened 2017-12-05
Request: xcode


davidklebanoff:
The xcode icon can be found in official SVG format here:
https://developer.apple.com/develop/images/icon-xcode.svg

Found on website here:
https://developer.apple.com/develop/

birjolaxew:
Merged in #714 - thanks @simonker!

PR #713, opened 2017-12-12
add xcode addresses #701

simonker:
svg taken from: https://developer.apple.com/develop/images/icon-xcode.svg

hex color is the default blue highlight color in xcode:
screen shot 2017-12-12 at 09 49 06


simonker:
Looks like an issue with jsonlint, will review

PR #716, opened 2017-12-13
add Known icon

simonker:
I re-drew this in illustrator, feel free to reject if that's not allowed:

screen shot 2017-12-13 at 13 41 16


ericcornelissen:
It is not preferred, mostly because we prefer an SVG from an official source. However there does not seem to be an official SVG provided. I think we can include this one as it seems you took the time to properly redraw the icon. However I'm not 100% if the brand qualifies as "popular" just yet 🤔 Thoughts @birjolaxew @davidklebanoff ?

birjolaxew:
Is the "K" from a font, or is it entirely handdrawn?

simonker:
It's all hand drawn.. I took a screen just in case:
screen shot 2017-12-13 at 12 58 05

dshanske:
I spoke to the co founder and he will speak to the current person running it about whether they have an svg.

birjolaxew:
:+1: Merging this for now. If @dshanske comes back with an SVG, we can always open a new PR. Nice work @simonker, much appreciated!

PR #723, opened 2017-12-16
Edited the example to the correct syntax: width and height

AhadCove:
There was a typo in the readme that had 2 width tags instead of a width and a height


davidklebanoff:
Good catch, thanks!

PR #724, opened 2017-12-18
Added nodemon

William-Owen:


ericcornelissen:
I think you can remove the `fill-rule="evenodd"` from the SVG 🤔 Also (if you change that) can you make it so the `.svg` is a single line instead of two?

ericcornelissen:
Just went ahead and quickly added "icon" to the `<title>` of the svg, missed it in my first comment. Still, good work, thanks for the contribution 👍

PR #725, opened 2017-12-21
Fix React, 500px, Coursera, Codecademy and Co-op

### LINKED ISSUE #588: Fix: React
React currently has a small kink in the lower-left of the icon. This should be fixed.

ericcornelissen:
Closes #588, and part of #616

Reviewed and update all icons discussed in #588 (see commit descriptions for more details).


davidklebanoff:
Hi @ericcornelissen, awesome job on the icons!

Can you elaborate on where the colors come from?

For example, one discrepancy I found... The logo at the top of https://www.coursera.org/ is `#4a89dc`, as extracted from the SVG. Which is different from the `#2a73cc` listed here.

ericcornelissen:
Oops 😅 Seems like I messed that one up, I took the colors of React, Codecademy and Co-op from the SVG (just double checked them). I will update the Coursera one 👍

davidklebanoff:
Always frustrating when the brands themselves can't keep things consistent... Thanks for the update!

PR #727, opened 2017-12-22
Add "icon outdated" label description to contributing guidelines

ericcornelissen:
Update the contributing guidelines for the "icon outdated" label I created (in hindsight this PR should have been first 😅).

I've got one related question if we decide to keep the "icon outdated" label: should the "icon" label be changed to "new icon" as well?


birjolaxew:
:+1: I think it would make sense to change "icon" to "new icon" if we're going to separate it into new/outdated.

Issue #735, opened 2018-01-02
Request: Aventrix


davidklebanoff:
Add icon for [Aventrix](https://www.aventrix.com)

Icon SVG and colors can be found at:
https://www.aventrix.com/press

PR #736, opened 2018-01-02
Added Aventrix icon.

davidklebanoff:
Addresses issue #735.

Icon and color from https://www.aventrix.com/press


birjolaxew:
:+1:

I will add that I'll request everyone to always disclose if you have conflict of interests (e.g. in this case owning the site being added) when requesting/adding new icons.

PR #739, opened 2018-01-03
Extract CSS and JavaScript from HTML

ericcornelissen:
Put all of the CSS and most of the JavaScript into their own file. The JavaScript that is left in HTML depends on some Liquid logic present in the HTML ([see here](https://github.com/ericcornelissen/simple-icons/blob/af10b32c243311193d571796db130a5547f3d9c4/index.html#L170-L175)). I tried to give the script file a name such that it cannot be confused with the NPM script file.

Mainly did all this because in my opinion the HTML file was becoming a bit full of stuff from the newer styling and scripts (it went from 623 lines to 233 lines!). Any feedback is welcome  🙂


davidklebanoff:
Lovely! I hadn't poked around in the main site's code before. I think this is a great step in keeping things clean.

I see a lot of redundancy in the CSS, especially all the duplicate media queries, but that's best left to another PR.

ericcornelissen:
> I see a lot of redundancy in the CSS, especially all the duplicate media queries, but that's best left to another PR.

Agreed 👍

birjolaxew:
:+1: Nice work! We should probably go through and rethink the entire structure (I'm not particularly happy with how our data is stored, and I'd like to move the GH Pages code to a folder by itself), but this is definitely a great start!

PR #748, opened 2018-01-30
Fix CDN usage snippet

purplecones:
Fix duplicate attribute on img element from width to height.


davidklebanoff:
Good catch! Thanks!

PR #749, opened 2018-02-01
Add Firebase icon

ericcornelissen:
Added the Firebase icon based on the provided [source](https://firebase.google.com/brand-guidelines/) URL, as requested by @tomap in #625

~~I choose for the color #ffffff because of _Do not use the flat knockout in any color but white_ listed in the [Common Errors](https://firebase.google.com/brand-guidelines/#common-errors) section.~~


davidklebanoff:
Looks good! I guess this may raise the question of whether we consider our brand colors the color of the _logo_ or the _brand_. For instance, if we did use the yellow color, the logo would still be white on a yellow background on SimpleIcons.org, still following their brand guidelines. With that said I believe users would not be aware of the guidance, and there's a 50/50 chance the would color the logo white on a yellow background vs coloring the logo yellow on some other color background.

I personally would lean toward including the yellow color over white, but I understand it's highly subjective and believe there is no right answer.

If no one comments or has any additional thoughts, I'll merge this in a couple of days.

tomap:
Hi, thanks for the icon, it looks nice.
I believe it should stay back & white, and it's up to users to use yellow or not.
I think there are the same guidance for nearly every brand icon, and if you use the icon while in a commercial relation with the brand, then you should obviously follow the relation, otherwise, you care if you want

ericcornelissen:
> With that said I believe users would not be aware of the guidance, and there's a 50/50 chance the would color the logo white on a yellow background vs coloring the logo yellow on some other color background.

I would actually suspect offering the icon in this way does not significantly increase those odds, and some users will still end up using (the wrong) yellow


> I believe it should stay back & white, and it's up to users to use yellow or not.

As @davidklebanoff said _the logo would still be white on a yellow background on SimpleIcons.org, still following their brand guidelines_, in addition the icon will not have any color/be white when it is downloaded because the `fill` value is not set.


I'm in favor of changing the color for Firebase to `#FFCA28`, but perhaps we should think about including caveats on the website that link to [Common errors](https://firebase.google.com/brand-guidelines/#common-errors) sections of the brand in question, if available (in a separate PR of course).

birjolaxew:
:+1: Merging this. Discussion on linking to a company's press kit/guidelines should probably happen in a separate issue. Nice work @ericcornelissen !

Issue #751, opened 2018-02-04
Webpack throwing errors when bundling Simple Icons in project


IainIsCreative:
Hi there! I have an issue with using simple-icons and webpack. I believe this is more of an issue on Webpack but I'm trying to get help there as well. Hopefully someone here may have come across the issue and may have a solution. This is to do with simple-icons via Node.

Here's the issue: I'm using `simple-icons` in my React code, and using Webpack to bundle it. Using Babel is fine, no problems there, but Webpack is throwing a hissy over the package. At first, it throws an error saying:
```shell
Module not found: Error: Can't resolve 'fs' in '/projects/react-simple-icons/node_modules/simple-icons'
```
So I fix that in Webpack using `node: { fs: 'empty' }` which somewhat solves the issue. But when it bundles, and I view in the browser, it throws the following error:
```javascript
Cannot find module "."
```
It turns out, that this is what causes that error:
```javascript
const dataFile = './_data/simple-icons.json';
const data = require(dataFile); // This line is the culprit!
const fs = require('fs');
```
I'm at a loss as to what to do as I can't seem to get it working no matter how many solutions I try. Has anyone here had any issues like this with `simple-icons`? Any suggested solutions? I will greatly appreciate the help!

<details>
  <summary>My Webpack Code</summary>
<pre>
import webpack from 'webpack';

export default {
  module: {
    loaders: [
      { 
        test: /\.json$/, 
        loader: 'json-loader',
      },
      {
        test: /\.svg$/,
        loader: 'svg-inline-loader'
      },
      {
        test: /\.jsx?$/, 
        exclude: /node_modules/, 
        loader: 'babel-loader'
      },
    ]
  },
  resolve: {
    extensions: ['.js', '.jsx'],
  },
  node: {
    fs: 'empty',
    path: true,
  }
}
</pre>
</details>

ericcornelissen:
> I'm a bit stumped as to why the JSON file string is in a variable, I see no need for it.

Same here 😅 But as you pointed out the main problem is that you (through the `simple-icons` package) are trying to read files from the server using `fs.readFileSync` and that won't work (even if `fs.readFileSync`  is known as a function).

If you look at [brandifyjs](https://github.com/greybax/brandifyjs) (which uses simple-icons as its main dependency), they compiled everything including the SVG source into a single `.js` file. It works (and should be possible using webpack/babel as well), but is not a very good solution IMO.

[...]

PR #752, opened 2018-02-05
[WIP] better npm package support for browser 

ericcornelissen:
I decided to quickly implement a solution I came up with for #751. I simply check if the script is being run in a browser and depending on that change what value is set for the `svg` key for each icon. The SVG source in a Node environment and a unpkg.com url in a browser.

:warning: Before this is merged, I believe it needs some more testing. I only did two quick tests 1) in the latest Mozilla Firefox browser with a script transpiled using webpack/babel, and 2) in a simple node example. :warning:

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,13 @@ Icons can be served from a CDN such as [JSDelivr](https://www.jsdelivr.com/packa
 The icons are also available through our npm package. To install, simply run:
 
 ```
-$ npm install simple-icons
+$ npm install --save simple-icons
 ```
 
-The API can then be used as follows:
+In a NodeJS environment the API can be used as follows:
 
 ```javascript
 const simpleIcons = require('simple-icons');
-
 console.log(simpleIcons['Google+']);
 
 /*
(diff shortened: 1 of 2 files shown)


IainIsCreative:
I'll give this a shot and see how it goes!

birjolaxew:
I'd much rather keep a consistent API across NodeJS and browser usage. I see a few different options:

### This solution (SVG string in NodeJS, CDN URL in browser):

| Pros | Cons |
| :-: | - |
Easy, requires no major restructuring | Difference in usage from NodeJS to browser (!)
~ | SVG's aren't preloaded, causing delay when used
~ | Endorses a specific CDN

### Generating a complete `index.js` in build step

Our Travis build step would generate an `index.js(on)` file, which includes all the data inside it. It would essentially be the same code as now, except in our build step and outputting to a file, not to `module.exports`. Usage would be the same as now, but consistent across NodeJS and browser

| Pros | Cons |
| - | - |
Consistent usage (!) | Requires addition to build step
SVG's are preloaded, as browsers get the SVG string, not a CDN link | *Massive* file size if not tree shaken

### Generating a folder of export files in build step

Our Travis build step would generate a folder of `[icon-name].js(on)` files, which can be included from Webpack/NodeJS. Usage would change from `const icons = require("simple-icons"); console.log(icons["Google+"]);` to `const icon = require("simple-icons/google-plus"); console.log(icon);`.  
We could additionally generate an `index.js` that simply requires all the individual files and re-exports them under their proper name, so that we avoid breaking changes when implementing this.

| Pros | Cons |
| - | - |
Consistent usage (!) | Requires big addition to build step
SVG's are preloaded, as browsers get the SVG string, not a CDN link | ~
Only bundles the used files | ~

ericcornelissen:
> I'd much rather keep a consistent API across NodeJS and browser usage.

First off, I totally agree with that! But as I said this was just a quick (and dirty) solution I came up with. The main reason I went with this solution over your first suggestion (_Generating a complete index.js in build step_) is because of the massive file size. I like your last suggestion because it avoids the file size problem, but I didn't know you can generate new files for an npm package in Travis 🤔

Issue #111, opened 2013-05-10
Microformats


eugenb:
Request for microformats icon.

http://microformats.org/

PR #753, opened 2018-02-07
Add Unsplash icon

chrisssycollins:
Added icon for [unsplash.com](https://unsplash.com/) 📷


birjolaxew:
Where is the color `#111111` from? The official website seems to use `#000000` (or rather, no defined fill), with [some subsites](https://madewith.unsplash.com/) using [other colors](https://source.unsplash.com/).

chrisssycollins:
@birjolaxew Apologies, I took `#111111` from their body copy. I've changed it to `#000000` now.

birjolaxew:
@chrisssycollins The color is intended to be the brand color - in this case I think that would be `#000000` (so you had it right the second time :) ). The icon will be rendered in white on the site, but that's intended (you can also see e.g. AngelList for another example of this)

chrisssycollins:
@birjolaxew ahhh, should be good now! 🎉

birjolaxew:
@chrisssycollins I think it should be `#000000` instead of `#111111` - `#111` is simply the color of the text they use.

Unfortunately Unsplash don't have a press kit, so it's impossible to know what *they* think. I'd rather just play it safe and use the color of their logo (`#000`)

birjolaxew:
:+1: Looking good. Nice work @chrisssycollins !

PR #762, opened 2018-02-14
Add Cloudflare icon

EvasivePangolin:
Cropped down then recolored. Original source available [here](https://Cloudflare.com/logo).


birjolaxew:
Nice work! It looks like you have a couple of extra attributes in your SVG - namely `width` and `height` on your `<svg>`, and `class` and `fill` on your `<path>`. Remove those and it should be ready for merging!

ericcornelissen:
Also, can you specify why you picked `#F69259`? I can only seem to find that somewhere in the rules-list for the button at the top of the website but for me it uses `#F48120` by default. While in their logo they use `#F38020` and `#FAAE40` (from the SVG).

EvasivePangolin:
Fixed issues

birjolaxew:
:+1: Thanks for the contribution! It is very much appreciated.

EvasivePangolin:
No problem :)

Issue #733, opened 2018-01-02
Update: Hulu Icon


davidklebanoff:
We currently use the full logo as the icon, but their favicon appears to only be the "h", which leads me to believe we should update our icon to not use the full logo.

The color was just updated, so it should be good.

davidklebanoff:
Done.

PR #767, opened 2018-02-15
#733 update hulu icon

gmachure:
Hi ! 

I changed the hulu icon as requested. I downloaded the hulu favicon (png) and converted it into svg. Then, i followed the following instructions [https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md](url) to be sure i'm doing nothing wrong.

Note: It's my first contribution, so if there is something i did wrong, please tell me :)

Thank you !


davidklebanoff:
Hi @gmachure,

This looks great! My only hesitation is the `h` has a little bump in the top of it... it's not the smoothest. I went to the official website and on the homepage they had a logo that was an SVG. Extracting the h from that and using it as the source should result in cleaner icon. Would you be able to use that? Other than that, everything looked great - an impressive first contribution!

https://www.hulu.com/static/hitch/static/logos/hulu-logo-dark.svg

gmachure:
Thanks for your quick answer !

Oh, indeed there is a little bump, i didn't notice ! I fixed this.
And I used a SVGO precision of 3, i don't see another bump, tell me if it's ok right now ! :)

davidklebanoff:
Could you please:

1. Scale the icon to fit the viewbox, while preserving the icon's original proportions. This means the icon should be touching at least two sides of the viewbox.
1. Center the icon horizontally and vertically.

Other than that it should be good. Bumps are gone!

gmachure:
Ok it's done !

davidklebanoff:
Perfect - Congrats on your first contribution! 🎉

gmachure:
Thank you ! :+1:

PR #772, opened 2018-02-16
Added Blender icon

Josh-Powlison:
The Blender icon and info is in this branch! Let me know if any changes need to be made!

I chose the color by grabbing it from the logo EPS file. I chose the orange because even though blue is in the icon, where Blender is showed and displayed the orange is generally shown more prominently for more important things (for example, their [fundraising rocket](https://www.blender.org/) is orange). Also, the orange is surrounding the blue and pops more.

PR #763, opened 2018-02-15
Added Wattpad, Tapas, Aurelia, Gumroad icons

Josh-Powlison:
Hey guys,

The icons have been added and the json updated! I may keep adding stuff to this until it's merged in. :)

Thanks for all your help figuring this out @birjolaxew, you've been fantastic! :D

PR #773, opened 2018-02-17
Update CONTRIBUTING.md

ericcornelissen:
Add instructions to branch off of the `develop` branch and work on a separate topic branch to avoid situations such as #763 -> #772 (look at the commits). Feel free to suggest changes for the wording and location of the new text 😄

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,8 +3,9 @@
 Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below. If you decide to contribute anything, please do the following:
 
 1. Fork this repository
-1. Hack on a separate topic branch from the latest `develop`
-1. Commit and push to the topic branch
+1. Create a new branch from the latest `develop` (read more [here](https://guides.github.com/introduction/flow/))
+1. Start hacking on the new branch
+1. Commit and push to the new branch
 1. Make a pull request
 
 ## Table of contents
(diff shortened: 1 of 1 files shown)


davidklebanoff:
Looks good to me. My only change would be to change "topic" to "new". But I think it's good either way.

CONTRIBUTING.md
```diff
@@ -1,6 +1,11 @@
 # Contributing To Simple Icons
 
-Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below.
+Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below. If you decide to contribute anything, please do the following:
+
+1. Fork this repository
+1. Hack on a separate topic branch from the latest `develop`
```
birjolaxew: Should probably be made clearer for people who are new to git. Possibly "Create a new topic branch from the latest `develop`", as GitHub's interface uses "Create branch: [name] (from [base])" when creating new branches through their GUI.

ericcornelissen: In that case I would say we choose for @davidklebanoff suggestion (replacing "topic" by "new"). The GitHub GUI is not the only GUI, and other GUIs may choose a different wording. Agreed?

birjolaxew: I'm more worried by people understanding the "Hack on a ..." part. If I have no idea what a branch is, "Hack on a separate branch" won't help me understand it at all. "Create a separate branch" indicates that it's an action in and of itself (i.e. you should create your new branch, then start coding). I am indifferent to the topic/new thing - whichever one you guys feel is better is fine by me.

birjolaxew: :+1: Seems good to me

birjolaxew:
:+1: Nice work @ericcornelissen !

PR #777, opened 2018-02-20
Replace jsonlint by jsonlint2

ericcornelissen:
[jsonlint2](https://www.npmjs.com/package/jsonlint2) is a fork of [jsonlint](https://www.npmjs.com/package/jsonlint) that is still being updated. Relevant discussion: https://github.com/zaach/jsonlint/pull/105

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "author": "Simple Icons Collaborators",
   "license": "CC0",
   "devDependencies": {
-    "jsonlint": "^1.6.2"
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Let's hope this one sticks around

ericcornelissen:
Just want to point out that the original jsonlint has been updated (see [here](https://github.com/zaach/jsonlint/commit/ac74f816472d7ad178c4171b2ed727c6bcc84b85)), so we might consider switching back 🤔

PR #779, opened 2018-02-20
Add Disroot logo to json data

EvasivePangolin:
Disroot doesn't seem to have an actual color or branding guide, so the color code #50162d is from the original SVG itself.


EvasivePangolin:
Adding the SVG, one moment

EvasivePangolin:
Did a mistake, please ignore

PR #780, opened 2018-02-21
Add icon for Disroot.org

EvasivePangolin:
The color #50162D is from the source vector. Disroot doesn't seem to have defined branding colors as far as I know. 

For those that aren't familiar with Disroot: It's a privacy-oriented provider of email and other productivity services for activists, journalists, and everyone else.


birjolaxew:
Nice contribution! You seem to have left in `width="24" height="24"` on your `<svg>` - those'll have to go before we can merge :)

EvasivePangolin:
Fixed @birjolaxew

birjolaxew:
:+1: Nice work! Thanks for the contribution!

PR #784, opened 2018-03-04
Fix Cloudflare icon

EvasivePangolin:
When I put the file through SVGOMG, the output was distorted. Changing the precision to 4 (instead of 3) fixed this.
Not sure how I didn't notice it before committing it here. My bad.
  
[Here's a comparison](https://imgur.com/a/YTgqD) for future reference. [Source file](https://pastebin.com/wK1mBHQR).


ericcornelissen:
Good catch, too bad we also didn't notice it... Thanks anyway!

Issue #787, opened 2018-03-08
REQUEST: UI apps


rdricco:
- Figma
- SketchApp

davidklebanoff:
Sketch is being discussed in #788.

abenjamin765:
I have an open PR that includes a Sketch icon used frequently for community branding alongside icons for Framer, Principle, Figma, and Flinto. Looking to add more shortly.

louistsaitszho:
Any suggestion on what color to choose for [Figma](https://blog.figma.com/figmas-new-icon-49324fa00298)? Their icon is made up of a bunch of colors.

- `#1abcfe`
- `#0acf83`
- `#ff7262`
- `#f24e1e`
- `#a259ff`

And in the blog post above, the background looks like `#2B2D35`

birjolaxew:
Closing this as Figma has been added (#1014), and Sketch is currently incompatible with our format (#788)

Issue #755, opened 2018-02-07
Linking to a company's press kit/guidelines/common errors


ericcornelissen:
tl;dr: Should we **a)** include a link to either the brand guidelines of an icon or **b)** include a link to a _Common Errors_ section of an icon (if available)

* * *

From #749 we found that it can be somewhat unclear what color should be linked to an icon. From the [ Firebase brand guidelines](https://firebase.google.com/brand-guidelines/) it is clear that the primary color for Firebase is `#FFCA28`, but the [Common Errors](https://firebase.google.com/brand-guidelines/#common-errors) section it is specified that you're not supposed to use the logo in any other color then white (i.e. `#FFFFFF`).

> Do not use the flat knockout in any color but white

It was then decided to use `#FFCA28` on the website since 1) when the SVG is downloaded has no color associated with it and 2) that "is" the brand color of Firebase.

However, it may be an idea to include a link to either the brand guidelines of an icon (for all icons) **or** include a link to a _Common Errors_ section of an icon (if one exists/is known). So that user can check the documentation for themselves when using an icon.

ericcornelissen:
Just throwing out there that we can take some inspiration from [BrandColors](https://brandcolors.net/) which provide a link to the brand URL and the source URL as well as the date it was updated. E.g. for [500px](https://brandcolors.net/b/500px):

![image](https://user-images.githubusercontent.com/3742559/38525641-5d3bcc30-3c53-11e8-9215-c05a8b87d5ac.png)

ericcornelissen:
> It's worth noting that many brands have exceptions to these guidelines. For example:
Brands that stipulate that a solid white icon on a colored background may still use this treatment in the Windows App store, the MacBook touch bar, an app icon, a social media avatar, or other places.

If that's the case, it should be possible to find that in the guidelines, so I see no problem there.


> I'm curious what the importance or value is in following brand guidelines beyond the shape of their mark. Could those guidelines be cited and left to the discretion of the user to style them correctly? I'm not sure many users are utilizing Simple Icons for official branded material.

I won't be surprised if not everyone using SimpleIcons is going to look at the brand guidelines, but some folk might be concerned about this stuff and for them a simple link can help a ton. And since we already have a lot of these links available to us (as the `source` of an icon) I see no reason to just incorporate them to the site. But, as you said, I do think that just citing them and let that be it is enough.

PR #788, opened 2018-03-08
Add flattened version of the Sketch icon

ericcornelissen:
Relevant issues: #625, #755, #787

Color is taken from the middle part of the bottom of the icon, as their [press page](https://www.sketchapp.com/press/) does not specify any colors to be used.

Note that their [press page](https://www.sketchapp.com/press/#press-media) does specify the following...

> We would appreciate it if you don’t modify or embellish Sketch brand assets in any way, such as by changing the design or color of our symbol, thanks.


birjolaxew:
I think this one is tricky. I feel that the big black blob in this PR loses almost all of the brand identity. A better solution would be [to keep the inner lines](https://www.shareicon.net/brand-logos-brands-app-logo-sketch-878818).

<p align="center">

</p>

But as you mention, they explicitly ask us not to modify the logo, and even give the above (with green color) as an example of what not to do.

Still, I think that if we *have* to change the logo (which we do, if we want to include it), we should aim to maintain as much of the logo's identity as possible - and I think the right version above (black lines) is better suited for that.

ericcornelissen:
> I think this one is tricky. I feel that the big black blob in this PR loses almost all of the brand identity. A better solution would be to keep the inner lines.

Agreed


> There is also the option of not including them, since their brand guidelines are at odds with the logo format we use. 

We either do that, or we should get started on #755 to warn people about potential issues using the logo like this (since the same goes for the Firebase logo, see #749).

Potentially we can choose to not include the logo itself, but do include the "brand" and instead link to the press page/provide the not flattened icon?


> Perhaps we should consider emailing them and asking?

We could do that, but I think their press page is explicit enough

birjolaxew:
Closing this for now; we should revisit it when we implement a way to handle icons with guidelines / icons that aren't easily monochromized.

PR #789, opened 2018-03-08
Update AddThis icon

ericcornelissen:
Updated the SVG which was slightly distorted (see image ⬇️ left is old, right is new). Also updated the color, taken from the [coloured SVG found in the header](http://www.addthis.com/www/18021452bd192/style/images/wrapper/logo.svg) and updated the outdated link (the old link gave a 404)

beforeafter_addthis


davidklebanoff:
Good stuff!

Issue #478, opened 2017-07-08
A humble contribution


yammadev:
Hi there! 

Thanks for this icon collection. I've written this implementation using this svg icons and auto generating fonts + sass + css. Please refer to [https://github.com/yammadev/brand-icons](https://github.com/yammadev/brand-icons) for those who need an extensible and simple solution for usage on web.

Regards 👍

davidklebanoff:
Cool tool!

I'm closing this issue as it's not really an issue and doesn't include an action item. :)

Issue #454, opened 2017-05-11
I've created brandifyjs tool based on icons from this repo


greybax:
Hi Dan,

Thank you for this repo. It rocks! :rocket: :sunny:

I'd like just to suppose 2 things:

* :thumbsup: for npm package
* I've created tool based on icons from this awesome repo. You can check it out on [brandifyjs playground](https://greybax.github.io/brandifyjs). I've also noticed issue https://github.com/danleech/simple-icons/issues/443 by @peterpme and this is great tool too. We can use it for automation process of creating new icons. I'd be really appreciate if you can do [brandifyjs](https://github.com/greybax/brandifyjs) part of this repo f.e. as sample of app where icons from this repo can be used.

:v:

ericcornelissen:
@greybax I'm going to close this issue as I don't think its relevant to keep it open. Even though its an interesting project, I don't think we should include this project (or projects like it) in for example the README. Especially since the [Github dependency graph](https://github.com/simple-icons/simple-icons/network/dependents) already shows which projects are using Simple Icons.

One option could be to add repos like yours as organization repos, but I don't think your project necessarily adds any value to the organization 🙂

Issue #443, opened 2017-05-08
I Created a Simple Icon Generator


peterpme:
:wave:

Thanks for putting together this awesome list of icons!

[I've created a simple icon generator](https://github.com/peterpme/create-simple-icon)

I was hoping for two things:
- Being able to expose the json as an npm package
- Linking my generator to your repo

Right now I scrape the json and it works fine, but I'm not sure if this'll ever fail. If you have any questions for how the generator works or find any glaring bugs, I'd love to be able to work this out together!

danleech:
@peterpme This is a popular request. I’ve never published on NPM so will need to do some reading.

peterpme:
Hey @danleech! Thank you for considering this option! I'd to open a PR and help do this :)

danleech:
Please do!

RusinovAnton:
@danleech Hello, great job with library! Is there some problems with publishing it to npm? Can I help on it?  
I'm interested in npm package too because I want to make a react library based on it.  
There is already such lib [react-simple-icons](https://github.com/jxnblk/react-simple-icons), but I don't like the approach where all icons being hardcoded into JSON file and paths are read from it tho. I want make icon components more dynamically.
Looking forward for progress on this issue. Thanks

ericcornelissen:
Going to close this issue since its main purpose, create a npm package, is fulfilled (see [here](https://www.npmjs.com/package/simple-icons))

Issue #388, opened 2017-04-14
Icons fonts


Soullivaneuh:
Are you planning to add font icons support as FontAwesome for example?

danleech:
No, but if you want an icon font with Simple Icons glyphs you can generate one with [IcoMoon](https://icomoon.io/app/#/select). There are good reasons why I won’t be offering an icon font:

**Performance issues with icon fonts**

- Your user needs to download every single glyph before one can be rendered. For Simple Icons that means 300 brand icons (~100kB), most of which you’re probably not using.
- Your user is caching the entire glyph set in a single file, so changing one glyph invalidates the whole cache.
- HTTP/2 (now at [80% support](https://caniuse.com/#search=http%2F2)) favours multiple, small files (like SVGs).

**Accessibility issues with icon fonts**

- If an icon font can’t or won’t load, there’s no graceful fallback (unless you use custom ligatures).
- You have no control over users with user stylesheets, which may override your icon fonts (e.g. dyslexic people using weighted fonts).
- Screen readers will not read icon font glyphs (unless you add `aria-label` attributes to your icon elements).

ericcornelissen:
I'll just point out [this project](https://github.com/yammadev/brand-icons) (as seen in #478) for anyone interested in this

PR #796, opened 2018-03-21
Add WordPress plugin to README.md

### LINKED ISSUE #715: I made a Simple Icons WordPress plugin
https://wordpress.org/plugins/simple-icons/

Let me know if there is anything I need to change or if you want the plugin removed from the WP repository all together.

ericcornelissen:
Closes #715

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ console.log(simpleIcons['Google+']);
 */
 ```
 
+### WordPress Usage
+
+Icons are also available on WordPress through a simple plugin, which you can find [here](https://wordpress.org/plugins/simple-icons/).
+
 ## Status
 
 [![Build Status](https://img.shields.io/travis/simple-icons/simple-icons/develop.svg)](https://travis-ci.org/simple-icons/simple-icons)
--- a/README.md
+++ b/README.md
@@ -47,9 +47,11 @@ console.log(simpleIcons['Google+']);
 */
 ```
 
-### WordPress Usage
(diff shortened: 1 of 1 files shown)


birjolaxew:
I think I would prefer to have a separate section for third party extensions. That way we separate officially supported (Node, CDN and file) usage from those that we have no control over. Something along the lines of

<blockquote>
...

## Node usage

The icons are also available through our npm package. To install, simply run:

```
$ npm install simple-icons
```

The API can then be used as follows:

```
...
```

## Third party extensions
### WordPress
Icons are also available through a simple WordPress plugin created by @tjtaylo - it can be found [here](https://wordpress.org/plugins/simple-icons/).
</blockquote>

We should also have a look at #478 (possibly pointing out #388) while we're at it.

ericcornelissen:
I added the "Third party extensions" suggestion. However I'm hesitant to add the projects you pointed out because in my opinion they're a bit different from this WordPress plugin, except maybe for #478.

#454 seems more like a project *using* SimpleIcons, and #443 seems more like a fun side project rather then something people would use often (as the [npm page](https://www.npmjs.com/package/create-simple-icon) also suggests)

birjolaxew:
:+1: I agree with that assessment. I've also discovered that #478 doesn't automatically pull the newest release from here, and the last update was 9 months ago - so it can be considered abandoned.

I think that this is fine for now. Nice work @ericcornelissen!

PR #848, opened 2018-05-09
Update colors of Gitter, Ghost and Evernote

ericcornelissen:
Related #616


birjolaxew:
There's an interesting discussion relating to the Evernote color on whether the color we specify is the *brand* color (which would be green) or the *logo* color (which would be gray).

PR #850, opened 2018-05-14
Update: Roots Color

knowler:
Roots now uses `#525ddc`. See [their website](//roots.io).


ericcornelissen:
Thank you for the update @knowler, it is much appreciated!

PR #862, opened 2018-06-02
Add icon for Riseup.net

EvasivePangolin:
Original: https://riseup.net/about-us/images/button-crow.svg
Removed the background and text, then removed the color value from the crow.


birjolaxew:
Looks like you have an extra `#` in your `"hex"` value, and an extra space in your before the `>` in your `<svg ... >` - fix those, and it's looking good :smile:

EvasivePangolin:
Fixed @birjolaxew

birjolaxew:
Looks like the logo usage is fairly inconsistent from Riseup - they use [this star](https://riseup.net/about-us/images/riseup.net-inline.svg) for their favicon, they use [this star](https://riseup.net/about-us/images/riseup-rainbow.svg) on their homepage, and they use [this logo (without star)](https://riseup.net/about-us/images/riseup.net-inline.svg) in their header.

But I cannot find anywhere that they use a crow to identify themselves. Seems to me like we should use the same star as their favicon, but I don't know much about Riseup. Got any insigts @EvasivePangolin ?

EvasivePangolin:
The star can't really be made monochrome, so I went with the crow from their buttons and [Twitter profile.](https://twitter.com/riseupnet)

birjolaxew:
We want to maintain the identity of the brand when we add an icon to Simple Icons; in this case Riseup have chosen to identify themselves using a star (which star it is is something they apparently haven't quite decided on).

I think it would be better if the [star from their favicon](https://riseup.net/about-us/images/riseup.net-inline.svg) was turned monochrome by turning the red part transparent. This would maintain the spirit of their identity the best we can.

EvasivePangolin:
I'll go ahead and email them about it to get their input.

EvasivePangolin:
Still no response from Riseup, I'm going to try a different email.

taggart:
I agree with the idea of the star from the favicon but with the red transparent.

EvasivePangolin:
Got a response from Riseup, they suggested going with the crow since it's a newer design.

birjolaxew:
@EvasivePangolin If you could forward their response to `johanringmann@gmail.com`, I'll get the icon merged in 😃

EvasivePangolin:
Beyond using some of their services, I'm not affiliated with Riseup.

PR #864, opened 2018-06-03
Add icon for TAILS Linux

EvasivePangolin:
[TAILS (The Amnesic Incognito Live System)](https://tails.boum.org) is a Linux live system designed for high-risk situations where anonymity is vital.


birjolaxew:
:+1 Thanks for the contribution! Great work.

PR #888, opened 2018-07-20
Limit destructive filename mapping

### LINKED ISSUE #882: Find alternative to destructive title→filename mapping
This issue popped up in relation to #875, and is explained [here](https://github.com/simple-icons/simple-icons/pull/875#issuecomment-405777365).

The TL;DR is that our mapping removes some symbols (e.g. dots, hyphens and exclamation marks) when translating a title into a filename. This leads to some issues, mainly that .NET's filename becomes `net.svg` which isn't super clear, and potentially (although very unlikely) filename collisions in the future (if a brand was to be named e.g. `-NET`).

Possible solutions as I see it:

1. Leave it as-is. Although not ideal, it could be argued that `net.svg` is a reasonable representation of `.NET` especially since the filename doesn't have huge requirements for clarity - only for uniqueness. This would leave up the potential collision issue.

2. [Change mapping to not lose information](https://github.com/simple-icons/simple-icons/pull/875#issuecomment-405777365) - this would involve changing e.g. `"." → ""` to `"." → "dot"`. This would be a breaking change as some filenames would have to be changed (e.g. `aboutme.svg` becomes `aboutdotme.svg`) breaking CDN usage of those icons - the JS API would be unaffected. This would leave up, but greatly diminsh, the potential collision issue.

3. Implement explicit mapping somehow (e.g. as key in `_data`). This wouldn't be a breaking change as no external API would change, but would be a bit ugly code-wise (some icons having extra metadata while other's don't, and the metadata being a bit out of place in `_data`). It also seems to me like making a mountain out of a molehill.

I'm personally fondest of option 2 here. Any opinions @ericcornelissen / @davidklebanoff ?

---

*Blocks #875*

birjolaxew:
Closes #882.
Changes our title -> filename mapping to be less destructive, by not removing `"-"` and renaming `"."` to `"-dot-"` - this gives file names such as `last-dot-fm.svg` and `co-op.svg`.

There is still some information lost (namely "!" and "`").

**Publishing this is a breaking change**


davidklebanoff:
I like the idea, I do worry about future edge cases, where potentially the `.` is at the end of a filename for instance.

So, `foo.com` would become `foo-dot-com` - good; but `bar.` would become `bar-dot-` - bad.

Additionally, I think removing "!" and "`" makes sense in expected cases, think "Yahoo!" or "Denny's", so that seems ok to me. However, I would expect a brand named "Yahoo." to be treated the same as a brand named "Yahoo!". So all this to say, perhaps we should treat punctuation, or at least the dot, differently if it's at the end of a brand name vs in the middle.

birjolaxew:
@davidklebanoff I agree. Unfortunately Jekyll's [`replace` filter](https://shopify.github.io/liquid/filters/replace/) doesn't support regexs, so implementing handling for end-of-string-placement is a bit more cumbersome than it would appear.

I'd suggest that we should rewrite the website to use something other than Jekyll soon-ish. It would both give us a better structured codebase, and would allow us more freedom in the logic we use (so we don't end up with Jekyll's templating language blocking us like this). That is obviously a bit beyond the scope of this PR though.

birjolaxew:
@davidklebanoff Good idea. I hadn't even considered that.

One issue I foresee with it is that attempting to predict how names are pronounced doesn't work well - e.g. `Yahoo!` is pronounced `yahoo`, while `Google+` is pronounced `google-plus`. We might run into issues in the future with brands that would expect to have ending punctuation kept.

Overall it's a bit of a non-issue as long as the filenames are largely recognizable as the brand they represent. The main danger in that regard is *removing* information (e.g. if our mapping had been more aggressive, `Google+` could have conceivably ended up being mapped to `google.svg` which would be a no-go), so I'd prefer being a bit careful on removing information from the title.

ericcornelissen:
I agree with @birjolaxew that it is a bit of a non-issue, but I do like @davidklebanoff suggestion to ignore punctuation at the end of a name. This wouldn't affect `Google+` as `+` is not punctuation. The only problem I foresee with this approach is the inconsistency between different characters and different places, but this was true before as well...

birjolaxew:
@ericcornelissen The point was more that future brands might include punctuation at the end of their name that is expected to be pronounced (like a reverse `.NET`).

Regardless, I'll be merging this in a few hours if there are no further comments. Although I like the idea of more advanced mapping, it should probably be implemented when we eventually (hopefully) move to a non-Jekyll build process (#898). As it is now, the `index.html` site is already too big.

PR #894, opened 2018-07-21
Updating nodejs.svg

DoronBrayer:
Loyal to the official source:
https://nodejs.org/static/documents/foundation-visual-guidelines.pdf
https://nodejs.org/en/about/resources/


ericcornelissen:
I'm personally of the opinion that the icon we're currently using better suites our needs, as the full logo has too much clutter (especially at the size these icons might be used). I am aware that this is true for some other icons as well, e.g. `AmericanExpress`, `AMD`, and `MySQL`, but these have clear guidelines on how their logo should be used, where NodeJS does not (at least in my opion).

DoronBrayer:
@ericcornelissen I understand the concept of making simple logos, hence the name 'Simple icons'.
In that case, the following variations are more suitable:
1. _nodejs-v2
2. _nodejs-v3

In my opinion, it's better than the [existing hexagon/gemstone](https://github.com/simple-icons/simple-icons/blob/develop/icons/nodejs.svg) drawing.

What do you think?

ericcornelissen:
@DoronBrayer I see where you're coming from with your suggestions (and I think both are pretty good), but unfortunately the NodeJS guidelines are clear enough to determine that neither of them would be valid. It's either the current icon, or the icon you're adding in this PR.

I intentionally didn't close the PR so other people can give their say as well 🙂

birjolaxew:
We have the choice between either the "primary logo" (i.e. this PR) or the "primary icon" (i.e. the current SVG). Given that simple-icons is a collection of icons, it would seem straight forward to pick the "primary icon". However, since we only use monochrome icons the "primary icon" loses pretty much the entirety of the brand identity, which makes it a poor choice.

So the question really is if the symbol version loses enough brand identity that we *must* use the logo instead. It should be noted that the symbol version is used by Node itself across favicon, avatars on e.g. Medium/YouTube, and pretty much everywhere else that a symbol is used.

I'm personally in favor of keeping the symbol version. The rounded hexagon still has *some* brand identity, and the logo contains a lot of very small details that aren't really suitable for icon usage. Neither option is good, and it is unfortunate that Node.js picked an icon that is so hard to monochromize, but such is the current state of the logo.

davidklebanoff:
@birjolaxew @ericcornelissen 
I agree with you two as well, so I'm going to go ahead and close this issue.

@DoronBrayer 
Thanks for the ideas and input, however it seems as though the current icon best fits our Simple Icons goals and guidelines. It's not always black and white on how a brand's icon(s) should be represented so we try to our best to be consistent in our decisions. I hope you understand. With that said, we're very happy with your recent contributions to the project. :)

PR #900, opened 2018-07-23
[New icon] Qualcomm

DoronBrayer:
Title: Qualcomm
Hex: 3253DC
Source: https://www.qualcomm.com


davidklebanoff:
It would appear their brand guidelines might be locked behind their brand portal website - https://www.qualcomm.com/logos . We can inquire, but I personally don't think it's necessary.

It appears their favicon is just their Q icon and their social media (e.g. [Instagram](https://www.instagram.com/qualcomm/) & [Facebook](https://www.facebook.com/Qualcomm/)) is also just the Q icon. So with that said, I think the icon should just be the Q icon. Should be easy to just derive it from the logo SVG itself.

DoronBrayer:
@davidklebanoff Done.
 Edited to look like the [favicon/avatar](https://pbs.twimg.com/profile_images/950419109069631488/3-4MNT0E_400x400.jpg), also optimized via SVGOMG.

PR #903, opened 2018-07-24
[New icon] Samsung

DoronBrayer:
Title: Samsung
Hex: 1428A0
Source: https://www.samsung.com/us

Brand guidelines: https://samsunghvac.com/extranet/bl_guide/Samsung%20Logo%20Usage%20Guidelines_112916.pdf


ericcornelissen:
Good work @DoronBrayer. Again, thanks for providing the link to the brand guidelines! Could you set it as the source in `_data/simple-icons.json`?

birjolaxew:
@ericcornelissen I think the current source in the `.json` is fine. The point of the source is that we can later go back and find any logo updates from the link; if Samsung have an official page where they publish brand guidelines (e.g. https://brand.linkedin.com/) then that would be optimal, but linking directly to a PDF doesn't let us find updates to their logo.

We should instead implement #755 for linking to guidelines.

ericcornelissen:
> if Samsung have an official page where they publish brand guidelines (e.g. https://brand.linkedin.com/) then that would be optimal, but linking directly to a PDF doesn't let us find updates to their logo.

Fair point 👍

DoronBrayer:
@ericcornelissen /@birjolaxew
1. I did not find such webpage, so the PDF is better than nothing.
2. in _data/simple-icons.json, what does the word 'source' means? - Does it mean where did I download the SVG file from? - Or does it mean the guidelines of the brand? - It's really unclear.
3. I used 5 in the 'Precision' of SVGOMG. I do not trust '3' or '4', because it's below the middle, and sometimes it provides sloppy results. Therefore, You should consider using '5' all the time.

davidklebanoff:
@LogoLover18 - If would like to make icon requests, please submit an issue. This is not the appropriate place.

DoronBrayer:
@ericcornelissen / @birjolaxew / @davidklebanoff
Do you have an answer redarding my previous comment/question?

birjolaxew:
@DoronBrayer

1. As Samsung don't seem to have a brand website, the best location for checking for logo updates should be used; since [their official website](https://www.samsung.com/) uses an SVG logo, that is probably preferable - exactly as you done in this PR.

2. `source` is intended to mean "where you got the logo from". It serves a purpose beyond this though: it also is a place where we can later go back and check if the logo has been updated. This means that if a brand has a place where they host brand guidelines as PDFs then you *shouldn't* put the PDF URL as the `source` - instead, go one step up and use the URL to where they host the PDFs. Since their official page has the logo as an SVG, your current `source` is perfect.

3. In general the precision should be chosen on a logo-by-logo basis. It's better to be safe than sorry (so a precision of 5 is better than a precision of 2 in this case), but I didn't see any artifacts when dropping the precision to 3. Can you verify this?

DoronBrayer:
@birjolaxew Thanks for the extended review.

I took the inline SVG from the header of Samsung's website, indeed.
Using '3' gave me sloppy results in various logos (at least 5 logos). Therefore, I only use 4-5, and you should embrace this solution as well.

I guess this PR is valid/qualified now, right?
If so, please approve it.

birjolaxew:
@DoronBrayer Yes, some icons definitely get messy at `3` precision. It very much so is a logo-by-logo thing; we try to find the lowest precision that keeps the logo free of any perceivable artifacts, mostly to keep the file size as small as possible.

In this case a precision of `4` works fine - you are right in that a precision of `3` is too low though:

<p align="center"></p>

I've gone ahead and lowered the precision to 4, which makes the PR great.

:+1: Thanks for the contribution! It is very much appreciated.

PR #906, opened 2018-07-26
Update index.js filename mapping to match site

### LINKED ISSUE #904: [Bug] npm package broken, using old filenames
&nbsp; | version
:--: | :--:
npm | `v6.2.0`
node | `v10.7.0`
os | `Windows 10`
simple-icons | `v1.7.0`

npm package is currently broken,
the package seems to still be looking for the old filenames:
```console
will update with error message when not on mobile
```
I believe this is due to changes in https://github.com/simple-icons/simple-icons/commit/d5585894dfe1e37db75889aa084c195d8d375f0c not being applied to index.js

https://github.com/simple-icons/simple-icons/blob/fc1efcc5e0ace5d2a11540fb53e82e18a6e24997/index.js#L10

birjolaxew:
Updates index.js filename mapping to match the Jekyll implementation.
This was missed in #888 (we should implement tests to avoid this in the future).
The JS implementation is a direct translation of the Jekyll one.

Closes #904

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,10 @@ const icons = {};
 data.icons.forEach(i => {
   const filename = i.title.toLowerCase()
     .replace(/\+/g, "plus")
-    .replace(/[ .\-!’]/g, '');
+    .replace(/^\./, "dot-")
+    .replace(/\.$/, "-dot")
+    .replace(/\./g, "-dot-")
+    .replace(/[ !’]/g, '');
   i.svg = fs.readFileSync(`${__dirname}/icons/${filename}.svg`, 'utf8');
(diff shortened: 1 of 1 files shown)


birjolaxew:
Tested by pulling and running

```js
const simpleIcons = require("./index.js");

console.log(simpleIcons['Google+']);
console.log(simpleIcons['.NET']);
```

As this is a rather critical bug I'll be merging this in 30 mins even if no other maintainer has reviewed by then.

ericcornelissen:
Looks like this works as intended, good work @birjolaxew 👍

PR #909, opened 2018-07-27
Page search find .NET Google+ etc

### LINKED ISSUE #902: Page search does not find .NET
The search on our website appears to be based on [file names](https://github.com/simple-icons/simple-icons/blob/924632bc6f737d35f584a7ceb04e269036f54bb1/index.html#L182) - it should be based on titles. This clashes with our recent changes to filenames, leading to e.g. ".NET" not finding .NET, while "dot-net" does ([as correctly pointed out by @ericcornelissen](https://github.com/simple-icons/simple-icons/pull/875#pullrequestreview-139680729))

RedSparr0w:
closes #902
Allow search to find `.NET`, `Google+` etc.

### DIFF
--- a/index.html
+++ b/index.html
@@ -177,7 +177,12 @@
     </script>
     <script>
       function normalizeSearchTerm(value) {
-          return value.toLowerCase().replace(/ /g, '');
+          return value.toLowerCase()
+              .replace(/\+/g, "plus")
+              .replace(/^\./, "dot-")
+              .replace(/\.$/, "-dot")
+              .replace(/\./g, "-dot-")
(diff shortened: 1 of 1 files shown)


birjolaxew:

adsingh14:
Why not we add dotnet as alias ?

adsingh14:
@LitoMore What about .env?

LitoMore:
See https://github.com/microsfot/dotnet. They only use `dotnet` for the repository name and URL path name.

But Dotenv is different. They call themself `Dotenv`.

PR #912, opened 2018-08-01
Remove Code School icon

ericcornelissen:
Code School was aquired by Pluralsight (for which an icon is already present). See [their site](https://www.pluralsight.com/codeschool)


birjolaxew:
:+1: Good catch.

Issue #905, opened 2018-07-26
Broaden tests ran by CI


birjolaxew:
We should test:

- [x] NPM package still works (to avoid e.g. #904)
- [x] Page builds (to avoid e.g. incorrect filenames) (#913)
- [x] SVGLint (#485)
- [x] JSONLint

Issue #898, opened 2018-07-23
Move to a CI-powered release process


birjolaxew:
Although it is nice to be able to simply let GitHub worry about building our website, it does limit our choice of build tools significantly. Jekyll - the only static page generator GH supports - puts rather strict restrictions on data structure (this is also why the `_data` folder is named as it is) and Liquid, the templating language, is rather underwhelming.

I would prefer moving to [a build process that is powered by our CI](https://docs.travis-ci.com/user/deployment/pages/) (currently Travis). This would allow us to build using whatever build tool we want (even a custom one), and allow us far greater freedom over how the project is structured.

We could additionally perform automatic git tagged releases, which would give us a formal changelog (#747).

This is all a bit abstract for now, as this issue is mostly created in order to allow discussion on the topic, but it is a future plan for the project (along with #485)

ericcornelissen:
In the light of  #942, I just wanted to add to this that we can and should also do this with the NPM release, more information [here](https://docs.travis-ci.com/user/deployment/npm/)

birjolaxew:
We are actually already doing this for NPM releases. All merges into master will trigger an NPM publish through Travis, and a GitHub Pages rebuild through GitHub (and will spam my email with updates on the NPM publishing :smile:)

ericcornelissen:
My bad 😓 For some reason I just completely missed [this snippet](https://github.com/simple-icons/simple-icons/blob/57e66ef99af40f21416a0df0bd534151cb8822eb/.travis.yml#L16-L21) of code...

runxel:
Closed with #77

ericcornelissen:
Not sure how #77 resolved this issue, but I'm fine closing this as it is more or less resolved with https://github.com/simple-icons/simple-icons-website

PR #913, opened 2018-08-01
Add page buid to CI script

ericcornelissen:
For now, simply let Jekyll build the page to see if no errors occur as requested in #905. This may need to be changed, or can be omitted, when #898 is implemented.


birjolaxew:
:+1: Very nice! This should cover a lot of issues that could pop up.

PR #915, opened 2018-08-01
Update Plurk icon

jmlntw:
[Plurk](https://www.plurk.com/) renewed their brand logo today, so I updated the SVG logo and color.


birjolaxew:
Good catch! I think we should probably use their logo without the circular background now, as that seems to be what they're using as their primary logo. Everything other than that looks good!

birjolaxew:

Issue #873, opened 2018-06-26
Icon Request: Mathematica & Wolfram Language 


SumNeuron:
Mathematica
name: "spikey"
where to find: https://www.wolfram.com/mathematica/
logo: http://www.wolfram.com/common/framework/img/spikey.en.png
colors:
RGB(213, 27, 41)
RGB(34, 34, 34)

Wolfram Language
name: "wolf" / "wolfram"
where to find: https://www.wolfram.com/language/
logo: https://www.wolfram.com/language/img/wolfram-language-logo-home.png
colors:
RGB(194, 32, 30)
RGB(14, 14, 14)
RGB(255, 255, 255)

birjolaxew:
Official SVG sources are available in their [press center](http://company.wolfram.com/press-center/), embedded in PDFs. Note that the logo you linked is for Wolfram the company, not for Mathematica the program. Mathematica use [two different logos](http://company.wolfram.com/press-center/mathematica/).

Extracted the SVGs are:

- **Wolfram (the company)** (inconsistent coloring, use `#DD1100` as with the rest of their logos)

    ```svg
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 838 716">
[...]

PR #917, opened 2018-08-02
[New icon] Wolfram (organisation)

ericcornelissen:
As per #873


_data/simple-icons.json
```diff
@@ -2045,6 +2045,11 @@
             "hex": "FAAD4D",
             "source": "http://www.wix.com/about/design-assets"
         },
+        {
+            "title": "Wolfram",
```
birjolaxew: Although the technically accurate name is `"Wolfram Research"`, it is never used except where distinguishing between Wolfram and Wolfram products is important. As such I feel that `"Wolfram"` - which is used practically everywhere that the company refers to itself - is acceptable.

PR #922, opened 2018-08-05
Add SVGLint

### LINKED ISSUE #485: Linter
In an effort to make it easier to follow [CONTRIBUTING.md](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md), we could add a linting tool as a precommit hook.

It would test the files being committed, and block the commit with a message similar to the following if something is wrong:

```
✔ icons/valid.svg
✖ icons/invalid.svg
    Expected viewBox to be 24x24, was 512x512
```

This could of course also be ran in a CI to integrate it into GitHub.

birjolaxew:
Adds [`svglint`](https://github.com/birjolaxew/svglint) as a dev dependency, and runs it as part of our CI steps. Note that the Node version ran by our CI has also been upgraded from 6 to 8, as SVGLint requires support for async functions.

Note that this adds a folder called `scripts`, which is intended to be where we put meta-code (e.g. SVGLint config, scripts to build the site for #898, other scripts to be ran by CI) - the name of this folder is up for debate if anyone has a better idea.

Currently tests for the presence of an `<svg>` and a `<title>` tag, and disallows all `<g>` tags. It further checks that the attributes we require in our contribution guidelines are there.

Closes #485


birjolaxew:
The build log when a file fails to lint:

TravisCI log

ericcornelissen:
> Note that this adds a folder called scripts, which is intended to be where we put meta-code (e.g. SVGLint config, scripts to build the site for #898, other scripts to be ran by CI) - the name of this folder is up for debate if anyone has a better idea.

I can kinda see why you added the `svglintrc.js` file in the scripts folder, but it feels more like a configuration file comparable to [.jsonlintschema](https://github.com/simple-icons/simple-icons/blob/develop/.jsonlintschema) and therefor I would put it in the root.

Other then that it looks good to me 👍 

If I had to give a nitpick it was on the script you added to the `package.json`, which I would write as "svglint icons/* --ci --config scripts/svglintrc.js". This in order to be consistent with the script for jsonlint, and because I like to separate my named arguments from my unnamed arguments.

birjolaxew:
@ericcornelissen Could you please verify that this work on Windows, and squash merge it if so? I'll be gone for the next week, so I won't be able to merge it myself.

Issue #932, opened 2018-08-14
Update Evernote icon


augbog:
Evernote just launched a new brand and has a slightly different logo now. See details here:

https://evernote.com/blog/ceo-notebook-evernote-brand/

I can try to work with a designer and submit a PR for an updated one :)

PR #933, opened 2018-08-14
#932 update Evernote icon for new brand

augbog:
I went and updated the Evernote icon to align with the new brand :) Let me know if there are any issues.

This is a preview of what it looks like locally.

screen shot 2018-08-14 at 1 30 48 pm


ericcornelissen:
Where did you find the new SVG @augbog ?

augbog:
@ericcornelissen I work at Evernote so I resized and modified it to fit simple-icons straight from the source! I simply took it from one of the brand assets we were given :P 

For reference: 
https://evernote.com/blog/ceo-notebook-evernote-brand/

If you need a confirmed source of where the SVG itself is being used, I can also find one.

ericcornelissen:
@augbog we prefer a link to a page where the icon source (and colour) can be found (if one exists), I found [this](https://brandfolder.com/evernote) linked on your [press page](https://evernote.com/press), I am okay with either 😄

augbog:
Oh nice thanks! (Embarassing you found it easier than I did even though I work here 🤣. I'll update the link in the JSON

birjolaxew:

PR #935, opened 2018-08-20
Add Envato icon

### LINKED ISSUE #85: Envato
How about an icon for the Envato marketplaces?

Some stuff can be found here : http://themeforest.net/make_money/banners_and_logos

birjolaxew:
Color taken from the official SVG on [their website](https://envato.com/)
Closes #85 

Note that the icon is kind of poor quality, but since it is available as an official SVG on their website I have avoided touching it beyond running it through SVGO.


ericcornelissen:
Looks good 👍 The official icon does indeed seem to be of a somewhat low quality 😕

PR #936, opened 2018-08-20
Add Renren

### LINKED ISSUE #309: Request: Renren
Color : #005EAC
Website : renren.com

birjolaxew:
`#217DC6` from their official .jpg logo from renren.com
Note that the .jpg is very noisy - as such there are multiple shades of blue in it. The chosen one is from the biggest consistent area I could find.
Alternative color choice: `#005EAC` from the .png logo from [Wikipedia](https://en.wikipedia.org/wiki/Renren#/media/File:Renren.png), which claims to be sourced from renren.com, although I couldn't find the .png on the official site.

As no official SVG source is available, one was taken from third party sources. This was verified to match the official .jpg as closely as possible.

Closes #309


ericcornelissen:
The SVG looks good to me 👍 Extracting colors from JPGs is always tricky, I got `#267ec5` in Photoshop, but it doesn't really matter...

PR #942, opened 2018-08-25
Add prepublish script

### LINKED ISSUE #751: Webpack throwing errors when bundling Simple Icons in project

### LINKED ISSUE #941: Make simple-icons npm package compatible with browsers
Hi,

Because the current implementation use `fs`, I cannot use `simple-icons` in a browser without passing through a CDN.

How about introducing webpack in this project to be able to build a static index.js with all svg strings already injected in the bundle ?
https://github.com/webpack-contrib/svg-inline-loader

birjolaxew:
Compiles icons into both static js files (icons/{filename}.js) and into a static index.js. The latter is not tree-shakeable.

This will run `scripts/prepublish.js` before the package is published to NPM. Said script will:

1. Compile `index.js` - it will from now on only contain a static JS object, and can therefore be used in browsers (#751, #941)
2. Compile each icon into its own `.js` file, located in `icons/{filename}.js`. This can be imported directly, as importing `index.js` puts the *entire* JS object into your compiled package. By importing the icons directly, only the used icons are included.

The published project structure will be:

```
simple-icons
├ icons
│ ├ 500px.svg
│ ├ 500px.js
│ ├ about-dot-me.svg
│ ├ about-dot-me.js
│ └ ...
│...
└ index.js
```

which can be used by doing:

```js
// not tree-shakeable, but now browser usable
const googleplus = require("simple-icons")["Google+"];
// tree-shakeable, also browser usable
const googleplus = require("simple-icons/icons/googleplus");
```

As the API has not changed, this is not a breaking change.

Closes #941 
Closes #751


scripts/prepublish.js
```diff
@@ -0,0 +1,31 @@
...
+ * and are tree-shakeable.
+ * The static .js files go in icons/{filename}.js.
+ * Also generates an index.js that exports all icons by title, but is not tree-shakeable
+ */
+
+const dataFile = "../_data/simple-icons.json";
+const indexFile = `${__dirname}/../index.js`;
+const iconsDir = `${__dirname}/../icons`;
+const data = require(dataFile);
+const fs = require('fs');
```
ericcornelissen: Nitpick, but other then this you're always using double quotes

scripts/utils.js
```diff
@@ -0,0 +1,14 @@
...
+     * Converts a brand title into a filename (not a full path)
+     * @param {String} title The title to convert
+     */
+    titleToFilename: title => (
+        title.toLowerCase()
+            .replace(/\+/g, "plus")
+            .replace(/^\./, "dot-")
+            .replace(/\.$/, "-dot")
+            .replace(/\./g, "-dot-")
+            .replace(/[ !’]/g, '')
```
ericcornelissen: Nitpick, but other then this you're always using double quotes

README.md
```diff
@@ -47,6 +47,23 @@ console.log(simpleIcons['Google+']);
 */
 ```
 
+Alternatively you can import the needed icons individually.
+This is useful if you are e.g. compiling your code with webpack and therefore have to be mindful of your package size.
```
ericcornelissen: `webpack` could be a link, just like "JSDelivr" and "Unpkg"

package.json
```diff
@@ -23,6 +23,8 @@
   },
   "scripts": {
     "jsonlint": "jsonlint _data/simple-icons.json -q -V .jsonlintschema",
-    "svglint": "svglint icons/* --ci"
+    "svglint": "svglint icons/* --ci",
+    "prepublishOnly": "node scripts/prepublish.js",
+    "postpublish": "rm icons/*.js && rm index.js"
```
ericcornelissen: I believe `rm icons/*.js index.js` also works, why did you write it like this?

scripts/prepublish.js
```diff
@@ -0,0 +1,31 @@
...
+
+const icons = {};
+data.icons.forEach(i => {
+    const filename = titleToFilename(i.title);
+    i.svg = fs.readFileSync(`${iconsDir}/${filename}.svg`, "utf8");
+    icons[i.title] = i;
+    // write the static .js file for the icon
+    fs.writeFileSync(
+        `${iconsDir}/${filename}.js`,
+        `module.exports = ${JSON.stringify(i)};`
```
ericcornelissen: Maybe change this to `module.exports=${JSON.stringify(i)};` (removing the space) to reduce the output size?

scripts/prepublish.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+});
+
+// write our generic index.js
+fs.writeFileSync(indexFile, `module.exports = ${JSON.stringify(icons)};`);
```
ericcornelissen: Same as [above](https://github.com/simple-icons/simple-icons/pull/942/files#diff-0f96cf093958bcaa0a9e684fd9e6b9a4R26), change this to `module.exports=${JSON.stringify(icons)};` (removing the space) to reduce the output size?

scripts/prepublish.js
```diff
@@ -0,0 +1,31 @@
...
+const dataFile = "../_data/simple-icons.json";
+const indexFile = `${__dirname}/../index.js`;
+const iconsDir = `${__dirname}/../icons`;
+const data = require(dataFile);
+const fs = require("fs");
+
+const { titleToFilename } = require("./utils");
+
+const icons = {};
+data.icons.forEach(i => {
```
ericcornelissen: Rename the variable `i` to `icon` for readability

birjolaxew:
First version with this is now out, and I have confirmed that it works in both node and webpack :smile:

PR #948, opened 2018-08-26
Update the travis build to work in stages

ericcornelissen:
Improve the CI (test, build, and deploy) process by using [build stages](https://docs.travis-ci.com/user/build-stages/#examples) rather than a (somewhat hard to understand and arbitrary) [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#build-matrix). For now it just improves the readability of the `.travis.yml` file and the understandability of the [build status page](https://travis-ci.org/simple-icons/simple-icons).

For clarity, I added two commits showing the build report when something goes wrong. The first is when the [website cannot be build](https://travis-ci.org/simple-icons/simple-icons/builds/420875745?utm_source=github_status&utm_medium=notification), and the second is when there is a [linting error](https://travis-ci.org/simple-icons/simple-icons/builds/420876383?utm_source=github_status&utm_medium=notification). As seen in the images below:

build website error

linting error

* * *

To summarize, Travis will now first lint & build the website (in parallel), and only if both succeed will it try to deploy. That only happens if the commit is on `master`. So it is ignored most of the time, as seen [here](https://travis-ci.org/simple-icons/simple-icons/builds/420876894?utm_source=github_status&utm_medium=notification).

In the future, with #898 in mind, this would make it easier to build to website in one step, and have it be deployed in another one (if that is even needed 🤔), Possibly we could even load external files (e.g. [simple-icons-pdf](https://github.com/simple-icons/simple-icons-pdf)), and include them in the website. Also, extra test steps (such as described in #905) are now easier to add.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ jobs:
       name: "NPM Package"
       language: node_js
       node_js: 8
+      if: branch = master
 
       deploy:
         provider: npm
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,24 @@
-matrix:
+jobs:
   include:
-    - language: node_js
-      node_js:
-        - "8"
-
+    - stage: "Test"
+      name: "Lint"
+      language: node_js
+      node_js: 8
       script:
         - npm run jsonlint
         - npm run svglint
+    - name: "Build website"
+      language: ruby
+      rvm: 2.4.1
+      install:
+        - gem install jekyll
+      script:
+        - jekyll build
 
-      notifications:
-        email:
-          on_success: never
-          on_failure: change
+    - stage: deploy
+      name: "NPM Package"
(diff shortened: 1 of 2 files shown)


ericcornelissen:
> Minor thing, but could we add a conditional branch = master to the deploy stage, so that PRs targetting develop don't run an empty stage?

Good idea, I added the conditional and it seems [to work](https://travis-ci.org/simple-icons/simple-icons/builds/421149730?utm_source=github_status&utm_medium=notification). I left the conditional in the `deploy:` as well just to be sure ([this one](https://github.com/simple-icons/simple-icons/pull/948/files#diff-354f30a63fb0907d4ad57269548329e3R28)), but I would be happy to remove it 😄

birjolaxew:
:+1: Nice work on this!

PR #944, opened 2018-08-26
Add Go (the programmin language)

ericcornelissen:
As requested in #469

PR #951, opened 2018-08-29
Update [go] icon

RedSparr0w:
Update go programming language icon.

It was updated 3 days ago:
https://blog.golang.org/go-brand
They also have a new primary color:
[go branding](https://golang.org/s/brandbook)


birjolaxew:
There actually was [some discussion on this in the previous PR](https://github.com/simple-icons/simple-icons/pull/944#pullrequestreview-149532388). The consensus seemed to be that, since it's been 4 months since they published the new design and they haven't really started using it anywhere, we'd stick with the gopher for now and update at a later time when they start using the new design as their actual brand.

I'm open for discussion on the topic though.

RedSparr0w:
Ah, didn't see that!
Thanks for the link :+1:
Yeah, that definitely makes sense, i'll close this for now until they actually start using the logo also :laughing: 
>it's been 4 months since they published the new design

For some reason i read it as 26 August instead of April :man_facepalming:

PR #953, opened 2018-08-29
Add CoffeeScript

ericcornelissen:
Add icon for the CoffeeScript language as requested in #469 

The color, `#3E2723` is taken from the navigation bar of their website (`<nav>`). This is surely debatable, the main icon on the homepage (not the one in the navigation bar) uses `#2F2625`, and they use `#2F2526` in their code blocks (`.code-example`). These are just some colors I considered using...


ericcornelissen:
> I got the color from their logo as #2F2625, based on the big SVG logo on their frontpage:

My bad, I looked at the `color` rather then the `fill` for that one 😅 I updated the original comment. Do you think we should use `#2F2625` rather then `#3E2723`?

birjolaxew:
Yes. They apparently prefer to use `#2F2625` for their logo, so we should too.

birjolaxew:

Issue #893, opened 2018-07-21
Rename 'OneDrive' to 'Microsoft OneDrive'


DoronBrayer:
1. Rename title: 'OneDrive' to 'Microsoft OneDrive'.
2. Rename filename: onedrive.svg to microsoftonedrive.svg

PR #957, opened 2018-08-29
Update OneDrive (#893)

metaa:


birjolaxew:
:+1: Good catch! Thanks for the contribution.

PR #960, opened 2018-08-30
Add Mixer

metaa:
# Added Icon
Mixer

## Website
https://mixer.com/

## Logo
https://github.com/mixer/branding-kit/blob/e08cd4a165cc8a40e17bde384c7e664ec43dd155/svg/MixerMerge_Black.svg

## Color
`#002050`

You may wonder why I chose this color instead of the light blue (or maybe not) as there are two main colors on the logo (for dark backgrounds):

I've been a user on Mixer for quite a long while now and I was almost certainly sure that it could be the light blue, but I still wasn't sure.

I got in contact with Microsoft employees and their marketing team and they told me to use the navy color.

## What it will look like
image


metaa:
I did put it through SVGO with precision `3` (as the contribution guide states) and I only see very few vertices that seem necessary though.

Should I reduce the precision? 🤔

ericcornelissen:
> I did put it through SVGO with precision 3 (as the contribution guide states) and I only see very few vertices that seem necessary though. (I repeated this process after your comment but apart from the header changing and the title missing, there are no changes at all)

I might have been to quick with replying, it seems you're right 😅 My apologies. Thanks again @metaa!

PR #974, opened 2018-09-09
propose design update

abenjamin765:
Using a border color on each icon rather than a background color makes the icons and labels more visible. 

This also resolves the need for dynamic coloring for light and dark brand colors.

### DIFF
--- a/index.html
+++ b/index.html
@@ -215,7 +215,7 @@
             </li>
             {% for icon in iconsArray %}
                 {% assign iconArray = icon | split: "," %}
-                <li class="grid-item {{ iconArray[6] }}" style="background-color: #{{ iconArray[4] }}" order="{{ iconArray[7] }}">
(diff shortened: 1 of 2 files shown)


ericcornelissen:
I'm okay with an update to the design, and I like the idea of using the border to convey the brand color. However, I'm not the biggest fan of your proposal aesthetically, but maybe that is just because I'm so used to the current design 😅 I feel like your design is a little too bright, if that makes sense?

One critical remark I have though is that the text for the hex-valueis currently somewhat hard to read for all icons (rather then just some of them), so this text should be made darker in your current design.

For anyone wondering what this will look like, here is a screenshot:

image

abenjamin765:
Thanks for the feedback! I agree that further improvements to readability could be made. Might be interesting to explore a "light mode/dark mode" toggle. Also, maybe bringing the color into closer context with the hex value could work better.

birjolaxew:
This should be discussed in an issue before moving to a PR; opened #980 for this purpose.

Reopening this PR if we find that we're going with this design.

PR #973, opened 2018-09-09
Add Apple Pay, AT&T, T-Mobile, Sprint, Verizon, Boost, Framer

abenjamin765:
- add icon for AT&T
- add icon for T-mobile
- add icon for Sprint
- ~~add icon for Cricket~~
- add icon for Verizon
- add icon for Boost
- add icon for Framer
- ~~add icon for Sketch~~
- Update _data/simple-icons.json with new icon data

- add Apple Pay

Issue #770, opened 2018-02-16
Change project structure


birjolaxew:
I'm not entirely happy with how the folder structure is currently set up. I see a couple of different possibilities for a new structure:

1. **Keep our current structure**
    ```
    _data
      └ simple-icons.json
    icons
      ├ 500px.svg
      ├ aboutme.svg
      ├ acm.svg
      └ ...
    ```

1. **Move our JSON out of its folder**
    There's little point having a folder for a single file.

    ```
    simple-icons.json
    icons
      ├ 500px.svg
      ├ aboutme.svg
      ├ acm.svg
      └ ...
    ```

1. **Move our JSON to the `icons` folder**
    The one upside of having the JSON in its own folder is that it is kept separate from our meta files (`CONTRIBUTING.md`, `package.json`, website files, etc.). We could instead move it to the `icons` folder, so that all of the actual icon files are in the `icons` folder.

    ```
    icons
      ├ _data.json
      ├ 500px.svg
      ├ aboutme.svg
      ├ acm.svg
      └ ...
    ```

    or

    ```
    icons
      ├ data.json
      └ svgs
        ├ 500px.svg
        ├ aboutme.svg
        ├ acm.svg
        └ ...
    ```

1. **Create self-contained folder for each icon**
    Each folder would be self-contained, meaning the current structure of `_data/simple-icons.json -> icons/[name].svg` would become `icons/[name]/icon.json -> icons/[name]/icon.svg`.

    I personally think this solution adds too many files, although I do like avoiding the monolithic JSON file we have currently.

    ```
    icons
      ├ 500px
      │  ├ icon.json
      │  └ icon.svg
      ├ aboutme
      │  ├ icon.json
      │  └ icon.svg
      ├ acm
      │  ├ icon.json
      │  └ icon.svg
      └ ...
    ```

Note that our build step will convert whatever structure we choose into one tailored for use in Webpack/NodeJS before publishing to NPM, so that is not a concern.

ericcornelissen:
I agree that a separate folder for the icon data is not necessary (I guess it was done initially for Jekyll). I personally prefer option 2. Option 3b, with a nested directory for the SVG icons, since that could help with #743 (pdf formatted icons) .

Have you also considered restructuring code related to the NPM package and code (CSS, JS, HTML) related to the website?

> Note that our build step will convert whatever structure we choose into one tailored for use in Webpack/NodeJS before publishing to NPM, so that is not a concern.

[...]

PR #975, opened 2018-09-09
add iconjar file

abenjamin765:
- Add `.iconjar` file
More info here: https://geticonjar.com/

Note: Automation is available here: https://github.com/IconJar/iconjar-exporters - Would require an installation of the PHP library


birjolaxew:
The icon addition has a couple issues:
- IconJar is missing from `simple-icons.json`
- [IconJar has a >500k Alexa rank](https://www.alexa.com/siteinfo/geticonjar.com), which [usually makes it too unpopular to be included in Simple Icons](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon). Since it is a downloadable program and not just a website this might be a bit misleading though - we could consider merging it in if anyone has a source that indicates that it's used by a wide audience.

More importantly though, the PR has some issues:
- It includes the commits from #973 - branching should happen from `develop` so we don't end up with commits in two PRs :) This can be fixed by creating a new branch off of `develop` and then cherry-picking c4f7d64 onto it.
- The commit that adds IconJar (c4f7d64) includes some apparent restructuring (?) - all icons have been moved to `assets/simple-icons.iconjar/icons/`. This definitely shouldn't happen.

abenjamin765:
I wasn't super clear in my commit. Apologies :). I'm looking to add an `.iconjar` file that contains all of the simple-icons and their metadata so that designers who use IconJar can use simple-icons in their projects. You can read more about IconJar in the link.

abenjamin765:
image

ericcornelissen:
> I wasn't super clear in my commit. Apologies :). I'm looking to add an .iconjar file that contains all of the simple-icons and their metadata so that designers who use IconJar can use simple-icons in their projects. You can read more about IconJar in the link. My recommendation would be to host the .iconjar file and made it available to users for download.

I'm definitely in making SimpleIcons available for as many people as possible, however I don't think the correct approach for this is to include all the data for IconJar in this repository. For one, all the now appear twice in this repository. This also means that updating or adding new files becomes more of a hassle. 

There is an ongoing discussion on the structure of the project in #770. Personally I think it would be best to create a separate repository for the IconJar files and have it somehow be updated automatically (similar to [simple-icons-pdf](https://github.com/simple-icons/simple-icons-pdf). However, I'm not familiar with IconJar so this might not be the best approach.

For now I will close this PR, but we can discuss further here or in a separate issue :slightly_smiling_face:

abenjamin765:
Sounds good. I linked a PHP library above (https://github.com/IconJar/iconjar-exporters) that would automate the duplication process to the jar directory with the correct metadata. It would be ideal to only have a single source, and that may be possible with further investigation. IconJar is one of the most widely used Icon managers for designers, so having an `.iconjar` file available would be valuable.

PR #977, opened 2018-09-10
Update filemapping, ampersand

ericcornelissen:
As encountered in #973 with AT&T, this should resolve any problems with brands including an ampersand (`&`) in their name. @birjolaxew, I hope I didn't forget anything?


birjolaxew:
> @birjolaxew, I hope I didn't forget anything?

:+1: Neither do I 😉 Hopefully the rewrite from earlier means that we only have to change it in two places (and once we rewrite the website, in one place only), so this should be good to go.

PR #1004, opened 2018-10-07
at-and-t.svg ampersand character

geoctrl:
concerning **at-and-t.svg**:

the title has an ampersand (not used as a character) which makes it invalid xml.

can't open it in illustrator or use svgo like this.

changing the `&` to `&amp;` fixes this


ericcornelissen:
That is a bit silly of us 😅 Thank you so much for fixing this @geoctrl!

Issue #792, opened 2018-03-12
REQUEST: Money transfer apps


jonhoo:
- [x] Venmo
Logo available from [Brand page](http://brand.venmo.com/styleguide#section-logo-variations) (see "Profile image or app button").
- [x] Google Pay
Surprisingly hard to track down. May be available through the [GSuite logo pack](https://gsuite.google.com/setup/resources/logos/).
- [x] Circle Pay
Logo kit available on [Press site](https://www.circle.com/en/news).
- [x] Square Cash
Logo inlined on [website](https://cash.me/):
```svg
<svg class="app-icon" viewBox="0 0 64 64">
 <g fill-rule="nonzero" fill="#FFF">
  <path d="M41.7 0c6.4 0 9.6 0 13.1 1.1a13.6 13.6 0 0 1 8.1 8.1C64 12.7 64 15.9 64 22.31v19.37c0 6.42 0 9.64-1.1 13.1a13.6 13.6 0 0 1-8.1 8.1C51.3 64 48.1 64 41.7 64H22.3c-6.42 0-9.64 0-13.1-1.1a13.6 13.6 0 0 1-8.1-8.1C0 51.3 0 48.1 0 41.69V22.3c0-6.42 0-9.64 1.1-13.1a13.6 13.6 0 0 1 8.1-8.1C12.7 0 15.9 0 22.3 0h19.4z" fill="#00D632">
  </path>
[...]

PR #1006, opened 2018-10-09
Add Venmo icon

ericcornelissen:
Related to #792 

Icon and metadata comes straight from the guidelines. I also sneaked in a little reordering to the `_data/simple-icons.json` file


birjolaxew:

PR #1007, opened 2018-10-09
Add Cash App icon

ericcornelissen:
Related to #792

The icon is imported directly from the press kit. The color is taken from the logo, which has a gradient from `#00C244` (which I picked) to `#00D64B`.


birjolaxew:
Various other color variants they use:

- `#28C101` for their CtA on their website.
- `#00D632` for their logo SVG on their website.

They clearly aren't very consistent with their color usage - since the one in this PR is taken from their official press kit, I'm fine with using that one.

PR #1017, opened 2018-10-21
Add Google Pay

louistsaitszho:
Again, #792. Let's hope I got it right this time 👉👈

comparegif

There is just 1 issue: What's the color? Right now I just put `#ffffff` as a placeholder


ericcornelissen:
The SVG is almost there. I think we should omit the circle and have it just be the letter `G` and the word `pay`. Either way you should remove the `fill` (makes sure the SVG is easier to style) and `fill-rule` (has some problems with different rendering engines iirc) from the SVG.

As for the color, I would go with the gray color they're consistently using [the guidelines](https://developers.google.com/pay/api/web/guides/brand-guidelines) on light background, but I'm unsure as well  😅

ericcornelissen:
One last thing about the SVG, could you make it so the `G` and `y` touch the sides of the box @louistsaitszho?

ericcornelissen:
Thanks a bunch @louistsaitszho 🎉

PR #1019, opened 2018-10-22
Added Flattr & Jabber icons

atomicjeep:
#86 #1011


birjolaxew:
Good to see you resolved most of my comments from the old PR (#1011) :)

One still remains: On [Flatter's Twitter](https://twitter.com/flattr) it seems that they use `#000000` as their brand color, and `#FFFFFF` is used on the website because the logo is displayed on top of black. This is up for debate however, and if you feel that white better represents their brand then I'm probably able to be convinced

ericcornelissen:
I would vote for `#000000`, but its certainly up for debate. If you're going to change that, maybe also remove the extra space at the start of `icons/jabber.svg` 😜

atomicjeep:
that should be both things resolved :-)

ericcornelissen:
It is, thank you very much @atomicjeep 🎉

PR #1023, opened 2018-10-23
Update Github templates

ericcornelissen:
Changes:
- Moved the templates to a `.github` folder.
- Added Pull Request template with some helpful reminders for contributors (such as that the icon should be optimized using SVGO or SVGOMG)
- Created multiple issue templates for different issue types, which will look like ⬇️ 

create issue preview


birjolaxew:
I like the general idea, however I do have concerns in regards to over-formalizing the contribution process.

In my opinion it is important that the first-time contributor meets the least amount of friction possible. By requiring people to read a lengthy issue template, or by requiring a specific format for issue titles, we introduce a fair bit of mental overhead. Strictly formalizing issue formats also (needlessly, imo) removes flexibility, making us appear less... human. It's less fun to contribute to a repo ran by a bot ;)

Personally I'd go with something along the lines of

```diff
+ When requesting a new icon we need information such as:
+ * Name and website of the brand
+ * Official resources for icon and color (e.g. media kits, brand guidelines, SVG files, ...)
+
+ Before opening a new issue, please check that an existing one doesn't exist (or has been closed)

- Before opening a new issue search for duplicate or closed issues.
-  When requesting a new icon include helpful information such as:
- * The requested icon’s brand name in the title.
- * Links to official sources for the brand's icon and colors (e.g. media kits, brand guidelines, SVG files etc.).
- * Make sure the title starts with "Request:".
-  In addition to that remember that:
- * We won't add generic icons (such as email or arrows) since our goal is to provide brand logos.
- * We won't add anything related to illegal services (e.g. piracy, malware, threatening material, spam, etc.).
```
This is significantly shorter, with less "noise". Note that I removed some things - we receive requests for generic icons or illegal services so rarely (once and never, in my time of being a maintainer) that I'm fine with just closing the issue with an explanation if it pops up.

ericcornelissen:
I agree that we shouldn't over-formalize the contributing process, I mostly just iterated on existing content (i.e. the current [Issue template](https://github.com/simple-icons/simple-icons/blob/47c74a1f0a0b588279c3076e8b0bf449b8443f73/ISSUE_TEMPLATE.md) and [Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/dd97f13ff47798169435ec6593f993fa3a67afa1/CONTRIBUTING.md)) for the initial PR. 

With regards to removing the formatting requirements from the templates, should we also remove them from the Contributing guidelines? And related, maybe we can add the generic icon case to the contributing guidelines?

My goal with the different templates was specifically to prevent "requiring people to read a lengthy issue template" by only providing relevant information to the type of issue. Maybe I went too far by including the notes on generic icons or illegal services, and I also don't mind closing such issues either. On the other hand I don't think it hurts to include them regardless, but maybe in a shorter fashion, something like:

```diff
+ <!-- Before opening a new issue, please check that an existing one doesn't exist (or has been closed) -->
+
+ <!-- When requesting a new icon we need information such as: -->
+ **Name:**   
+ **Website:**   
+ **Official resources for icon and color:** (e.g. media kits, brand guidelines, SVG files, ...)  
+
+ <!--
+ We won't add generic icons or anything related to illegal services.
+ If in doubt open an issue anyway.
+ -->

- Before opening a new issue search for duplicate or closed issues.
-  When requesting a new icon include helpful information such as:
- * The requested icon’s brand name in the title.
- * Links to official sources for the brand's icon and colors (e.g. media kits, brand guidelines, SVG files etc.).
- * Make sure the title starts with "Request:".
-  In addition to that remember that:
- * We won't add generic icons (such as email or arrows) since our goal is to provide brand logos.
- * We won't add anything related to illegal services (e.g. piracy, malware, threatening material, spam, etc.).
```

_Note here that I used HTML comments to hide the tips when the issue is published. And that I moved the reminder about duplicate issues back to the top, as I think that is quite important._

I like the idea of formatting the templates such that they can be filled out rather then having them be description of what we expect 👍 I will make changes accordingly

birjolaxew:
Nice detail with the comments!

Yes, I do think that our current `CONTRIBUTING` is a fair bit too long for us to reasonably expect that people read it. Splitting it out into issue templates is great, and definitely resolves most of the problems with our `CONTRIBUTING`, but they should be as minimal and easily parseable as possible with `CONTRIBUTING` acting as a backup with more details if we need it.

I think that the current way we describe issue format in `CONTRIBUTING` is fine; it doesn't explicitly define any format, but does suggest one by using it for examples. I like this approach of implicitly nudging readers towards the format we like without making any hard rules about it.  
The sections on adding a new icon could definitely be rewritten to be shorter though.

In general I think that the templates should be rewritten to be shorter, to be more human/friendly, and to be easily parseable. To that end I like the format from your comment, especially with the less important notes (on generic/illegal icons) in their own block so that they do not interfere with the rest of the template. The only suggestion I have would be to change "open an issue anyway" into "open an issue and we'll have a look" - which needless to say is just personal preference.

As for adding the case of generic icons to `CONTRIBUTING`, I think that's already covered by requiring the icon to be of a brand. The one time an issue with it was opened was based on a misunderstanding of the project. Opening issues that don't relate to the repo doesn't seem like something we have to explicitly forbid. While I don't particularly hate adding a mention of it, it does seem a bit redundant.

ericcornelissen:
I updated the issue templates to be a _form to fill out_ rather being _a description of what we need_. Note that the Pull Request template still needs to be updated.

I'm completely fine with removing the `<!-- include this in the title as well -->` comments ([here](https://github.com/simple-icons/simple-icons/pull/1023/files#diff-44c7924885501a49e748a79c45019698R11) and [here](https://github.com/simple-icons/simple-icons/pull/1023/files#diff-d17059e92d80dc61098e36b42e299316R11)) as most people will do that anyway 👍 

I'm not quite sure about the package/website templates yet. For one, I put the enters between the `### Description` header and its comment so people don't have to manually insert them, but it might look a bit disconnect right now 🤔 Also they're very similar, which I don't think is necessarily a problem (they won't be updated often so keeping them consistent is not a problem) but we might merge them...

.github/ISSUE_TEMPLATE/package.md
```diff
@@ -0,0 +1,22 @@
...
+about: For issues about or improvements to the NPM package
+---
+
+<!-- Before opening a new issue search for duplicate or closed issues -->
+
+
+### Kind of issue <!-- Change the one that applies to `[x]`  -->
+  - [ ] Improvement
+  - [ ] Bug
+  - [ ] Other, namely: <!-- 1 to 3 words -->
```
birjolaxew: I think I'd rather leave out the `<1-- 1 to 3 words -->`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+Before opening you Pull Request have a quick look at our Contributing Guidelines.
```
birjolaxew:
```diff
+ your pull request, have
- you Pull Request have
```

```diff
+ our contribution guidelines (https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md)
- our Contributing Guidelines

File: .github/PULL_REQUEST_TEMPLATE.md [hunk 3]
```diff
@@ -0,0 +1,7 @@
...
+
+When adding a new or updating an icon remember the following points:
```
birjolaxew:
```diff
+ adding a new icon, or updating an existing one, remember
- adding a new or updating an icon
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
...
+* Include the icon’s brand name in the title.
```
birjolaxew: Think this could be left out, similar to the comments from the other templates

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
...
+* Tell us why you chose the color you picked.
+* Make sure you optimized the icon with SVGO or SVGOMG.
+* Reference the issue, if any, the Pull Request addresses.
```
birjolaxew: Perhaps add an "make sure the icon has been resized to have a viewbox of `0 0 24 24`", as that's a problem that keeps popping up

.github/ISSUE_TEMPLATE/package.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Description
+
+
+<!--
+Anything that is even remotely relevant.
```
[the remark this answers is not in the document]
ericcornelissen: I see where you're coming from, will update 👍

ericcornelissen:
I addressed all your comments @birjolaxew 👍 I now also updated the PR template to in the _form to fill out_ style (while also including your comments on it). What are your thoughts on the "Checklist"?

birjolaxew:
:+1: Good stuff @ericcornelissen!

Issue #1027, opened 2018-10-23
Update Tencent QQ icon


birjolaxew:
A new icon has appeared [on Wikipedia](https://en.wikipedia.org/wiki/File:Tencent_QQ.svg) for Tencent QQ - they claim that this one is "extracted from the executable". It also matches their small logo as seen on https://im.qq.com.

biswaroop1547:
@birjolaxew I am trying to fix it 🙂

birjolaxew:
Closed in #1033

PR #1028, opened 2018-10-23
Replaced tencentqq icon with the new updated tencentqq icon #1027

biswaroop1547:
#1027 Replaced tencentqq icon with the new updated tencentqq icon


ericcornelissen:
Hi @biswaroop1547, thank you for taking the time to contribute. For a number of reasons (consistency, accessibility, etc.) we try construct our SVGs in a certain way. For details you can consult the [Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md), but I'll give some initial comments here.

The most important issue with your submission (which is actually not covered by Travis CI) is that it is not monochrome. Have a look at the rich diff for the icon [here](https://github.com/simple-icons/simple-icons/pull/1028/files?short_path=ad65c0e#diff-ad65c0e2d90e38f9c8963355f07cc37f) and that sentence should make sense.

Other then that icon should be constructed in a 24x24px viewbox without specifying an explicit `width` and `height` (i.e. the SVG shouldn't have these attributes).

This might seem like a lot of work, but if you're at all familiar with editing SVG I encourage you to try it out, but to be fair `Good first issue` label might have been a bit misleading.

birjolaxew:
You can see [why it failed in the build log](https://travis-ci.org/simple-icons/simple-icons/jobs/445133967#L478-L492). In this case you have a couple issues mostly because the SVG doesn't follow [our contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md):

- The `<svg>` element should have `aria-labelledby` and `role` set
- The icon should be resized so the viewbox is `0 0 24 24`, not `0 0 120.99969 143.88988`
- The `height` and `width` on the `<svg>` should be removed
- The `<g>` should be removed
- The `fill` on the `<path>` should be removed
- The icon should be ran through [SVGOMG](https://jakearchibald.github.io/svgomg/) with a precision of ~3

biswaroop1547:
@ericcornelissen Thanks! , I will try to fix it 🙂

birjolaxew:
Closed in #1033

Thanks for the effort though @biswaroop1547 , it is much appreciated!

PR #1029, opened 2018-10-23
Adobe premiere

atomicjeep:


ericcornelissen:
It seems you're missing the `role="img"` attribute on the SVG as you can see [here](https://travis-ci.org/simple-icons/simple-icons/jobs/445135552#L480) @atomicjeep 😉

atomicjeep:
doh! got it this time, will create a template going forward

ericcornelissen:
Looking good @atomicjeep, thanks once again 👍 


> doh! got it this time, will create a template going forward

Yea, I tend to copy the initial part of the SVG (`<svg ...><title ...>...</title>`) from an existing icon to be extra sure I don't make mistakes

atomicjeep:
same here I'm doing it before I run it through SVGOMG which removes the title and role tag by default, grrrr.  Thanks for the help

ericcornelissen:
I know, that's why I'm planning on implementing a feature to export/import configuration in SVGOMG so that we can provide a predefined configuration to contributors 😃

PR #1033, opened 2018-10-24
Updated the Tencent QQ icon

AceKnight:
This is my first time working with an SVG file. I hope I did everything right.
Since I used inkscape to convert the image to monochrome it added a bunch of annotations, i didn't know if they would be a problem.
I followed the guideline and did everything else accordingly.


AceKnight:
Done :D

birjolaxew:
:+1: Looking good! Thanks for your work on this!

Just for the records, I've also gone ahead and updated the color in this PR; the blue that was chosen earlier (by me, ironically) was from their website, but a quick google search indicates that the red scarf is the most dominant color when people think "Tencent QQ"

birjolaxew:
Again for the record: I messed up when I created the SVG I gave you previously: that one doesn't have the scarf in it. I've fixed it in #1034.

PR #1037, opened 2018-10-30
Added R-script,Type script & Node-js logo's  

sarithay:
Added R-script,Type script & Node-js logo's


birjolaxew:
There are quite a few issues with this PR. Please [read our contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md).

**General:**
- We already have Node.js, no need to add it again

**JSON:**
- The `.json` isn't sorted - each entry should be placed in alphabetical order according to its title
- The indentation is slightly messed up
- Curly brackets should not appear on the same line - each entry should be formatted similar to all the other entries
- All the `color`'s are set to `#000000` - they should be the brand color of the brand
- `R-script` is not the official name - that would simply be `R`
- Each of the sources simply link to the official website. Neither R nor TypeScript has an official SVG on their website. We use the `source` link when we go back to check if new versions of the logo has come out, so we'd prefer it pointing to the most official place where we can find updates (for TS that would probably be [this issue](https://github.com/Microsoft/TypeScript/issues/1375), for R that would be [this official page](https://www.r-project.org/logo/))

**SVGs:**
- Both the Node and R SVGs are pretty messed up, and do not represent the logos they are meant to
- None of the filenames are correct. `TS.svg` should be `typescript.svg`, `Rlogo.svg` should be `r.svg`
- All logos appear to be created by potrace. We want SVGs from as official sources as possible, and in this case both TypeScript and R actually have official vector versions of their logo. There is no need to potrace a raster.
- None of the logos are formatted as we ask in our contribution guidelines. They should be ran through SVGOMG, should have a `<title>` and a11y attributes added, and be cleaned to avoid extra filesize. All of this is detailed rather thoroughly [here](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#adding-or-updating-an-icon)

birjolaxew:
Closing this one off as abandoned. Feel free to open a new PR if you want to take this one up again!

PR #1039, opened 2018-10-30
Adding VLC mediaplayer icon

atomicjeep:
### Checklist

### Description
  - Why did you pick the hex value?
    **Picked colour used in the windows RT icon SVG which was the only flat coured icon in the source, there is many many shades of orange used in the source codee and website but nothing definetive that I could find**
  - Did you manually vectorize the logo?
 **no icon taken from most current SVG version in source as noted in data json file**
  - Have you used multiple sources?
**no**


ericcornelissen:
- You typed "VNC" rather than "VLC" in both the `simple-icons.json` and the SVG itself
- Could you update the source to link to the repository (or folder in the repository) containing the file rather then the file itself?

atomicjeep:
Argg not the first time I've done that either!!

atomicjeep:
might be an idea to update the contrib guidelines, the example given indicated linking directly to a SVG and does say be specific?

_data/simple-icons.json
```diff
@@ -2280,6 +2280,11 @@
             "hex": "6383A8",
             "source": "https://vk.com/about"
         },
+        {
+            "title": "VLC media player",
+            "hex": "FF8800",
+            "source": "http://git.videolan.org/?p=vlc.git;a=blob_plain;f=extras/package/macosx/asset_sources/"
```
ericcornelissen: For me this link doesn't work but this one does?

```suggestion
            "source": "http://git.videolan.org/?p=vlc.git;a=tree;f=extras/package/macosx/asset_sources"
```

ericcornelissen:
> might be an idea to update the contrib guidelines, the example given indicated linking directly to a SVG and does say be specific?

That is a bit complicated as this mostly only concerns git repositories. Personally I feel more like just asking for an update to contributors. But perhaps that might be frustrating to you (or whoever is contributing)?

birjolaxew:
I think we should definitely try to give contributors all the info they need prior to our review. It's frustrating to submit a PR that is then rejected/requires updating due to something you had no chance of knowing.

The `source` URL for an icon is usually something that's up for interpretation; I think we should add our general rule of thumb of "a URL we can later use to find the logo if it should update in the future" to the contribution guidelines.

ericcornelissen:
Either way, thanks for your contribution @atomicjeep :tada:

PR #1040, opened 2018-10-30
Add note about source URL to contribution guidelines

birjolaxew:
Adds small note on the `source` URL, detailing our rule of thumb for selecting it. This was discussed in #1039 (thanks @atomicjeep for pointing it out).

Also updates the JSON example to one that uses an optimal URL. This could be considered detrimental since it doesn't help if the logo being added doesn't have a similar optimal URL - but I generally feel that it represents what we want better.


ericcornelissen:
Good clarification, I like it :+1:

PR #1046, opened 2018-10-31
A higher quality Grafana logo.

davidklebanoff:
@birjolaxew After seeing your comments about the icon quality, I looked at some of the other SVGs on the Grafana website. I got lucky and it appears that the icon used for Grafana Enterprise had less artifacts, so I remade the icon with that one. In the diff you can see it's noticeably smoother.


birjolaxew:
Looks like the tip has been messed up a bit - it has become thinner and gotten an outwards curve, neither of which I can find in their big logo on [their Grafana Enterprise page](https://grafana.com/enterprise) - did you use another logo, or are those compression artifacts?

davidklebanoff:
@birjolaxew Commited that version. I'd say it's slightly better than the original, not a big difference and still not perfect, but it may just be the best we can get.

birjolaxew:
:+1: Yep, looks about as good as we can hope for, given their official SVGs.

PR #1047, opened 2018-11-01
Add elementary

louistsaitszho:
**Issue:** Add elementary LLC/OS/whatever it is

### Checklist

### Description

This one is super easy thanks to all the resources all available in the right format in 1 place 👍


birjolaxew:
Everything looks good, except I can't seem to find the color you've chosen here anywhere official - the [official blue logomark](https://github.com/elementary/brand/blob/master/SVG/logomark-blue.svg) uses `#64baff`, which is listed as "Blueberry 300" on their brand page.

ericcornelissen:
I think using [the webpage](https://elementary.io/brand) as source rather then the [github repo](https://github.com/elementary/brand) makes more sense, and in that case I would say the color we use should be `#64baff` as @birjolaxew pointed out

birjolaxew:
> And in their website [...] `#3892e0` is Blueberry 500

Wait, are we not seeing the same colors on their brand page? I'm getting `#3689e6` for Blueberry 500

<p align="center"></p>

Regardless, I think using the actual color from their logo (which is also on their brand page) would be more correct. As noted at the top of their README:

> *This is a copy of the Brand Page from our website. Please also check the [actual page](https://elementary.io/brand) for information, this copy is just informational.*

birjolaxew:
I also agree with @ericcornelissen that the `source` should probably be [their brand page](https://elementary.io/brand), which they seem to prefer (and which links to the GitHub repo)

birjolaxew:
:+1: Looks good! Thanks for your work on this!

PR #1053, opened 2018-11-04
Remember users ordering preference

### LINKED ISSUE #999: Remember user preference for ordering
As suggested by @xdesro in #980, we should remember the ordering selected by the user in local storage (e.g. cookies). So that when the user returns to the SimpleIcons website it automatically uses Color-based sorting or Alphabetical sorting, rather then always defaulting to Color-based sorting.

❗️ If we decide to use cookies, we also have to live with the fact that [cookie laws](https://en.wikipedia.org/wiki/HTTP_cookie#EU_cookie_directive) exist , and that we probably have to put up a banner...

ericcornelissen:
**Issue:** closes #999 

### Description
When this PR is merged the website will remember what sorting order the user last selected (either "by color" or "alphabetically"). It does this through the [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) API. Even though it is [widely supported](https://caniuse.com/#search=localStorage) I added if-statements checking if it is available before using it.

_Misc: I also renamed the concept of "sorting" to "ordering" in the script, seems clearer to me_ 😃

### DIFF
--- a/site_script.js
+++ b/site_script.js
@@ -9,6 +9,7 @@
       $orderByRelevance    = document.getElementById('sort-relevance');
 
   var queryParameter = 'q',
+      orderingPreferenceIdentifier = 'ordering-preference',
       previousQuery  = null,
       previousOrdering  = $orderByColor;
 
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Looks good! Thanks for implementing this

Issue #991, opened 2018-09-28
Change title id to class on the CDN [WordPress Plugin]


tjtaylo:
Hey guys, I'm the developer for the Simple Icons WordPress plugin: https://wordpress.org/plugins/simple-icons/

One issue I'm having is that I'm loading and caching the icons from the jsdelivr CDN and these icons include a title snippet like so:

`<title id="simpleicons-twitter-icon">Twitter icon</title>`

This is all fine until a user wants to include more than one twitter icon on a page. For example, having social icons in the header a footer. The result gives an W3C error for having duplicate ID elements.

Now, I could probably hack this away with PHP. But I wanted to reach out here to see if we could change the id to a class instead.

Thank you!

birjolaxew:
The ID is used for targetting by the `aria-labelledby` property of the SVG itself. Unfortunately this requires an ID, and cannot be done with classes (see [Accessible SVGs](https://css-tricks.com/accessible-svgs/#article-header-id-6) for a run-down)

[The spec](https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements) says that we don't need the `aria-labelledby`, putting the `<title>` as the first element is enough. Browser support for this used to be non-existent ([Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=231654&q=SVG%20%20title%20attribute&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified), [Firefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1151648)), but my own testing indicates that both Firefox and Chrome *do* take the name from the `<title>` even without `aria-labelledby`.

I think we should test if this is the case with popular screen readers too (at least those we can get access to), and if so just remove the whole `aria-labelledby` and `id` all-together.

ericcornelissen:
I will try to test the possibility of removing `aria-labelledby` and the ID from `<title>`s as soon as possible. In the meantime, here's a simple script to remove both from all icons in the repository (based on [`script/prepublish.js`](https://github.com/simple-icons/simple-icons/blob/develop/scripts/prepublish.js)):

```js
const dataFile = "./_data/simple-icons.json";
const iconsDir = `${__dirname}/icons`;
const data = require(dataFile);
const fs = require("fs");

const { titleToFilename } = require("./scripts/utils");

const ariaRegex = /aria-labelledby="[^"]*"\s/;
[...]

PR #1062, opened 2018-11-06
Remove aria-labelledby and ids from SVGs

ericcornelissen:
**Issue:** #991

### Description
This Pull Request removes the `aria-labelledby` attribute from all SVGs and the `id` attribute from all SVG `<title>`s. I updated the [SVGLint](https://www.npmjs.com/package/svglint) configuration and documentation (i.e. README.md and CONTRIBUTING.md) accordingly.

As mentioned in the issue I tested this on Windows 10 using: _Windows 10 build in Narrator_, _[NVAccess](https://www.nvaccess.org/)_, _[ChromeVox](https://chrome.google.com/webstore/detail/chromevox/kgejglhpjiefppelpmljglcjbhoiplfn)_. Additionally we might want to test on MacOS, smartphone OSes and Linux (e.g. Ubuntu)(?).

Another thing we may consider doing is informing (existing) contributors about this change, as they will probably still include the two attributes when they submit a new icon 🤔


birjolaxew:
:+1: Great! Thanks for putting in the work to test and update this!

As for current contributors, I'm fine with informing them if they submit a PR with the old format - the build step should fail in those cases regardless.

ericcornelissen:
> As for current contributors, I'm fine with informing them if they submit a PR with the old format - the build step should fail in those cases regardless.

I agree. Just figured a heads-up in the template might be an idea. I also don't think adding temporary messages to the template is a particularly good idea.

PR #1074, opened 2018-11-19
Update Auth0 Icon

pzygielo:
Missing 't' in title was added.

### Checklist
(I did not change the path.)


davidklebanoff:
Good catch!

PR #1078, opened 2018-11-20
Fixed Google Drive color.

davidklebanoff:
Google Drive's color was slightly off. Corrected it so it's now the same as the primary blue color as seen on the official brand guidelines and in the Google and Google Chrome icons.

Also, confirmed we are using the correct color for Google. It's the same hex as seen in other official resources:
https://design.google/library/evolving-google-identity/

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -817,7 +817,7 @@
         },
         {
             "title": "Google Drive",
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Thanks for the fix, small details matter 💯

PR #1105, opened 2018-12-06
Add Icon for XD Adobe

NoSilver2018:
**Issue:**
Add Icon
When I set the SVG `viewbox` to `0 0 24 24`, then Its part is cutted.
How could I fix this problem?
(current setting is `0 24 24`)

### Checklist
  - [ o ] I optimized the icon with SVGO or SVGOMG
  - [ x ] The SVG `viewbox` is `0 0 24 24`


ericcornelissen:
Hi @NoSilver2018, thank you for taking the time to contribute, but it looks like there are some issues with your Pull Request. For a detailed explanation I recommend you to have a look at our [Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md).

In short, you should use some SVG editing software to resize the SVG to fit into a 24px by 24px viewbox (that is what the `0 0 24 24` means). Once you have done that you should use [SVGOMG](https://jakearchibald.github.io/svgomg/) to reduce the filesize of the SVG. Also, make sure the JSON file is valid (you removed the opening `{` of the file) and formatted properly (you're indentation is off from the rest of the file).

One last thing, if you have a look at [all the Adobe logos](http://simpleicons.org/?q=adobe) already available on the site you can see that the product logos include the square and I think the same should be done for Adobe XD.

ericcornelissen:
There is no need to close this Pull Request @NoSilver2018, you can continue working on this branch and the Pull Request will be updated automatically 👍

PR #1112, opened 2018-12-07
With bad xml travis should fail the build

pzygielo:
I propose following change in travis configuration, for sanity test of icons.

**Issue:**
Currently - if SVG with icon is not well-formed XML, travis build succeeds. It could fail fast instead.

1st commit of PR - to trigger build failure on bad xml.
2nd - to remove test icon, and leave config change only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,11 @@ jobs:
     - stage: "Test"
       name: "Lint"
       language: node_js
+      before_install:
+        - sudo apt-get install -y xmlstarlet
       node_js: 8
       script:
+        - xmlstarlet validate --well-formed icons/*.svg
         - npm run jsonlint
         - npm run svglint
     - name: "Build website"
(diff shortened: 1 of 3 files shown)


birjolaxew:
:+1: I would prefer if a PR was made to [svglint](https://github.com/birjolaxew/svglint) to make it fail on malformed XML - but this is a nice stop-gap measure. Thanks for the work @pzygielo (and sorry for the delay in merging)

pzygielo:
@birjolaxew 
1. Thanks for accepting this proposal
2. I don't know svglint (yet) - and xmlstarlet seemed to me a fast enough for sanity test. When svglint gets adopted for similar check - this new step of course will be discard-able.
3. Judging by recent icons submitted - I think this new step will probably never report failure. But just for 1 case of 100 - I think it's good to have it :smile:

Issue #1121, opened 2018-12-13
[Idea]: Improve the pr template


gabrieledarrigo:
### Kind of issue

  - [X] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns

  - [ ] The NPM Package
  - [ ] The website
  - [X] Contribution experience

### Description

One of the six points of the contributing guidelines on the "Adding or Updating An Icon" section, is the
["Update The JSON Data For SimpleIcons.org"](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#6-update-the-json-data-for-simpleiconsorg).

I propose to add another check to the checklist in the pull request template that reminds the user to have properly added the entry in the JSON file.

ericcornelissen:
Closed with #1122

PR #1122, opened 2018-12-13
Add a point to the pr template checklist

gabrieledarrigo:
**Issue: https://github.com/simple-icons/simple-icons/issues/1121**

### Description

Add a point to the checklist of the pull request template that reminds the user to have properly added the entry in the JSON file.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,7 +8,7 @@ Before opening your pull request, have a quick look at our contribution guidelin
 ### Checklist
   - [ ] I optimized the icon with SVGO or SVGOMG
   - [ ] The SVG `viewbox` is `0 0 24 24`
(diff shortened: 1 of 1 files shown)


birjolaxew:
Forgetting to edit the JSON isn't something we actually see people forgetting, and it'll be caught in our linting step if they do. That being said, I vote for merging this PR: it'll update the checklist so that it gives a better idea of what's needed for a PR, and it'll at the very least give contributors one more checkmark to check and feel good about ;)

Waiting for @ericcornelissen's feedback before merging.

gabrieledarrigo:
Ciao @birjolaxew, what you say makes total sense, wait to hear @ericcornelissen!
In any case, thank you so much for the feedback!

ericcornelissen:
I agree that most people will probably not forget it, but I still think its worth including like @birjolaxew said. Thanks for the suggestion and the PR @gabrieledarrigo 😄

Issue #1124, opened 2018-12-14
[Idea] Add css hover effect


eunjin-oh:
### Kind of issue 
  - [ ] Improvement
  - [ ] Bug
  - [x] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [x] The website
  - [ ] Other, namely: 

### Description

I suggest adding a simple CSS hover effect when selecting an icon.

ezzzin:
Revised title

ericcornelissen:
Thank you for the suggestion @ezzzin, this is kind of a sub-request for #980. I prefer to spend time on finding a suitable hover effect until after (or as part of) the redesign, but if you want to spend time on it before that I am okay with that.

As for #1125, I think the approach you choose (scaling the cards) is overkill for Simple Icons and I think going with a more subtle effect is better, but that is just my opinion. If we do decide to use the scaling, I would suggest decreasing the amount by which a hovered item is scaled.

ericcornelissen:
Closing this for now in favor of #980

PR #1125, opened 2018-12-14
Css hover

ezzzin:
**Issue:**#1124

### Checklist


ericcornelissen:
For anyone interested in what this looks like here is a preview (tip for you @ezzzin, when you change something visually, include a screenshot or video of the change 😉):

preview

RedSparr0w:
👍 

It looks pretty good, but maybe a slightly more subtle effect will look nicer?
`scale(1.1)`:



`scale(1.05)`:

shields

Or just scale the icons & name:

ericcornelissen:
Thanks for mocking up different styles @RedSparr0w, I like your last idea, but I would limit it to just the icon (or scale up the hex-value as well)

ericcornelissen:
If anyone wants to add hover effects, I suggest reviewing the discussion here and perhaps first discussing your idea in #1124, I'm closing this pull request for now.

Issue #1129, opened 2018-12-14
Request: PostgreSQL Icon


chris48s:
**Name:** PostgreSQL
**Website:** https://www.postgresql.org/
**Official resources for icon and color:**
https://wiki.postgresql.org/wiki/Logo

I'm interested in contributing this. I've had a go, but I'm a bit stuck on minifying it to match the SimpleIcons requirements. The [2-colour icons](https://wiki.postgresql.org/images/9/90/PostgreSQL_logo.1color_blue.svg) are essentially implemented as a group of lines with a `stroke-width`. Looking through the existing icons, none of them use a `stroke-width` (I assume this is an [extraneous attribute](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#5-check-the-icon) ).

Does anyone have any advice on how to appropriately minify this to match the guidelines?

gabrieledarrigo:
Ciao @chris48s, why don't you use the single color version here:

https://wiki.postgresql.org/images/9/90/PostgreSQL_logo.1color_blue.svg

obtain a single object and fill it in black?

chris48s:
Thanks for the quick response. One of the issues I had is that the lines all visually join up with that thickness, but once you remove the `stroke-width`, there are gaps, so you don't get a single joined-up shape. Maybe I should just freestyle it to join up the gaps?

When you say fill it in, are you thinking of removing the "internal" detail so its just presented as a filled outline?

gabrieledarrigo:
Hi @chris48s,

>  When you say fill it in, are you thinking of removing the "internal" detail so it's just presented as a filled outline?

Yes or, sort of:
I use Inkscape to work on SVG, so what I mean is that you can try to combine all the different lines into a single shape (except for the eyes I guess) and fill it to black; doing so probably the stroke attributes will be removed and the resulting object should be simplified.

Let me know if I can help you!

gabrieledarrigo:
Glad to help, @ericcornelissen, I really like this project and if you are happy I'd like to continue to contribute when I can 🙂
In any case, your explanation was really better than mine!

chris48s:
Thanks for the help, both. "convert stroke to path" was what I was looking for :) I've now submitted #1131 . we can continue the conversation there.

PR #1132, opened 2018-12-17
add postgresql icon (based on 3 colour)

chris48s:
### Checklist

### Description

refs #1129

screenshot at 2018-12-17 21-35-50


ericcornelissen:
Thanks for your work on both this and #1131 @chris48s Having both preview I see why you would prefer the one based on 2 colours but I still feel like the one based on 3 colours is more in line with what PostgreSQL itself uses.

Either way, you work is much appreciated 🎉

PR #1136, opened 2018-12-22
add Azure DevOps

chris48s:
### Checklist

### Description

Azure DevOps logo appears as used at:

https://twitter.com/azuredevops
https://azuremarketplace.microsoft.com/en-us/marketplace/apps/aad.vsonline
https://docs.microsoft.com/en-gb/azure/devops/

Name is generally styled as DevOps not Devops

Possible colours: `#FFFFFF`,  `#0078D7`

azuredevops

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -201,6 +201,11 @@
             "source": "https://www.aventrix.com/press"
         },
         {
+            "title": "Azure DevOps",
+            "hex": "0078D7",
+            "source": "http://azure.com/devops"
+        },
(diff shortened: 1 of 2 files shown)


chris48s:
I wasn't able to track down an SVG from an official source.

My source for this is https://github.com/badges/shields/blob/master/logo/azuredevops.svg but I'm unclear if the original source for that is a vectorization of a raster image or if it came from some non-public source (e.g: somewhere you have to be a Microsoft customer to get it).

PR #1138, opened 2018-12-22
add Dependabot

chris48s:
### Checklist

### Description

https://github.com/dependabot/website/blob/master/src/images/dependabot-logo-square.svg

I've taken the robot face in `#FFFFFF` here. Another option would be to cut the face out of the hexagon (so its a transparent hole) and make the hexagon `#0025FF` .

dependabot


chris48s:
Yep - no probs, updated

dependabot2

birjolaxew:
@ericcornelissen Given the [widespread usage](https://github.com/search?q=is%3Apr+author%3Aapp%2Fdependabot), and the fact that using the service doesn't mean visiting the website, I'm good with adding Dependabot despite the Alexa rank.

ericcornelissen:
Good thinking on searching for GitHub repositories that use Dependabot 👍 I think we're good adding it to the collection. Thanks @chris48s for your work! 🎉

Issue #791, opened 2018-03-12
REQUEST: Liberapay


jonhoo:
[Liberapay](https://liberapay.com/) is a payment site used for recurring contributions to free and open-source software. An SVG version of their logo is in [their GitHub repository](https://github.com/liberapay/liberapay.com/blob/master/www/assets/liberapay/icon-v2_black.svg).

jonhoo:
Oh, and brand color should be pretty straightforward to get from https://github.com/liberapay/liberapay.com/blob/master/www/assets/liberapay/icon-v2_yellow-r.svg.

PR #1139, opened 2018-12-22
add Liberapay

chris48s:
### Checklist

### Description

As used at https://liberapay.com/

liberapay


ericcornelissen:
We closed an issue requesting Liberapay (#791) earlier on the basis that [its Alexa ranking](https://www.alexa.com/siteinfo/liberapay.com) was too low. It still is and its ranking does not seem very stable.

However, I'm guessing this PR is related to https://github.com/badges/shields/issues/2510 (?) In which case I'm fine with adding it. Thoughts @birjolaxew or @davidklebanoff ?

chris48s:
No worries. No pressure to merge it if it doesn't meet your guidelines.

Understanding that there are some logos we want to use aren't suitable for inclusion in simple-icons is still a useful data point for us at this stage.

birjolaxew:
@ericcornelissen I don't really like the idea of making exceptions for the sake of an external project. What I *do* like is taking a discussion on whether our current rules should be kept. Personally, I think the spirit of our (soft) Alexa rank is more along the lines of "don't peddle your own low-popularity projects", and rewording it to reflect that would solve issues like this - but then we're pretty close to maintainers playing doormen, which I don't particularly like.

ericcornelissen:
@birjolaxew 

> I don't really like the idea of making exceptions for the sake of an external project.

I can understand that, my thinking was that the badges project would not support any brand icons we wouldn't support even if the Alexa rank is too low.


> I think the spirit of our (soft) Alexa rank is more along the lines of "don't peddle your own low-popularity projects", and rewording it to reflect that would solve issues like this - but then we're pretty close to maintainers playing doormen, which I don't particularly like.

I agree that the Alexa rank is becoming a bit awkward in many cases, but the spirit you describe is a good one. Perhaps asking for a "proof of popularity" in icon requests/pull requests is an idea? This could by anything from an Alexa rank, to a Google/GitHub/whatever search like [the one you gave](https://github.com/simple-icons/simple-icons/pull/1138#issuecomment-449874851), to maybe even a Wikipedia article stating how popular something is.

Either way, not to burden @chris48s too much with our discussion(s), are you okay with merging this particular Pull Request?

PR #1159, opened 2019-01-05
[build] refine xmlstarlet options 

pzygielo:
**Issue:**
xmlstarlet could be more verbose on validation errors :arrow_forward: --err
xmlstarlet could print only files that do not pass validation :arrow_forward: --list-bad


ericcornelissen:
Very helpful thanks a bunch @pzygielo :tada:

Issue #1162, opened 2019-01-08
Request: GoG.com


tryton-vanmeer:
**Name:** GoG
**Website:** https://www.gog.com/
**Official resources for icon and color:**
+ https://www.cdprojekt.com/en/media/logotypes/
+ https://www.cdprojekt.com/en/wp-content/uploads-en/2016/06/gog_com.zip

PR #1163, opened 2019-01-08
Add GOG icon

ericcornelissen:
**Issue:** #1162 

### Checklist

### Description
The name is written as "GOG" as this is how it appears in their website title and on their social media (even though the logo is clearly lowercase), alternatively we could go with "GOG.com".

The color `#86328A` was picked from the website (see ⬇️, no official source for the color was found) because GOG clearly has a purple aesthetic, although the logos found at the source all use `#333333`.

color on website

### Preview
preview


_data/simple-icons.json
```diff
@@ -840,6 +840,11 @@
             "hex": "A22430",
             "source": "https://www.gnu.org/graphics/social.html"
         },
+        {
+            "title": "GOG",
```
birjolaxew: Seems like they prefer "GOG.com" when referring to their website (which is what the logo is for). I haven't seen them use plain "GOG".

ericcornelissen: Updated (also in the SVG `<title>`) 👍

birjolaxew:
:+1: Looking good! Thanks for the work!

Issue #1089, opened 2018-11-29
SimpleIcons.org search is slow


davidklebanoff:
The Simple Icons search feature is slow. It performs search queries immediately after each character is typed, causing performance issues. It needs to be [debounced](https://levelup.gitconnected.com/debounce-in-javascript-improve-your-applications-performance-5b01855e086).

ericcornelissen:
Actually the search is debounced, see ⬇️, increasing the wait time (currently 50 millisecond) will help at least a bit.

https://github.com/simple-icons/simple-icons/blob/ef9d3b4fb5f4783fd1572426509aea1107877640/site_script.js#L175

Furthermore, I will look into optimizing the search function in other ways, currently the search function (potentially) goes over all icons 4x 😅

PR #1170, opened 2019-01-16
Improve search performance

ericcornelissen:
**Issue:** #1089

### Description
Improve performance of search on website

By reducing the number of times a search query loops over all icons or
all icons that match the search query (from 3 to 1 and 2 to 0
respectively) the search on [simpleicons.org](https://simpleicons.org/) should now be noticeably
faster (if it was slow before). Especially in conjunction with #1169.

This was achieved by utilizing CSS (which operates faster then plain
JavaScript) to do most of the heavy lifting. A class on `<body>` now
dictates the order of the icons rather then a JavaScript for-loop (found
in the removed function `orderIcons`).

It was also achieved by more cleverly utilizing the fuzzy search logic
and recognizing that the CSS `order` property also works if numbers are
not consecutive. Less relevant icons will have a higher value assigned
by fuzzy search, which is what the `order`-value of the icon will
become, removing the need for sorting and an additional loop.

Some resources that helped me:
- https://css-tricks.com/css-attr-function-got-nothin-custom-properties/
- https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables
  - Specifically the "Values in JavaScript" section
- https://caniuse.com/#search=css%20var

* * *

PS. there are a few minor unrelated changes (specifically [this](https://github.com/simple-icons/simple-icons/pull/1170/files#diff-50f66c9d98e3555ff8ba6eba727f2e5cL269), but more), my apologies


birjolaxew:
:+1: Interesting use of CSS variables!

PR #1195, opened 2019-02-06
Update Dependabot

feelepxyz:
### Checklist

### Description
We've just update the Dependabot logo so created a custom export following the guidelines and used our official logo colour.


davidklebanoff:
Hi @feelepxyz, It's nice to have an official contribution!

Visually it looks good! However, it seems there are some items that need correcting:

- All fills and related attributes should be removed.
- It needs a `role` attribute.
- The SVG should be minified to a single line.

It's possible I may have missed other minor issues as well. Can you please make the above corrections and also verify that the icon meets all the [guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md)? Thanks!

feelepxyz:
@davidklebanoff ah yes clearly didn't read those guidelines properly, should be all fixed up now!

Thanks!

ericcornelissen:
@feelepxyz, although not listed in the Contributing Guidelines the `fill-rule` attribute should also be removed*. Unfortunately, often times when this is removed the icon no longer works and that is also the case here. If you have access to SVG editing software this can be solved, but I cannot give you a step-by-step way of fixing it.

* If I remember correctly we want to avoid using `fill-rule` because it is not always rendered correctly.

feelepxyz:
@ericcornelissen awesome! Think I fixed this.

FYI: Was using Figma which always seems to export using fill/clip-rule if the shape has been created there. Re-exported using Sketch which seems to have fixed the issue. This new version from Sketch also works if imported in Figma (maybe useful for someone else in the future).

Always nice to learn something new 💯

davidklebanoff:
Looks great! Thanks for getting it fixed up so quickly and of course thanks for the contribution!

Issue #1193, opened 2019-02-05
Tidal icon


Proudspark:
Checked

**Name:**

Tidal

**Website:**

http://tidal.com

**Official resources for icon and color:**
  

https://en.wikipedia.org/wiki/File:Tidal_(service)_logo.svg
- looks like black on white is best option

davidklebanoff:
There is an official SVG on the bottom right of this page:
https://listen.tidal.com/login

(we like to pull SVGs from official sources)

ericcornelissen:
The Tidal icon will be available with the next release thanks to @davidklebanoff 💯

PR #1202, opened 2019-02-07
Add Tidal

davidklebanoff:
**Issue:** #1193 

### Checklist

Icon pulled from the official website. Specifically from the bottom of the [login page](https://listen.tidal.com/login).


davidklebanoff:
I could be persuaded either way. 🤷‍♂️

Thinking out loud here... In all official uses I have seen the icon is always white and the background is always black. So does this mean the brand is white or black?

Typically when we see white icons, it's because they are using the "light version" on top of a brand color (e.g. facebook). So one could make that argument using the same logic that the brand color is black.

But, if the logo was all black and sitting atop a white page, we would also assume the brand is black. 🤔

ericcornelissen:
> In all official uses I have seen the icon is always white and the background is always black. [...] Typically when we see white icons, it's because they are using the "light version" on top of a brand color (e.g. facebook). So one could make that argument using the same logic that the brand color is black.

That I agree with. The main reason I think the brand color in this case should be black (or dark gray) is because the entire website is black(/dark).


> And, if the logo was all black and sitting atop a white page, we would also assume the brand is black.

That is a fair point and a "bias" I hadn't noticed myself yet 🤔 Although, personally, I feel like having white as a brand color is quite rare and shouldn't be selected unless it is abundantly clear it is. Mainly because white is the default background color for both the web and paper.

davidklebanoff:
Updated. I chose black instead of a dark gray because it was consistently used across their homepage, social media profiles, and favicon.

ericcornelissen:
Thanks for the work on this @davidklebanoff 👌

PR #1203, opened 2019-02-09
update svgo invocation

mr-c:
`svgo`'s syntax has changed


ericcornelissen:
Thanks for the fix @mr-c 👍 I usually use the web version so I wouldn't have spotted this issue...

PR #1205, opened 2019-02-09
Update dependencies and license information

ericcornelissen:
**Issue:** n/a

### Description
- Update the `jest` and `svglint` dependencies.
- Update the "license" in `package.json` to be in the SPDX format.
- Update the version number in `package-lock.json`. Which should be done in subsequent releases to avoid issues as seen in #1185.


davidklebanoff:
👍

Issue #1098, opened 2018-12-04
Request: DocuSign


mmallis87:
**Name:** DocuSign
**Website:** https://www.docusign.com
**Official resources for icon and color:**
  
SVG: https://developers.docusign.com/img/Logo-DS-White-Bg.svg
Color: HEX 005DB9

davidklebanoff:
It's worth noting that their official favicon is the arrow logo, which I think may be a better fit:
https://www.docusign.com/sites/all/themes/custom/docusign/favicons/mstile-310x310.png

mmallis87:
@davidklebanoff I agree with you that the favicon is most suitable for the use case of simple-icons. Let's start with that and once the favicon is added to simple-icons then I will share it with the whole DocuSign product development team and get it updated in case I get either a strong resistance or better suggestions.

davidklebanoff:
@mmallis87 - There is also the [Docusign Github page](https://github.com/docusign) which has a slightly more stylized version with arrow inside the position "marker". That may be an option as well.

It sounds like you are either part of the Docusign team or very close. Any official SVGs of the arrow icon you can supply would be much appreciated. I've only seen PNGs at a cursory glance.

davidklebanoff:
@ericcornelissen - I understand the concern, but it's still _their_ brand icon... Having used their product, it is referenced a number of times in the product and I do associate it with their brand, but the yellow and red coloring definitely plays a role in that and without the colors it certainly does become more generic.

This does bring up an interesting philosophical debate for our project. Ultimately I feel like it's the brand's prerogative to make the icon capture their brand's message as they see fit and I'd hesitate for us to pass "judgment" on if the icon is up to our "standards" (e.g. too generic).

For a thought experiment, if a popular brand's icon was simply a solid square, would we reject it? I'm currently thinking no... b/c despite it being generic we are still providing users with additional, usable data about the brand: "This is the brand's icon and this is the brand's color."

I'd love to people's thoughts on the matter. Also, feel free to chime in @mmallis87.

mmallis87:
I agree with @davidklebanoff. If you take a popular brand like Verizon for instance and open their icon SVG it's nothing but a check mark!

Like you said it's the combination of the icon with the name and the color that makes the magic of "simple" icons 😄

ericcornelissen:
[...]

PR #1223, opened 2019-02-19
Add DocuSign.

davidklebanoff:
**Issue:** #1098

### Checklist

### Description
Color and icon supplied by a DocuSign employee in #1098.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -576,6 +576,11 @@
             "source": "https://www.docker.com"
         },
         {
+            "title": "DocuSign",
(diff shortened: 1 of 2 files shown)


davidklebanoff:
I know it's a somewhat generic looking icon - I state my justification for inclusion in the original issue #1098.

birjolaxew:
:+1: I agree that, although generic, it *is* the icon of the brand. In that sense it is similar to (but perhaps more extremely generic than) Node, Garmin, Buffer and Flickr, all of which we already have icons for.

Issue #1146, opened 2018-12-26
Request: Pivotal Tracker


bergren2:
**Name:** Pivotal Tracker
**Website:** https://www.pivotaltracker.com/
**Official resources for icon and color:** https://www.pivotaltracker.com/branding-guidelines

PR #1230, opened 2019-02-22
Added Pivotal Tracker.

davidklebanoff:
**Issue:** #1146 

### Checklist

### Description

Icon and color are from an SVG icon found in the [official media kit](https://www.pivotaltracker.com/branding-guidelines). The brand has two colors `#517A9E` and `#DE752C`. I ended up using `#517A9E` because it's more prominently featured on their website, but I don't think there is a right or wrong answer.

It's also worth noting that the SVG logo in their website header does not use the exact same colors as the logos in the media kit.


birjolaxew:
:+1: Nice! I agree with all the decisions you've made here (also worth pointing out that judging by the screenshots found on their landing page, their product features blue header/highlights, and barely any orange comparatively).

PR #1244, opened 2019-02-22
Add CEVO icon

KovuTheHusky:
### Checklist

### Description

The brand guide only has rasters available so I used image trace in Adobe Illustrator and tweaked some of the anchors to keep the edges perfectly straight. I picked the color from their logo. Their Alexa rank is ~197K.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -336,6 +336,11 @@
             "source": "http://supertop.co/castro/press/"
         },
         {
+            "title": "CEVO",
+            "hex": "1EABE2",
+            "source": "https://cevo.com/event/cs-globaloffensive/downloads/"
+        },
+        {
             "title": "Circle",
             "hex": "8669AE",
             "source": "https://www.circle.com/"
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -338,7 +338,7 @@
         {
             "title": "CEVO",
             "hex": "1EABE2",
-            "source": "https://cevo.com/event/cs-globaloffensive/downloads/"
+            "source": "https://cevo.com/"
         },
         {
             "title": "Circle",
--- a/icons/cevo.svg
+++ b/icons/cevo.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" role="img" viewBox="0 0 24 24"><title>CEVO icon</title><path d="M3.5 6h12l6 10.3L24 12 18 1.6H6L3.5 6zm-.9 1.5L0 12l6 10.4h12l2.6-4.5-2.5-4.3-2.6 4.5h-7L5 12l2.6-4.5h-5z"/></svg>


birjolaxew:
To clarify @ericcornelissen's comment, the main issue is that Illustrator's vectorization function made the corners rounded, where-as the actual icon has sharp corners. Other than that it looks pretty similar.

KovuTheHusky:
> Thank you for taking the effort to manually vectorize the logo, however I don't think it is true to the real logo. If you visit their website (by the link you provided as source) you can find the logomark in the header encoded as a font icon. To get a good look at it you can set the `font-size` of the icon to a (very) large value.
> 
> Also, the fact that it is part of a font means that we can somehow extract an SVG from it, either through an SVG font or indirectly from the `.woff` (or whatever format). I have no experience with this myself thus far so I cannot help you with that unfortunately.
> 
> That logomark also gives us a more appropriate brand color (I assume you extracted the color from a rasterized image?), namely `#1EABE2` instead of `#1CB3E5`.
> 
> Can you update the SVG and the color? @KovuTheHusky ?

I did notice their logo in the header was based on a custom font, but I'm not very familiar with how fonts and vectors relate to each other so I had no idea to even pursue that route. The color was indeed sampled from their press kit. Odd that they use a different color. I've updated the logo and color as asked. Hopefully all is good! Thanks!

> To clarify @ericcornelissen's comment, the main issue is that Illustrator's vectorization function made the corners rounded, where-as the actual icon has sharp corners. Other than that it looks pretty good.
> 
> [Edit] I converted the font file from CEVO to an SVG. The logo is extracted as
> 
> ```
> <path d="M1042 149l-120 211l-170 295h-1l-1 1h-579l120 208h582l291 -504zM998 73l-125 -217h-582l-291 504l127 219h241l-126 -219l170 -295h340l125 217z" />
> ```
> 
> This is somewhere far off on the x-axis, so you might have to hunt for it a bit if you open it in Illustrator ;)

Thank you very much! If you don't mind me asking, what software did you use to extract a vector from a font? Thanks again!

KovuTheHusky:
Actually, the extracted vector from the font appears to be rotated. Let's see if I can fix that.

birjolaxew:
@KovuTheHusky I just grabbed the woff2 file from their CSS (it was encoded as a data URL, which I threw into an `<a href="data-url-goes-here">Clicky</a>`), and ran it through a [woff2 to svg converter](https://everythingfonts.com/woff2-to-svg). This gave ~400 glyphs, but by checking their CSS it's clear that `\f029` is their logo mark. Copy/paste the `d="..."` from that glyph, wrap it in a `<path>` and let the friendly contributor do his thing on it.

As you've noticed the extracted SVG was rotated at some point. I'll merge once you've gotten it right side up.

KovuTheHusky:
I had no problem getting the woff itself, but I didn't realize dumping the glyphs to svg was that simple. Anyway, should be all set now. Thank you for your patience!

birjolaxew:
:+1: Thanks for the work @KovuTheHusky ! It is much appreciated

ericcornelissen:
> The color was indeed sampled from their press kit. Odd that they use a different color.

For your information @KovuTheHusky, rasterized images typically have different colors from brand guidelines, most likely due to compression/optimization.

Issue #1237, opened 2019-02-22
Request: RubyGems


ericcornelissen:
**Name:** RubyGems.org
**Website:** https://rubygems.org/
**Official resources for icon and color:** https://rubygems.org/pages/about (at the bottom)
  

RubyGems.org is the primary package manager for the Ruby programming language ([that icon is already available](http://simpleicons.org/?q=ruby)).

davidklebanoff:
Added. Thanks @ericcornelissen!

PR #1247, opened 2019-02-27
Add RubyGems icon

ericcornelissen:
**Issue:** #1237

### Checklist

### Description
he SVG can be downloaded at the bottom of [the specified source](https://rubygems.org/pages/about), the color is taken from the `background-color` property on the `<body>` of [the homepage](https://rubygems.org/).

It could be argued that their official name is "RubyGems.org" rather than "RubyGems", I'm open to changing that if preferred 👍

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1986,6 +1986,11 @@
             "source": "https://www.ruby-lang.org/en/about/logo/"
         },
         {
+            "title": "RubyGems",
+            "hex": "E9573F",
+            "source": "https://rubygems.org/pages/about"
+        },
+        {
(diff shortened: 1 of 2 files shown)


davidklebanoff:
Looks good to me.

After investigating, it think the software/product itself is referred to as RubyGems and the repository/website that hosts them is RubyGems.org. I think the logo is intended to reference both and thus keeping the name simply `RubyGems` seems fine to me.

I wasn't able to identify any official color hexes, so I think the one chosen is fine. The red PNG logo in the logos.zip is a slightly different color red, but since it's in a lossy format, I prefer your approach anyway.

Issue #1180, opened 2019-01-27
Request: WebComponents


alxpez:
**Name:** webcomponents
**Website:** https://www.webcomponents.org/
**Official resources for icon and color:** https://www.webcomponents.org/

ericcornelissen:
The icon or WebComponents has been added and will be available with the next release 💯

PR #1253, opened 2019-03-02
Add Web Components

davidklebanoff:
**Issue:** #1180 

### Checklist

### Description

Everything about this particular pull request was non-ideal. I'll walk you through my decisions, but I don't think there are good answers for any of them.

- **Name** - I chose `Web Components`, but one could are just as easily that the name should be `WebComponents.org` or `webcomponents.org`
- **Icon** - The Web Components logo is largely composed of gradients, which doesn't translate well to a clear monochrome icon when everything was merged together (I tried). It has weird little cutouts that I don't think represent the brand well. I ended up extracting an icon from the [homepage](https://www.webcomponents.org/) that doesn't have the cutouts, but otherwise looks like the logo.
- **Color** - I wasn't able to identify an official brand color, so I choose the dark, almost black color, in the logo as it was the largest portion of the logo. Additionally, the other colors (blue and green) were gradients without a clearly identifiable color. The color I used was extracted from the official SVG logo.


birjolaxew:
> - **Name** - I chose Web Components, but one could are just as easily that the name should be WebComponents.org or webcomponents.org

Looks like [they refer to themselves as either "webcomponents.org" or "WebComponents.org"](https://www.webcomponents.org/about), but most often full lowercase. The [GitHub organization](https://github.com/webcomponents) is named "Web Components", but that's just about the only place I could find it without .org. I'd personally prefer using the lowercase .org variant.

> - **Icon** - [...] It has weird little cutouts that I don't think represent the brand well. I ended up extracting an icon from the homepage that doesn't have the cutouts, but otherwise looks like the logo.

I personally think the folding ribbon is an important part of the logo, and removing it would make it more difficult to associate.

<p align="center"></p>

There definitely is better ways to monochromize a folding ribbon, but I'd rather stick to the official sources.

<sup>(on a sidenote I couldn't find the version you talked about without the cutouts)</sup>

davidklebanoff:
Updated! I also corrected the official SVG's tiny gap you see in the bottom left.

ericcornelissen:
Looks good in the end @davidklebanoff 👌 I agree with your final choice of color as their website uses more blue than green.

Issue #1073, opened 2018-11-19
Riot.im


hThoreau:
**Name: Riot**
**Website: riot.im**

ericcornelissen:
For anyone willing to pick this up, [this repository](https://github.com/vector-im/logos) is probably the best source. Thank you for the request @hThoreau 

_Riot.im currently has an [Alexa ranking](https://www.alexa.com/siteinfo/riot.im) of ~80.000_

ericcornelissen:
The Riot.im icon has been added and will be available with the next release thanks to @davidklebanoff  💯

PR #1264, opened 2019-03-05
Add Riot

davidklebanoff:
**Issue:** #1073 

### Checklist

### Description
The logo is from their official [GitHub Repository](https://github.com/vector-im/logos/tree/master/riot), and is consistent with their black and white version.

The color was extracted from the SVG logo on the homepage. It’s the dark portions of the logo and it’s also used throughout the website as a text color for links and some headers. `#368BD6`


davidklebanoff:
Hmm, I would disagree. Hear me out.

The different colored/striped logos that you are linking to are part of the "dynamic logos" that were added in order to generate an animated version of the logo seen here: https://github.com/vector-im/logos/blob/master/riot/riot-im-logo-animated-480px.gif

I think it's just a fun representation of the logo (throwback to the old MTV animated logo?), but not necessarily the "official" version by which we should base our icon. The monochromatized version of the logo I am attempting to add is based on their "official b&w riot logo" (that's the git comment anyway) [here](https://github.com/vector-im/logos/blob/master/riot/riot-im-logo-bw.svg). I feel that since it's an official monochromatized version and more similar to their homepage logo than a simple silhouette, it seems like a better choice.

Another option is to monochromatize their homepage logo... but part of me wonders why not simply accept one they have already outlined as official vs creating another variant.

ericcornelissen:
Fair enough, maybe I was a bit too quick in my analysis. Anyways, thanks for adding this @davidklebanoff! 🎉

PR #1270, opened 2019-03-07
Removed labels from Contribution Guidelines

davidklebanoff:
Back in 2017, I added a table linking to and describing our labels in the contribution guidelines. Since then, it seems GitHub had a similar thought and on 2/22/18 they added support for descriptions on the  [labels page](https://github.com/simple-icons/simple-icons/labels), rendering our table redundant. As a result, I added our existing descriptions to the labels page and created this PR to remove the duplicate label descriptions from the Contribution Guidelines.

We still need to add descriptions to some of the previously "unlisted" labels such as `invalid`, but I wasn't confident in my ability to do that since I didn't add them and didn't want to misrepresent their intended purpose.

Additionally, we probably should update the wording on some of them seeing as they are now intended to be shared between issues and PRs.

Github Blog introducing the feature:
https://github.blog/2018-02-22-label-improvements-emoji-descriptions-and-more/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,23 +147,6 @@ Here is the object for The Movie Database as an example:
 
 Once you've completed the previous steps, create a pull request to merge your edits into the *develop* branch.
 
-## Labeling Issues
-
-We use several labels to help organize and identify issues. You can find all labels [here](https://github.com/simple-icons/simple-icons/labels). Here's what they represent and how we use them:
-
-| Label Name | Description |
-| :---- | :---- |
-| [new icon](https://github.com/simple-icons/simple-icons/labels/new%20icon) | Issues for adding a new icon. |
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Good points all around. I'll have a look through the existing labels

ericcornelissen:
Nice change, I like it 👍 


> Additionally, we probably should update the wording on some of them seeing as they are now intended to be shared between issues and PRs.

I think @birjolaxew already did this? The descriptions look good to me now 👍 


> I'm considering removing the invalid label. Although it has the valid use of indicating PRs/issues that do not follow our project guidelines, it seems a tad hostile to me. It's never fun to have "invalid" slapped on your issue/PR, and I think we are better off adding an "out of scope" label for issues and simply waiting for PRs to be considered abandoned.

I think what you say makes sense, but I would (indeed) not remove the label altogether as it definitely serves a useful purpose. Renaming it to "out of scope" seems like the perfect solution 👌

Issue #1286, opened 2019-03-13
ReasonML


weswhite:
Name: ReasonML
Website: https://reasonml.github.io/

This is the latest and greatest from Facebook akin to React. Get on it! :)

davidklebanoff:
The "RE" icon can be easily extracted from https://reasonml.github.io/img/reason.svg

weswhite:
I can give it a stab @davidklebanoff

PR #1288, opened 2019-03-13
Adding Reason icon

weswhite:
**Issue:**
#1286

### Checklist

### Description
Clipped the RE section of the Reason logo


ericcornelissen:
Thanks for picking this one up @weswhite! However, it seems the icon is not properly displayed ⬇️ I suspect this is due to the change in 7df50d1 🤔 Note that if you edit the `viewBox`, you also have to update the actual dimensions of the SVG path. (also note that the problem might be something else...)

If you need help with this, let me know what SVG editor you're using and I'll try to guide you through the process 😄 

image

weswhite:
@ericcornelissen ! Thanks! This was more involved than i thought it would be 😅. Ill google how to properly set viewbox with Inkscape. If I have any issues with it I will reach out. Looks like I can adjust that i properties.

weswhite:
@ericcornelissen ok i tried. I need help. I am using Inkscape and i am struggling

ericcornelissen:
Alright, no problem 😄 I'm also new to Inkscape but I think I will be able to help you.

If you open the SVG from 3fdb9b2 in Inskape, the first thing I recommend is zooming out so you can see what is going on. To do this you can select the zoom tool (<kbd>F3</kbd> to select it, and <kbd>-</kbd> to zoom out).

Then you can change the dimensions of the SVG using the Select and Transform tool (<kbd>F1</kbd>) from the toolbar in the top (see ⬇️). Select the graphic first and change the width and height to 24 so that it matches the viewBox, you can use the little lock you see in the screenshot to keep the height-to-width ratio the same.

image

Hope that helps, let me know if you're still stuck!

Issue #1287, opened 2019-03-13
Icon Update: Spotlight


davidklebanoff:
**Name:** Spotlight
**Official resources for icon and color:**
The [homepage](https://www.spotlight.com/) logos are SVGs.

The current Spotlight SVG has an artifact on the 'T' and should be replaced with a higher quality version.

PR #1297, opened 2019-03-17
Fixed icon artifacts in Spotlight

davidklebanoff:
**Issue:** #1287

### Checklist

### Description

Fixed artifacts in the icon by redo-ing it. Pulled the icon is from their [homepage](https://www.spotlight.com/
) SVG logo.


ericcornelissen:
Thanks for fixing this :ok_hand: 

it seems that at some point the icon was optimized to much (i.e. to low precision) but the artefacts where not noticeable on the scale GitHub usually displays the icon :thinking:

PR #1301, opened 2019-03-19
Added Stadia icon

metaa:
# Added Icon
**Stadia**, a new gaming platform Google [just announced at GDC](https://youtu.be/nUih5C5rOrA?t=1562).

## Website
https://stadia.com/
https://stadia.dev/

## Color
`#CD2640`

Now this was something I was not sure about.
Stadia has two main colors:
* `#FF4C1D`
* `#9B0063` (`stadia.dev` also seems to use that as the hover color on buttons)

What I did is, I just took the color inbetween those two.

## What it will look like
image


birjolaxew:
Boy that was fast.

Main issue I can see here is the popularity, which obviously hasn't gotten time to build up. Given that it's ran by Google, I'm ok with adding it regardless. I'll wait for another contributor to agree before merging in.

I'm good with your color choice. Gradients are always messy to boil down to a single color, and in this case the mid point seems like a good representation.

metaa:
> Boy that was fast.

Always at your service.



> Main issue I can see here is the popularity, which obviously hasn't gotten time to build up. Given that it's ran by Google, I'm ok with adding it regardless.

Yeah, I had that in mind as I was adding it. But since it's gotta be quite a big thing, I hope it won't be vanishing again any time soon.

> I'm good with your color choice. Gradients are always messy to boil down to a single color, and in this case the mid point seems like a good representation.

Glad to see I made the right choice.

Speaking of the color, I start seeing a new kind of favorite color pattern over at Google 🤔

 
image
(Debian. Stadia runs on Debian and Google has been supporting Debian for a while now, even basing their internal OS they use on Debian)

ericcornelissen:
> Boy that was fast.

I'm not even done watching the announcement yet 😅 


> Main issue I can see here is the popularity, which obviously hasn't gotten time to build up. Given that it's ran by Google, I'm ok with adding it regardless. That being said, Google do have a tendency to kill off unpopular projects, which is a slight fear here. I'll wait for another maintainer to agree before merging in.

I share your concern, though usually those kind of project don't get their own announcement session at a conference like GDC. Also, they already have a lot of buy-in from major gaming companies who benefit greatly from the increase of customers.

birjolaxew:
:+1: Thanks for your work on this @metaa !

Issue #1273, opened 2019-03-07
Add CiviCRM logo


sluc23:
Please Add CiviCRM logo
https://civicrm.org/trademark

davidklebanoff:
Alexa Ranking: ~230k

ericcornelissen:
The SVG for CiviCRM has been added in #1313 and will be available in the next release 😄

sluc23:
thanks @ericcornelissen

PR #1313, opened 2019-03-25
Add CiviCRM icon

birjolaxew:
**Issue:** #1273

### Checklist

### Description

Color is toss up between blue `#086287` and green `#81C459`
Green was chosen because it seemed dominant to my eyes


davidklebanoff:
Icon looks good to me! As for color, I would have chosen the blue because it's featured both in the wordmark and icon itself. 🤷‍♂️

Not that there is a "right" answer.

birjolaxew:
@davidklebanoff True. The thought went through my head, but I eventually disregarded it because the two blues are different, and the blue text didn't really seem very impactful. But then again I'm slightly colorblind so ¯\\\_(ツ)_/¯

ericcornelissen:
judging from their website and social media I feel like we could go with either. I would personally pick green in this case, but as @davidklebanoff said there does not seem to be a "right" answer. So I'll just go ahead and merge this, thanks @birjolaxew 👍

Issue #1257, opened 2019-03-03
Request: SpaceX


timisenman:
**Name:** SpaceX
**Website:** SpaceX.com
**Official resources for icon and color:**
  
https://www.spacex.com/media

ericcornelissen:
Not sure, but I think the logo should be just the 'X' rather than the full wordmark 'SpaceX' 🤔

birjolaxew:
SpaceX are a bit interesting. They use the full "SpaceX" wordmark on their rockets, on their website, and most places where they are referred to (e.g. Wikipedia). They use the "X" logomark as an avatar (for Twitter, YouTube, etc.), in their favicon and other places where the available space is circular/square (e.g. on their landing pad). They seem to identify themselves equally by either (e.g. they [have](https://shop.spacex.com/mens/t-shirts/spacex-circle-graphic-t-shirt.html) [both](https://shop.spacex.com/mens/t-shirts/spacex-front-logo-t-shirt.html) in their shop)

Given our format I think that the "X" logomark makes most sense. The full wordmark would obviously have a stronger relation to the brand, but I agree that the logomark should be used (similar to our icon for Netflix)

davidklebanoff:
I'm not finding any official SVGs on their website. If anyone can locate one that would be helpful.

metaa:
> I'm not finding any official SVGs on their website. If anyone can locate one that would be helpful.

Whoop-de-doo, [there's something](https://www.spacex.com/sites/all/themes/spacex2012/images/logo.svg) in my shoe.

Issue #477, opened 2017-06-24
Align size/center of mass


dy:
Hey @danleech , awesome project, I use it in a couple of websites.

But I noticed a small problem - some of the icons are a bit displaced or have lesser visual mass, as in the example for my friend's site:

![image](https://user-images.githubusercontent.com/300067/27509789-747d435c-58d2-11e7-8ed7-3f46505ab7b0.png)

G+ and twitter look smaller than others.

Recently I created [optical-properties](https://github.com/dfcreative/optical-properties) tool for balancing alignment/mass of graphical symbols.

![image](https://user-images.githubusercontent.com/300067/27509824-05bccffe-58d3-11e7-9813-10d2bbd2744f.png)

It worked really well in [font-atlas-sdf](https://github.com/dfcreative/font-atlas-sdf/blob/master/index.js#L87).

It could be used to align the whole icon set - making sure the visual center and mass of icons is same.

What are your thoughs?

Cheers!

birjolaxew:
Started implementing a test at [birjolaxew/simple-icons/optical-sizing](https://github.com/birjolaxew/simple-icons/tree/feature/optical-sizing), but there are some issues:

- `optical-properties` seems to return different radius' if run multiple times. I still haven't tracked down why.
- The difference in radius returned by `optical-properties` between G+ and Instagram is minimal (308.13 vs. 306.79 at 512px size)

birjolaxew:
Pinging @danleech for opinion on above solution.

dfcreative:
@birjolaxew ok, I will try to tackle these issues down this weekend.

* [ ] make work in nodejs
* [ ] take alpha/custom channel
* [ ] concave height before detecting radius

davidklebanoff:
Any progress @dfcreative? This sounds very useful.

styfle:
> have lesser visual mass

I noticed this myself. On [my website](https://www.ceriously.com/), I used the StackOverflow and Keybase icons from `simple-icons` and they look a little thinner than the other icons.

PeterShaggyNoble:
Closing this out due to inactivity. If anyone would like to continue working on this then please feel free to open a new issue or PR.

PR #1314, opened 2019-03-25
Add SpaceX

birjolaxew:
**Issue:** #1257

### Checklist

### Description

Big thanks to @metaa for finding the official SVG.

The "X" logomark is chosen over the full wordmark for reasons explained in #1257 
Now that I see it, I am starting to doubt that decision though. This is mostly due to the very off-center nature of the logomark, which doesn't mesh well with how Simple Icons is usually used. Two ways to solve this: use a version cropped to a square (like SpaceX use in their favicon) or use the full wordmark (my preferred solution) - or just use the logomark as in this PR.

<p align="center"><br /><br/><br/></p>


davidklebanoff:
I think you chose the correct icon for the PR. It's consistent with their [Instagram](https://www.instagram.com/spacex/), [Twitter](https://twitter.com/spacex), [LinkedIn](https://www.linkedin.com/company/spacex/), etc. and I think keeps in line with the rest of Simple Icons.

It is a little odd in its positioning, but so are a lot of other icons. It's a known issue and this is where https://github.com/simple-icons/simple-icons/issues/477 comes into play. But even after #477 it may be off, as their brand seems to consider the proper center of the icon to be the middle of the X.

As for the color, I don't think white is correct. If one were to base the decision on their current website and social media, and followed the discussion in #1202, it should be black. However, as someone who has been following the company for a while and watched a bunch of launches, I would argue their brand color may actually be blue (unless it's changed recently). The logo on their rockets, capsule, and launch facility are all blue. So I'm not sure what it should be.

metaa:
> As for the color, I don't think white is correct. [...] it should be black. However, [...] the logo on their rockets, capsule, and launch facility are all blue. So I'm not sure what it should be.

I just dug out this [SVG on Wikipedia](https://upload.wikimedia.org/wikipedia/commons/d/de/SpaceX-Logo.svg) (could've considered checking Wikipedia first 😅) and that one is blue aswell, though the X also has some gray.

ericcornelissen:
Not sure how official this resource is (at least it matches the SVG on Wikipedia) but [SchemeColor.com](https://www.schemecolor.com/spacex-logo-colors.php) also lists `#005288`, at least for the _SpaceX Logo_ (they also have an [entry for the SpaceX website](https://www.schemecolor.com/spacex-website.php)).

Another resource I found is [this pdf](http://stotion.com/spacex/SpaceX-logo-corrections.pdf) which also uses blue (doesn't specify a hex though...), but again I'm not sure how official it is.

metaa:
The PDF uses the same blue by the way.

ericcornelissen:
Since I don't think there is going to be response from Elon anytime soon... Do we stay with the current color or switching to blue (probably `#005288`) @birjolaxew / @davidklebanoff ?

metaa:
I'd use the blue since their logos use blue.

birjolaxew:
I've updated the color to blue. I'll let another maintainer merge this in.

ericcornelissen:
Thanks for your work on this way back @birjolaxew 👍

PR #1315, opened 2019-03-25
Add Itch.io

schweller:
**Issue:** n/a

### Checklist

### Description
The icon is from Itch.io's [official press kit](https://itch.io/press-kit). The color was extracted from the press kit SVG: `#000000`


ericcornelissen:
Looking at [the color palette in the press-kit](https://itch.io/press-kit#color-palette) I would expect the color to be `#fa5c5c` (note that the color should reflect the brand rather than the logo itself). 

Would you mind changing it or explain why you went with `#000000`? I'm not too familiar with itch.io so you might be able to convince me that's better 😉 

Also, I went ahead and quickly changed the `<path d="..."></path>` into `<path d="..."/>`. Other than that this is a great Pull Request!

schweller:
@ericcornelissen The reason I did it is that when you download their press kit, SVG has a color `#000000`. That being said, I agree we should expect `#fa5c5c` and will make the change. Appreciate the feedback!

ericcornelissen:
Thanks for working on this @schweller, and thanks for the response on the color. Your contribution is much appreciated! 🎉 

With this PR merged the icon will be available in the next version (which should be coming up rather quickly as there have been quite a few new icons since the [last release](https://github.com/simple-icons/simple-icons/pull/1305)) 😄

metaa:
This introduced some ugly whitespace after the comma in the JSON file 😢

ericcornelissen:
Oops :sweat_smile:

davidklebanoff:
It's not a big deal, but we might as well take the opportunity to prevent it from happening again. I created issue #1324 to track it.

Issue #1309, opened 2019-03-22
Icon Request: Chase


etiennecoutant:
**Name: Chase**
**Website: https://www.chase.com/**
**Official resources for icon and color:**

birjolaxew:
For anyone who wants to contribute this, the official SVG can be extracted from the icon font used on their official website. It becomes (viewBox approximate, path is official):

<p align="center"><img src="https://svgshare.com/i/Bwe.svg" height="128px" /></p>

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" viewBox="0 0 1000 1000">
    <path d="M338 901q-7 0 -12.5 -2.5t-10.5 -6.5t-7 -10t-3 -13l1 -224h592l-269 256h-292h1zM921 577q-1 7 -3 12.5t-7 10t-10.5 7t-12.5 2.5h-224l-1 -593l257 270v291h1zM596 -5q6 0 12.5 2.5t10 6.5t6.5 10t2 13l1 224h-594l271 -256h290h1zM14 319q-1 -7 2 -12.5t7 -10t10 -7 t12 -2.5h225l-1 593l-255 -270l-1 -291h1z"/>
</svg>
```

ericcornelissen:
The icon for Chase has been added by @metaa and will be available in the next release 👌

PR #1316, opened 2019-03-25
Added Chase icon.

metaa:
# Added Icon
Chase (#1309)

## Website
https://www.chase.com/

## Logo
image

## Color
`#117ACA`

## What it will look like
image


ericcornelissen:
Could you specify where you got the color @metaa? Was it perhaps [wikimedia](https://commons.wikimedia.org/wiki/File:Chase_logo_2007.svg), if so can you update the source to that URL as well?

metaa:
Yup, the original source URL must've slipped through somehow 😅

Issue #1292, opened 2019-03-16
Icon Request: Typescript


weswhite:
{
    "title": "TypeScript",
     "hex": "006FC5",
     "source": "https://github.com/remojansen/logo.ts/blob/master/ts.svg"
}

weswhite:
I have a commit ready to go for this one

metaa:
How would this go? Just the `TS` letters centered in the 24x24 viewbox?
Otherwise it would look ugly I guess

PR #1318, opened 2019-03-25
Added TypeScript icon.

metaa:
# Added Icon
TypeScript (#1292)

**NOTE:** This is a *community* logo, not [the *official* one](https://github.com/Microsoft/TypeScript/blob/master/doc/logo.svg). Though, it is widely used, even by Microsoft officials. Read more [here](https://github.com/remojansen/logo.ts/blob/master/README.md).

## Website
https://www.typescriptlang.org/
https://github.com/remojansen/logo.ts

## Logo
image

## Color
`#007ACC`

## What it will look like
image

---

Supersedes #1293


ericcornelissen:
I agree that this is the logo to use, note also that [the website](https://www.typescriptlang.org/) uses the same (or a similar?) version of the logo as favicon.

Either way thanks for updating #1293 @metaa 👌

PR #1317, opened 2019-03-25
Added jsDelivr icon.

metaa:
# Added Icon
jsDelivr (#1217)

## Website
https://www.jsdelivr.com/

## Logo
image

## Color
`#E84D3D`

## What it will look like
image

PR #1319, opened 2019-03-25
[Sanitary 🚽] Removed ugly whitespace after comma of Itch.io JSON block.

metaa:
#1315 contained some ugly whitespace after the comma in its JSON block.

image

It would be better to merge this single commit than all following PRs (in which code optimization would attempt to fix this) conflicting with each other from here on.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1229,7 +1229,7 @@
             "title": "Itch.io",
             "hex": "FA5C5C",
             "source": "https://itch.io/press-kit"
-        },        
+        },
         {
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Seems this was fixed in #1317, correct @metaa?

metaa:
> Seems this was fixed in #1317, correct @metaa?

Yeah, even though I wanted to prevent that it's being fixed where it's not mentioned 😅
But ok, stuff happens.

ericcornelissen:
It happens, not a big deal in this case :wink:

Issue #1276, opened 2019-03-08
Request: Creative commons


OpportunityLiu:
**Name:** Creative commons
**Website:** https://creativecommons.org/
**Official resources for icon and color:**
  
![](https://creativecommons.org/wp-content/themes/cc/images/cc.logo.white.svg)

https://creativecommons.org/wp-content/themes/cc/images/cc.logo.white.svg

ericcornelissen:
The Creative Commons icon has been added and will be available with the next release (i.e. when #1347 gets merged) 😉

PR #1322, opened 2019-03-25
Added Creative Commons icon.

metaa:
# Added Icon
Creative Commons (#1276)

## Website
https://creativecommons.org/

## Logo
image

## Color
`#EF9421` (see [their wiki](https://wiki.creativecommons.org/wiki/Colors#Color_Table) and the discussion below)

### Original color choice
~~`#049CCF`~~

This, I am actually not sure of, if that's right. They use that as their main color for links, the search button and a lot of panels **on their main website**, even though there is a lot of orange too, while the logo is usually either white or black.

[This site](https://stateof.creativecommons.org/) also has an awful lot of different light blue shades:
* `#01AED1` (**primary**? also the link hover color)
* `#01ADD1` (hero image)
* `#00C9F2` (overlay text, gauges and some quote)
* `#0194B2` (dark shade 1, link color)
* `#0197B5` (dark shade 2, )
* `#03ADD1` (thank you image all the way down)
* then further down wild petrol colors are spawning

- [X] Discussion 🤔

## What it would look like
image

**or with the original color choice:**
image


ericcornelissen:
I would personally go with orange instead of blue, but I agree it's not clear cut. If one of the other maintainers thinks blue is fine they can go ahead and merge.

If we decide to go with orange, [this](https://creativecommons.org/about/downloads) seems a pretty good source for the color (see the **Logos** for an orange SVG version of variant of the logo).

metaa:
> If we decide to go with orange, [this](https://creativecommons.org/about/downloads) seems a pretty good source for the color (see the **Logos** for an orange SVG version of variant of the logo).

[This link](https://wiki.creativecommons.org/wiki/Colors) on the page mentioned above also has blue colors, but they don't match any of those listed in the first post 😐

Screenshot (because the site seems to be very slow):
image

ericcornelissen:
> This link on the page mentioned above also has blue colors, but they don't match any of those listed in the first post 😐

It isn't even loading for me 😕 

Anyway, they are very inconsistent with their colors. I double checked in the SVG I mentioned actually uses `fill: red;` and the color listed on the page I linked is `#ABB3AC` and does not appear in the list of colors you mentioned.

I could be persuaded to go with the blue from the last source `#4374B7` (rather than a blue by @metaa, this seems at least somewhat official) or the orange from the last source `#EF9421`.

Again, _I personally associate orange with Create Commons more than blue_. In the end it doesn't matter that much, we can always update if needed 😄

birjolaxew:
I agree that orange seems to be the way to go - they seem to use it significantly more than blue.

PeterShaggyNoble:
For what it's worth [here](https://brandcolors.net/b/creative-commons) are the colours BrandColors have listed for Creative Commons.

ericcornelissen:
> For what it's worth [here](https://brandcolors.net/b/creative-commons) are the colours BrandColors have listed for Creative Commons.

Thanks, it seems those colors are coming from [the page with the Color Table](https://wiki.creativecommons.org/wiki/Colors) 🤔 

* * *

@metaa, what's your opinion, are you okay with changing to `#EF9421`?

metaa:
> > For what it's worth [here](https://brandcolors.net/b/creative-commons) are the colours BrandColors have listed for Creative Commons.
> 
> Thanks, it seems those colors are coming from [the page with the Color Table](https://wiki.creativecommons.org/wiki/Colors) 🤔

That's what I took the screenshot of above 🙂

> @metaa, what's your opinion, are you okay with changing to `#EF9421`?

Yeah, no problem. I just still don't see where they actually declare one color as the main one 🤔

ericcornelissen:
> Yeah, no problem. I just still don't see where they actually declare one color as the main one 🤔

As far as I can tell they don't and you could have changed the blue (probably to `#4374B7`) as well. From the limited discussion in this thread it seems orange is a better fit. But it can also be updated 👍 

Anyway, thanks for adding this icon @metaa 🎊

Issue #1277, opened 2019-03-08
Add Travis CI Check To Verify Number of SVG Paths


davidklebanoff:
As we make our way toward restricting icons to have a single path, I wanted to make sure we don't forget to add a Travis CI check to enforce the new rules.

This may coincide with updates to our contribution guidelines.

Related to:
- #1275 Batch optimize SVGs
- #1272 Return raw path data in NPM
- svg/svgo/pull/1084

ericcornelissen:
This can be achieved pretty quickly with [SVGLint](https://github.com/birjolaxew/svglint) by just adding `"svg > path": 1` like so:

```javascript
module.exports = {
    rules: {
        elm: {
            "svg": 1,
            "svg > title": 1,
            "svg > path": 1, // <--------------------------------------------------------
            "g": false,
        },
        attr: [
            { // ensure that the SVG elm has the appropriate attrs
                "role": "img",
                "viewBox": "0 0 24 24",
                "xmlns": "http://www.w3.org/2000/svg",

                "rule::selector": "svg",
                "rule::whitelist": true,
            },
            { // ensure that the title elm has the appropriate attr
                "rule::selector": "svg > title",
                "rule::whitelist": true,
            }
        ]
    }
};
```

PR #1275, opened 2019-03-08
Batch optimize SVGs

ericcornelissen:
Batch optimize all icons present in the repository (if applicable) using [this SVGO config (zip)](https://github.com/simple-icons/simple-icons/files/2946392/svgo-config.zip):
- In light of #1272 I compounded all icons in c3c426f (99 additions changed, see ⬇️)
- Removed newlines at the end of files and trailing spaces in SVGs in 0f536e5 (trivial changes, no review required)
- Converted shapes to `<path>` (and also compounded the `<path>`) in f711cea (6 files changed)

## Compounding `<path>`s
This is a pain to review manually, but I went through all icons and I think everything is okay except for the ones listed here. Pinging @goyney before trying to resolve this myself...

- Google Podcasts: n/a, for some reason GitHub has problems loading the icon... 🤔 (fixed in 2aa178b)
- Laravel: [before](https://github.com/simple-icons/simple-icons/blob/develop/icons/laravel.svg), [after](https://github.com/simple-icons/simple-icons/blob/f711ceab81ae7fb153434cea87e0de8b01283964/icons/laravel.svg) (Thanks to @davidklebanoff, fixed in 55960f5)
- Stackshare: [before](https://github.com/simple-icons/simple-icons/blob/develop/icons/stackshare.svg), [after](https://github.com/simple-icons/simple-icons/blob/c3c426fdc8ac97e4333e816bab8c941830d5ff46/icons/stackshare.svg) (fixed in f5e173a)
- Telegram: [before](https://github.com/simple-icons/simple-icons/blob/develop/icons/telegram.svg), [after](https://github.com/simple-icons/simple-icons/blob/c3c426fdc8ac97e4333e816bab8c941830d5ff46/icons/telegram.svg) (fixed in 2261968)
- Travis CI: [before](https://github.com/simple-icons/simple-icons/blob/develop/icons/travisci.svg), [after](https://github.com/simple-icons/simple-icons/blob/c3c426fdc8ac97e4333e816bab8c941830d5ff46/icons/travisci.svg) (fixed in ce088b1)

Issue #1272, opened 2019-03-07
Return raw path data in NPM 


goyney:
### Kind of issue 
  - [x] Improvement

### This issue concerns 
  - [x] The NPM Package

### Description

Currently, the NPM package returns a data object that looks something like:

```js
{
    title: 'Simple Icons',
    hex: '111111',
    source: 'https://simpleicons.org/',
    svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>'
}
```

It would be a welcome improvement if it also returns the raw, combined path data for the icon. Something like:
```js
{
    title: 'Simple Icons',
    hex: '111111',
    source: 'https://simpleicons.org/',
    svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>',
    path: 'M....Z'
}
```

This would allow the `simple-icons` library to be used very easily in conjunction with other libraries, such as [Material Design Icons](https://materialdesignicons.com/), who offer more generic node packages that work off of SVG path data.

For example, with the path data added, I could do something like this:

```jsx
import React from 'react';
import Icon from '@mdi/react';
import { mdiAccount } from '@mdi/js';
import google from 'simple-icons/icons/google';

function App() {
  return (
    <div>
        <Icon path={mdiAccount} />
        <Icon path={google.path} />
    </div>
  );
}
```

birjolaxew:
How would this work with [non-`<path>` icons](https://github.com/simple-icons/simple-icons/blob/develop/icons/highly.svg) or multi-`<path>` icons?

We don't currently enforce the content of the SVG beyond having a `viewBox="0 0 24 24"`, so exposing the content in a non-SVG format is going to be messy.

goyney:
You mean you don't compound your icons all down to a single path? In that specific icon example, you could shave off 30 bytes by compounding the path.

That seems like another improvement that could be made to benefit the entire library. Should I open another request to have that happen?

---

I would love to see the icons be compound down and this functionality be added. As a core contributor to the MDI project, we've starting to send people over to this library for brand icon requests as this seems to be a more appropriate library for them. It would be great if we could work together to make our tooling benefit each other. 😄

ericcornelissen:
> You mean you don't compound your icons all down to a single path? In that specific icon example, you could shave off 30 bytes by compounding the path.

[...]

PR #1335, opened 2019-03-28
Update SVG linting. Only single path SVGs allowed.

davidklebanoff:
**Issue:** #1277 

### Description
I updated the SVG linting rules to enforce the new rule that all SVGs must be a single path (no multi path SVGs and no shape based SVGs). The two new rules are:

- `"svg > path": 1` - ensures that at least one and only one `path` element may exist
- `"*": false` - ensures only the `title` and `path` elements may exist. This will catch shape elements like `circ` and `rect` in addition to `g` elements.

I also updated the Ionic icon because it had 2 paths and failed the new linting rules.

This PR is in conjunction with the efforts of #1272 and #1275.


ericcornelissen:
Works as expected, thanks for updating this :+1:

Issue #1324, opened 2019-03-25
Update or add a JSON linting step in Travis CI to check for formatting issues


davidklebanoff:
Update the [JSON lint schema](https://github.com/simple-icons/simple-icons/blob/develop/.jsonlintschema) to check for formatting issues. This can include extraneous whitespace, improper indentation, etc.

davidklebanoff:
After a little reading, it would appear that JSON schema validation doesn’t have the ability to check formatting, and the linter we use `jsonlint` currently does not support formatting checks either. So this task may have a little more depth than it initially appears (not a quick toggle or schema update).

Additionally it's worth noting that there are two versions of `jsonlint`, there’s the original [jsonlint](https://www.npmjs.com/package/jsonlint)
which appears to be [abandoned](https://github.com/zaach/jsonlint/issues/100) and there’s [jsonlint2](https://www.npmjs.com/package/jsonlint2) (which we are currently using) which is simply a fork that updates the `nomnom` dependency.

#527, #528, #529 are where json linting first came into the project.

ericcornelissen:
I've looked around a bit but there seems to be no available linter that just lints white space no matter the filetype/syntax. I've tried some things with [eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json) and [jshint-json](https://www.npmjs.com/package/jshint-json) but to no avail.

This tool, [grunt-wslint](https://github.com/bbyars/grunt-wslint) comes closest to what I think we want. Unfortunately it is a Grunt plugin 😞 I personally feel a bit weird adding a `Gruntfile.js` just for this plugin...

Also, I want to add to this that we might consider adding some kind of linting that checks if the icon SVGs do not have a trailing newline.

ericcornelissen:
Update, it seems it is possible to achieve what we want using [editorconfig-checker](https://www.npmjs.com/package/editorconfig-checker) in which case we have to add an `.editorconfig` file (read more about EditorConfig [here](https://editorconfig.org/)). Personally I prefer that over adding Grunt as a devDependency and adding a `Gruntfile.js`, and it is more convenient than building a linter from scratch, but that is of course debatable.

I'm currently considering the following EditorConfig configuration. This should correctly lint the `_date/simple-icons.json` file for trailing whitespace (and some other things). However, it does not seem to want to lint SVG files (for the trailing newline issue), but I contacted the maintainers about that.

```
root=true

[*]
charset=utf-8
indent_style=space
indent_size=2
[...]

PR #1349, opened 2019-04-05
Lint file formatting (primarily whitespace)

ericcornelissen:
**Issue:** #1324

### Description
Addresses the issue regarding formatting of files using [EditorConfig](https://editorconfig.org/) and [editorconfig-checker](https://www.npmjs.com/package/editorconfig-checker).

Some comments about my choices in the `.editorconfig`, of course this is open for discussion:
- I set the `insert_final_newline` to `true` for all files mainly for consistency. Currently this is missing from `index.html`, `.gitpod.yml`, and `scripts/utils.js`. This setting should be negated (`insert_final_newline=false`) for SVGs [once those are checked as well](https://github.com/simple-icons/simple-icons/issues/1324#issuecomment-480431538).
- The `LICENSE.md` `indent_size` is `unset` as it uses inconsistent indentation, but I feel like it is more important that it stays true to the original copy.
- The GitHub templates include some trailing whitespace in some places so that it is easier to fill out, therefor the trailing whitespace shouldn't be trimmed.


ericcornelissen:
Expected error in the build for this change is [this one](https://travis-ci.org/simple-icons/simple-icons/jobs/516800074#L500), seems to be working correctly 👌

Issue #1364, opened 2019-04-18
Rename 'Adobe Lightroom' to 'Adobe Lightroom Classic'


nexus0212:
**Current name:** Adobe Lightroom
**New name:** Adobe Lightroom Classic
**Website:** https://www.adobe.com/products/photoshop-lightroom-classic.html
**Official resources for icon and color:** https://upload.wikimedia.org/wikipedia/commons/5/56/Adobe_Photoshop_Lightroom_Classic_CC_icon.svg

---
**Related issue:** #1363
Adobe has created new application witch names '**Adobe Lightroom CC**' (more about name here: #1363). The old version '**Adobe Lightroom**' was changed to '**Adobe Lightroom Classic**'. In addiction, new 'Adobe Lightroom CC' has a different icon (with rounded corners).

![image](https://user-images.githubusercontent.com/7335471/56342230-b7149680-61b7-11e9-9738-cd7bcc061716.png)

PR #1365, opened 2019-04-18
Renamed 'Adobe Lightroom' to 'Adobe Lightroom Classic' #1364

nexus0212:
Issue: #1364


ericcornelissen:
Thanks for fixing this one @nexus0212 😄

PR #1369, opened 2019-04-23
Capitalize XD in Adobe XD

davidklebanoff:
This puts it in line with how [Adobe](https://www.adobe.com/products/xd.html) refers to the product.


ericcornelissen:
Good point, I think that's better :+1 : thanks @davidklebanoff

Issue #1160, opened 2019-01-07
Request: Linux distros


YoshiRulz:
These are the 15 most popular distros of 2018 based on the right sidebar on [distrowatch.com](https://distrowatch.com), plus SteamOS, minus Kali. And I know ReactOS isn't Linux.

* [x] Arch Linux
* ~~Antergos~~ [apparently discontinuing](https://antergos.com/blog/antergos-linux-project-ends/)
* [x] CentOS: [ArtWork on the CentOS Wiki](https://wiki.centos.org/ArtWork)
* [x] Debian
* [x] elementaryOS
* [x] Fedora: [Logo on the Fedora Project Wiki](https://fedoraproject.org/wiki/Logo)
* [x] [Linux Mint](https://linuxmint.com)
* [x] [Manjaro](https://manjaro.org/), two devs have listed themselves as artists on the [wiki](https://wiki.manjaro.org/index.php?title=About_Manjaro#Website_and_Artwork_Team)
* [x] MX Linux: [branding page](https://mxlinux.org/art)
* [x] openSUSE: [Portal:Artwork on the openSUSE Wiki](https://en.opensuse.org/Portal:Artwork)
* [x] [ReactOS](https://reactos.org/), [wiki](https://reactos.org/wiki/Steve_Cockeram) helpfully lists logo designer as MIA.
* [x] Red Hat
* [x] Solus: [branding page](https://getsol.us/branding)
* [x] SteamOS
* [x] Ubuntu
* [x] Zorin OS: [press kit](https://zorinos.com/press)

metaa:
I'd like to add RedHat here.

hylickipiotr:
I will take care of some linux distributions' logo 😄

metaa:
[Alpine Linux](https://alpinelinux.org/) would be great, too ☺️

Scotty-Trees:
Can Solus get some love btw? It's one of the best and most stable rolling distros I've ever tried.

davidklebanoff:
@Scotty-Trees - Solus was added by @metaa 2 days ago and will be available in the next release.

avylove:
Any progress on CentOS?
Can [EPEL](https://fedoraproject.org/wiki/File:Logo_name.png) be added?

ericcornelissen:
> Any progress on CentOS?

CentOS has been added and will be available in the next release :+1: 


> Can EPEL be added?

[...]

PR #1423, opened 2019-05-22
Added Fedora icon

hylickipiotr:
**Issue:** #1160 

### Checklist

### Description
I've use the oficial Fedora logo, but to display in only black-white, I've removed brighter line inside the logo, so it is simple to see is the 'f' like the Fedora. If I colorized this line white, it would be infinity sign.

Color was got from [official Fedora logo guidline](https://fedoraproject.org/wiki/Logo/UsageGuidelines) `#294172`

image


birjolaxew:
I could've sworn there was an existing discussion on the Fedora logo, but I can't find it for the life of me. Must be something I've imagined ¯\\\_(ツ)_/¯

Monochromization of this one is difficult. There are a couple different solutions to this in other icon packages (e.g. [FontAwesome uses hashing](https://fontawesome.com/icons/fedora), while [font-logos](http://lukas-w.github.io/font-logos/) use the solution also chosen in this PR). There's some good discussion on [Fedora's own issue](https://pagure.io/design/issue/620), including a suggestion that might work with the current logo (although to match the current logo this would need both ends of the "f" marked):

<p align="center">Credit to @duffy on pagure.io</p>

It should also be noted that [they have a new logo](https://pagure.io/Fedora-Council/tickets/issue/248) which was approved 2 months ago. They have not yet started using it, so we should stick with the old logo (the one in this PR) until they do.

I am ok with the monochromization this PR currently has. Will wait for another maintainer (@davidklebanoff @ericcornelissen) before merging.

ericcornelissen:
I personally feel like the approach taken by @hylickipiotr is more recognizable than the one shown in [@birjolaxew comment](https://github.com/simple-icons/simple-icons/pull/1423#issuecomment-494871074) (but I'm not super familiar with the logo), so I went ahead and merged it. 

If anyone disagrees they can continue a discussion in this thread or open a separate issue :slightly_smiling_face:

Issue #1104, opened 2018-12-06
Request: PyPI


1313e:
**Name: PyPI
**Website: https://pypi.org
**Official resources for icon and color: https://pypi.org in general or https://pypi.org/static/images/logo-small.6eef541e.svg

I tried looking for official guidelines, but although I can easily find the ones for the Python language, finding them for PyPI is much harder.

birjolaxew:
This one is going to be difficult to monochromize.

gajus:
This would be nice. It could be just an outline of the blocks.

PR #1435, opened 2019-05-25
Add PyPI icon

birjolaxew:
**Issue:** #1104

### Checklist

### Description

Icon is monochromization of their official SVG, with minor touchups
The lines unfortunately do not line up, but that is a property of the official SVG
Color is chosen from the front-facing blue of their official SVG 

This one is difficult to monochromize, especially if we wish to keep the distinction between the orange and blue colors; after trying out a few different approaches I ended up just coloring them both in (similar to how [our Python icon](https://simpleicons.org/?q=python) is monochromized)

It is unfortunate that the lines of the background blocks are so thin. They are likely to be problematic to display at small sizes. I nonetheless chose to use the official outlines.


ericcornelissen:
Thanks for adding this one @birjolaxew!


> It is unfortunate that the lines of the background blocks are so thin. They are likely to be problematic to display at small sizes. I nonetheless chose to use the official outlines.

That is indeed unfortunate. Alternatively it would be possible to have just the block representing the Python logo as SVG. But I agree that it is better to stick with the official version.

Issue #1093, opened 2018-12-03
Elsevier Tree Logo


pkeir:
Elsevier Tree Logo
https://cdn.elsevier.io/matrix/includes/img/logo-tree.svg

davidklebanoff:
Alexa Rank: [473](https://www.alexa.com/siteinfo/elsevier.com)

takingmynetback:
Hey I'm looking for my first issue to solve. Do I just do it and submit it?

hylickipiotr:
I'll take care of it.

birjolaxew:
This has been added and will be in the next release. Thanks @hylickipiotr!

PR #1448, opened 2019-05-29
Added Elsevier icon

hylickipiotr:
**Issue:** #1093 

### Checklist

### Description
I've extracted the SVG logo from the Elsevier website footer. I've also picked the color `#FF6C00` from the extracted SVG logo. 

I've correctly optimized the icon, but unfortunately its size is 138 KB.


davidklebanoff:
This brand is certainly unique in its complexity. Recognizing that, I'm not sure the tree logo is the most appropriate icon for Simple Icons. The current [website's](https://www.elsevier.com/) favicon is the stylized 'E' from the logo which may be more appropriate? With that said, their social media profile icons on [Facebook](https://www.facebook.com/ElsevierConnect), [Twitter](https://twitter.com/ElsevierConnect), and [LinkedIn](https://www.linkedin.com/company/elsevier/) all make use of the tree icon combined with the wordmark.

I'm not familiar with Elsevier, but I imagine the tree logo itself is more recognizable than the stylized 'E'. However, at the sizes that Simple Icons are often used, I'm not sure the tree would even be recognizable. That said, we don't have any sort of non-anecdotal metrics demonstrating how Simple Icons is used in the real world.

Given that the official website makes use of the 'E' icon for it's favicon, coupled with the fact that the 'E' icon functions much better at varying sizes (some of which are commonly used for buttons, social media badges, etc), my personal vote is to use the 'E' icon instead.

I would love to hear additional thoughts from @ericcornelissen and @birjolaxew.

ericcornelissen:
tl;dr I think we should use the tree logo

* * *

I'm not necessarily against using the 'E' icon, but personally I think we should stick with the tree logo. Even though the 'E' icon is more suitable for different sizes, I think the tree logo is "their preferred choice" as logo. It is not up to us to decide that.

As for official sources, I found [this site](https://design.elsevier.com/accounts/login/?next=/) that should have brand guidelines, but for some reason you need a log in. An alternative is [this case study](https://www.brandingbusiness.com/case-studies/elsevier/) which seems to have some pictures of official brand guidelines (scroll down about 3/4 of the page).

Also, there are some SVGs with a similar level of complexity (e.g. [Flask](https://simpleicons.org/?q=flask) and [GNU](https://simpleicons.org/?q=gnu)).

birjolaxew:
I agree that, although the tree logo is definitely a complex logo, it is more appropriate than the stylized "E" - at very small sizes it definitely becomes unreadable, but with the official website using it at a size of 54x48 I think it is applicable to our use case.

I'll merge this in for now. If anyone has anything to add to the discussion, feel free to keep commenting - we are open to changing the logo if there is a valid reason.

PR #1472, opened 2019-06-14
Add "path" key with raw path data to NPM package

ericcornelissen:
**Issue:** #1272

### Description
This PR updates the prepublish script to add a new attribute to the JavaScript files for NPM called "path". The path is just the string that is inside the `d` attribute of the (single, see #1275) `<path>` element of the SVG.

The PR also includes an update to the [Furry Network icon](https://github.com/simple-icons/simple-icons/blob/develop/icons/furrynetwork.svg) which used an `E` command in its path, which is not officially supported (as far as I can tell).


birjolaxew:
:+1: Thanks for taking care of this @ericcornelissen! It's good to finally get it implemented. Should be a minor version bump on release, agree?

ericcornelissen:
> Should be a minor version bump on release, agree?

Since it doesn't affect any existing usages of the package, a minor version bump should be fine yes.

PR #1476, opened 2019-06-16
Add NPM script to perform SVGO optimization from the repository

ericcornelissen:
**Issue:** n/a

### Description
Added a NPM script to run SVGO (the CLI tool) more easily and included a predefined configuration that aligns with our guidelines. I also updated the Contributing Guidelines accordingly.


davidklebanoff:
@ericcornelissen - Can you elaborate on the differences in output when running SVGO via the command line with `svgo --precision=3 -i icon.svg -o icon.min.svg` versus what you have specified here?

ericcornelissen:
The primary benefits are (compared to just just `svgo --precision=3 -i icon.svg -o icon.min.svg`):
- The `<title>` won't be removed ([here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R18)).
- The `role="img"` attribute won't be removed ([here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R24) and [here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R26)).
- Arcs will be converted to paths ([here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R11)).
- Once https://github.com/svg/svgo/pull/1084 gets merged, the SVG paths will get compounded ([here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R15)).

The attribute order ([here](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-a2081c9fe125af3758bf59003f07b0e9R4)) is just a small extra thing, not really necessary.

Besides that:
- Having to run just `npm run svgo -- [ICON]` is a bit easier to remember.
- If the configuration ever changes people using that command won't notice.
- This does not require you to install SVGO globally, which I can imagine might annoy some people. (though that would be quicker as you don't need to install the other dependencies for simple-icons, but then again we don't have that many dependencies)

davidklebanoff:
Thanks! This certainly demonstrates the added value provided.

I'm not terribly familiar with the workings of SVGO's configuration file, but I noticed the lack of a _precision_ attribute. Perhaps the default is `3`, but even so I think it may be of value to explicitly indicate it as such.

ericcornelissen:
> Thanks! This certainly demonstrates the added value provided.

Coming back to this after a couple of days I realized I should've explained it better 😅 

> I'm not terribly familiar with the workings of SVGO's configuration file, but I noticed the lack of a precision attribute. Perhaps the default is 3, but even so I think it may be of value to explicitly indicate it as such.

Me neither and I didn't get it to work with the configuration, hence I [specified it in the script itself](https://github.com/simple-icons/simple-icons/pull/1476/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R36).

davidklebanoff:
Ah, sorry, I missed that! Yeah, I'm sure we both would prefer having the _multipass_ and _precision_ attributes in the configuration, but if that's how SVGO works then so be it. I think it may be worth submitting an issue on their repos or at least digging into why that design decision was made.

Looking at our current icons, it would appear that most (all?) have the attributes in the order of `role`, `viewbox`, `xmlns` (alphabetical). I ran the script and it placed the attributes in order of `xmlns`, `role`, `viewbox`, which is consistent with your configuration, but perhaps not what we want unless you believe it's better?

Oddly enough, I changed the order of the attributes in the configuration file to match our current order and re-ran the script and it still put the attributes the in the order of `xmlns`, `role`, `viewbox`. Desired sort-order aside, are you sure that the script is following the sort order as specified in the configuration?

ericcornelissen:
> Ah, sorry, I missed that! Yeah, I'm sure we both would prefer having the multipass and precision attributes in the configuration, but if that's how SVGO works then so be it. I think it may be worth submitting an issue on their repos or at least digging into why that design decision was made.

Good point, I'll look into that before we merge this 👍 


> Looking at our current icons, it would appear that most (all?) have the attributes in the order of role, viewbox, xmlns (alphabetical). I ran the script and it placed the attributes in order of xmlns, role, viewbox, which is consistent with your configuration, but perhaps not what we want unless you believe it's better?
Oddly enough, I changed the order of the attributes in the configuration file to match our current order and re-ran the script and it still put the attributes the in the order of xmlns, role, viewbox. Desired sort-order aside, are you sure that the script is following the sort order as specified in the configuration?

I had the same problem, that is why I put the order it currently has. If I recall correctly the `xmlns` is treated as a special attribute and therefor is always placed first. I will get back to that as well if I can find it again.

ericcornelissen:
> Good point, I'll look into that before we merge this 👍

Found this issue: https://github.com/svg/svgo/issues/171 and it seems that `--precision` automatically sets a couple of configurations for you, but since we're only using one of them in the end I added that to the configuration file in 9ebe4c94485c984abd9f66841eb032653c3a1467.


> I had the same problem, that is why I put the order it currently has. If I recall correctly the xmlns is treated as a special attribute and therefor is always placed first. I will get back to that as well if I can find it again.

Looking at the source cod I released that the `xmlns` attributes are sorted separately and can be configured to be the first attributes or the last attributes, so I updated the configuration to have them as the last attributes in e353c307624bc9908bb5e702d1a6cc1047507335.

davidklebanoff:
Awesome! Thanks for the updates. This looks good to me!

Issue #1416, opened 2019-05-16
Dynatrace


DeadManPoe:
**Name:** Dynatrace
**Website:** www.dynatrace.com
**Official resources for icon and color:** https://www.dynatrace.com/company/press-kit/

hylickipiotr:
I'll take care of it!

hylickipiotr:
I've almost done, but I have a problem choosing the color. In the logo are five different colors: 
* symbol: `#b4dc00` `#73be28` `#6f2da8` `#1496ff` 
* typography: `#1a1a1a`

I've also checked the Dynatrace website. The primary color of the website is `#0f1419`.

---
I think, better option is to use the color from the website `#0f1419`.

@DeadManPoe @birjolaxew What is your opinion?

ericcornelissen:
This is a difficult one, I'm not sure but I do want to add that given the styling on their website `#6F2DA8` and `#1496FF` are also good options since they're used in the accent gradient, e.g.:

![image](https://user-images.githubusercontent.com/3742559/57999082-c485af00-7adc-11e9-87a3-688d14994a70.png)

DeadManPoe:
I believe the best one is the colored one (green, blue, purple), it's the logo used everywhere except when the background is of a dark color

PR #1498, opened 2019-06-28
Added Dynatrace icon

JoaoPedroMoraes:
**Issue:** #1416 

### Checklist

### Description
From the official SVG I was able to create a monocolor version. https://www.dynatrace.com/


birjolaxew:
Although the logo looks good, I don't agree with the brand color chosen for this PR. As [discussed in the issue](https://github.com/simple-icons/simple-icons/issues/1416#issuecomment-493795391), they seem to use 4 colors for their brand. I agree with @ericcornelissen that we would probably want either `#1496FF` or `#6F2DA8` - my personal preference is the blue one. Thoughts @ericcornelissen @davidklebanoff ?

davidklebanoff:
I vote `#1496FF` because out of the four brand colors, it appears to be used the most.

It may be worth noting that the monochrome version of their logo in their presskit uses `#1A1A1A`.

ericcornelissen:
> I vote #1496FF because out of the four brand colors, it appears to be used the most.

I personally agree.

I would like to get back to @DeadManPoe comment in the original issue where they prefer green (I missed the comment at the time...), may I ask why @DeadManPoe? I looked around a bit more and it seems they used to have a logo that only has green as color, is this why you associate green with the brand, or something else?


> It may be worth noting that the monochrome version of their logo in their presskit uses #1A1A1A.

I don't think that is anything to go by, it is just that: "a monochrome version" and doesn't really tell us anything about the brand color.

DeadManPoe:
@ericcornelissen In the original issue I did not mention my preference for green, I just underlined that the original logo is the multicoloured one. For what concerns me, I agree with your colour choice.

ericcornelissen:
Okay, sorry for the confusion @DeadManPoe and many thanks once again @JoaoPedroMoraes!

PR #1502, opened 2019-06-30
Increase specificity of source URL pattern for JSONLint

ericcornelissen:
**Issue:** https://github.com/simple-icons/simple-icons/pull/1494#discussion_r298812951

### Description
Update the JSONLint scheme to be more specific about what constitutes a valid source URL, based on the discussion [in this Stackoverflow thread](https://stackoverflow.com/a/3809435/4132379).

Perhaps such a specific pattern is undesirable as I already had to modify it (adding `\\(\\)';!` near the end) to accept all existing URLs 🤔 I'm open to other suggestions.

### DIFF
--- a/.jsonlintschema
+++ b/.jsonlintschema
@@ -23,7 +23,7 @@
                     "source": {
                         "description": "The website from which the icon originated",
                         "type": "string",
-                        "pattern": "^https?://",
+                        "pattern": "^https?:\/\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=\\(\\)';!]*)$",
                         "required": true
                     }
                 },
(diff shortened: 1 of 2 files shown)


birjolaxew:
I think keeping it short and maintainable might trumph having 100% validated URLs.

We usually want to validate that the URL is absolute (i.e. includes `https://`) and that there aren't any difficult-to-notice errors (e.g. the whitespace padding). Unfortunately I don't really know enough about the URL format to know if there are other errors we want to stop, but since whitespace is the only one that's been a problem so far I think something like `^https?:\/\/[^\s]+$` might work.

davidklebanoff:
I'm in complete agreeance with @birjolaxew.

ericcornelissen:
That is totally fair, I considered both options and wasn't sure which to go with so I just opened the PR with one of the two options. I updated the pattern to the pattern @birjolaxew proposed in [his comment](https://github.com/simple-icons/simple-icons/pull/1502#issuecomment-507048966).

Issue #1330, opened 2019-03-27
[Icon Request] Apache Flink


garyelephant:
**Name: Flink**
**Website: https://flink.apache.org/**
**Official resources for icon and color:https://flink.apache.org/img/flink-header-logo.svg**

ericcornelissen:
Alexa rank is about ~1300 :+1: It is not obvious how the logo should be monochromatized though :thinking:

JoaoPedroMoraes:
We can leave the icon like this, what's up? but we have to decide the color!
Can I open the Pull Request with the icon like this and the color # E65270? @ericcornelissen 
![Captura de Tela_selecionar área_20190630130655](https://user-images.githubusercontent.com/46731099/60399239-007a4e00-9b38-11e9-987b-55b19807b9fc.png)

davidklebanoff:
@JoaoPedroMoraes - I found an official monochrome version of the SVG [here](https://flink.apache.org/img/logo/svg/black_outline.svg). Please use it instead and use https://flink.apache.org/material.html as the source page.

As for color, I think the squirrel's body color is the most appropriate. I extracted `#E65270` from an official [colored SVG](https://flink.apache.org/img/logo/svg/color_black.svg).

ericcornelissen:
Go ahead with creating a PR @JoaoPedroMoraes, consider @davidklebanoff comment tho 😉

JoaoPedroMoraes:
> Vá em frente com a criação de um PR @ JoãoPedroMoraes , considere @davidklebanoff comentário tho

Pull Request Ok :+1:

PR #1507, opened 2019-06-30
Added Apache Flink icon

JoaoPedroMoraes:
**Issue:** #1330 

### Checklist

### Description
From the official SVG I was able to create a monocolor version. https://flink.apache.org/

the svg logo was removed from https://flink.apache.org/material.html#scalable-vector-graphics-svg


davidklebanoff:
I went ahead and updated the source to the logo page.

As for the logo itself, I was thinking the "Black Outline" version of the logo would be the most appropriate, but I realise the one you have included here might be equally as good. Perhaps it's just a matter of opinion. I'd like some additional opinions. Thoughts @ericcornelissen @birjolaxew?

See all versions here: https://flink.apache.org/material.html

ericcornelissen:
I personally think the black outline version is a better option since it's a little easier on the eyes. Can you elaborate why you picked this version @JoaoPedroMoraes?

JoaoPedroMoraes:
> Eu pessoalmente acho que a versão de contorno preto é uma opção melhor, pois é um pouco mais fácil para os olhos. Você pode explicar por que escolheu esta versão @JoaoPedroMoraes ?

I thought that when the icon was small it would be better to visualize. But you who deny what is best

ericcornelissen:
Hmm that's an interesting hypothesis. Care to create a comparison between the two (maybe on a few different scales)? If not, I'll try to do it myself later.

ericcornelissen:
Here is a quick comparison between the two options ⬇️ I'm not convinced either way, neither version is very legible at a small scale. Personally I still prefer the outlined version, which I think looks better at larger scales, but in my opinion both are fine.

### Small

| Outline | Filled |
| --- | --- |
| image | image |

### Large

| Outline | Filled |
| --- | --- |
| image | image |

JoaoPedroMoraes:
> Here is a quick comparison between the two options  I'm not convinced either way, neither version is very legible at a small scale. Personally I still prefer the outlined version, which I think looks better at larger scales, but in my opinion both are fine.
> 
> ### Small
> Outline	Filled
> image	image
> ### Large
> Outline	Filled
> image	image

All in all, the outline version looks more beautiful

JoaoPedroMoraes:
and, what version will it be?

ericcornelissen:
Let's let others way in before we make a final decision :)

davidklebanoff:
I prefer the black outlined version.

Issue #1362, opened 2019-04-17
URL- and code-friendly slugs


paulmelnikow:
Hi! Thanks so much for this package. It's been a huge help to be able to use this package in Shields.

### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [x] The NPM Package
  - [x] The website
  - [ ] Other, namely: 

### Description

It would be helpful to have a canonical slug for each icon that users could see on the website and could be used by consumers of the package to pull the corresponding logo programmatically.

It looks like the package only supports pulling logos by name. The website shows the name, though it exposes the filename when you download the svg, which is basically a slug.

In Shields we accept a logo in the query string and use that to look up the logos by a canonicalized slug. The [current transform code](https://github.com/badges/shields/blob/d1f00b93e8ac2adfcbdb92a8e6286cbdcfd1a194/lib/load-simple-icons.js#L8) looks like this:

```js
Object.keys(simpleIcons).forEach(key => {
  const k = key.toLowerCase().replace(/ /g, '-')
  ...
})
```

This is somewhat brittle. It isn't working well on Let's Encrypt (also Macy's) which use single right quotes rather than apostrophes. (badges/shields#3297)

Presenting a canonical slug on the simple-icons website seems like it would be helpful not only for Shields, but also for other applications which may consume simple-icons.

I'd suggest making the slug the default way to look up the icons, because it encourages people to switch to the slug, which may be a a more stable identifier. (Consider the Let's Encrypt example, which may cause confusion outside of Shields, too.) A `withName()` function could provide the old behavior.

Though that would be a breaking change, and maybe adding a `withSlug()` function would be fine, too.

I'm really curious to hear your thoughts!

ericcornelissen:
Thanks for the suggestion and I think it is very important for us to consider. The most important thing that would prevent us from implementing it directly is backwards comparability, although I feel like we would've seen a report like this earlier if someone actually tried to use e.g. Let's Encrypt with the NPM package (like you said) 🤔 

We need some time to figure out how we can best solve this. What I can say is that it is unlikely it will be a function-like approach. Most likely it will be a change to the key itself, or an additional field on the object associated with the key. The former is more usable but worse for backwards compatibility, the former is good for backwards compatibility.

---

As for the suggestion about the changes to the website, I just want to note that #980 and #989 are two relevant issues.

paulmelnikow:
[...]

PR #1520, opened 2019-07-12
Add URL friendy/slugged name as property to package icon objects

ericcornelissen:
**Issue:** #1362 

### Description
This adds a new key to the icon object called `name` (this is open for discussion) that has a URL friendly/slugged name (that is basically the SVG's filename. E.g. for **Let's Encrypt**:

```diff
 {
   title: "Let's Encrypt",
   svg: "<svg>...</svg>",
   path: "[PATH]",
   source: "[URL]",
   hex: "XXXXXX",
+  name: "letsencrypt"
 }
```

### Testing

Currently I only test if every icon has the new property, I'm not sure how to make it more advanced without using the exact code that is used to set the property... 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ console.log(simpleIcons['Simple Icons']);
 /*
 {
     title: 'Simple Icons',
-    name: 'simpleicons',
+    slug: 'simpleicons',
     hex: '111111',
     source: 'https://simpleicons.org/',
     svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>',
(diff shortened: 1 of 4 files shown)


birjolaxew:
Thanks for handling this! The issue's been sitting for way too long.

I'd probably prefer using `slug` as the key instead. I feel that `name` is a bit ambiguous, and easy to mistake for `title`.

ericcornelissen:
> I'd probably prefer using slug as the key instead. I feel that name is a bit ambiguous, and easy to mistake for title.

That is totally fair. I wasn't sure if slug was desirable either so I just went with something and hoped for some discussion in the Pull Request 😃 I will update it.

PR #1519, opened 2019-07-12
Reduce package size by replacing JSON.stringify

ericcornelissen:
**Issue:** n/a

### Description
I updated the code that generates the package JavaScript, replacing `JSON.stringify` by custom stringify method that avoid putting a whole lot of `"` and `\"`in the resulting JavaScript which are unnecessary. Major changes in the output:

- Keys inside the icon object no longer use `"`
- Values inside the icon object now use `'` instead of `"`, removing the need for `\"`. This does introduce the need for `\'`, but that is far less common. Alternatively we could use ` instead, but that is not supported by older versions of JavaScript/NodeJS so I choose not to.

Note that the keys on `const simpleicons = require('simple-icons')` are still surrounded by `"` as of this PR there is no brand name with a `"` and a number of brand names have a space in them so the `"` is required. I didn't change them into something that doesn't require surrounding it by qoutes for backwards compatibility.

### Output size comparison

Here is a table comparing file sizes for the latest `develop`.

| What | Size before (in bytes) | Size after (in bytes) |
|---|---|---|
| Just `index.js` | 1,972,191 | 1,960,438 |
| All `icons/*.js` combined | 1,974,786 | 1,963,033 |

### Example of an icon

The icon in question is **500px**.

#### Before

```javascript
[...]

PR #1521, opened 2019-07-12
Reduce impact of path value on package size

ericcornelissen:
**Issue:** continues on #1519 

### Description
This PR replaces the hardcoded value for the path data by a getter for the path `{ get path() { ... }, ... }` which calculates the path data in realtime based on the icons SVG value.

I implemented and committed a version using getters and `Object.defineProperty`, a size comparison for both solutions can be found below. In the end I used getters as they seem to have better support in older browsers, but it doesn't matter for NodeJS(?).

### Size comparison with `get path`

> For compatability see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#Browser_compatibility

| What | Size before (in bytes) | Size after (in bytes) |
|---|---|---|
| Just `index.js` | 1,972,191 | 1,087,327 |
| All `icons/*.js` combined | 1,974,786 | 1,089,922 |

### Size comparison with `Object.defineProperty()`

> For compatability see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Browser_compatibility

| What | Size before (in bytes) | Size after (in bytes) |
|---|---|---|
| Just `index.js` | 1,972,191 | 1,048,291 |
| All `icons/*.js` combined | 1,974,786 | 1,122,572 |

### DIFF
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -20,7 +20,7 @@ function iconToKeyValue(icon) {
   return `'${icon.title}':${iconToObject(icon)}`;
 }
 function iconToObject(icon) {
-  return `{title:'${icon.title}',svg:'${icon.svg}',path:'${icon.path}',source:'${icon.source.replace(/'/g, "\\'")}',hex:'${icon.hex}'}`;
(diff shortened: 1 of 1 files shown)


birjolaxew:
:+1: Pretty big cut! Nice idea.

It surprises me a bit that `get` has better support, but it seems to predate the method definition shorthand from ES6 despite looking an awful lot like it.

I'm personally ok with either. Not supporting IE8 isn't exactly a big loss; but then again, might as well keep the support as wide as possible unless there is a good reason.

PR #1522, opened 2019-07-13
Generic get icon function for easier access to brands with "difficult" names

ericcornelissen:
**Issue:** related to #1362, continues on #1520

### Description

This pull request adds a function to the exported object in `const simpleIcons = require("simple-icons"');` that allows for easier access to icons. Beforehand you had to write the exact brand name, which, as pointed out in #1362, is quite annoying for some brands. Now you can use both that and the SVG filename to get the icon.

To make the template for the generated `index.js` file understandable after this change I moved it into a separate file. The prepublish script will read the file and format it using [NodeJS's `util.format`](https://nodejs.org/api/util.html#util_util_format_format_args) (similar to C's printf) and finally minify it using [Uglify-JS](https://www.npmjs.com/package/uglify-js) before finally writing it.`

I also updated how "you should" use the package in the README to use `.get` rather then the bracket notation. Note that this is fully backwards compatible as the bracket notation still works.

### Implementation

- The function is named just `get`, I think that is fine and understandable. I don't expect the API will be expanded in the feature so the name doesn't need to be more clear, say `getIcon`. The only downside is that there might some day be a brand that names itself "get"...
- The search is case insensitive, it doesn't have to be but I think it is more developer friendly. I'm open to discussing this for sure.
- It is not immediately obvious what are valid inputs to the `.get` function, but I'm not sure how to adequately document it. One option would to have a automatically generated list of valid inputs in a new `docs/` folder in the repository?

### Example

```javascript
const simpleIcons = require("simple-icons");

// Before
let letsEncrypt = simpleIcons["Let's Encrypt"];

// After
letsEncrypt = simpleIcons.get("Let's Encrypt"); // same as before
letsEncrypt = simpleIcons.get("let's encrypt"); // case insensitive
letsEncrypt = simpleIcons.get("letsencrypt"); // slugged name
```

### Testing

I added some tests to `index.test.js` that test if it can find every icon given its "title" and "name" property.


birjolaxew:
:+1: I'm good with all the decisions for this PR.

It's a shame that using square notations for everything requires either `Proxy` (thus no IE support, which is too strict for us to impose) or a significantly larger package size.

I'll be leaving this open until #1520 is merged, as this might need to be updated to use `slug` instead (pending discussion in the other PR)

ericcornelissen:
> It's a shame that using square notations for everything requires either Proxy (thus no IE support, which is too strict for us to impose) or a significantly larger package size.

Definitely agree but this is the best compromise...


> I'll be leaving this open until #1520 is merged, as this might need to be updated to use slug instead (pending discussion in the other PR)

I will update this as soon as possible 👍

PR #1526, opened 2019-07-14
Audit dependencies

ericcornelissen:
**Issue:** n/a

### Description
Audited the package dependencies using `$ npm audit fix`, going down from 1110 vulnerabilities to 0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4030,9 +4030,9 @@
       }
     },
     "mixin-deep": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/mixin-deep/-/mixin-deep-1.3.1.tgz",
-      "integrity": "sha512-8ZItLHeEgaqEvd5lYBXfm4EZSFCX29Jb9K+lAHhDKzReKBQKj3R+7NOF6tjqYi9t4oI8VUfaWITJQm86wnXGNQ==",
+      "version": "1.3.2",
+      "resolved": "https://registry.npmjs.org/mixin-deep/-/mixin-deep-1.3.2.tgz",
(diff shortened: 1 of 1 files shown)


birjolaxew:
> 1110 high severity vulnerabilities

Yikes. Nice job catching this!

ericcornelissen:
> Yikes. Nice job catching this!

It's honestly not such a big deal because it's all devDependencies and the way were using them barely effects our package. But it's still nice to stay up-to-date :smile:

davidklebanoff:
Hmm, I see bot potential here.

ericcornelissen:
> Hmm, I see bot potential here.

Sure, didn't think it was necessary since dependencies are not super relevant (everything is a `devDependency`). Anyhow, Ihave experience with [Greenkeeper](https://github.com/marketplace/greenkeeper) and [Renovate Bot](https://github.com/marketplace/renovate/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW45NTk=#pricing-and-setup) (both are free for open source), I would definitely recommend the latter over the former. What do you think @davidklebanoff ?

davidklebanoff:
I don't have experience with either of those tools, so I can't say one way versus another. The task is definitely low priority since as you mentioned they are devDependencies. We could trial the Renovate Bot to see how it goes. I like that they allow you to throttle how often it runs - given that these are only devDependencies we don't have to rush to get things updated ASAP. Perhaps a cadence of biweekly or monthly would be appropriate for a project such as Simple Icons.

ericcornelissen:
> We could trial the Renovate Bot to see how it goes. I like that they allow you to throttle how often it runs - given that these are only devDependencies we don't have to rush to get things updated ASAP. Perhaps a cadence of biweekly or monthly would be appropriate for a project such as Simple Icons.

I like that idea, I didn't know you can throttle how often it runs but that seems like a great feature for our case indeed! I will try to look into this as soon as possible as it isn't that hard to set up (though I never configured a Renovate Bot before) and I'll open an issue as well just to be sure :wink:

Issue #1397, opened 2019-05-06
Multiple Icon Requests


h1nk:
~~**Name:** Unreal Engine
**Website:** [unrealengine.com](https://www.unrealengine.com/) ([Alexa Ranking ~2K](https://www.alexa.com/siteinfo/unrealengine.com))
**Official resources for icon and color:** https://www.unrealengine.com/en-US/branding & https://cdn2.unrealengine.com/BrandingMedia%2FLogos%2FUnreal_Engine_Logos-194aec037564e2e849650161de30d5a294a592ee.zip~~

---

~~**Name:** OpenVPN
**Website:** [openvpn.net](https://openvpn.net/) ([Alexa Ranking ~20K](https://www.alexa.com/siteinfo/openvpn.net))
**Official resources for icon and color:** https://openvpn.net/community-resources/openvpn-logos-and-icons/ & https://openvpn.net/wp-content/themes/openvpn/assets/images/logo.svg & https://openvpn.net/wp-content/themes/openvpn/assets/images/openvpn-mark.svg~~

---

Resolved in PR #1906 

~~**Name:** StrongSwan~~
~~**Website:** [strongswan.org](https://www.strongswan.org/) ([Alexa Ranking ~180K](https://www.alexa.com/siteinfo/strongswan.org))~~
~~**Official resources for icon and color:** https://lists.strongswan.org/pipermail/users/2012-November/003925.html & https://www.strongswan.org/images/strongswan_square_large.png~~

---

~~**Name:** WireGuard
**Website:** [wireguard.com](https://www.wireguard.com/) ([Alexa Ranking ~163K](https://www.alexa.com/siteinfo/wireguard.com))
**Official resources for icon and color:** https://www.wireguard.com/img/emblem.svg & https://www.wireguard.com/img/wireguard.svg~~

---

**Name:** Source Engine 
**Website:** [valvesoftware.com](https://www.valvesoftware.com/) ([Alexa Ranking ~18K](https://www.alexa.com/siteinfo/valvesoftware.com))
**Official resources for icon and color:**  https://upload.wikimedia.org/wikipedia/commons/d/d9/Source_engine_logo.svg (unable to find official resource)

---

Resolved in PR #1893 

~~**Name:** AlliedModders~~
~~**Website:** [(forums.)alliedmods.net](https://forums.alliedmods.net/) ([Alexa Ranking ~53K](https://www.alexa.com/siteinfo/alliedmods.net))~~
[...]

PR #1535, opened 2019-07-16
Add OpenVPN & WireGuard Icon

fatihmeh:
Issue: #1397

### Checklist

### Description
- Used Inkscape.
- Optimized icons using https://jakearchibald.github.io/svgomg/
- The color come from the SVG found in the source.
- OpenVPN original logo comes from the official page looks like it is not circular, has some dents and a spike in the middle keyhole shape which I didn't understand if this is on purpose. So I used it as is.


fatihmeh:
@ericcornelissen About OpenVPN logo:

**I was working on "openvpn-mark.svg".** After your mention, I looked at "logo.svg" and noticed what is different and wrong.

You are correct about hex value. But two different files have two different values:
"https://openvpn.net/wp-content/themes/openvpn/assets/images/logo.svg" -> Hex value is EA7E20.
"https://openvpn.net/wp-content/themes/openvpn/assets/images/openvpn-mark.svg" -> Hex value is ED7F22.


I am talking about this weird shape:
image


image
And as you can see from upper image, **red colored shape is a circle I added**. Even though their centers are not concentric, their edges should follow same path as below image. This doesn't occur on "logo.svg"-except one problem I showed at bottom.


This is what it should look like when 2 circles come on top of each other. (I made this just for an example)
image


Right outside edge of "logo.svg", circle shape ends with a short vertical line and continue.
image


Hope I didn't miss anything.

ericcornelissen:
Thanks for the extensive analysis @fatihmeh, great job! I saw the problems you mention with `openvpn-mark.svg` 😉 I didn't noticed the issues with the version I mentioned though. It is unfortunate that one doesn't look perfect either, but I think it is the best we can do. Since it is not jagged or anything, I don't feel comfortable changing it to a perfect circle.

For now I think it is best if you update the SVG in this Pull Request to the `logo.svg` version and I'll leave the final decision on this to either @davidklebanoff or @birjolaxew.

fatihmeh:
@ericcornelissen I worked on `logo.svg` and released masks that are already there, logo has become full circle. I hope this is ok for you.

oleg-nenashev:
Thanks for the OpenVPN icon!

icons/openvpn.svg
```diff
@@ -0,0 +1 @@
+<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>OpenVPN icon</title><path d="M12 .715C5.385.715 0 6.047 0 12.61c0 4.36 2.358 8.154 5.896 10.205l.77-5.076a7.046 7.046 0 0 1-1.846-4.719c0-3.897 3.18-7.076 7.13-7.076 3.948 0 7.126 3.179 7.126 7.076 0 1.847-.717 3.488-1.846 4.77l.77 5.078c3.59-2.051 6-5.899 6-10.258C24 6.047 18.615.715 12 .715zm-.05 8.156a3.786 3.786 0 0 0-3.796 3.795 3.738 3.738 0 0 0 2.461 3.54L9.13 24h5.64l-1.435-7.795c1.385-.564 2.41-1.898 2.41-3.539a3.786 3.786 0 0 0-3.795-3.795z"/></svg>
```
birjolaxew: Looks like the icon isn't properly centered vertically in the viewport - could you move it up so it's centered?

fatihmeh: It's should be ok now. Thanks.

birjolaxew:
:+1: Looks good! Thanks for your work on this, it is much appreciated! (and sorry again for the slow review)

Issue #1415, opened 2019-05-16
Cassandra icon


DeadManPoe:
**Name:** Cassandra
**Website:** http://cassandra.apache.org/
**Official resources for icon and color:** none for what i'm aware of, png taken from the website can be found here http://cassandra.apache.org/img/cassandra_logo.png

Svg icon taken from Wikimedia can be found here: https://upload.wikimedia.org/wikipedia/commons/5/5e/Cassandra_logo.svg

PR #1541, opened 2019-07-17
Add Cassandra Icon

fatihmeh:
Issue: #1415

### Checklist

### Description
Used Inkscape.
Optimized icons using https://jakearchibald.github.io/svgomg/
The color come from the SVG found in the source.
I tried 2 ways on how to display eyelash shadows and decided not to use them.
image


fatihmeh:
I am not sure about color also, I respect your selection. 

Sorry for not mentioning this, I used logo on left, same stuff. @ericcornelissen

davidklebanoff:
I don't believe there is an official color or really an obvious choice. So with that said I think the iris color  `#1287b1` which was chosen by @fatihmeh should work just fine.

birjolaxew:
:+1: I agree with @davidklebanoff.

Thanks for your work on this @fatihmeh!

PR #1547, opened 2019-07-21
Refactor prepublish script to use template file for icon objects

ericcornelissen:
**Issue:** n/a

### Description
Refactor `prepublish.js` script to use template file for icon objects to hopefully increase readability. Arguably this is less readable as the arguments inside `util.format` inside `iconToObject` are somewhat arbitrary, on the other hand that function is meant to abstract that away.

Let me know if you disagree that this is easier to understand/more maintainable as well 😄


birjolaxew:
:+1: I like separating the file structure away from the generating code. I do question some of the code design (e.g. using defining `UTF8` instead of using `"utf8"`), but that's pretty minor and not worth refactoring a 25 day old PR for.

Good job on this! And sorry for the late review.

ericcornelissen:
> I do question some of the code design (e.g. using defining UTF8 instead of using "utf8"), but that's pretty minor and not worth refactoring a 25 day old PR for.

I would have been (and still are) open for a discussion on such things, I would say the PR could've waited a few more days :wink:

I started using `UTF8` over `"utf8"` because the error you get when mistyping `UTF8` is clearer than the one you get when mistyping `"utf8"` but it's  a minor thing and arguably not necessary for a script of this size.

PR #1555, opened 2019-07-24
Don't enumerate the get method

### LINKED ISSUE #1554: Provide API endpoint to iterate over all icons
### Kind of issue 
  - [x] Improvement _(or a bug since it was (unofficially) possible before?  🤔)_
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [x] The NPM Package
  - [ ] The website
  - [ ] Other, namely: 

### Description
With the addition of [the generic get method](https://github.com/simple-icons/simple-icons/pull/1522) to simplify access to brands with "difficult" names [added in v1.12.0](https://github.com/simple-icons/simple-icons/pull/1549) it has become more of a pain to iterate over all icons in the set.

Before, most likely this was achieved using:

```javascript

Object.values(simpleIcons).forEach(icon => {
  // Do something with the icon
});
```

However, now the last "icon" will be the `get` function that was added. Therefor I think we should look into a way to provide users of the package with an iterable list of icons. Here are some possible approaches.

- Add a `.forEach` method that works like [Array.prototype.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach). However, then we also need a `.map` function, and a `.filter` function, etc. So this is not ideal...
- Add a new value (getter) or function that will just return an array of all keys. I.e. it would return `Object.values(simpleIcons)` minus the `get` method and this getter/function.
- _Suggestions?_

ericcornelissen:
**Issue:** Resolves #1554

### Description
Adds, and tests, functionality that hides the `.get` method (introduced in #1522) when iterating over the values (and keys) of `const simpleIcons = require('simple-icons');`.

### DIFF
--- a/scripts/templates/index.js
+++ b/scripts/templates/index.js
@@ -1,17 +1,21 @@
 var icons = {%s};
 
-module.exports = icons;
-module.exports.get = function(targetName) {
-  if (icons[targetName]) {
-    return icons[targetName];
-  } else {
-    var normalizedName = targetName.toLowerCase();
-    for (var iconName in icons) {
-      var icon = icons[iconName];
-      if ((icon.title && icon.title.toLowerCase() === normalizedName)
-       || (icon.slug && icon.slug === normalizedName)) {
-         return icon;
+Object.defineProperty(icons, "get", {
+  enumerate: false,
+  value: function(targetName) {
+    if (icons[targetName]) {
+      return icons[targetName];
+    } else {
+      var normalizedName = targetName.toLowerCase();
+      for (var iconName in icons) {
+        var icon = icons[iconName];
+        if ((icon.title && icon.title.toLowerCase() === normalizedName)
+         || (icon.slug && icon.slug === normalizedName)) {
+           return icon;
+        }
       }
     }
   }
-}
+});
+
+module.exports = icons;
--- a/tests/index.test.js
+++ b/tests/index.test.js
@@ -44,3 +44,11 @@ icons.forEach(icon => {
     expect(found.title).toEqual(icon.title);
   });
 });
+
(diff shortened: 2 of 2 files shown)


davidklebanoff:
@ericcornelissen @birjolaxew 

Hmm, unless I'm missing something, the proper key name should be `enumerable`. That said, the default value for that key is `false`, so even if the key name was correct, the code would act exactly the same. Long story short, you could just remove the `enumerate: false,` altogether!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Description

ericcornelissen:
> Hmm, unless I'm missing something, the proper key name should be enumerable. That said, the default value for that key is false, so even if the key name was correct, the code would act exactly the same. Long story short, you could just remove the enumerate: false, altogether!

You are absolutely right @davidklebanoff (and I did test it now). I didn't look at the documentation and just copied @birjolaxew's [snippet](https://github.com/simple-icons/simple-icons/issues/1554#issuecomment-514578072) 😅

In that case I agree we can just remove `enumerate: false`. Ideally there would be a comment stating that we rely on the default value of `enumerable` being `false` though. That way it is clear why we use `Object.defineProperty` and we safe a few bytes off of `enumerable: false` (or we just use `enumerable: false`...).

birjolaxew:
Whoops, my bad :disappointed: 

I'd probably prefer replacing it with `enumerable` instead of removing it; that way the code is more self-documenting, and the comment isn't needed.

ericcornelissen:
> I'd probably prefer replacing it with enumerable instead of removing it; that way the code is more self-documenting, and the comment isn't needed.

That is fine as well 😄

Issue #1427, opened 2019-05-22
Request: Publish to Composer/Packagist


joshuabaker:
It’d be great to have this resource available via [Packagist](https://packagist.org/).

ericcornelissen:
I don't see why we wouldn't add this :thinking: 

Do you have any guidelines/examples/help on how we can best approach that, I'm not a PHP developer myself (I don't know if any of the other maintainers are) so I'm not super familiar with Composer/Packagist

joshuabaker:
Thanks for being open to this.

The required `composer.json` file is pretty similar to the existing [`package.json`](https://github.com/simple-icons/simple-icons/blob/e9f76351eceb00baba53696dfc0b45820637dae3/package.json) file.

```json
{
  "name": "simple-icons/simple-icons",
  "description": "SVG icons for popular brands",
  "homepage": "https://simpleicons.org/",
  "keywords": [
    "svg",
    "icons"
  ],
  "support": {
    "issues": "https://github.com/simple-icons/simple-icons/issues"
  },
  "license": "CC0-1.0"
}
```

Then a maintainer just needs to [submit the package to Packagist](https://packagist.org/packages/submit).

joshuabaker:
Closing as this isn’t moving.

For those needing to install via Composer, at the time of writing, you can install via [asset-packagist.org](https://asset-packagist.org/).

ericcornelissen:
> Closing as this isn’t moving.

I feel like that is something for the people maintaining a repository to decide 🤔 I consider this issue still as a TODO and would therefore like to keep it open. Is the issue being open bugging you as the opener, @joshuabaker?


> For those needing to install via Composer, at the time of writing, you can install via asset-packagist.org.

That is actually a great link, thanks for sharing it!

joshuabaker:
That wording might have sounded passive aggressive. It wasn’t meant to. 😅 

In my mind, I have as much responsibility to contribute to open source as repo owners (i.e. opening/closing/owning issues and submitting PRs/solutions). If I can’t, as in this case, it just sits as a perpetual to do in my issues.

There’s a couple of solutions out there right now that support similar packages (i.e. front-end assets, not PHP packages/libraries) so I saw fit to close.

ericcornelissen:
> That wording might have sounded passive aggressive. It wasn’t meant to. 😅

I wouldn't say aggressive, I just didn't see why you'd close it 😉But that's just a difference in style of using GitHub.


[...]

PR #1559, opened 2019-07-26
Add a .gitattributes files

ericcornelissen:
**Issue:** Related to #1427

### Description
The Composer/Packagist issue convinced me to implement this (though it turned out to not be strictly necessary from an `export-ignore` point of view), but I have been thinking about adding a [`.gitattributes` file](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes) to the project for a long time. The primary things it will provide us with is better diffing and more consistency across platforms (e.g. line-endings).

The contents of the file are obviously open for discussion 🙂

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,13 +11,7 @@ package-lock.json  -diff
 *.png              binary
 
 # Don't export/archive these files
-images/            export-ignore
-scripts/           export-ignore
-tests/             export-ignore
-.*                 export-ignore
+.github            export-ignore
+.gitpod.yml        export-ignore
+.travis.yml        export-ignore
 CNAME              export-ignore
-CONTRIBUTING.md    export-ignore
-package-lock.json  export-ignore
-index.html         export-ignore
-site_script.js     export-ignore
-stylesheet.css     export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,23 @@
+*                  text=auto
+
+# SVGs are treated as binary by default
+*.svg              text
+
+# Don't diff machine generated files
+package-lock.json  -diff
+
+# Treat images as binary
+*.ico              binary
+*.png              binary
+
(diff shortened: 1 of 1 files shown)


birjolaxew:
My only question is what impact the `export-ignore` lines have on our project. Does NPM or GitHub use the `git archive`-generated export?

ericcornelissen:
> My only question is what impact the export-ignore lines have on our project. Does NPM or GitHub use the git archive-generated export?

On GitHub it affects the content so the zip file you get when you download the repository as zip ([repository homepage](https://github.com/simple-icons/simple-icons) `>` "Clone or download" `>` "Download ZIP").

As far as I know(/can tell) it doesn't affect NPM releases and running `npm publish` just looks at the files in the directory, ignoring only files specified in `.npmignore` (or `.gitignore`);

> Also sorry for the super late review.

Not a problem :wink:

birjolaxew:
I think that in that case I'd prefer not excluding files from it. When clicking the download button I would expect to get a ZIP of all the files I see in the web interface too.

ericcornelissen:
> I think that in that case I'd prefer not excluding files from it. When clicking the download button I would expect to get a ZIP of all the files I see in the web interface too.

I can agree with that to some extend. To clarify, the `export-ignore` is intended to be used to manage what is included in the archive when you run the `git archive` command. The fact that GitHub uses that to generate the zip file may lead to awkward situation when a user expects that the download the whole repository. However, I think that is primarily GitHub's issue, not ours.

Regardless, I do agree that if you use it for that purpose the number of files ignored in the export can be far less. For now I have reduced it to:
- `CNAME`: only relevant for our GitHub pages site.
- `.github` (folder): is only relevant to GitHub, not the project itself.
- `.gitpod.yml`: is only relevant on GitHub, not to the project itself.
- `.travis.yml`: is only relevant on GitHub, not to the project itself.

PR #1576, opened 2019-08-05
added shell icon

vigonotion:
**Issue:**

### Checklist

### Description

I chose yellow as the dominant color, it's more prominent on the gas stations and the website [https://www.shell.com/](https://www.shell.com/), but red could also be the dominant color.


_data/simple-icons.json
```diff
@@ -2430,6 +2430,11 @@
             "hex": "525DDC",
             "source": "https://roots.io/"
         },
+        {
+            "title": "Royal Dutch Shell",
+            "hex": "FFD500",
+            "source": "https://de.wikipedia.org/wiki/Datei:Royal_Dutch_Shell.svg"
```
ericcornelissen: In this case I think it is more appropriate to use the English wikipedia 😉 

```suggestion
            "source": "https://en.wikipedia.org/wiki/File:Shell_logo.svg"
```

_data/simple-icons.json
```diff
@@ -2430,6 +2430,11 @@
             "hex": "525DDC",
             "source": "https://roots.io/"
         },
+        {
+            "title": "Royal Dutch Shell",
```
ericcornelissen: Not something that necessarily requires a change, just want to point out that just "Shell" would be fine, too. From [its Wikipedia page](https://en.wikipedia.org/wiki/Royal_Dutch_Shell):

> Royal Dutch Shell [...], commonly known as Shell [...]

vigonotion: I like that! I was thinking about just Shell too, as I would always refer to it as just Shell, but unsure because the Wikipedia article and Google used the full name. But I think as the full name is rarely used, just Shell is the better choice.

ericcornelissen:
Thank you very much for your work on this @vigonotion, it is much appreciated! 🎉

PR #1580, opened 2019-08-09
Updated StackShare's hex value

jdorfman:
**Issue:**

We got a couple of complaints about our logo not being A11y friendly on [our Shields.io badge](https://img.shields.io/badge/Follow%20on-StackShare-blue.svg?logo=stackshare&style=flat)

### Checklist

### Description

All I did was update `_data/simple-icons.json`. I'm hoping this will make the StackShare icon white on shields.io for those who have trouble with contrast.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -2737,7 +2737,7 @@
         },
         {
             "title": "StackShare",
-            "hex": "0690FA",
+            "hex": "000000",
             "source": "https://stackshare.io/branding"
         },
         {


davidklebanoff:
Hi @jdorfman,

The color within Simple Icons is the brand's primary color. In this case, the current color, `#0690FA `, is correct and so we will not be approving this pull request.

It sounds like your issue may best be raised with the shields team. Our project, Simple Icons, is not affiliated with Shields.io, they just happen to make use of our icon and brand color library. Looking at their [documentation](https://shields.io/), it would appear users can optionally override the icons default color via the `logoColor` attribute. I would advise user's concerned with the color contrast to go that route.

Another idea would be to raise the issue with Shield.io and perhaps recommend they modify the default background color to take the icon's color into consideration. The default background color could be made dynamic, insuring it is always in high contrast to the icon.

PR #1611, opened 2019-09-04
Add composer.json

ericcornelissen:
**Issue:** #1427

### Description

Add `composer.json` so that Simple Icons can be made available as a package for PHP on 
[Composer](https://getcomposer.org/)/[Packagist](https://packagist.org/).

Also pinging @joshuabaker for feedback/approval.

### DIFF
--- a/.github/ISSUE_TEMPLATE/website_and_package.md
+++ b/.github/ISSUE_TEMPLATE/website_and_package.md
@@ -1,5 +1,5 @@
 ---
-name: Website & NPM Package
+name: Website & Packages
 about: For general issues and improvements
 ---
 
@@ -13,6 +13,7 @@ about: For general issues and improvements
 
 ### This issue concerns <!-- Change the one that applies to `[x]`  -->
   - [ ] The NPM Package
+  - [ ] The Packagist Package
   - [ ] The website
   - [ ] Other, namely: 
 
@@ -23,7 +24,9 @@ about: For general issues and improvements
 Anything relevant, for example:
   - For bugs: "Steps to reproduce" and "Expected behavior"
   - For improvements: An example of a use case
-  - Package issues: What is your node version ($ node --version)
+  - Package issues: Version information
+    - For JavaScript/NodeJS: $ node --version
+    - For PHP/Packagist: $ composer --version
   - Website issues: What browser do you use
   - etc.
 -->
--- a/README.md
+++ b/README.md
@@ -76,6 +76,25 @@ console.log(icon);
 */
 ```
 
+### PHP usage
+
+The icons are also available through our Packagist package. To install, simply run:
+
+```
+$ composer require simple-icons/simple-icons
+```
+
+The package can then be used as follows:
+
+```php
+<?php
+
+echo file_get_contents('path/to/package/icons/simple-icons.svg');
+
+// <svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>
+?>
+```
+
 ## Third Party Extensions
 
 ### WordPress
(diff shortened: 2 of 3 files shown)


birjolaxew:
How does the PHP usage look like, once the package is installed using Composer/Packagist?

ericcornelissen:
> How does the PHP usage look like, once the package is installed using Composer?

Good question, @joshuabaker care to give a brief description?

joshuabaker:
A call to [`file_get_contents`](https://www.php.net/manual/en/function.file-get-contents.php).

```php
echo file_get_contents('path/to/package/icons/github.svg);
```

ericcornelissen:
> How does the PHP usage look like, once the package is installed using Composer?

More generally, I believe you will be able to do the following in your Composer configuration file.

```json5
//  ./composer.json
{
    "require": {
        "simple-icons/simple-icons": "^1.15.0"
    }
}
```

And then a simple PHP demo would be

```php
<!-- ./index.php -->
<html>
  <head>
    <title>SimpleIcons Test</title>
  </head>
  <body>
    <h1>Simple icons</h1>

    <?php echo file_get_contents('vendor/simple-icons/simple-icons/icons/github.svg'); ?>
  </body>
</html>
```

Then if you run the following command and visit `localhost:8000` you will see [the GitHub SVG](https://github.com/simple-icons/simple-icons/blob/develop/icons/github.svg):

```bash
$ php -S localhost:8000
```

The thing that is not so nice about this, at least compared to Node, is that if you're in a child directory you need to change the file path. I'm sure there is ways to solve this on our side (maybe [autoloading](https://getcomposer.org/doc/01-basic-usage.md#autoloading?)), or maybe even just a better way to use it.

Another thing is that we have [metadata](https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json) that developers can use (mainly the color), and I'm not sure how that can be used in PHP.

Do you have any suggestions for us in these regard @joshuabaker?

* * *

@birjolaxew and @davidklebanoff what are your opinions? I will look into updating documentations etc.

joshuabaker:
> The thing that is not so nice about this, at least compared to Node, is that if you're in a child directory you need to change the file path.

Yes, you would, however, that’s a pretty standard paradigm for PHP development. To interact or output the SVG you’d have to load it from disk regardless. I don’t see this as a cause for concern.

A fair number of frameworks create a global constant for such use cases. For example, [craftcms/cms](https://github.com/craftcms/cms/) sets [`CRAFT_VENDOR_PATH`](https://docs.craftcms.com/v3/config/php-constants.html#craft-vendor-path).

> Another thing is that we have [metadata](https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json) that developers can use (mainly the color), and I'm not sure how that can be used in PHP.

I’m not sure how the metadata feature works here. Looks like magic via a Node script. Is that right? 

In PHP you’d probably just load the JSON from the file, I think. Anything beyond this is likely out of scope for this issue (i.e. importing via composer).

ericcornelissen:
> Yes, you would, however, that’s a pretty standard paradigm for PHP development. To interact or output the SVG you’d have to load it from disk regardless. I don’t see this as a cause for concern.
>
> A fair number of frameworks create a global constant for such use cases. For example, craftcms/cms sets CRAFT_VENDOR_PATH.

Thanks for the insight, I think that is out of scope for now and we can add something like that if we get an issue for it.


> In PHP you’d probably just load the JSON from the file, I think. Anything beyond this is likely out of scope for this issue (i.e. importing via composer).

That makes sense, then I don't think we should worry about it for now :+1: 

Again, I will wait for the other maintainers to have a look at this before we move onwards.

PR #1682, opened 2019-09-30
No longer require no final newline in SVGs

ericcornelissen:
**Issue:** [this comment](https://github.com/simple-icons/simple-icons/issues/1324#issuecomment-536188025) and [this comment](https://github.com/simple-icons/simple-icons/pull/1677#issuecomment-536188278)

### Description

This PR updates our linting so that a final newline is no longer required in `.svg` files and update the build script to prevent build failures due to these newlines.

This should make it easier for people to contribute as they're not force to figure out how to remove that final newline their editor keeps adding to the file. In the end the difference it makes on our side is, I think, negligible.


birjolaxew:
:+1: Thanks for taking care of this! I agree, the difference it makes on our side is minimal.

Issue #1690, opened 2019-10-03
CoinBase Icon


Paulsky:
**Name:**: CoinBase
**Website:**: [https://www.coinbase.com](https://www.coinbase.com)
**Official resources for icon and color:**
  
https://www.coinbase.com/press
https://www.coinbase.com/img/favicon/favicon-228x228.png
https://www.coinbase.com
`#0667d0`

phatbhoy67:
Closed with #1845 by @el1f 👍

PR #1695, opened 2019-10-04
Add Coinbase icon

omgmog:
**Issue:** #1690

### Checklist

### Description
This is a redraw of the [Coinbase favicon](https://www.coinbase.com/img/favicon/favicon-228x228.png)

This is using the blue mentioned in #1690 that appears in the logos provided on the [Coinbase press page](https://www.coinbase.com/press).


birjolaxew:
@ericcornelissen I definitely agree that a square around it or the full wordmark would make it more recognizable; I think I would personally prefer the full wordmark, for the simple reason that they don't seem to use the single "C" anywhere but their favicon, so it doesn't have much brand recognizability (the wordmark is available as SVG in the header of their website).

omgmog:
Thanks for the feedback; I had initially drawn the icon including the enclosing box, but wasn't sure that would follow the PR guidelines.

I'll update this PR with a couple of variations (box around the C, full wordmark) when I am back at my computer.

Cheers

Issue #1346, opened 2019-04-04
Stackbit icon


gatarelib:
**Name:** Stackbit icon Request
**Website:** [stackbit.com](stackbit.com)
**Official resources for icon and color:**

ericcornelissen:
Thanks for the request @gatarelib, unfortunately Stackbit does not currently meet [our popularity threshold](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon) (Their Alexa rank is [about 1.3m](https://www.alexa.com/siteinfo/stackbit.com)) and the project is also still in beta.

We'd be glad to add their icon once the project has gained more popularity. You don't have to keep an eye on the Alexa rank, but maybe alert us through this issue once it is out of beta 😉

phatbhoy67:
Resolved by PR #1699

PR #1699, opened 2019-10-05
Add Stackbit icon

foo-dogsquared:
**Issue:** #1346

### Checklist

### Description

The logo can be found at the header (and footer) of the [homepage](https://www.stackbit.com/) as an SVG file. 
It is extracted from the vector file. 

The brand color is chosen simply from the vector file. 

NOTE: There is a variant of the logo found at the favicon (here's a [sample](https://www.stackbit.com/images/favicon/safari-pinned-tab.svg)). 
I think the logo I extracted is the correct version since it is also featured on [their GitHub org page](https://github.com/stackbithq).


birjolaxew:
I agree, this is definitely the right logo to choose.

Thanks for your work on this @foo-dogsquared! It is much appreciated.

PR #1718, opened 2019-10-08
[Patch] Fix potential display issues with 11 icons

ericcornelissen:
This patch fixes potential display issues with 11 icons added over the last 3 versions. These icons could not be displayed by Adobe Illustrator (and potentially other software).

I also included an update to the SVGO configuration file that I used to fix these problems. It seems one aspect of the optimization process (which was introduced in [v1.3.0](https://github.com/svg/svgo/releases/tag/v1.3.0)) is a bit too much for some (older?) software.

* * *

*Note*: this Pull Request merges directly into `master`. I'm not sure if this is how we want to release patches. My idea was to merge this directly into `master` and then merge `master` into `develop`. Alternatively we can merge this branch into `develop` separately, but in that case I'm not sure how the next merge from `develop` into `master` will look like...


ericcornelissen:
> Nice catch. Did you discover this from a bug report or just stumbling upon it?

I noticed the problem already when reviewing one of the PRs for the latest release but at that time I thought it was something I did wrong (besides, the particular SVG that I tested was replaced with another version before it was merged into `develop`). Then when issues occurred in https://github.com/simple-icons/simple-icons-font/pull/24 I decided to investigate further and found the problem.


> Cherry-pick the commits from this PR onto a new branch based off of develop, merge that into develop then merge that into master. This follows our normal flow.

These commits actually branch off of `develop`. The reason I didn't take this approach is because there are already a few new things on `develop` since [the last release](https://github.com/simple-icons/simple-icons/pull/1701) and I didn't want to include them in the patch.

Issue #1391, opened 2019-04-30
LaTeX Logo Request


europ:
**Name:** LaTeX
**Website:** https://www.latex-project.org/
**Official resources for icon and color:** https://www.latex-project.org/img/latex-project-logo.svg
**GitHub:** @latex3

hylickipiotr:
I'll be happy to take care of it, but I have to know what kind of logo is requested.

Latex logo is built with two parts, the sign and text which contain LaTeX phrase. @europ Would you choose between this two propositions **[1]** - sign (humming-bird 🐦), or **[2]** - LaTeX phrase.
![Web 1280 – 1](https://user-images.githubusercontent.com/7335471/57313658-590d0c00-70f0-11e9-8c3f-fa4589010436.png)

sa-tasche:
The bird should have an eye.
I think second should do be used. Because it is well-known.

davidklebanoff:
Hi @hylickipiotr,
We try to avoid using wordmarks as icons in Simple Icons. The bird would be the more appropriate icon for this project.

Could you create a PR with the bird icon? Also, the option here appears to be missing some details. I haven't seen an official monochromatic version of the LaTex logo, but looking at the [official logo](https://www.latex-project.org/img/latex-project-logo.svg), I would:

- Remove the black shadow.
- Cut out the white details such that they are transparent.
- Convert the green body and the green eye into a colorless single path.

europ:
I'm sorry that I did not add the direct link to the logo.

There is a separate repository for this purpose https://github.com/latex3/branding.

Hummingbird (SVG):
https://github.com/latex3/branding/blob/master/latex-project_logo/material/hummingbird.svg

LaTeX phrase (SVG):
https://github.com/latex3/branding/blob/master/latex-project_logo/material/latex-project-logo_old.svg

birjolaxew:
It should be mentioned here that the logo people generally associate with LaTeX is the wordmark. This is mostly because the wordmark can be easily inserted into LaTeX documents, so it is what is used to refer to LaTeX pretty much everywhere. This is reflected on e.g. [their Wikipedia page](https://en.wikipedia.org/wiki/LaTeX) and the search results:

<p align="center"><img src="https://user-images.githubusercontent.com/4542461/57625686-a677f600-7594-11e9-89e7-484e09b3c87c.png" width="50%" /></p>

[...]

PR #1722, opened 2019-10-08
Added LaTeX

phatbhoy67:
**Issue: #1391 **

### Checklist

### Description

Simplified official logo hummingbird logo. (https://github.com/latex3/branding)

Hex value is from full logo and confirmed in colour pallette here: (https://github.com/latex3/branding/blob/master/latex-project_color-palette/color-palette.png?raw=true).


phatbhoy67:
Thanks @ericcornelissen I also have the wordmark only version ready. Should I do a separate PR for that just in case @birjolaxew or @davidklebanoff go that route? Let me know.

I also looked at simplified version of the official logo, below:
 LaTeX Logo

ericcornelissen:
> Should I do a separate PR for that just in case @birjolaxew or @davidklebanoff go that route?

No preferably not, if we decide we want the other icon I would prefer it to be in this Pull Request anyway :slightly_smiling_face:

If you don't want to lose the SVG file of the wordmark you can add it to a `.zip` file and include it in a comment here.


> I also looked at simplified version of the official logo, below:

Having a combination of wordmark and logomark is never a good idea for the intended purpose of the SVGs we provide, at the scale we expect our users to use the logos (i.e. pretty small) having both makes the icon more difficult to read while not adding any value.

phatbhoy67:
[latexwordmarksvg.zip](https://github.com/simple-icons/simple-icons/files/3708393/latexwordmarksvg.zip)

Thanks @ericcornelissen I will wait and update this PR if required. Also attaching wordmark version as ```.zip``` as suggested 😃

birjolaxew:
Although I personally see the LaTeX wordmark as their primary logo, I do agree with @ericcornelissen that "If the brand guidelines bad (in this case the actual logo is not recognizable) that is the problem of the brand and not ours" - with that in mind, I think that the choice of logo in this PR is the right one.

Thanks for your work on this @phatbhoy67! It is much appreciated.

Issue #1729, opened 2019-10-09
Issuu


PeterShaggyNoble:
**Name:** Issuu
**Website:** https://issuu.com/
**Official resources for icon and color:** https://issuu.com/press

There's an SVG logo in the header of that press page or you can scroll down to grab a selection of logos and icons in various formats.

Their [brand book](https://issuu.com/issuu/docs/issuubrandbook/28), which is featured on that page and was published in 2015, has `#f36d5d` as their primary colour but the colour they use on their site is `#f26f61`.

PeterShaggyNoble:
Took a stab at it: https://github.com/simple-icons/simple-icons/files/3707780/issuu.zip

[![issuu](https://user-images.githubusercontent.com/15157491/66489325-53549180-eaa7-11e9-83a3-cc2229575dd4.png)](https://github.com/simple-icons/simple-icons/files/3707780/issuu.zip)

ericcornelissen:
Looks good @PeterShaggyNoble, feel free to open a Pull Request for it :+1:

PeterShaggyNoble:
Working on it now, @ericcornelissen. What do you reckon for the colour?

PR #1733, opened 2019-10-09
Add Issuu icon

PeterShaggyNoble:
First PR here; hopefully I've done this right.

**Issue:** #1729
issuu

### Checklist

### Description
See #1729 for a discussion on the chosen colour.


PeterShaggyNoble:
Looking at #1718: are unspaced `0`s causing issues in some software? If so then the path data for this will need to be updated.

ericcornelissen:
> are unspaced 0s causing issues in some software? If so then the path data for this will need to be updated.

This is probably not the only SVG in recent contributions that have the same issue as #1718. I still need to investigate what exactly is causing the problem in the software and even then the solution is probably going to be with SVGO and/or SVGOMG. If people use SVGOMG the problem from #1718 is likely to occur, if you use SVGO through our local command (i.e. `npm run svgo -- icons/file.svg` then it should now work correctly).

Anyway, I was planning on just manually optimizing each new icon in `v1.18.0` using the command I just mentioned in a single commit. As long as Hacktober is going on (and a bunch of not-very-experience people start contributing) and we have no way to determine what SVGs are problematic and which are not I'm not going to request people to change things if our current CI system approves the PR.

PS. if you want to investigate the problem further, the problem also occurs when building https://github.com/simple-icons/simple-icons-font with SVGs that cannot be loaded by Adobe Illustrator. That is perhaps an easier (cheaper) way to figure out what in the SVGs is causing the issue. Whatever way you investigate this, it must be related to the `noSpaceAfterFlags` option in [SVGO](https://github.com/svg/svgo)s [convertPathData](https://github.com/svg/svgo/blob/master/plugins/convertPathData.js) plugin.

Issue #1660, opened 2019-09-25
Seagate Icon


rodrigoalmeidanet:
**Name: Seagate Inc
**Website: www.seagate.com
**Official resources for icon and color: www.seagate.com

Issue #1616, opened 2019-09-08
TensorFlow


SumNeuron:
**Name:** TensorFlow
**Website:** TensorFlow.org
**Official resources for icon and color:** https://www.tensorflow.org/extras/tensorflow_brand_guidelines.pdf

ericcornelissen:
For anyone interested in adding this to the collection, sn SVG of the logo can be found through the "Download brand assets" button in the PDF. As for the color, I think `#FF6F00` (the light orange) is the best choice to represent the brand, but it can be that `#FFA800` (the dark orange) is better if someone finds a credible source arguing for that.

alexakasanjeev:
Hi @ericcornelissen, I would like to work on this issue, I downloaded the assets and there is a white svg file just  of the logo, but there is some extra space around the logo, what should I do, should I directly upload the svg to the repo or crop first?

Screenshot of how the logo appears in svg
![TF_FullColor_Icon](https://user-images.githubusercontent.com/13250741/64669187-33b55380-d47e-11e9-8bbf-c02a9d42f9a7.png)

ericcornelissen:
Cool @alexakasanjeev! Looking forward for your Pull Request. Let me know if you run into any problems and I (or maybe someone else) will try to help you :smiley:

ericcornelissen:
Any updates on this @alexakasanjeev? I assigned the issue to you, but if you decided to stop working on it I will unassigned you so other can pick up this issue 🙂

alexakasanjeev:
Sorry, been bit buzy, I will surely complete this by the end of 6 oct, if I missed, feel free to assign to someone less lazy 😅

alexakasanjeev:
Hi @ericcornelissen @birjolaxew I'm able to create a SVG of 24 by 24 but as mentioned in contributing guideline 
> - Contain only a single path element.
> - Not contain extraneous elements.
>     - This includes: circ, ellipse, rect, polygon, line, g, etc.
> - Not contain extraneous attributes.
>    - This includes: width, height, fill, stroke, clip, font, etc.

but the svg that is exporting out of Adobe illustrator contains multiple `path` attribute and other attribute like `g`, `defs`, `use`

any help? :disappointed:

ericcornelissen:
[...]

PR #1740, opened 2019-10-10
Adding Seagate Icon

ngrosso:
**Issue:#1660**

### Checklist

### Description

Credits for [freepik](https://www.flaticon.com/authors/freepik)

Got the simplified hex value from SVGOMG optimization


birjolaxew:
Thanks for the PR @ngrosso!

Unfortunately it looks like something went wrong with the PR; nothing is showing up. Usually this means that the actual path is somewhere outside of the 24x24 viewbox.

<p align="center"></p>

Could you look into fixing this? If you need any help you can always ask here and we'll see if we can help you out.

birjolaxew:
@ngrosso Thanks for fixing this! The SVG looks good to me now, except for one detail: the logo seems to be just a bit less than 24pt wide. We want the logos to fill as much as they can, so the biggest side (either height or width) should take up the entire 24pt viewbox.

Could you resize the logo slightly? Your editor should have a way to set the size of a path directly, instead of having to fiddle with resizing it using the mouse (e.g. see [this comment](https://github.com/simple-icons/simple-icons/issues/1616#issuecomment-531005241) if you're using Illustrator) - that should make it easier to make the fit perfect.

birjolaxew:
:+1: Looks good! Thanks for your work on this @ngrosso, it is much appreciated!

Issue #1720, opened 2019-10-08
Octave (Matlab alternative)


foo-dogsquared:
**Name:** Octave 
**Website:** https://www.gnu.org/software/octave/
**Official resources for icon and color:** https://www.gnu.org/software/octave/img/octave-logo.svg
  

A software that offers an environment for mathematical computing. 
Also one of the leading alternatives to Matlab. 

The brand color could be `#0790c0` but I'm not sure.

JoaoPedroMoraes:
I take this pr

PR #1743, opened 2019-10-10
Added Octave icon

JoaoPedroMoraes:
**Issue:** #1720

### Checklist

### Description
From the official SVG I was able to create a monocolor version. https://www.gnu.org/software/octave/

the svg logo was removed from header https://www.gnu.org/software/octave/img/octave-logo.svg


birjolaxew:
:+1: Looks good! Good job on the monochromization!

In case someone comes back to look at this PR in the future, it is worth noting that the color is taken from their website, where the title and buttons use `#0790C0`.

Thanks for your work on this @JoaoPedroMoraes, it is much appreciated!

Issue #1515, opened 2019-07-12
Google Maps Icon


PeterShaggyNoble:
**Name:** Google Maps
**Website:** https://www.google.com/maps
**Color:** *Seems* to be the standard Google blue: `#4285f4`
**Icon:** I haven't been able to find a source for an "official" SVG yet but they do include a Google Maps icon in their [Material Icons Extended font](https://fonts.googleapis.com/icon?family=Material+Icons+Extended) and, for reference, the MDI one can be found [here](http://materialdesignicons.com/icon/google-maps).

PeterShaggyNoble:
[![Google Maps](https://user-images.githubusercontent.com/15157491/66498727-7175be00-eab6-11e9-89db-871b13051e51.png)](https://github.com/simple-icons/simple-icons/files/3708334/googlemaps.zip)

https://github.com/simple-icons/simple-icons/files/3708334/googlemaps.zip

PeterShaggyNoble:
It has been suggested, in #1750, that a monotone version of [this logo](https://commons.wikimedia.org/wiki/File:GoogleMaps_logo.svg) be used instead, if anyone wants to take a crack at it.

Prior to the icon above Google used the following icon for Map notifications, which is a bit closer but not exactly the same and was also [tweaked](https://github.com/Templarian/MaterialDesign/issues/1006) for inclusion in [MDI](https://materialdesignicons.com/): 
![](https://user-images.githubusercontent.com/15157491/66591004-dfd78080-eb89-11e9-82eb-c5fe063b3039.png)

PR #1750, opened 2019-10-10
Add Google Maps

PeterShaggyNoble:
**Issue:** #1515
Google Maps

### Checklist

### Description
I'm not 100% on the colour so I'm open to suggestions on that, as well as a better source URL.


birjolaxew:
Thanks for the PR!

Unfortunately this looks like it isn't the same as the [official Google Maps logo](https://www.google.com/permissions/trademark/logos-list/)<sup>1</sup> - we always want to use the official logo, and an official vectorization if available.

In this case the official logo might be a bit of a challenge to monochromize.

<sup>1. Unfortunately Google only provide it as PNG. The most reputable site I could find with a vectorization is [Wikimedia Commons](https://commons.wikimedia.org/wiki/File:GoogleMaps_logo.svg)</sup>

PeterShaggyNoble:
@birjolaxew, note that the source link is to an icon font maintained by Google themselves so the icon in this PR can be considered somewhat official.

The previous version they had was a *bit* closer to the logo but missing the `G` and with a modified pin:
Google Maps

birjolaxew:
Hmm, I still do think that we want to use the [normal Google Maps icon](https://www.google.com/permissions/trademark/logos-list/). Pinging @ericcornelissen.

ericcornelissen:
I think I agree with @birjolaxew, even though the SVG currently in the PR is more recent(?), Google seems to be primarily using the version @birjolaxew linked to... That said I'm not necessarily against adding the legacy icon since it matches the one Google uses pretty well and it is from an official source.

Your call @birjolaxew on whether we add the [legacy icon](https://github.com/simple-icons/simple-icons/pull/1750#issuecomment-540684958) or the "normal Google Maps logo".

PeterShaggyNoble:
Let me double check that the legacy MDI icon came from an official source.

PeterShaggyNoble:
The old one was used by Google for app notifications (and has since been replaced by the new one) but was manually recreated and tweaked for MDI. Original issue is [here](https://github.com/Templarian/MaterialDesign/issues/1006).

birjolaxew:
@PeterShaggyNoble In that case we would definitely prefer having a vectorization of their current logo in Simple Icons instead. I haven't been able to find an official resource, but [Wikimedia has a vector available](https://commons.wikimedia.org/wiki/File:GoogleMaps_logo.svg).

PeterShaggyNoble:
Sorry for the delay. If that's the one ye would prefer to go with, it's going to take someone with a bit more skill than I to recreate!

Issue #1481, opened 2019-06-19
Request: Minetest


MuhdNurHidayat:
**Name:** Minetest
**Website:** https://www.minetest.net
**Official resources for icon and color:** https://github.com/minetest/minetest/blob/master/misc/minetest.svg

Minetest is an open-source game and game engine, the website itself is ranked [less than 200k on Alexa](https://www.alexa.com/siteinfo/minetest.net) which is obviously lower than the required `less than 500k`.

I linked to the actual logo as the official resource because they did not have media kit yet (but I've opened an issue to discuss on possibility to have media kit on their website - minetest/minetest.github.io#175).

PR #1751, opened 2019-10-10
Added Minetest icon

phatbhoy67:
**Issue: #1481 **

### Checklist

### Description

Removed colour from the [official logo](https://github.com/minetest/minetest/blob/master/misc/minetest.svg) and cut the shapes to form a single path.

Hex value (53AC56) is from the [Header Stripe on the website](https://www.minetest.net/) which is also used for button colour and heading text colour.


phatbhoy67:
> That said, there does seem to be some artefacts in the cave as shown below.

From the original .svg I took the "cave" to be representing a socket/plug/connector and those "artefacts" to be representing pin points of such plug 😕 so I left them in. If it is a cave, I suppose it might be demon eyes? They are the red dots on the [original coloured .svg](https://github.com/minetest/minetest/blob/master/misc/minetest.svg)

Include below some options based on your feedback. Let me know what you think.

mintest-update

ericcornelissen:
> from the original .svg I took the "cave" to be representing a socket/plug/connector and those "artefacts" to be representing pin points of such plug confused so I left them in. If it is a cave, I suppose it might be demon eyes? They are the red dots on the original coloured .svg

I actually did not see the red things inside the cave (or whatever it is) at all! As I thought it where artefacts, I think it is better to leave them out.

I do think the version without the lines is a better fit for the purposes of this project. I think that is our best option. I personally like the filled version as well, but I'm not sure if those changes are too arbitrary... @birjolaxew what do you think?

birjolaxew:
I think that if the eyes become distracting due to being monochromized, then they should be removed as that wouldn't match the intent of the original logo (where they are almost invisible - unless that is just my color blindness speaking). If we can monochromize it without the eyes becoming distracting (i.e. if they only become noticable when zoomed way in), then I think we should keep them; this is mostly for the sake of staying true to the original source.

As for the logo itself, I personally prefer the filled version you showed above @phatbhoy67 - nice work on that! It is significantly more readable at small sizes than the outlined version.

An important thing to note though: it looks like the stroke you've used is a lot thinner than the stroke on the original (see below). This usually happens when the logo is resized without converting the stroke to paths - a 1pt stroke looks a lot bigger on a 4x4 cube than on a 400x400 cube ;) The solution is to convert the stroke to paths before any work is done on it. How you do it depends on your editor (e.g. in Illustrator it's `Object > Path > Outline stroke`).

<p align="center"></p>

Could you please update the SVG to use the same stroke width as the official version? As mentioned I'm personally a fan of your filled version, but @ericcornelissen might disagree. Pinging for his opinion.

phatbhoy67:
minetest-updated

I've updated the stroke width @birjolaxew @ericcornelissen new versions below the dotted line.

B or C for me.

birjolaxew:
@phatbhoy67 Great! Both B and C look good to me too. Guess that leaves @ericcornelissen to pick.

Thanks for updating it!

ericcornelissen:
Good catch on the width of the lines @birjolaxew :+1:

I like both **B** and **C** as well. As **C** is more of an artistic impression then **B** I think we're better of choosing **B** for the project (i.e. it stays truer to the original in my opinion).

Once you updated @phatbhoy67, we can merge this in :ok_hand:

PR #1756, opened 2019-10-12
Add Glitch icon

monika:
### Checklist

### Description
- Logo from official brand assets `SVG` scaled down to 24x24
- Logo paths were combined in Illustrator and then optimized in SVGOMG
- Full-color logo is made up of three different gradients, I chose the blue in the center of the colored icon and also seen in Glitch's default headings, as the background color.
Screen Shot 2019-10-11 at 9 20 40 PM


birjolaxew:
Looks good to me! Considering the color ambiguity I'll let @ericcornelissen merge this in - but I agree with everything I see in this PR. Nice work!

ericcornelissen:
I agree with @birjolaxew their colours are all over the place but blue seems to be somewhat dominant.

Thanks for adding the Glitch icon @monika it is much appreciated! :tada:

PR #1759, opened 2019-10-12
Add hp icon

dennib:
**Issue:**

### Checklist

### Description
Manually vectorized the icon using **`hp.com`** as source.


birjolaxew:
Thank you very much for your PR! The vectorization looks good, especially if you only had the small logo from their header to work off of. I do have a few things to mention tho:

- Very minor detail, but it looks like they use the color `#0096D6` instead of `#0099D8`, according to their CSS. We prefer using the official version where possible, even if the difference isn't exactly visible to the naked eye.
- HP have a high-res version of their logo [here](https://www8.hp.com/us/en/hpi-news/corporate-images.html). If we compare this to your vectorization, we unfortunately see that the vector is a bit off (in the below gif your vector is red, the official version is blue).

<p align="center"></p>

If you want to save yourself the trouble of vectorizing it all over again, we generally accept Wikimedia as a reliable source - and in this case [their HP logo vector](https://commons.wikimedia.org/wiki/File:HP_New_Logo_2D.svg) seems to match the official version pretty well. It would simply have to be resized and annotated.

Thank you for the PR though! We really do appreciate the work that goes into it.

dennib:
Thanks @birjolaxew! 

I will work on the wikimedia version as soon as possible :)

ericcornelissen:
Thanks for your work on this @dennib :tada: Again, it is much appreciated :smile:

PR #1758, opened 2019-10-12
Publish 24 new icons and 1 updated icon, plus dependency updates

birjolaxew:
New icons:

- JetBrains (#1714)
- Mattermost (#1712)
- Gradle (#1724)
- Anaconda (#1728)
- Swagger (#1715)
- HTML Academy (#1716)
- Issuu (#1733)
- Palantir (#1732)
- Sonos (#1739)
- LaTeX (#1722)
- Stackbit (#1699)
- Node-RED (#1727)
- HackerOne (#1723)
- Octave (#1743)
- Lumen (#1741)
- Gitea (#1742)
- Groovy (#1735)
- Kaspersky (#1738)
- Seagate (#1740)
- Quarkus (#1749)
- RadioPublic (#1752)
- Visual Studio (#1757)
- Nokia (#1754)
- Glitch (#1756)

Updated icons:

- Android (#1717)

This also contains the dependency updates from #1703.

---

Lots of activity this week!  
Thanks to all the contributors who submitted these icons, without you we wouldn't be here.

<sup>On a sidenote to @ericcornelissen, please copy the relevant content above into the commit message when you merge this in. That way it'll show up on [our releases page](https://github.com/simple-icons/simple-icons/releases)</sup>

PR #1760, opened 2019-10-12
Patch icons with potential display issues

ericcornelissen:
Solving the same issues as in https://github.com/simple-icons/simple-icons/pull/1718 in preparation of https://github.com/simple-icons/simple-icons/pull/1758


birjolaxew:
:+1: Looks good!

Is this something that we should add to our linting checks, or as an automatic update? I am afraid I'm not entirely sure what this issue is all about, but it looks like it might be an issue for future PRs too?

ericcornelissen:
I still have to investigate the issue but indeed it will affect future PRs as well. As this is being caused by SVGO(MG) it is probably worth opening an issue for it there as well (if there isn't already). 

But like you I can't say exactly what the issue is so I don't know how to check it in CI. Besides, if SVGOMG gives us "incorrect" SVGs then I don't want to bother our contributors (especially with Hacktober going on) with the issue. The best solution would be to encourage contributors to use our `npm run svgo` script (which doesn't have the problem because of the SVGO configuration change in https://github.com/simple-icons/simple-icons/pull/1718)...

PR #1772, opened 2019-10-13
Add Lenovo icon

dennib:
**Issue:**

### Checklist

### Description
Added Lenovo icon from Wikimedia, a bit uncertain about the correct HEX color, I used the one found here: [Scheme Color](https://www.schemecolor.com/lenovo-logo-red-color.php)


birjolaxew:
Thanks for PR!

After a bit of googling I found that they actually have [a page for their logo downloads](https://www.lenovopartnernetwork.com/us/branding/) that we would probably prefer to use as a source. They also have [a visual identity guide](https://www.lenovopartnernetwork.com/us/resource-search/?keyword=visual+identity), which tells us that the official "Lenovo red" color is `#E2231A`.

Could you please update the color and source to the above? I unfortunately won't be able to merge this week due to being on vacation, so I'll let @ericcornelissen check if the logo matches the official version. Nonetheless, thanks for the work you've put in already!

dennib:
Totally miss that page, sure, I'll work on this soon

dennib:
Added new version of the icon and updated the "red" color. :)

dennib:
Mmmh I had, in fact, problems with the precision (the icon would result very off), but at last it seemed it was looking good.

I will look into it (and eliminate the blank line) :)

dennib:
It seems to be a problem with Adobe Illustrator export (not with the optimization part).
I tried to work with it, do you think it's ok now?

ericcornelissen:
Thanks for the updated @dennib, looking good now :smile:

* * *

> It seems to be a problem with Adobe Illustrator export (not with the optimization part).

Do you know (roughly) what the issue was? Care to share it here?

dennib:
> Thanks for the updated @dennib, looking good now 😄
> 
> > It seems to be a problem with Adobe Illustrator export (not with the optimization part).
> 
> Do you know (roughly) what the issue was? Care to share it here?

Sorry, I wasn't able to clearly identify the problem, it seemed like every different combination of svg export options (even from different Adobe Illustrator export commands) would produce the same svg with misplaced "O centers". 
I had to manually move them and export the icon again and again until it seemed correct.

PR #1776, opened 2019-10-14
Added Nim logo

neverix:
### Checklist

### Description

I have used the [Nim logo](https://nim-lang.org/assets/img/logo.svg) from the website. It had two colors, but I picked the top one (`#f3d400`) because it appeared on the page.

I used SVGOMG to optimize the logo and then manually formatted it and joined the paths.


phatbhoy67:
Thanks for your work on this PR @neverix - the [read.me](https://github.com/nim-lang/assets) here in the Nim language repo shows that there are two versions of the logo for use on either a light or dark background. The [version](https://nim-lang.org/assets/img/logo.svg) you chose from the Nim website is the option for dark backgrounds I believe. With that in mind it may be better to modify your version to match the website [favicon](https://nim-lang.org/assets/img/logo_bw.png). The back "crown" has fill and the front "crown" is stroked. That would involve removing the fill from the front "crown" in the version you submitted. Other than that they appear to be the same. What do the @simple-icons/maintainers think?

The colour you chose `#F3D400`, I can only find that in the logo, not within the front page of the site. The hex value `#FFE953`, which is the colour of the front "crown" is used in the styling on the website for buttons and heading text as far as I can see.

[Alexa rank](https://www.alexa.com/siteinfo/nim-lang.org) 267,669

neverix:
@phatbhoy67 Thanks for replying!

Yes, the colors were a mistake on my side. I considered both `#F3D400` and `#FFE953`, I believe I picked the wrong one from a list, but I remember seeing one of those colors on the page.

I will fix the issue right now.

birjolaxew:
@phatbhoy67 Nice research! Usually I would agree that we should use the white-background version, but there are a couple things that make me unsure in this particular case:

- The dark-background version seems to be used by the brand themselves for their website header, social media accounts, etc.
- The dark-background version is (imo) the clearer of the two when it is monochromized.

I do not think we should use the favicon, as that variant seems to be more or less unused elsewhere, and doesn't have as much brand recognition as the other version. I can see arguments for both the white- and the dark-background versions though, so I think I'll leave this one up to one of the other @simple-icons/maintainers to merge.

phatbhoy67:
Hey @birjolaxew my thought process was along the lines of:
- all 3 versions make a distinction between the front and back part of the logo with a colour change of some sort
- when a single colour version was required, the brand made and used the favicon
- the favicon appears to be the dark version with the fill removed from the front part

nim-options

ericcornelissen:
Many thanks for adding the Nim logo to the collection @neverix, it is very much appreciated! :smile:

Issue #1622, opened 2019-09-11
genius icon is missing


haloboy777:
**Name:** Genuis.com
**Website:** https://genius.com
**Official resources for icon and color:** [Resource on MegaUpload](https://mega.nz/#!lxNknKxD!2mLOjTERrQQelXB08hhYKCm9PjrAGn1vT37kS7mRao0)

giovannilima262:
what`s the problem?

ericcornelissen:
It's just a request for the logo of [Genius](https://genius.com) to be added to the collection. If you want to work on it, you can find a vectorized version of the logo in the resource link provided above.

phatbhoy67:
![genius-option](https://user-images.githubusercontent.com/18670428/66779180-7034fe80-eec5-11e9-96d5-87e1dd5010db.png)

I used the wordmark logo from [wikipedia](https://upload.wikimedia.org/wikipedia/en/a/ad/Genius_website_logo.svg) as source, isolated the "G" and rotated it -90 degrees.

Suggest the colour should be `#FFFF64`

If @haloboy777 @ericcornelissen are happy with that, I'll submit the PR.

ericcornelissen:
Looking good @phatbhoy67, I agree with the colour :+1: Did you verify the "G" is turned exactly -90°?

phatbhoy67:
![geniusninety](https://user-images.githubusercontent.com/18670428/66809966-3fd17c80-ef26-11e9-82fe-4252b7c4a0ca.png)

The image on right is the .png version from their twitter profile @ericcornelissen

ericcornelissen:
In that case go ahead and open a Pull Request @phatbhoy67 :+1:

PR #1777, opened 2019-10-15
Add Genius icon

phatbhoy67:
**Issue: #1622 **

### Checklist

### Description

Isolated the letter "G" from the wordmark logo [.svg from wikipedia](https://upload.wikimedia.org/wikipedia/en/a/ad/Genius_website_logo.svg) and rotated it counter-clockwise 90°.

Hex value `#FFFF64` is from the [Genius website](https://genius.com/) header.


ericcornelissen:
> I think that the way it is in this PR is how we want it. The negative-space-G is not centered in their Twitter avatar either, so I don't think we should take it into account. Even if it was, I think we would still want to stick to our current format.

In that case I'm fine with this PR as is, it can always be updated if needed. Thanks for your work on this @phatbhoy67 :tada:

Issue #1086, opened 2018-11-28
Request: Ripple & XRP


anteksiler:
Name: Ripple
Website: https://ripple.com/
Icon: https://commons.wikimedia.org/wiki/File:Ripple_logo.svg

Name: XRP
Website: https://ripple.com/xrp/
Icon: https://commons.wikimedia.org/wiki/File:XRP_Logo.svg

NoSilver2018:
I will take #1086 issue. I'm ready about both icons.

phatbhoy67:
Possible icons for XRP (left) and Ripple (right).

![xrp-ripple-options](https://user-images.githubusercontent.com/18670428/66874613-9d57de80-efa3-11e9-952f-e176d1fa6adf.png)


XRP:
Resized Logo from [website navbar](https://xrpl.org/)
[Alexa rank](https://www.alexa.com/siteinfo/xrpl.org) 213,144
Suggest hex value `#FFFFFF` or `#000000` based on [website](https://xrpl.org/) branding.

Ripple:
Isolated the Ripple Triskelion from the [ripple-logo.svg](https://www.ripple.com/media-kit/).
Suggested hex value `#0085C0` from the [brand guidelines .pdf](https://www.ripple.com/files/brand-guidelines.pdf) "solid blue".

Will this require separate PRs for each icon?

PR #1783, opened 2019-10-16
Add XRP icon

phatbhoy67:
**Partially closes Issue: #1086 **

### Checklist

### Description

Resized Logo from [website navbar](https://xrpl.org/assets/img/XRPLedger_DevPortal-black.svg).

Hex value `#25A768` is from the from [XRP website manifest theme_color](https://github.com/ripple/xrpl-dev-portal/blob/master/assets/favicons/site.webmanifest) .


birjolaxew:

I took the liberty of shortening the source URL so that it points to the main homepage instead of the actual image file. We mostly use the source URL as a way to check for updates to the logo, and in this case I believe the homepage URL is less likely to change in the future while still being fairly easy to extract the logo from. That being said, if you (or any of the other @simple-icons/maintainers) disagree with this decision, please do comment - I am open to changing it back.

Regardless, the contribution is very much welcome!

PR #1799, opened 2019-10-19
added fix position to navbar

norbertzajac:
**Issue:**

Didn't add any new icon, the matter of change is to fix navbar, to have an ability to quicky return to the top or just not get lost while scrolling down.

### Checklist


stylesheet.css
```diff
@@ -316,6 +317,9 @@ a.share-button {
 
 .navbar {
```
ericcornelissen: Making the `.navbar` fixed results in the rest of the page being shifted up by the height of the `.navbar` (see below for an illustration of the problem). Fixing this requires some creativity but I think the easiest way to solve it is to update the padding on the `.hero` element (as I did below). What do @simple-icons/maintainers think? 

| Current | Changes |
|---|---|
| current | changes |

I think the snippet below fits best with the current layout (you can just copy the whole thing and replace it with the existing styling for `.hero`):

```css
.hero {
  font-size: 1.5rem;
  line-height: 2rem;
  margin: 0;
  padding: 7.5rem 1.5rem 1.5rem;
  text-align: center;
}
@media (min-width: 45rem) {
  .hero {
    padding: 9rem 3rem 1.5rem;
  }
}
```

birjolaxew:
I'll be closing this for now. I don't think there's much reason to fix the navbar, since it isn't used for any functionality that you're likely to need after scrolling down. If anyone disagrees with this decision, feel free to comment and I'll reopen.

Issue #980, opened 2018-09-11
Website redesign proposal


birjolaxew:
The current design has a few issues:
- Difficult to see some icons/text (e.g. JavaScript)
- Have to style dark/light icons differently to mitigate this - still not perfect

A few design proposals:
- [Variant proposed by @abenjamin765](https://user-images.githubusercontent.com/3742559/45262339-90ba9b80-b41d-11e8-883f-85c9eed1bd77.png) in #974 (see that PR for code)

- [Variant with drop shadows behind icons](https://user-images.githubusercontent.com/4542461/45362191-78d94800-b5d4-11e8-8b48-1d6a07fd32ab.png)

- [Variant with separated title + subtitle](https://user-images.githubusercontent.com/4542461/45362525-57c52700-b5d5-11e8-89cf-3b56a875fbc8.png)

- [Variant with separated title](https://user-images.githubusercontent.com/4542461/45362604-8cd17980-b5d5-11e8-9e7f-52fa26ffefd8.png)

This issue is for discussing whether we should implement a different website design - and if so, which one.

birjolaxew:
I am personally not entirely happy with any of the proposals, and prefer the current design (although that might just be me being used to it). Other design proposals are *very* welcome.

abenjamin765:
Some things we could explore:
- Since the brand colors are usually applied to the logo, should we explore a treatment that showcases them that way?
- Are there any thoughts about the size of the icons displayed? 
- How can we best care for color contrast to be inclusive of those with low vision or colorblindness?
- Should we make more affordance for the download action? 
- How do you currently gather feedback from users outside of github? 
- What else could we improve?

ericcornelissen:
I do think it is worth discussing a redesign, with the main issue of the current design being color contrast/visibility of some of the text. Also with issues like #755 in mind (and the affordance of the download action, as pointed out by @abenjamin765) it will be useful to redesign the icon blocks.

Of the proposed designs so far I like the [Variant with separated title + subtitle](https://user-images.githubusercontent.com/4542461/45362525-57c52700-b5d5-11e8-89cf-3b56a875fbc8.png). Either way I think using drop shadow to get rid of the dark/light icons problem is a good start, as I think we should keep a colored background behind the SVG.

To answer some of the questions by @abenjamin765:
- _Since the brand colors are usually applied to the logo, should we explore a treatment that showcases them that way?_ Actually, the SVGs do not inherently have any color, and displaying them with a color may be misleading I think.
[...]

PR #1800, opened 2019-10-19
Nzajac add grid item resize

norbertzajac:
Didn't add any new icon, just added grid item resize for better icon inspection by user

### Checklist

### DIFF
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -277,8 +277,8 @@ a.share-button {
             transform: scale(1);
   }
   100% {
-    -webkit-transform: scale(1.5);
-            transform: scale(1.5);
+    -webkit-transform: scale(1.45);
+            transform: scale(1.45);
   }
 }
 @keyframes scale-up-center {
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Thanks for having taken the time to make a contribution! Terribly sorry for the (very) late review but it seems your work slipped through the overwhelming amount of contribution last Hacktober.

This feels very similar to what was requested in #1124. As I was then, I'm not a huge fan of scaling the icons (given the current design of the website) and I would prefer to leave a change like this to people working on #980. What do other @simple-icons/maintainers think?

On the practical side, I think it would be better to implement this with `:hover` and [`transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/transform) rather than CSS animations.

birjolaxew:
I agree. I think that this is an issue that should be tackled with a proper redesign, rather than a bit of a bandaid solution like this. There is definitely merit to being able to see the icons at a larger scale, but I too am not a big fan of the scaling in the current website design.

PR #1801, opened 2019-10-19
add mobile menu

ENeska:
Add mobile menu.

**Issue:**

### Checklist

### DIFF
--- a/stylesheet.css
+++ b/stylesheet.css
@@ -37,10 +37,12 @@ svg {
   height: 100%;
   width: 100%;
 }
+
 #carbonads a {
   color: #263238;
 }
-#carbonads > span {
+
+#carbonads>span {
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


adityasinghbaghel:
### Mobile menu bar.
[menu.txt](https://github.com/simple-icons/simple-icons/files/3760349/menu.txt)
**Please check out this one.**

birjolaxew:
I'll be closing this off for now. I think our current website works decently on mobile (unless the phone is less than 350px wide, which is a rarity), and it looks like this PR has a lot of formatting changes. I think a cleaner approach would be to update the website so the navbar wraps on <350px, instead of collapsing it into a mobile menu - after all, the navbar isn't particularly important on our website, and having it in a menu might give it a bit more attention than I feel is necessary.

If anyone disagrees with this decision, feel free to comment and I'll reopen.

PR #1802, opened 2019-10-19
Analogue

louistsaitszho:
**Issue:** N/A

### Checklist

### Description

If you go to the website, each page has it's own icon color. But since all of their social network icon seems to be just black, I just use that as the hex value.

Also I moved Arduino down. For some reason it not being sorted pissed me off more than it should.


ericcornelissen:
> If you go to the website, each page has it's own icon color. But since all of their social network icon seems to be just black, I just use that as the hex value.

I think black (`#000000`) is a fine choice, their physical products seem to be pretty much black. That said, the logos on their social media seem to be using a colour that is not quite black (I got `#1a1a1a` when picking the colour using Gimp from [their Instragram](https://www.instagram.com/analogue.co/)). @simple-icons/maintainers, what do you think?


> Also I moved Arduino down. For some reason it not being sorted pissed me off more than it should.

That is fine, it is a welcome change :smile:

birjolaxew:
@ericcornelissen `#1a1a1a` also seems to be the color of [their favicon](https://www.analogue.co/assets/favicon-96x96.png) - I think we would prefer using that over `#000000`. Could you please update the color @louistsaitszho?

birjolaxew:
:+1: Looks good! Thanks for your work on this @louistsaitszho!

PR #1807, opened 2019-10-21
Add C#  (#625)

alisen:
**Issue:** #625 

### Checklist

### Description
I picked sourced from Wikipedia, made some certain types of changes in order to tasks.


ericcornelissen:
Thank you for taking the time to contribute @alisen! The Pull Requests looks good overall, you're just missing a `role="img"` attribute on the `<svg>` tag :wink: 

That said, do you have any source (other then Wikipedia) that this round version is some sort of official version? Given the [Google Image result for C#](https://www.google.com/search?q=c%23&client=ubuntu&hs=0u8&channel=fs&sxsrf=ACYBGNQBU3AehD-T5okQ-zDBtoI0VNPrdA:1571636181903&source=lnms&tbm=isch&sa=X&ved=0ahUKEwjvl9qC0azlAhWQfZoKHb0LD08Q_AUIESgB&biw=1866&bih=979) and the [C++ logo we already have](https://simpleicons.org/?q=C++) I would expect the "C#" to be inside a rounded hexagon :thinking: @simple-icons/maintainers, what are your opinions?

alisen:
Thanks for pointing out the missing attribute. @ericcornelissen Actually, I take it from Wikipedia because on this site https://docs.microsoft.com/en-us/dotnet/ they are using the same version like Wikipedia.

alisen:
@ericcornelissen I made the changes. Is there anything to else do it?

ericcornelissen:
@alisen I'm just waiting for feedback from other @simple-icons/maintainers about the style of the logo (circle or hexagon), please be patient :slightly_smiling_face:

birjolaxew:
When it comes to programming languages without official logos, I think we should defer to whatever the most authoritative source we can find is. In the case of C++, that was [the Standard C++ Foundation](https://isocpp.org/) (also used on Wikipedia). In the case of C#, I think that would be Microsoft themselves, who use the circle logo (also used on Wikipedia).

That being said, we'd definitely want the circle to be included. As the PR is now, the "C#" text is simply floating in the middle of the viewbox.

phatbhoy67:
I agree with @birjolaxew, the wikipedia version links back to the [dotnet docs](https://github.com/dotnet/docs/blob/cb475ed45f881e9462e34764480d3b0ebce85e91/docs/images/hub/csharp.svg) so that is probably the right one to use, with the circle.

ericcornelissen:
Thanks for the update @alisen, glad this is merged now 😃 Thanks a bunch for your work on this!

alisen:
I'm glad we made it! @ericcornelissen 😃👾

Issue #1382, opened 2019-04-27
Provide list of integrations/third party extensions on the website


ericcornelissen:
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [x] The website
  - [ ] Other, namely: 

### Description
Provide a list of integrations/Third Party Extensions on the website. E.g. the README currently already lists [the WordPress extensions](https://wordpress.org/plugins/simple-icons/),. Also, recently [simple-icons-font](https://www.npmjs.com/package/simple-icons-font) has been published, and [vue-simpleicons](https://github.com/sh7dm/vue-simpleicons) is also in the works to be released.

<sup>Relevant for #980</sup>

ericcornelissen:
Closing this in favor of https://github.com/simple-icons/simple-icons-website/issues/16

PR #1808, opened 2019-10-21
Add comments

kkaushalya97:
**Issue:**

### Checklist

### DIFF
--- a/index.html
+++ b/index.html
@@ -210,6 +210,7 @@
 </head>
 
 <body>
+    <--start navigation contents-->
     <header class="navbar" role="banner">
         <ul class="navbar__nav" role="navigation">
             <li class="navbar__nav-item navbar__nav-item--home"><a href="/">Simple&nbsp;Icons</a></li>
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Hey @kkaushalya97, thank you for taking the time to contribute! May I ask what your goal is of adding the comments? Since we're using (semantic) HTML5 the description of the comments should more or less be covered by the tag we use. E.g. for the footer you write "_start footer section_", but the tag already says `<footer>`.

Could you elaborate on the intent of your changes?

ericcornelissen:
> I wrote comments for increase the readability

Okay that makes sense. However, for the "start navigation contents" I would also say the HTML itself is expressive enough. The `<header>` has a class `navbar` (short for "navigation bar") and the `<ul>` inside the `<header>` has the attribute _role_  set to _navigation_ (as actually only that unordered list inside the header is about navigation, we might want to use the `<header>` for something else in the future).

We appreciated your efforts but I don't think we need them 😄 You may consider adding [a new icon](https://github.com/simple-icons/simple-icons/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) or perhaps do something else with the side (propose a new issue or work on https://github.com/simple-icons/simple-icons/issues/1382).

More generally I think it is better to make the code itself expressive rather then adding comments for that, this is because you might forget to update the comment when you change the code.

PR #1812, opened 2019-10-22
Revert "Publish 24 new icons and 1 updated icon, plus dependency updates"

birjolaxew:
Reverts simple-icons/simple-icons#1758

The previous release was merged as a squashed commit. This means that the commits on `develop` are not present on `master`, giving us some trouble when we want to merge new releases in. This reverts the previous release so we can merge in more cleanly. Since this is a new commit on the history, the old release should still be available.


birjolaxew:
@ericcornelissen Pretty much. The next `develop -> master` merge will contain all the old commits (from `1.18`) in addition to the commits from next release. The `master` branch will then (assuming git is smart about it) contain all those commits + the two extra commits 8c2925d and fcc6e42. Since 8c2925d would still be in the history of `master`, the old release should still work as planned.

ericcornelissen:
By "_the old release should still work as planned_" you mean the git tag/[GitHub release](https://github.com/simple-icons/simple-icons/releases/tag/1.18.0) (and by extension the release on Packagist, I don't think NPM release care much about what we do on GitHub)? If so, I think your statement is correct and you can merge this 👍

PR #1813, opened 2019-10-22
Gimp icon (simple-icons#1709)

reginareynolds:
**Issue:** simple-icons#1709

### Checklist

### Description

While GIMP has an official logo, it does not seem to have an official color. I found an email thread from several years ago discussing the possibility of creating a design document, but as of now, it doesn't seem to exist. The site also does not have a primary color theme. I used a color picker on the colorized version of the GIMP logo to choose the hex value.

The logo used for the final .svg is located in the header and footer of the page linked in the JSON information. I did not link the image directly in the event that it changes locations on their server.


birjolaxew:
One minor detail (and the cause of the build failing): we use lowercase file names, so `GIMP.svg` should be `gimp.svg` instead. Other than that, this looks good!

birjolaxew:
:+1: Thanks for the update and your work on this @reginareynolds! We appreciate it a lot.

PR #1818, opened 2019-10-23
Mojang icon (simple-icons#1708)

reginareynolds:
**Issue:**

### Checklist

### Description

The hex value is the color value of the logo. I checked the Mojang website for any design documents, but the official source linked in simple-icons#1708 was only about the terms and conditions of using the logo.


birjolaxew:
It actually looks like the header at [mojang.com](https://www.mojang.com/) contains an SVG; the website itself is no longer used, but it looks like the logo is the current one.

reginareynolds:
@birjolaxew You're right, the logo off of [mojang.com](mojang.com) is perfectly square while the previous version is ever so slightly rectangular. The shades of red are also slightly different. I updated the image source and the hex value to reflect the new source .svg. I've also updated the icon based off of the new image source

birjolaxew:
:+1: Great to hear! Thanks for the speedy update and the work put into this PR, we appreciate it a lot.

Issue #1693, opened 2019-10-03
Would love a C and C sharp icon


kchateau:
**Name:**
**Website:**
**Official resources for icon and color:**

nil3sh99:
Name: C and C++ icon
Official Resource: 
1. C++ - https://www.dropbox.com/s/yyg1hqo4ndbjmlz/C%2B%2B.svg?dl=0
2. C - https://www.dropbox.com/s/pxf4aszkoor5pfn/C-Programming.svg?dl=0

ericcornelissen:
Thanks @nil3sh99, where did you find those?

nil3sh99:
Hey @ericcornelissen , i took a lil help from svg icons and modified them to suit the needs of website.

ericcornelissen:
> i took a lil help from svg icons

I'm sorry but I'm not quite sure what you mean @nil3sh99?

el1f:
The C logo proposed in this issue seems to be an alteration of the the C++ logo endorsed by Standard C++, you can find it being used on the [ISO website for C++](https://isocpp.org/).
It seems like the correct C logo has been already added by @PokiDokika anyways so that's great.
The water's a little muddier with the C# logo. There seems to be no actual official logo for C# as referred in [this stackoverflow question](https://stackoverflow.com/questions/2100109/is-there-an-official-logo-of-c) and this [issue in another repo](https://github.com/exercism/meta/issues/27) which seems to leave us with two possibilities regarding which logo to use: [this one](https://camo.githubusercontent.com/0617f4657fef12e8d16db45b8d73def73144b09f/68747470733a2f2f646576656c6f7065722e6665646f726170726f6a6563742e6f72672f7374617469632f6c6f676f2f6373686172702e706e67), which uses the C++ style and has been featured on seeveral other websites or [this other one](https://upload.wikimedia.org/wikipedia/commons/thumb/0/0d/C_Sharp_wordmark.svg/1024px-C_Sharp_wordmark.svg.png) which is being used in Wikipedia.

I am open to vectorizing the one that we decide to end up using.

PokiDokika:
@el1f I submitted a c logo that was based off the c++ logo, but everyone wanted me to remake it in the style of the K&R book. I still have the original if you wish to have it.

ericcornelissen:
@PokiDokika that is correct, the related discussion can be found in that Pull Request for anyone interested. As for the C# logo, you can follow the progress in #1807 @el1f.

[...]

PR #1819, opened 2019-10-23
C Programming Language

PokiDokika:
**Issue:**

### Checklist

### Description

I picked the hex value because the C++ logo (which is really similar to the C logo) uses it. I found the vectorized logo from dropbox (I found the link from issue #1693) and I decided to make it usable for this project. This is my first attempt to do anything in inkscape so I hope that it'll be good enough.


birjolaxew:
This is a tough issue to start off with, because it requires a bit of judgement on which logo to pick ;) 

The good news is that the SVG looks pretty good! There are some minor problems (e.g. the logo looks to be a bit more than 24pt high - did you [resize using the toolbar](https://www.tutorviacomputer.com/inkscape/change-the-objects-size/) which lets you input the size directly?) and a misspelling in the `name` of the JSON which is causing the build to fail, but we'll let that rest for now.

The more pressing problem that we'll have to figure out is what the "official" logo of the C programming language is. This is a bit harder than e.g. C++ which has [Standard C++](https://isocpp.org/) as an authority, or C# which has [Microsoft](https://docs.microsoft.com/en-us/dotnet/) as an authority - I don't think that C really has a an authority that uses a logo. As far as I can tell, two logos have become popular: the logo of [the K&R book](https://en.wikipedia.org/wiki/The_C_Programming_Language), and then the one used in this PR (which apparently is based on the C++ one?).

I think it's difficult to really decide on one to pick. For my money, I'm fine with the version used in this PR. It does not, however, have an official resource, which is kind of unique in Simple Icons. Any thoughts on this @simple-icons/maintainers?

ericcornelissen:
Although I'm personally fine with the version in the Pull Request (and the SVG looks great, good job @PokiDokika!), I do think that the "K&R" version can be consider "official" as [the book it comes from](https://en.wikipedia.org/wiki/The_C_Programming_Language) is (co)written by (according to [the Wikipedia page for C](https://en.wikipedia.org/wiki/C_(programming_language))) the "creator" of the language.

One other thing is that I think the name should be just "C" rather then "C Programming Language" (also, there is a type in "language" in the JSON data file) since we don't use the "Programming Language" suffix for any other programming language.

PokiDokika:
Sorry about misnaming it and all!
I hope that it works now.
Thank you for your feedback!

birjolaxew:
@phatbhoy67 I think that quote refers to the style. It is also written on e.g. [their touchpad icon](https://icons8.com/icon/M9xNyIu2wL3e/touchpad).

That being said, it looks like the consensus is to use the [logo of the K&R book](https://commons.wikimedia.org/wiki/File:The_C_Programming_Language_logo.svg) for C instead of the version in this PR. Could you please update the SVG @PokiDokika?

PokiDokika:
I updated the SVG to match the one on the K&R book.
I hope that it's what you're looking for.

ericcornelissen:
Thanks for your contribution and cooperation on this icon @PokiDokika :tada:

PokiDokika:
No problem! I'm happy to help.

Issue #1619, opened 2019-09-09
Roundcube


TheSeanU:
Roundcube
https://roundcube.net/
https://roundcube.net/favicon.ico

ericcornelissen:
For anyone interested, the header of the website contains an SVG version of their logo :+1:

louistsaitszho:
What would be a good way to simplify it? And what color should it be?

![roundcube_union](https://user-images.githubusercontent.com/3873011/64635703-894b1b00-d400-11e9-9000-196c8c039df6.png)

IMO just merging everything doesn't really work. This doesn't look like "Roundcube" anymore 🤷‍♀

birjolaxew:
I think we'll have to play with some negative space. Here's two variants:

<p align="center"><img height="128" src="https://svgshare.com/i/Ett.svg" />&nbsp;&nbsp;&nbsp;<img height="128" src="https://svgshare.com/i/EuJ.svg" /></p>

The left one is probably the most fitting in this case (the right one would need a bit of cleanup to make all strokes equal width and centered; the above version is mostly to show the concept).

phatbhoy67:
![rc-option](https://user-images.githubusercontent.com/18670428/66783958-0de1fb00-eed1-11e9-966a-6acd5629472e.png)

Possible variation on the left option above by @birjolaxew

PR #1821, opened 2019-10-24
add roundcube icon

jmmendivil:
**Issue:** #1619 

### Checklist

### Description
Created custom shapes based on one from the discussion because the [actual logo](https://roundcube.net/images/roundcube_logo_icon.svg) don't work on monochromatic.
Primary color based on the blue one from the original logo.


birjolaxew:
Thanks a lot for the PR @jmmendivil! This is an excellent monochromization! We'd also like to say thanks for following our contribution guidelines.

The only minor thing I'd like changed before merging this in is that the brand seems to style their name as "Roundcube" when it's written in text, while this PR uses "roundcube" - could you update this PR to use the official capitalization (both in the JSON and in the `<title>`)?

birjolaxew:
:+1: Thanks a lot for the update @jmmendivil, and for all your work on this!

PR #1833, opened 2019-10-25
add Asus icon

xloli20:
**Issue:**

### Checklist


birjolaxew:
Hi @xloli20, thanks a lot for the contribution! Unfortunately there are a few things about this PR that we'd like to update so they follow our [contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#4-annotate-the-icon). Namely:

- It looks like the brand stylizes their name as "ASUS", while this PR uses "Asus". Could you update it (in both the JSON and `<title>`)?
- Could you describe where you found the brand color? Judging by their favicon I would expect a more blue color than `#273347`, e.g. `#00539B` as used on [Wikipedia](https://commons.wikimedia.org/wiki/File:ASUS_Logo.svg). Thoughts @simple-icons/maintainers?
- The `viewBox` should be `0 0 24 24`, with the icon fitting within that box (centered, taking up as much space as possible - in this case that means having a width of 24pt)
- All extra whitespace (e.g. newlines) should be removed. We do this to hopefully keep the icon files as small as reasonably possible.

Other than that, this PR looks pretty good! Thanks a lot for following (most of) our contribution guidelines!

Issue #1671, opened 2019-09-25
Hitachi Icon


rodrigoalmeidanet:
**Name: Hitachi Icon
**Website: www.hitachi.com
**Official resources for icon and color: www.hitachi.com

el1f:
According to this Hitachi [design document](http://www.multimedialab.be/doc/tech/chartes/hitachi.pdf) there doesn't seem to be an easy way to translate the logo into a squarable glyph.

birjolaxew:
I think the approach here would be to simply take their full wordmark and center it in the 24x24 viewbox - similar to [Kaspersky](https://github.com/simple-icons/simple-icons/blob/develop/icons/kaspersky.svg)

PR #1846, opened 2019-10-28
Add Hitachi icon

el1f:
**Issue:** #1671 

### Checklist

### Description
Got the color hex from the website as the design document I found [here](http://www.multimedialab.be/doc/tech/chartes/hitachi.pdf) seems to be set in CMYK mode and has a slightly different hue than the one used in digital media.


el1f:
I got the SVG from the Wikimedia commons [here](https://upload.wikimedia.org/wikipedia/commons/b/b1/Hitachi_Logo.svg).
As you can see their website only has the PNGs so I compared that version with the one used in their website, realized it was a perfect match and went on with that

birjolaxew:
@el1f In that case, could you please update the `"source"` to [the Wikimedia page](https://commons.wikimedia.org/wiki/File:Hitachi_inspire_the_next-Logo.svg)? Since this looks good other than that, I'll get this merged in afterwards :)

el1f:
Oh, you're right about that. Still, I think this is a grey area when it comes to how to approach this change:
The source in Readme is referred to as a link to a document that will enable to realize whether or not the current icon is representative of the brand. In this case the Wikimedia isn't helpful because if Hitachi goes through a rebrand that link isn't going to be useful to understand that the logo is outdated, nor is it going to provide an updated resource to the new logo.

What do you think about it?

birjolaxew:
@el1f While the `"source"` URL is mostly used so we can go back later and find updated versions of the logo, should it ever change, it also gives an idea of how official the logo source is (e.g. whether it's an official vector, a vector taken from Wikimedia, or a manual vectorization of an official raster). In this case the Wikimedia link further includes some important information about where the vector comes from (it appears to be an officially published PDF, where the logo has been extracted from).

I see your point that an eventual rebrand would reach Wikimedia with a bit of a delay, but I think that in this case the need to source the vector itself trumps that. Thoughts @simple-icons/maintainers?

ericcornelissen:
I agree with @birjolaxew, though as @el1f said it is (it almost always is) a gray area. But I think the Wikimedia link is more helpful for the project in this case.

Hence, @el1f once you update  the `"source"` to [the Wikimedia page](https://commons.wikimedia.org/wiki/File:Hitachi_inspire_the_next-Logo.svg) we can merge this 😃 

To the argument that the Wikimedia link is not going to tell us whether the brand details are outdated I would say that neither does a link to the website. If a brand is outdated we usually find out through contributions (be it from contributors or maintainers) rather then from someone going through the URLs in the metadata file and checking if everything is still up-to-date.

el1f:
Done! :)

birjolaxew:
Thanks for your work on this @el1f!

Issue #1854, opened 2019-10-29
arXiv (research distribution repo)


foo-dogsquared:
**Name:** arXiv 
**Website:** https://arxiv.org/
**Official resources for icon and color:** https://static.arxiv.org/static/base/0.15.2/images/arxiv-logo-web.svg 

The SVG can be found on the [support page](https://arxiv.org/help/policies), strangely. 

Suggested brand color is `#B31B1B`. 

Alexa ranking is at [~1.8k](https://www.alexa.com/siteinfo/arxiv.org).

phatbhoy67:
Hey @foo-dogsquared you made the favicon sad.................... 😢 
![sad-favicon](https://user-images.githubusercontent.com/18670428/67793389-a0a69c00-fa72-11e9-8863-eab82ccc8116.png)

foo-dogsquared:
Blame the organization for not featuring him/her/them enough. ;p

PR #1856, opened 2019-10-29
Add arXiv icon

foo-dogsquared:
**Issue:** #1854 

### Checklist

### Description

Seems like the wordmark is more appropriate here since it is pretty recognized and it is the face of their social media ([Twitter](https://twitter.com/arxiv) for a larger example).


phatbhoy67:
Hey @birjolaxew, there is an [official .svg](https://static.arxiv.org/static/base/0.15.2/images/icons/safari-pinned-tab.svg) in this ["about" page](https://arxiv.org/about) source.

birjolaxew:
Hmm. I personally still think that the wordmark is more appropriate than the logomark, solely because the logomark only seems to be used as their favicon and nowhere else. Thoughts @simple-icons/maintainers (maybe specifically @ericcornelissen)?

Issue #1861, opened 2019-10-30
Overleaf (online LaTeX editor)


foo-dogsquared:
Overleaf is an online editor for LaTeX documents allowing for a burden-free start on writing with LaTeX. 
By that reason, it is one of the popular (if not, the most popular) LaTeX-related tool. 

**Name:** Overleaf
**Website:** https://www.overleaf.com/
**Official resources for icon and color:** 

* https://www.overleaf.com/for/press/media-resources - The official media resources page. 
* https://www.overleaf.com/mask-favicon.svg - A more suitable vector. It is also composed only of one path (`<path>`) node so this is more suitable as the source. 

  

Alexa ranking is at [~1.8k](https://www.alexa.com/siteinfo/overleaf.com).

PR #1862, opened 2019-10-30
Add Overleaf icon

foo-dogsquared:
**Issue:** #1861 

### Checklist

### Description

Brand color is extracted from the SVGs linked in the [media resources page](https://www.overleaf.com/for/press/media-resources).


birjolaxew:
:+1: Looks good! Thanks for taking care of thsi one @foo-dogsquared!

birjolaxew:
I noticed just after merging that the `"source"` URL here seems to be a direct link to their favicon SVG. Since they are nice enough to provide an actual logo page, with official SVG download, I think we'd prefer using that as the source. I've pushed 281c19a directly to `develop` (instead of reverting this PR, updating it and re-merging). Feel free to revert if any @simple-icons/maintainers disagree with this decision.

Issue #1853, opened 2019-10-29
Mendeley (research management platform)


foo-dogsquared:
NOTE: I've seen an issue of this one (#25). 
What happened to that issue and to the icon? 
Is this issue still viable to be added? 

Mendeley is a platform for the research community where they can manage and discover research papers. 

**Name:** Mendeley
**Website:** https://www.mendeley.com/
**Official resources for icon and color:**  

There are two SVGs you can choose from the [homepage](https://mendeley.com/), thankfully. 

* https://cms-cdn.mendeley.com/prod/images/logo-full-white.svg
* https://cms-cdn.mendeley.com/prod/knowledge.svg 

Suggested brand color is `#9D1C21` if the search results is any indication.

foo-dogsquared:
Alexa ranking is at [~4k](https://www.alexa.com/siteinfo/mendeley.com).

PR #1864, opened 2019-10-31
Adding mendeley

bonniewhy:
**Issue:**
#1853

### Checklist

### Description

The hex value was provided in the issue, and looking around on the website and the Wikipedia page, that looks like it matches. The logo link that was provided was already an SVG, so I didn't have to manually vectorize it.


birjolaxew:
Thanks for the PR, and for following our contribution guidelines! That is very much appreciated.  
I do have a couple things to ask before I get this merged in:

- Could you update the `"source"` URL to `https://www.mendeley.com/`? It's a bit arbitrary, but we usually prefer having as general of a URL as possible if the SVG itself is easy enough to extract - in this case it seems to be available in the header of their website.
- It looks like the viewbox is just a tiny bit off what we want - the PR uses `viewBox="0 0.016 24 24"`, which makes our build system complain. Could you update that? Given the tiny difference it should be enough to just update the text directly, without bothering with the path data.
- And finally - and this is nonsensical enough that I definitely wouldn't mention it if I wasn't already sending you off to do edits, - based on the CSS of https://dev.mendeley.com/, it looks like the official color is `#9D1620`. Could you please update that too?

Other than those very minor details, good work on this PR so far!

bonniewhy:
Hopefully that's better! Let me know if anything else needs fixing. :)

birjolaxew:
:+1: Looks good! Thanks a lot for the contribution and the update!

Issue #1866, opened 2019-10-31
Request: Notion


ljvmiranda921:
**Name:** Notion
**Website:** https://www.notion.so/
**Official resources for icon and color:**
  
https://www.notion.so/

Sample logo:

![sample_logo](https://user-images.githubusercontent.com/12949683/67964228-e01fe480-fc3a-11e9-86cf-8c291bda13a3.png)

Hello! I think adding a Notion Icon is a great way for linking some Github repos to their respective public Notion pages! Thanks a lot and great work!

PR #1867, opened 2019-10-31
Add Notion icon

bonez0607:
**Issue:** #1866 

### Checklist

### Description

1. **Hex Value:** #000000 - selected from the logo itself. The website is also black and white. I didn't see a primary brand color located in their [media kit](https://www.notion.so/Media-Kit-205535b1d9c4440497a3d7a2ac096286)

2. Logo icon was taken from [home page header](https://www.notion.so/) bar via svg-grabber.


birjolaxew:
Thanks a lot for grabbing this one! It's a bit weird that the SVG in their media kit is actually a PNG embedded in an SVG, but as you point out they luckily have an actual SVG in their website header. The color used on their website is `#040402`, but the PNG from their media kit uses `#000000` so I am perfectly fine with using that.

The only change I'd ask for is to update the `"source"` URL to just `https://www.notion.so/`. This is probably the first time I'm asking to udpate it *away* from an official media kit, but considering that their media kit SVG isn't actually SVG, I think using the page with the SVG is preferable.

bonez0607:
Sounds good - will update this when I get back to my comp

birjolaxew:
:+1: Thanks for the update, and for taking care of this one!

Issue #1673, opened 2019-09-25
BMC Software Inc


rodrigoalmeidanet:
**Name: BMC Software Inc
**Website: www.bmc.com
**Official resources for icon and color: www.bmc.com

PR #1900, opened 2019-11-09
Add BMC Software icon

foo-dogsquared:
**Issue:** #1673 

### Checklist

### Description

The vector and color was extracted from the header logo of the website. 
More specifically, it is from [this link](https://www.bmc.com/etc/clientlibs/bmc/head/bmc_logo_header.svg). 

EDIT: Nevermind, I found the [logo usage guide](https://www.bmc.com/content/dam/bmc/corporate/bmc_logo_usage_guidelines.pdf). 
It seems that they preferred to keep the logo and the wordmark. 
I don't see a rule that allows the symbol to be used exclusively. 
What do you think?


birjolaxew:
Hmm, there's no obvious choice to make here. It looks like they use the full BMC wordmark fairly consistently, barring a few cases where the full wordmark wouldn't fit (e.g. their favicon). On the other hand, it could be argued that our icons are likely to be used exactly in those cases. Without any explicit guiding by the brand itself, it'll have to be a judgement call. Thoughts @simple-icons/maintainers?

birjolaxew:
Pinging other @simple-icons/maintainers again for opinions.

phatbhoy67:
Hey @birjolaxew sorry about the late reply, looking at the brand guideline the full logo is made up of a symbol and logotext. In most cases we isolate and extract the symbol, so I think that is what we should do again in this case.

birjolaxew:
Aight, let's remove the wordmark then. Could you update the SVG @foo-dogsquared?

foo-dogsquared:
The SVG has been updated (or rather reverted to its original commit), @birjolaxew .

PR #1907, opened 2019-11-10
Add Valve icon (#1397)

phatbhoy67:
**Issue: #1397**

### Checklist

### Description

Source is official logo .svg available from header in [Valve Software](https://www.valvesoftware.com/) website.

Hex value `#F74843` is from the logo background colour and is used throughout the [Valve Software](https://www.valvesoftware.com/) website.


foo-dogsquared:
This PR is pretty good for merging, @phatbhoy67 . 

Though, I'm not so sure about the name. 
I'm conflicted between "Valve Software" (the chosen name), "Valve Corporation", and simply "Valve". 
I'm leaning towards "Valve Corporation" since they expanded from creating software at this point. 
(But then again, the title of the logo for Microsoft is not "Microsoft Corporation" so the more likely chosen name is "Valve".) 

What do the other @simple-icons/maintainers think about this?

h1nk:
I'd say just "Valve" as Valve are Microsoft are both corporate organizations that use just their name on their logos and other branding. Searching for images with the query "Valve Corporation" or "Microsoft Corporation" will always lead to logo images with just the text "Valve" or "Microsoft" and it would be more consistent with the titles set for other corporation icons already in this project such as [Microsoft](https://github.com/simple-icons/simple-icons/blob/b56b9be7b3e71e34a168bd1edb3fcc417767f00f/_data/simple-icons.json#L2154-L2158).

birjolaxew:
I agree that just "Valve" is the choice here, since there's no risk of being confused with another variant of the brand (as was the case with Yamaha).

:+1: Thanks for taking care of this one @phatbhoy67!

PR #1908, opened 2019-11-10
Add Steamworks icon (#1397)

phatbhoy67:
**Issue: #1397**

### Checklist

### Description

Source is [official logo .svg](https://partner.steamgames.com/public/images/logo_steamworks.svg) used in the header on [website](https://partner.steamgames.com/). 

Several greys and black used on the (https://partner.steamgames.com/) website. Hex value chosen is `#1E1E1E` which is from the [.png version](https://partner.steamgames.com/public/images/steamworks_logo.png) of the logo. 


h1nk:
I'm not sure if the icon should include the logo text. Maybe it would be better just with the little bit blocks without the logo text like in the favicon: image

phatbhoy67:
> I'm not sure if the icon should include the logo text. Maybe it would be better just with the little bit blocks without the logo text like in the favicon: image

Hey @h1nk I did consider the favicon version but could not find anywhere else that the favicon version is used. 

Do you have any examples where the favicon version is used/currently the preferred option? 

The pattern of blocks in the favicon does not seem to match the blocks in the wordmark version either, and I thought nobody might recognise what it was to be honest. 🤔 

I had a good look around and the wordmark logo seems to be used exclusively, even down to quite small sizes on the Valve/Steam websites/forums/docs in places like footers. 

There is no specific mention of this logo (or favicon) in the (https://steamcdn-a.akamaihd.net/steamcommunity/public/images/steamworks_docs/english/Steam_Guidelines_1217.pdf) linked from this [page](https://partner.steamgames.com/doc/marketing/branding) that I could see so it was a decision made based on the above information to use the wordmark (which I agree isn't normally ideal) that I'm happy to change if that is the consensus.  What do @simple-icons/maintainers say?

h1nk:
I see. The typography and use of font weights in the "STEAMWORKS" wordmark logo is identical to that of "STEAM LINK" in the branding guidlines, so perhaps it would be safer to use the full wordmark logo even though it's not really an icon image. I also wasn't aware the blocks on Steamworks website favicon differed from the ones on the wordmark.

After searching briefly for standalone uses of the blocks graphic I was only able to find one instance, other than the website favicon, by a third party (the @SteamRE OpenSteamworks Organization avatar). So it doesn't seem to have any official usages.

I'd still like to hear from other maintainers about whether they think dropping the wordmark text would be appropriate or not though.

foo-dogsquared:
I think the full logomark with the words is a safer option. 
The logo is as displayed from the [developer documentation](https://developer.valvesoftware.com/wiki/Steamworks) and the [official development group on Steam](https://steamcommunity.com/groups/steamworks).

birjolaxew:
I definitely agree that the full wordmark is the choice here. I have a bit of experience with the dev side of Steam, yet wouldn't recognize Steamworks from just the logomark. This is largely because Valve use the full wordmark pretty much exclusively.

:+1: Thanks for taking care of this one @phatbhoy67!

Issue #1891, opened 2019-11-07
SmugMug


PeterShaggyNoble:
**Name:** SmugMug
**Website:** https://smugmug.com/
**Official resources for icon and color:** EPS & PSD available at https://help.smugmug.com/using-smugmug's-logo-HJulJePkEBf. As per their favicon, the icon should probably just be the smiley face. The green they use is `#6eb800`.

PeterShaggyNoble:
Took a stab at it:

[![SmugMug](https://user-images.githubusercontent.com/15157491/68665562-2bd76580-053a-11ea-8d54-b83776ba39a8.png)](https://github.com/simple-icons/simple-icons/files/3835718/smugmug.zip)

phatbhoy67:
Closed by #1918

PR #1918, opened 2019-11-14
Add Smugmug icon

PeterShaggyNoble:
**Issue:** #1891
SmugMug

### Checklist

### Description
Colour was retrieved from [website](https://smugmug.com/)'s stylesheet


phatbhoy67:
Hey @PeterShaggyNoble excellent work on the .svg 👍 , thanks for your work on this PR.

Happy to merge this one in if we can get to a consensus on the hex value. 

The colour is definitely a green. However, there seems to be a few choices available.

You picked `#6eb800`, which is used on the front page of their site as the background colour for the search icon and as the text colour (:hover) value on some `a` elements. The icon .pngs on the front page also use this colour. I also found `#7FCE69`, `#2ecc71` in other places.

Looking at the [files](https://help.smugmug.com/using-smugmug's-logo-HJulJePkEBf), the .eps (for print) uses `#62be32` for the logo, the .psd (for web) uses `#6db944` for the logo. The .png files for web use `#6db800` on the one with black text and `#6db944` on the one with white text. 🙀 

What do the @simple-icons/maintainers think? 🤔

PeterShaggyNoble:
Well, that's fun! I'd lean towards `#6db944` but only because only one that appears on the list twice.

`#7fce69` & `#2ecc71` are too different, visually from what's used in the logo so I'd eliminate them, at least.

birjolaxew:
I'd agree with `#6DB944`, although that is rather arbitrary. They don't seem super consistent in their color usage, so we probably shouldn't sweat it - but I still would prefer using a color from their logos, rather than their website. Could you update @PeterShaggyNoble?

PeterShaggyNoble:
To confirm: We're agreed on `#6db944`, correct?

If so, I'll get this updated in the morning (GMT).

birjolaxew:
@PeterShaggyNoble Yes, sorry for the confusing wording. `#6DB944` would be my preferred color.

PeterShaggyNoble:
Colour updated.

(Wasn't expecting such a quick reply!)

phatbhoy67:
Thanks for your work on this one @PeterShaggyNoble 👍

phatbhoy67:
Thanks for your work on this one @PeterShaggyNoble 👍

Issue #1915, opened 2019-11-13
Blackberry


PeterShaggyNoble:
**Name:** Blackberry
**Website:** https://www.blackberry.com/
**Official resources for icon and color:** Icon from website logo. Colours from https://www.blackberry.com/content/dam/blackberry-com/Documents/pdf/legal/blackberry-branding-guidelines-external.pdf

Two choices for the colour: `#000000` and `#1475dc`.

[![Blackberry](https://user-images.githubusercontent.com/15157491/68780559-63bfd500-062e-11ea-9abd-b8b74ce80de0.png)](https://github.com/simple-icons/simple-icons/files/3842139/blackberry.zip)

phatbhoy67:
Closed by #1927

PR #1927, opened 2019-11-15
Add Blackberry icon

PeterShaggyNoble:
**Issue:** #1915
Blackberry

### Checklist

### Description
Possible alternative colour: `#1475dc`.
Possible alternative source URLs:
- https://www.blackberry.com/
- https://www.blackberry.com/us/en/company/newsroom/media-gallery
- https://www.blackberry.com/us/en/legal/trademarks (which links to the PDF I used)


phatbhoy67:
Great work on this icon @PeterShaggyNoble 👍 

I agree with the hex value chosen, `#000000`, however since there is a decision to be made best to get the views of @simple-icons/maintainers .

Also, would it be better to cite the [Blackberry website](https://www.blackberry.com/us/en) as the source, there is an .svg in the header and this is more likely to be remain the best source for the future? 🤔

PeterShaggyNoble:
Added blackberry.com to the list of possible alternative source URLs.

birjolaxew:
I agree with `#000000`, but also would prefer just having blackberry.com as the source, given that an SVG is available in the header.

PeterShaggyNoble:
Source updated, checks still pending.

phatbhoy67:
Hey @PeterShaggyNoble nice work 👍

Issue #1851, opened 2019-10-29
Request: EPEL


avylove:
**Name:** EPEL
**Website:** https://fedoraproject.org/wiki/EPEL
**Official resources for icon and color:**
https://fedoraproject.org/wiki/Archive:Artwork/DesignService#EPEL
  

EPEL (Extra Packages for Enterprise Linux), a subproject under the Fedora umbrella, is one of the most popular Linux repos, providing additional packages for RHEL-derived distributions including RHEL, CentOS, Amazon Linux, Scientific Linux, and Oracle Linux.

This was originally requested under #1160

PR #1942, opened 2019-11-16
Add EPEL icon (#1851)

phatbhoy67:
**Issue: #1851**

### Checklist

### Description

Manual Vector of [logo .png](https://fedoraproject.org/w/uploads/9/92/Banner468x120.png).

Hex value `#FC0000` is from the logo.


birjolaxew:
The cut off on the right looks a bit odd to me. I realize that the official version is very low res and also flat there, but I think a cleaner curve into the flat portion would look better.

birjolaxew:
:+1: Thanks for the update @phatbhoy67, and for your work on vectorizing this!

PR #1977, opened 2019-11-23
Add Internet Archive (#1967)

### LINKED ISSUE #1967: Internet Archive (digital library)
A digital library where it hosts free and open access content from various fields. 

**Name:** Internet Archive 
**Website:** https://archive.org/
- https://openlibrary.org/static/images/ia-logo.svg
- A glyph of the icon from their [icon set](https://archive.org/includes/fonts/Iconochive-Regular.woff?-ccsheb) that is used from the header and the image in the hero banner. 
- Suggested brand color is `#000000`. 
  

Alexa ranking is at [~170](https://www.alexa.com/siteinfo/archive.org). 
No questions there.

runxel:
**Issue:**
Closes  #1967

### Checklist

### Description
See Issue.


foo-dogsquared:
The resulting SVG looks pretty off - specifically on the columns of the logo - as seen in the attached image. 



For context, it's just two vectors (yours and the vector from the suggested source) with different colors (pink and green, respectively) overlaid on top on each other. 

IDK how it became different so I advice to double-check if you are indeed using the source to create the SVG.

runxel:
> The resulting SVG looks pretty off - specifically on the columns of the logo - as seen in the attached image.
> IDK how it became different so I advice to double-check if you are indeed using the source to create the SVG.

Huh, that is indeed really odd. I just took it from the source linked in the issue and did not modify it at all. I would blame it on SVGOMG for the moment – I have no other explanation.
Thanks for noticing, I will have a look :)

runxel:
So it turns out I ran into an Illustrator bug/edge case.
I had to redraw the columns in the logo with less points (no problem here, the original had already way to many), otherwise the SVG output would be dented.

foo-dogsquared:
Nicely done, @runxel ! 
The improvements looking good to me. 

Many thanks for your contributions as well! :)

PR #1096, opened 2018-12-03
Add a itunes logo

ezzzin:
**Issue:**

### Checklist
  - [x] I optimized the icon with SVGO or SVGOMG
  - [ ] The SVG `viewbox` is `0 0 24 24`
(I didn't change 'viewbox'.)

PR #2019, opened 2019-11-28
Add iTunes icon (previously abandoned)

runxel:
**Issue:**
get #1096 (abandoned) going again

### Checklist


PeterShaggyNoble:
Was going to add this last week but then thought it might be covered by Apple Music. Perhaps not!

runxel:
> Was going to add this last week but then thought it might be covered by Apple Music. Perhaps not!

You're right, they are pretty similar, yet not the same. (I did miss that Apple Music was already on board)
I let the maintainers decide :)

birjolaxew:
I couldn't find any official guidelines on the actual iTunes logo. The [Apple Music guidelines](https://www.apple.com/itunes/marketing-on-music/identity-guidelines.html#apple-music-icon) (which uses a similar logo) specifically forbids using the circle version, but that seems to be for Apple Music app; the [iTunes Store guidelines](https://www.apple.com/itunes/marketing-on-itunes/identity-guidelines.html#itunes-store-icon) uses the app icon, which is significantly different from the logo of iTunes itself.

Thoughts @simple-icons/maintainers?

phatbhoy67:
The icon used on [iTunes](https://www.apple.com/itunes/) homepage is the one below, so I think the PR version is the correct one to use here. As previously mentioned, similar but not the same as Apple Music so I think adding it is a good idea. 

iTunes Logo

birjolaxew:
:+1: In that case this looks good. Thanks for reopening and taking care of this @runxel!

PR #1339, opened 2019-03-31
Add 48 new brand icons, exported from the MDI library

goyney:
As part of an effort by the [Material Design Icons community](https://github.com/Templarian/MaterialDesign) to direct people to SimpleIcons for their brand icon needs, we are proposing to add 48 missing brand icons that MDI currently has in their library, but SimpleIcons does not.

- [ ] 1Password
- [ ] Accusoft
- [ ] AdChoices
- [ ] Alexa
- [ ] Apache Cordova
- [ ] Babel (Closes #1327)
- [ ] Battle.net
- [ ] Beats by Dre
- [ ] Blackberry
- [ ] Box
- [ ] Creative Commons (Closes #1276)
- [ ] CRYENGINE
- [ ] DLNA
- [ ] Dolby
- [ ] Douban
- [ ] Ember
- [ ] Emby
- [ ] Fedora (Checks one off in #1160)
- [ ] Font Awesome
- [ ] FreeBSD
- [ ] Gentoo
- [ ] Gnome
- [ ] Hacker News (Closes #306)
- [ ] Home Assistant
- [ ] Kodi
- [ ] LastPass
- [ ] Linux Mint (Checks one off in #1160)
- [ ] Litecoin
- [ ] lumx
- [ ] Lyft
- [ ] Material-UI
- [ ] Microsoft Office (Relates to #72)
- [ ] MIDI
- [ ] Minecraft (Closes #280 (again))
- [ ] OpenID
- [ ] Pi-Hole
- [ ] Polymer
- [ ] QuickTime
- [ ] Slackware
- [ ] Square
- [ ] TeamViewer
- [ ] Unity (Closes #1323)
- [ ] Unreal
- [ ] Visual Studio
- [ ] Wunderlist
- [ ] Xamarin (Closes #1308)
- [ ] XDA
- [ ] XMPP

This merge also includes updated metadata with each brand's primary color (if applicable), or the primary color used on the homepage of their respective website. The source website is also included in the metadata.

PR #2030, opened 2019-11-30
Add 1password

runxel:
**Issue:**
From the previously abandoned #1339.

### Checklist

### Description
Logo is taken from their [presskit](https://1password.com/press/).  
Color however is kinda tricky. They always use a gradient and don't specify a genereal flat color.
Also they even never agreed on one gradient internally. Every logo in their presskit has a slightly different one. 🙄
One however seems very a bit more 'designed' with a gradient from `#00AAFF` to `#007DEA`. This means a calculated midpoint of `#0094F5`, which I propose.


birjolaxew:
I agree with the choice of color here. Only minor tweak I'd like made is that they stylize their name as "1Password", while this PR uses "1password". Once that's fixed I'll get this merged in.

runxel:
Good catch! I've updated the PR. Thanks :)

PeterShaggyNoble:
Thanks, @runxel - that's another one knocked off my own list.

Issue #1725, opened 2019-10-08
Quip (Productivity Software) 


ulasaydin:
**Name:** Quip
**Website:** https://quip.com
**Official resources for icon and color:** https://www.vectorlogo.zone/logos/quip/index.html

PeterShaggyNoble:
Took a quick gallop through their website and didn't notice a monotone version of their icon anywhere so the below is just a "best guess". For the colour, it should probably be the lighter orange `#f27557`.

![Quip](https://user-images.githubusercontent.com/15157491/66589542-a81b0980-eb86-11e9-831b-2da29101e743.png)

PR #2046, opened 2019-12-02
Add Quip icon

PeterShaggyNoble:
Quip
**Issue:** #1725

### Checklist

### Description
Still not sure if this is the right treatment of this; I couldn't find a monotone version on their site. Or I just didn't look hard enough!


ericcornelissen:
> Still not sure if this is the right treatment of this; I couldn't find a monotone version on their site. Or I just didn't look hard enough!

If there isn't an official version it is always an artistic interpretation of the logo. Luckily, in this case, you can find a monochrome version in the footer of the website :wink: 

I'm not entirely sure about the colour (@simple-icons/maintainers) but other than that this looks good to me :+1:

PeterShaggyNoble:
That *can't* have been there when I first checked the site - there's no way I could have missed it! ... Could I?!

Thanks, @ericcornelissen; I'll do a comparison between that version and mine tomorrow, see if an update is needed.

PeterShaggyNoble:
There were a couple of *hairline* thin differences. Updated.

Issue #2059, opened 2019-12-03
[COLLECT] Auto makers / car brands


runxel:
Tesla's already on board. So why not adding some more? (we already have petrol stations like Jet, Aral, Shell ... 🤷‍♂) 

- [x] Alfa Romeo (#2744)
- [x] Audi (#2065)
- [x] Bentley (#2892)
- [x] BMW (#2111, #2626, #2737)
- [x] Bugatti (#2745)
- [x] Caterpillar (#2807)
- [x] Chevrolet (#3172, PR in #4282)
- [x] Chrysler (#4357)
- [x] Citroen (#2359)
- [x] Dacia (#4338)
- [x] DAF (#2814)
- [x] DS (#2840)
- [x] Ferrari (#2360, #2431)
- [x] Fiat (#2652)
- [x] Ford (#2632)
- [x] GM (#2638)
- [x] Honda (#2642)
- [x] Hyundai (#2641)
- [x] Iveco (#2810)
- [x] Jeep (#2850)
- [x] John Deer (#2805)
- [x] Kia (#2813)
- [x] Lada (#4361)
- [x] Lamborghini (#2767)
- [ ] Lancia
- [x] Land Rover (#4476)
- [x] MAN (#2774)
- [x] Maserati (#2795)
- [x] Mazda (#2643)
- [x] Mercedes (#2114)
- [x] Mitsubishi (#2631)
- [x] Nissan (#2649)
- [x] Opel (#2361)
- [x] Oshkosh (#2777)
- [x] Peugeot (#2515)
- [x] Porsche (#2635)
- [x] Renault (#2651)
- [x] Scania (#2775)
- [x] Seat (#2351)
- [x] Skoda (#2657)
- [x] Smart (#2831)
- [x] Subaru (#2830)
- [x] Suzuki (#2630)
- [x] Tata (#2637)
- [x] Toyota (#2746)
- [x] Vauxhall (#2647)
- [x] Volvo (#2755)
- [x] VW (#2150)

\* FIN * END * That's it! *
**No more adding. This list is now exhaustive.**

PeterShaggyNoble:
Here's an SVG for Audi to get the ball rolling: https://www.audi.ie/bin/nemo.static.20191203080243/cms4i-nemo/assets/icons/favicon/favicon.svg

Colours can be found here: https://www.audi.com/ci/en/intro/basics/colours.html

PeterShaggyNoble:
[...]

PR #2065, opened 2019-12-03
Add Audi (#2059)

runxel:
**Issue:**
part of #2059

### Checklist

### Description
Icon and color `#BB0A30` are from their [brand manual](https://www.audi.com/ci/en/intro/basics/colours.html).
I chose the red one, because it's the only colorful and distinctive from their set. (Also this is the only color where I'd personally think of when thinking Audi.)

> The brand colours are equal in status.


phatbhoy67:
Thanks for your work on this one @runxel 🏎

Issue #1935, opened 2019-11-15
Quicktime


PeterShaggyNoble:
**Name:** Quicktime
**Website:** https://support.apple.com/quicktime
**Official resources for icon and color:** Haven't been able to find any brand guidelines or vector files for their new logo; all I could find was https://developer.apple.com/softwarelicensing/quicktime/, which is related to their old logo.

PeterShaggyNoble:
Came up with this, using [Icons8's version](https://icons8.com/icons/set/quicktime) as a starting point. Still need to figure out the colour - Icons8 use `#37474f` but they're obviously not an official source.

[![Quicktime](https://user-images.githubusercontent.com/15157491/68964287-e1bbe180-07d0-11ea-8f20-09b710829a03.png)](https://github.com/simple-icons/simple-icons/files/3852624/quicktime.zip)

PeterShaggyNoble:
Closed by #2066

PR #2066, opened 2019-12-04
Add QuickTime icon

PeterShaggyNoble:
QuickTime
**Issue:** #1935

### Checklist

### Description
Very uncertain about pretty much everything in this (see #1935 for more details) so could do with some input from the @simple-icons/maintainers.


birjolaxew:
The SVG itself looks pretty on point, except that the corners of the official version seems to be slightly rounded, while they're sharp in this PR.

<p align="center"></p>

I personally think I would prefer a blue for the color. I think that the primary color of the official logo is the blue center, which also matches well with the blue of the old logo. It's obviously a gradient (and a raster image) so it's hard to pick one color, but I think one of `#17D5FB` or `#1C69F0` would work (those are taken from the lightest part and the part just above the center respectively).

PeterShaggyNoble:
That's some eyesight you have, @birjolaxew! 😆 Leave it with me, I'll see if I can tweak that.

Personally, of the 2 blues, I'd lean towards the darker one.

PeterShaggyNoble:
Does this look a bit closer, @birjolaxew?

[QuickTime](https://github.com/simple-icons/simple-icons/files/3921555/quicktime.zip)

birjolaxew:
@PeterShaggyNoble That looks good! Thanks for updating it. I agree that the darker of the two blues is probably the way to go, so once that's been updated we'll get this merged in :)

PeterShaggyNoble:
Thanks for the help, @birjolaxew - those changes have been made.

birjolaxew:
:+1: Looks good! Since the color is pretty subjective, I'll let another @simple-icons/maintainers weigh in on it before merging this in.

PR #2084, opened 2019-12-04
Add Pi-hole

runxel:
**Issue:**
From previously abandonded #1339

### Checklist

### Description
Logo comes from their github. Color `#FF0000` is the lighter red from the logo.


PeterShaggyNoble:
I'd suggest using the repo or website itself as the source URL, should they ever (re)move that SVG file.

runxel:
Updated the source :)

phatbhoy67:
[Alexa Rank: 42,015](https://www.alexa.com/siteinfo/pi-hole.net)

[official .svg](https://github.com/pi-hole/web/blob/master/src/img/logo.svg)

[Brand Guidelines](https://pi-hole.net/trademark-rules-and-brand-guidelines/)

Hey @runxel the .svg is great, thanks for your work on this one. 👍 

Hex value chosen is red (`#ff0000`), which is the colour you would get if you picked it from any of the images on their website (including the one on the brand guide page 👇 😕)
logo

or indeed the .svg, however the brand guide states that "red" is actually supposed to be `#F60D1A` not `#ff0000`🤔 . 

Even though the image next to where this is stated appears to be `#f6241a` red 👀 

I think the brand guide has to win this one even though the visual difference is minimal and there is inconsistency in its application.

Can you update the hex value and I will merge this one in 👍

runxel:
Haha, good catch! Thanks! 👍

phatbhoy67:
Thanks for your work on this one @runxel 👍

Issue #2094, opened 2019-12-06
Vodafone


joshuah345:
**Name:** vodafone
**Website:** https://vodafone.com
**Official resources for icon and color:** 
`#E60000`
https://issuu.com/logobr/docs/vodafone_core_master_interactive_gu

PeterShaggyNoble:
Closed by #2096

PR #2096, opened 2019-12-06
Add Vodafone

PeterShaggyNoble:
Vodafone
**Issue:** #2094

### Checklist

### Description
Possible alternative source URLs:
- https://www.vodafone.com/, which doesn't have the SVG in the header
- https://brand.vodafone.com/, which requires registration


runxel:
I know you just took the logo from the website... but is it possible Vodafone didn't even themself use the right logo?
Like this thing here: this _can't_ be right! (I mean, it's their own logo, so who knows. But damn, nobody in their right mind would design a logo to be like this...)

what

PeterShaggyNoble:
Is that present in their original, too, @runxel? It may be my screw up - there were 2 paths in the SVG that seemed identical, but perhaps they weren't and I remove the wrong one. Either that or I simply set the precision too low in SVGOMG. Will take another look at this when I have a minute (which may not be for a few days!).

runxel:
Yes, the original is also screwed up, I think. 😄

PeterShaggyNoble:
Yeah, there's 2 layers in the original, neither of which seem to touch the circle properly:



The [SVG on WIkimedia](https://upload.wikimedia.org/wikipedia/en/c/cc/Vodafone_2017_logo.svg) doesn't have that issue so I'll need to update from that.

PeterShaggyNoble:
Updated, which means Wikipedia is now another possible source URL.

runxel:
Still a little bit off 🙈 
off

PeterShaggyNoble:
Dang! Had to zoom *way* in the Wikimedia one to see it but, you're right, @runxel, it breaks the fecking circle, too! 🙄

PeterShaggyNoble:
And, of course, then there are versions where the quotation mark doesn't touch the circle at all! <sigh!>

[](https://www.vodafone.com.au/sites/static/simplicity/assets/images/VF_Logo.svg)

PeterShaggyNoble:
Found a good few other versions where it doesn't touch the circle, either so have updated to match them.

Vodafone

birjolaxew:
It definitely looks to me like the one where the drop touches the circle is the one they use pretty much exclusively.

That being said, when I view the SVG from the website I get one where the teardrop doesn't overlap the circle (in fact it just barely doesn't touch it):

<p align="center"></p>

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
  <path d="M12 0C5.372 0 0 5.372 0 12s5.373 12 12 12 12-5.373 12-12S18.627 0 12 0zm.093 18.65c-3.276.01-6.686-2.785-6.701-7.277-.008-2.97 1.593-5.831 3.642-7.525 2.001-1.654 4.734-2.716 7.216-2.724.32 0 .654.025.859.095-2.172.45-3.898 2.47-3.891 4.761 0 .065.004.13.014.194 3.633.885 5.281 3.079 5.291 6.108s-2.384 6.355-6.43 6.368z"/>
</svg>
```

PeterShaggyNoble:
Thanks, @birjolaxew. Having gone through a few more of their sites, it does look like that is the most common treatment, with the tail not *quite* touching the circle so I've updated to match that now. The next question is, with all these inconsistencies across their sites, which is the best one to use for the source URL? I've gone with vodafone.com for now but the SVG I based this latest version on is from their Egyptian site.

Vodafone

birjolaxew:
Thanks for updating @PeterShaggyNoble!

I think that the appropriate source URL is one where the SVG can be found. It looks like they're *super* inconsistent across their sites (switching between SVG/PNG, switching between quality of SVG), but I'd say that the .nz from earlier (where I extracted my SVG from) or the egyptian one (where you extracted yours from) would fit the bill.

PeterShaggyNoble:
Thanks again, @birjolaxew. Source updated to their Egyptian site.

birjolaxew:
:+1: Looks good! Thanks for taking care of this one (that was a lot of revisions ;) ) @PeterShaggyNoble

Issue #1459, opened 2019-06-07
Icon Request: Dart


mpfaff:
**Name:** Dart
**Website:** https://dart.dev
**Official resources for icon and color:**
* [Color Guide](https://drive.google.com/open?id=1QBhMJqXyRt-CpRsHR6yw2LAfQtiNat4g)
* [Dart logos](https://drive.google.com/open?id=1oOHbcrJen3s9Okgsb5MeYtvr_QV7-ZH_) in AI format.
* [Dart logos](https://github.com/dart-lang/site-shared/tree/master/src/_assets/image/dart) in SVG and PNG format

davidklebanoff:
Given the absence of an official monochromatic version of this icon, I fear any conversion of the current/standard logo to meet Simple Icons guidelines would render it unrecognizable.

ericcornelissen:
That is a good point @davidklebanoff 😅 For anyone willing to pick this up, it might be a good idea to share your idea of how to make it monochromatic here first.

el1f:
I tried my hand at it. What do you guys think about it?
![image](https://user-images.githubusercontent.com/10723208/67685352-34a23600-f995-11e9-827c-db85265fee5d.png)

ericcornelissen:
I like the approach, nice work @el1f :+1: What source did you use exactly?

el1f:
I based it on the SVG in the official Dart repo (the 1080 version to be precise).
I then edited it using Affinity designer.
I'm not super happy about the loss of the transparency in the main side of the dart icon but it seems like a reasonable compromise to me.
Maybe I should thicken the outlines of the empty sections though?

ericcornelissen:
> I based it on the SVG in the official Dart repo (the 1080 version to be precise).
> I then edited it using Affinity designer.

Okay, thanks for the clarification


> Maybe I should thicken the outlines of the empty sections though?

This is about an artistic impression of the logo (since there doesn't seem to be any official guidelines on how to use the logo for the purposes we need). If you're unsure you can create a few version and post them here, see what other people think 😄

el1f:
Here they are!
1. thin lines
2. thick lines
3. thick lines w/ rounded corners

**512w**
![medium](https://user-images.githubusercontent.com/10723208/67992680-fe3b1400-fc3d-11e9-8e2e-29a1d44f2bb2.jpg)

**128w**
[...]

Issue #2061, opened 2019-12-03
SVGO(MG) misery


runxel:
### Kind of issue 
  - [x] Improvement
  - [x] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [x] Everything. 

### Description
I think most of us already know that there's some kind of trouble with SVGO(MG) optimized svgs.
I noticed this again when reviewing #2046: the icon looks fine and most browsers should have no problem at all, but Illustrator will just show an empty file, which to me indicates a problem.
(_It is_ in fact also a problem for us, as @ericcornelissen [already stated here](https://github.com/simple-icons/simple-icons-font/issues/29).)

I found [an issue at svgo](https://github.com/svg/svgo/issues/1164#issuecomment-542310044) where these problems are referenced.
First we should update our [svgo.yaml](https://github.com/simple-icons/simple-icons/blob/develop/.svgo.yml) file to incorporate the recommendations from the svgo issue and stop encouraging users to use the online tool. At least I didn't found any way to put a detailed config there.
Fixing SVGOMG could also be an alternative, but given that this thing has 5 year old issues and four year old PR's I doubt we will come very far this way.

ericcornelissen:
>  I found an issue at svgo where these problems are referenced.
> First we should update our svgo.yaml file to incorporate the recommendations from the svgo issue

I already updated the `.svgo.yaml` file in 5e57e8fef98cffb8363f482ab0860c954328b236


> stop encouraging users to use the online tool. At least I didn't found any way to put a detailed config there.

This would be the easiest solution for us. However, given one of the goals of this project - which is to make it very easy to contribute an icon if an SVG is readily available - it is preferable for us to keep that option open.

In some regards this would also requires us to look at #1532


> Fixing SVGOMG could also be an alternative, but given that this thing has 5 year old issues and four year old PR's I doubt we will come very far this way.

I get that this is cause of concern for you, but I got [a PR merged](https://github.com/jakearchibald/svgomg/pull/141) last year without much trouble and with help from the original author. If your PR is useful it will get attention (at least last year...).

That said, getting this to work would require two major updates

1. Add a toggle to change this setting
2. Add support to import a configuration file so that we can provide one with our desired configuration (as the standard configuration will probably not comply with https://github.com/svg/svgo/issues/1164#issuecomment-542310044)

runxel:
Hey @ericcornelissen, appreciating your fast answer! :)

[...]

PR #2106, opened 2019-12-08
Add Dart logo

el1f:
**Issue:** #1459 

### Checklist

### Description
Had to take a little creative liberty in making it monochromatic. Thanks to @runxel for the suggestions.


runxel:
Hey @el1f, thank you so much for this PR!
The Travis CI build is currently failing, since you forgot a `<title>` element in you svg. Remember: `<title>` should follow the format "[ICON_NAME] icon".

Also I would personally use `https://github.com/dart-lang/site-shared/tree/master/src/_assets/image/dart/logo` as the source. (The 'master' is the difference, so it always reference to the head of the repo)

One last question: How did you got the color?
I would expect it rather to be `#2CB7F6`.

el1f:
> One last question: How did you got the color?
> I would expect it rather to be `#2CB7F6`.

Got it from https://drive.google.com/file/d/1QBhMJqXyRt-CpRsHR6yw2LAfQtiNat4g/view.
Decided to go for the primary color.

runxel:
Great work! Thank you for the quick reply :)

One last thing: your svg still has some inconsistencies to the official svg (might be a precision error, try to bump the precision to 3).
Here it would expect the svg to be like this:
```svg
<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Dart icon</title><path d="M14.842,0.789C14.484,0.429,13.737,0,13.105,0c-0.543,0-1.076,0.109-1.421,0.316L4.105,4.105h0l0,0l0,0 L0.316,12C0.18,12.287,0,12.786,0,13.105c0,0.022,0,0.043,0.001,0.065c0.018,0.669,0.317,1.346,0.789,1.83l9,9h0h0h9.947v-4.263H24 h0V9.947L14.842,0.789z M12.131,1.209l0.034-0.017l0.033-0.02c0.144-0.086,0.48-0.173,0.907-0.173c0.297,0,0.805,0.271,1.029,0.495 l0.001,0.001l0.001,0.001l2.609,2.609H6.34L12.131,1.209z M18.737,23.001h-8.534l-4.73-4.73c-0.397-0.397-0.619-0.934-0.619-1.495 V4.855l13.883,13.883V23.001z"/></svg>
```

Maybe you can update it :)
Big thanks again for contributing!

el1f:
I actually did a little change when it comes to the paths in the original SVGs to try and balance the shape that the monochromatization process ended up producing.

image

As you can see in this screenshot of the original icon compared to mine, the left side border, which replaces the lighter shaded side of the logo, has a thinner border on the top side of the icon. This is due to the fact that when I simply apply a fill to represent the darker shade and an internal border to represent the lighter shade, it ends up producing a logo which loses much of the fidelity when it comes to representing the original logo geometry. Check out the following screenshot to see how overwhelming the filled area ends up looking like.

image
Here is what a more 1 to 1 representation, without the border compensation would end up looking like.
As you can see it kinda ends up misrepresenting the original geometry of the logo in a way that seems much graver to my eye. It may be just me though so I'm open to changes.

runxel:
I'm sorry, maybe I should have been more precise:
If you check your svg closely, you will see that your outline has some weird bumps, where as the original has flat/straight edges.
If you would please correct those issues I'm happy to merge this one in :)

(Also note that the online SVGOMG has some problems, see #2061, so I recommend either use the command line, or tidy up the svg by hand)

el1f:
Wow, I fear this will actually need much more than a simple check! I just tried optimizing it using the SVGO Cli, ditching SVGOMG. Opening the SVG on the browser results in no issue at all with a precision of 3 but look what happens when I open the SVG on Affinity designer...

image

Not good at all. I will try to look into this issue tomorrow in the morning.
Thanks for the feedback!

runxel:
That's why we have a [config yml](https://github.com/simple-icons/simple-icons/blob/develop/.svgo.yml) in the repo which you can use for the SVGO CLi.  :)
Using that one should yield no errors like that anymore (hopefully).

---
The technical explanation here is that both the online and the cli tool are greedy in their whitespace consumption:
`0 1 1.078` will be 
`011.078` after a treatment with svgo. (`0` and `1` are boolean flags, `1.078` is a coordinate)
Which is fine for some browsers, but most editors will fail to recognize that.

el1f:
Thanks for the explanation! I have re-optimized the icon using the config file and now everything works as it's supposed to :)

runxel:
Hey @el1f! Thank you for the quick change.
However, and that one is on me – I should have made this picture earlier, your outline is still wonky. See attached image:

precision

At the moment there are some weird bumps. Please straighten those and we can merge this one in :)

runxel:
Thanks for your work on this one, @el1f! 🎉

Issue #2075, opened 2019-12-04
Ember.js


PeterShaggyNoble:
**Name:** Ember.js
**Website:** https://emberjs.com/
**Alexa Rank:** [~80k](https://www.alexa.com/siteinfo/emberjs.com)
**Colour:** `#e04e39`
**Official resources for icon and color:** https://emberjs.com/logos/

[![Ember.js](https://user-images.githubusercontent.com/15157491/70162539-b6ccfb00-16b5-11ea-8c77-4a212fb7d89b.png)](https://github.com/simple-icons/simple-icons/files/3922992/ember-js.zip)

PR #2143, opened 2019-12-17
Add Ember.js

PeterShaggyNoble:
Ember.js
**Issue:** #2075

### Checklist


phatbhoy67:
@PeterShaggyNoble should there be a container as per the "icon" section of the [logos page](https://emberjs.com/logos/)?
>"In situations where the primary logos aren’t legible (e.g., avatars, favicons) you can use the simple “E” icon. Choose the container shape that best suits the context."

They give 3 different examples. I like the circle.

PeterShaggyNoble:
Good point, @phatbhoy67 - I had read that as the container also being optional.

I like the circle, too, but they use the square one as their favicon so maybe that would be the better option?

PeterShaggyNoble:
I've updated with the square version for now but here are previews of the other 2 versions:
Ember.jsEmber.js

ericcornelissen:
Looking god @PeterShaggyNoble, thanks for the contribution 👍 Also thanks for the nice review @phatbhoy67

PR #2164, opened 2019-12-18
Add AirPlay Audio & AirPlay Video

### LINKED ISSUE #2148: AirPlay & AirPlay Video
**Name:** AirPlay
**Website:** https://www.apple.com/airplay/
**Alexa Rank:** [~40](https://www.alexa.com/siteinfo/apple.com)
**Colour:** `#000000`
**Official resources for icon and color:** https://developer.apple.com/design/human-interface-guidelines/airplay/overview/icons/

I'm not *entirely* sure these are within the scope of SI.

[AirPlayAirPlay Video](https://github.com/simple-icons/simple-icons/files/3974077/airplay.zip)

PeterShaggyNoble:
AirPlay AudioAirPlay Video

**Issue:** Closes #2148

### Checklist


birjolaxew:
Hmm, the question of whether this is in scope is interesting. The question would also encompass other technologies with official logos like Bluetooth, Chromecast, etc.

I think I might be slightly in favor of it being out of scope, but that's a very slight preference. Would definitely want input from other @simple-icons/maintainers

PeterShaggyNoble:
I can definitely see the case *against* adding these and, therefore, other technologies.

Personally, though, I'd lean more in favour of allowing technology logo as we already include logos for standards (e.g., SVG), which also wouldn't, technically, be considered to be brands.

I've requested reviews from the rest of the team, to see if we can reach a consensus.

PeterShaggyNoble:
@phatbhoy67, is that a vote in favour of adding tech icons, so?

phatbhoy67:
> @phatbhoy67, is that a vote in favour of adding tech icons, so?

It's an official trademarked symbol for a technology so I guess that puts it on a par with other brand products which are included. Apple Pay is already added. So in for me. I would argue the same for Bluetooth and Chromecast tbh.

birjolaxew:
> Personally, though, I'd lean more in favour of allowing technology logo as we already include logos for standards (e.g., SVG)

That's a good point. I think you guys have convinced me to be on the include train too.

PeterShaggyNoble:
@runxel, as mentioned in #2061, that option can't (yet) be disabled in SVGOMG, which is what I use.

ericcornelissen:
There seems to be a consensus on whether to add icons like these and I agree 👍 Thanks for adding them @PeterShaggyNoble.

PR #2184, opened 2019-12-19
Add Stencyl (#2124)

### LINKED ISSUE #2124: Stencyl
**Name:** Stencyl
**Website:** stencyl.com
**Official resources for icon and color:** website header has a path
  
[Alea rank](https://www.alexa.com/siteinfo/stencyl.com): ~205k

runxel:
**Issue:** Closes #2124

### Checklist

### Description
Logo (the famous cog `c`) is from the press kit.
The color `#8E1C04` is the calculated midpoint of the colored logo version's gradient from `#BD2408` to `#5F1400`.


foo-dogsquared:
As per the [build job log](https://travis-ci.com/simple-icons/simple-icons/jobs/269036963), there is an extra attribute (`id`) on the `<path>`. 

Furthermore, based on quick inspection, there is a slight curving on some of the spikes of the cog with your PR compared to the source vector. 

Your PR is the red while the source vector is in blue. 
Both of them have half the alpha value so the combined color is basically the common area.

runxel:
Hey @foo-dogsquared, thanks for your review, much appreciating!
You're right, btw! In the original source the spikes look a bit different – they have additional handles which are _sometimes_ slightly bent. I removed those handles as this is clearly an unintended behaviour.
Good spotting, nonetheless! :)

foo-dogsquared:
Looks good to me! 
Nice job for the quick update, @runxel .

PR #2198, opened 2019-12-22
Add Das Erste

vigonotion:
**Issue:**

### Checklist

### Description
See the logo usage on the official website: [https://www.daserste.de/](https://www.daserste.de/)
I used Wikipedia as source because it had the svg version of the logo including the color (also used as primary color on the official website)


foo-dogsquared:
Alexa ranking: [~4.8k](https://www.alexa.com/siteinfo/daserste.de)

PeterShaggyNoble:
Thanks for your work on these, @vigonotion.

Where you've used Wikipedia as a source for an icon/logo, I'd suggest using the German page in the JSON file.

vigonotion:
Sure, I can change it to the german wikipedia, but can you elaborate on when to use the german and when to use the english wikipedia as source? In #1576, I had to change to the english wikipedia.
The wikipedias (https://en.wikipedia.org/wiki/Das_Erste, https://de.wikipedia.org/wiki/Das_Erste) have the exact same logo

PeterShaggyNoble:
@ericcornelissen has been involved with this project a lot longer than I, so I'll defer to his recommendation in this case.

ericcornelissen:
I generally advocate for the English page just so most people contributing to the project can immediately read the page (though in the case of Wikipedia I think most people will know how to change the language). I guess one can make the case that the German Wikipedia page makes more sense for a German brand 🤔

I could be persuaded either way, @PeterShaggyNoble what is your reasoning for preferring the German link?

---

@vigonotion in the case of Shell in #1576 either the Dutch or English (UK) Wikipedia would make more sense than the German one, I think

PeterShaggyNoble:
That was my reasoning, too, @ericcornelissen: using the German wiki for a German brand. In the case of a country-specific brand, I think we should endeavour to use the wiki in that country's language. In the cases of international brands *based* in non-English speaking countries, I think either/or would do but, for consistency, we should probably favour the English wiki.

ericcornelissen:
Another thing to consider is that we generally use the **Wikimedia** link (https://commons.wikimedia.org/wiki/File:Das_Erste_2014.svg in this case) rather than the **Wikipedia** link, which (I think) is always in English (thought a quick search in [simple-icons.json](https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json) shows we haven't been 100% consistent with that).

Hence, using the Wikimedia link will probably solve our problems, what do you think @PeterShaggyNoble ?

PeterShaggyNoble:
The problem, as I see it, with using the Wikimedia link is that if a logo is updated then the file name may be changed. By using the Wikipedia link instead, we're guaranteed that that will always display the latest version.

ericcornelissen:
> The problem, as I see it, with using the Wikimedia link is that if a logo is updated then the file name may be changed. By using the Wikipedia link instead, we're guaranteed that that will always display the latest version.

Good point, @simple-icons/maintainers (and in particular @birjolaxew) what do you say?

PeterShaggyNoble:
The SVG looks good to me and I don't think the source URL is a big enough issue to block merging this at this time, we can always come back to it.

Nice work, @vigonotion; thanks for the contribution.

PR #2201, opened 2019-12-22
Added RTL

vigonotion:
**Issue:**

### Checklist

### Description

This one was harder. It is a TV Channel and the logo seen in the program looks a little bit different (inverted): [Imgur](https://imgur.com/o1sBv7v)
The website logo has a ".de" for the web page at the end and the Wikipedia page (https://en.wikipedia.org/wiki/RTL_(German_TV_channel)) lists a 3-colored version.

I chose this version because it has the font that is present in all versions and was the one available as svg already.


foo-dogsquared:
Alexa ranking: [~8.9k](https://www.alexa.com/siteinfo/rtl.de)

PeterShaggyNoble:
Thanks for your work on this one, @vigonotion, the SVG looks good to me.

Given that the version you've used is from their official website, I think we're OK to use it. (Personally, I'd associate the version in your Imgur link more with RTL but that's probably a biased opinion from *watching* RTL.)

Although, just to throw a spanner in the works, their parent company do use the version from Wikipedia on [their website](http://www.rtlgroup.com/en/about_us/businesses/tv-broadcast.cfm) but that's not going to work as well in monochrome.

As there's a choice here, we'll need another one of the @simple-icons/maintainers to OK it too before we can merge it in.

runxel:
Hm, I would have expected it to be the boxes as solid with the letters knocked out....

ericcornelissen:
This is a rather difficult one as the branding is different in different countries... We could make this PR specifically for the German part (and change the name to `rtl.de`(?)) in which case the current SVG is fine I think. This, I think, is the way to go (in case we want to add the logo for RTL in another country later).

If we want to stay with RTL as a generic brand (separate from RTL group) I think we should, as @runxel points out, use the solid boxes with letters knocked out. This is based on what comes up when I search Google for "RTL" and the logos used on their social media.

Thoughts?

ericcornelissen:
Closing this for now due to a lack of activity. If you want to continue your work, @vigonotion, feel free to re-open the Pull Request!

For anyone else interested in working on this logo, feel free to use what is here as a source :slightly_smiling_face:

Issue #2116, opened 2019-12-11
[COLLECT] News Outlets


runxel:
- [x] Washington Post
- [x] Times
- [x] CNN
- [x] The Register
...

PeterShaggyNoble:
Closing this out as we now have PRs for the four requested brands.

PR #2224, opened 2019-12-24
New York Times

runxel:
**Issue:** #2116

### Checklist

### Description
They use the `T` as their favicon and also on the website you can find it at several places as icon.
I took it from the header of this [site](https://www.nytimes.com/interactive/2019/12/23/travel/germany-lost-luggage-auction.html) (Sometimes special stories get some special treatment regarding the layout. Here they just use the famous `T`, not even writing the full name.


phatbhoy67:
Thanks for your work on this one @runxel 👍

PR #2226, opened 2019-12-24
Add Bäderland

vigonotion:
**Issue:**

### Checklist

### Description

As there is no obvoius "small" logo variant and the whole word mark is too wide, I used the design they used as favicon on their website (https://www.baederland.de/). Colors and SVG taken from Wikipedia as stated  in source.


runxel:
Hey @vigonotion,
thanks for your PR. :)

While it technically fulfills the requirement (Alexa rank: ~460k) I'd still argue it is out of scope for this project. What I mean: it's not a global operating brand – it's a _swimming bath_. Why would anybody outside Hamburg find this useful?

vigonotion:
Hi, I can understand your point. I had to create these logos for a personal project, so I could as well share them with the community should somebody need it.

I don't rely on the icons being in upstream simple icons font, though I think it won't hurt anyone but help maybe some people (from Hamburg).

runxel:
That is very kind of you and we appreciate every effort to make this project better.
However since I am not sure here let's ask the @simple-icons/maintainers what they think about this one :)

PeterShaggyNoble:
I agree with @runxel that this is extremely niche and my knee-jerk reaction would be to not include it.

However, if we start down the road of rejecting brands that seem too niche, despite otherwise meeting our requirements for inclusion, we may run into problems as the "nicheness" of a brand is going to be largely subjective.

Therefore I'm actually leaning very slightly in favour of letting this one in with the caveat that, as it's Alexa rank is so close to our cut-off point, we give it a bit of time to see if there's any fluctuation - it's current rank could be a seasonal anomaly, for example.

ericcornelissen:
I do agree that this is a bit too niche as well, even though it has a high enough Alexa rank. Addressing @PeterShaggyNoble concern, I think we can always reconsider at a later point in time.

Thanks for the consideration and effort @vigonotion but I will close this for now. If anyone wants to continue the discussion feel free to do so here 🙂

vigonotion:
Just one thing: we should update the CONTRIBUTIONS file, because it says _icon requests_ should have the appropriate Alexa rank, when I first read it I thought there is no Alexa rank limitation on actual pull requests. So there should be some information about what icons are gonna be accepted (not just the icon requests).

Anyhow, I'm totally fine with the decision not including the icons in upstream.

runxel:
> Just one thing: we should update the CONTRIBUTIONS file, because it says _icon requests_ should have the appropriate Alexa rank, when I first read it I thought there is no Alexa rank limitation on actual pull requests. 

Interesting way of seeing this. Would have never guessed that the CONTRIBUTING file is misleading. While you're right, it's not explicitely stated that a PR should match, I personally think we are implying it quite heavily.
I could see the text restructured to take care of the fact that indeed anybody can make a PR without requesting it first.

PR #2245, opened 2019-12-28
Corona game engine

runxel:
**Issue:** _none_

### Checklist

### Description
Logo is as svg in the [website's](https://coronalabs.com/) header, color `#F96F29` included.


phatbhoy67:
Hey @runxel the .svg and colour are great 👍 

Should this be added as "Corona Game Engine" or "Corona Labs"?

PeterShaggyNoble:
Their [brand guidelines](https://coronalabs.com/presskit.pdf) allow for just "Corona" *but* it doesn't look like those guidelines allow for the icon by itself.

runxel:
So, _Corona Labs_ is the company behind the engine. I could see that we add it as `Corona engine`.

You, @PeterShaggyNoble, found the presskit regarding the company. They share the logo (and part of the name) with the engine. I'd argue while it's true that the name should be added for the company the same doesn't hold true for the _engine_. It should be okay to add the logo without the name since I want to add the _engine_ rather than the company.

phatbhoy67:
I think "Corona Engine" is a good choice @runxel, and I also think the graphic element without the wordmark is acceptable for use here @PeterShaggyNoble

engine

runxel:
Great! Thanks for reviewing :)
Updated the name to be `Corona Engine`.

phatbhoy67:
Nice work @runxel 👍

PR #2261, opened 2020-01-02
Add Pexels

### LINKED ISSUE #2156: Pexels (copyleft content hosting)
**Name:** Pexels 
**Website:** https://www.pexels.com
**Official resources for icon and color:** https://www.pexels.com (all I managed to find is the embedded SVG on the header)
  

Alexa ranking: [~429](https://www.alexa.com/siteinfo/pexels.com)

PeterShaggyNoble:
Pexels

**Issue:** Closes #2156

### Checklist

### Description
Icon & colour from logo in website header. Possible alternative, based on [this SVG](https://www.pexels.com/assets/icons/safari-pinned-tab-c3a22dad034351c9cda8c3bb257bc04b235e17785a945a3295ed21d0c340c7de.svg):

Pexels


PeterShaggyNoble:
I also haven't seen them using the unboxed version anywhere so I'd say it's safe to disregard it. Even moreso because I've since noticed that it was generated by [Potrace](http://potrace.sourceforge.net/) and, the more icons I see that are made with it, the more I'm becoming of the opinion that we shouldn't accept them as an official source.

It seems to me like there's something out there that's automatically running raster favicon files through Potrace and adding the `link` tag to the HTML of sites without anyone first reviewing them. While some of them turn out fine, most seem to be inconsistent with the branding or of very low quality or both.

ericcornelissen:
It seems to me that indeed the boxed version is the way to go 👍 


> While some of them turn out fine, most seem to be inconsistent with the branding or of very low quality or both.

It is probably okay that they're low quality because they're always displayed at a small size (namely as a pinned tab in the Safari browser, guessing from the file name). To be fair to Potrace, the standard raster-to-vector algorithms in Illustrator etc. have the same issue. I do agree that we should not use these SVGs as an official sources as the quality is usually too low.

PR #2310, opened 2020-01-08
Update Order of Icons

PeterShaggyNoble:
Over time, with so many people working on it, some of the icons in `_data/simple-icons.json` have become out of order, which affects the order icons are displayed in when sorting alphabetically on the site. I didn't want to commit this directly, in case I screwed something up - it needs another eye or 3 thrown over it.

The list of icons is up to date as of [commit `355afa7e55643c47be6d428d0121085d59b6d020`](https://github.com/simple-icons/simple-icons/commits/develop), which added the Just Eat icon, so if any other PRs are merged before this one, those icons will need to be added here too. The @simple-icons/maintainers should be able to commit directly to my branch for this.

We may also need to review all open PRs after merging this to resolve any conflicts it may create.

I wonder is there a way to automatically sort the list each time we merge from `develop` to `master`?


runxel:
Who knows about Travis CI?
Maybe it could already be checked in every submitted PR.

birjolaxew:
It should definitely be checked on every PR, but it unfortunately isn't right now. This is mostly because [jsonlint](https://github.com/zaach/jsonlint) (which we use for linting the file) doesn't have a feature for it. We would probably have to create a quick script in our `scripts` folder that lints it, and add that to the Travis build.

PeterShaggyNoble:
Our other option is, given that this primarily affects the alphabetical sort on the website, that we update the website to sort the array when pulling it in, before applying the `data` attributes. That way we don't need to worry if the JSON file gets out of order and can just fix it manually every once in a while.

PeterShaggyNoble:
Our other option is, given that this primarily affects the alphabetical sort on the website, that we update the website to sort the array when pulling it in, before applying the `data` attributes. That way we don't need to worry if the JSON file gets out of order and can just fix it manually every once in a while.

birjolaxew:
I still think I'd prefer having linting take place in our CI, just for the sake of keeping our JSON organized. I've implemented this in #2334.

Regardless, that is another issue and should be handled in a separate PR. Thanks for noticing and fixing this @PeterShaggyNoble!

PR #2316, opened 2020-01-08
Add the use of icons to react

wootsbot:
### Description
Add in the`README.md` the use of the [@icons-pack/react-simple-icons](https://github.com/mamut-dev/icons-pack/tree/master/packages/react-simple-icons) package to be able to expand the uses of the icons as components in `ReactJs`.


birjolaxew:
Thanks for implementing a React binding!

We currently only have officially supported usages in the *Usage* section of our README. Since this is a third party binding, we'd prefer having it listed under the *Third Party Extensions* section instead, with the same format as the others. Could you update to that instead?

PeterShaggyNoble:
Thanks for updating, @wootsbot, merging this in now,

PR #2318, opened 2020-01-09
Add FedRAMP (US government compliance program)

dacut:
* Add fedramp.svg

* Add FedRAMP data

**Issue:**

This is the logo for the US FedRAMP compliance program (fedramp.gov); that website has an Alexa rank of 192,588, but I'm actually using this for other federal agencies (e.g., nasa.gov, 1,352).

### Checklist

### Description
Logo was vectorized in Adobe Illustrator by tracing the PNG files from the FedRAMP logo guidelines (alas, they do not provide vectorized logos) using guides at the intersect points on pixel boundaries at 64000% zoom. (Auto vectorization was not so good.) See attached working sample.

The hex value was taken from the official FedRAMP logo guidelines PDF.

FedRAMP Vectorization


PeterShaggyNoble:
Thanks for your work on this one, @dacut 👍

Although, I personally prefer the version you've used and it is in an option they offer in their brand guidelines, I do think we should go with their boxed "official approved FedRAMP logo". What do the other @simple-icons/maintainers think?

dacut:
I have an objective reason -- and a selfish one :-) -- for choosing the unadorned version.

* The blue square/circle does not have a flat color version in either the guidelines or the kit, but is always a gradient (!). The red version is flat, but I have never seen it used outside of the branding guidelines (and I suspect we're probably amongst the dozen folks outside the GSA who have actually read the branding guidelines :-)). My thinking here: if someone wants to use Simple Icons to reproduce this accurately, they would render the square themselves and insert the FR logo in white.
* My selfish reason: I want to use this as a build badge via https://shields.io/ to indicate whether a package uses only FedRAMP-compliant services or not. Adding the square would make it somewhat less legible there.

But I'd be happy to redo it in the square/inverse version if that's the way folks sway.

runxel:
Yeah, it's clearly stated that a unboxed version is also okay to use. 
And if this is the case I'm always a big advocate on that :D

ericcornelissen:
Thanks a lot for the manual vectorization @dacut, it is much appreciated! 🎉

PR #2334, opened 2020-01-10
Add CI step for ensuring alphabetical order of icons

birjolaxew:
Implements linting to be executed by our CI to ensure that the list of icons is kept alphabetically ordered. Also see #2310.

The file itself is written to make it relatively easy to add new linting steps if we need any in the future, but for now it only contains the alphabetical linting.


birjolaxew:
Looks like GitHub removed the old commits after I force-pushed the cleaned branch, so I'll just go ahead and push a test commit again.

PeterShaggyNoble:
Nice work, @birjolaxew. Could probably do with being tested with multiple icons out of place, too.

I'd prefer if we could take care of the sorting on our end, rather than the contributors having to worry about it. But if that's not possible then this is the next best thing. We should document somewhere how we expect the array to be sorted as some people might ignore non-alphanumeric characters when manually adding entries (e.g., reading `.NET` as `NET`).

birjolaxew:
I suppose we could do something along the lines of having the CI push a commit with a fix if any commit on `develop` doesn't satisfy the linting. Would probably be a bigger task, but I could see it being useful for other things.

Alternatively we could just sort it ourselves on the website; I personally like having the JSON neat and sorted, but that might just be me ;)

runxel:
I think we can safely do all of it! 😇

- The linting is good as a quick fix.
- If we have the time and resources I can totally see the CI being able to push a commit so that's also automated.
- We definitely should say how we expect the thing to be sorted.
- And of course we can have a sorting method implemented at the website. [0]

---

[0] The sorting by color looks kinda weird to me btw. But I understand that this kind of two dimensional sorting is hard.

PR #2360, opened 2020-01-13
Ferrari Corp (#2059)

runxel:
**Issue:**
#2059

### Checklist

### Description
The wordmark was in the header of the website. We could add the horse, but that's just an extra, the wordmark stands on it's own normally.
Color is debatable. I think it _must_ be red: `#F20300` (eyedropped from website).
But it could also be the yellow `#ffff01` which is used as background, when you have the horse.

---

Addition: The scope is now slightly turned; it's about the Ferrari corporation as whole, not only the _cars_, that's why it is the wordmark and not the horse – which will be added with another PR.


PeterShaggyNoble:
From their websites' stylesheets, it looks like the colour should be `#eb2e2c`. Although a strong case could probably also be made for [*rosso corsa*](https://en.wikipedia.org/wiki/Rosso_corsa), `#d40000`.

PeterShaggyNoble:
@runxel, you might need to up the precision on the logo; there are a couple of hairline differences in the `e` and the `a` between your version (red) and the original (green).

Ferrari

However, in the absence of any brand guidelines that I could find, I think we should use the horse on its own, as they do in their favicon and splash screen - it lends itself better to a 24*24 icon.

runxel:
> 
> 
> @runxel, you might need to up the precision on the logo; there are a couple of hairline differences in the `e` and the `a` between your version (red) and the original (green).

Oh yeah, that was being expected. 
Their `a` has some problems I've taken care of. Also on of the handles at the `e` is also most probably false. 
I've to kindly decline to revert at least the change on the `a` 😅

phatbhoy67:
There is an official .svg version of the [horse](https://media.yoox.biz/ytos/resources/FERRARI/img/logo-scuderia-ferrari.svg) in the header of their [store website](https://store.ferrari.com/en-gb/?from=FT).

runxel:
Good find, @phatbhoy67!

I'd still say, the horse alone is not enough to present the brand :/
Maybe together with the wordmark?

phatbhoy67:
The wordmark is iconic. The "prancing horse" motif is their icon.

runxel:
I know, I know... So if everybody is for _just_ the horse I will do so!

phatbhoy67:
If you asked me to describe it without looking I'd probably include the shield. 🤔

PeterShaggyNoble:
The version with the `SF` in the shield is the logo for [Scuderia Ferrari](https://formula1.ferrari.com/en/).

But they do have many different versions of it. Looking at [a random car](https://www.ferrari.com/en-IE/auto/812-superfast) they have the shield version on the side panels, a rectangular version on the bonnet and the prancing horse on its own on the grille, which is the version they use on their website.

An option could be to add both, as was done with Yamaha; the shield version as Scuderia Ferrari and the version below (extracted from their website source) as Ferrari.

Ferrari

runxel:
> An option could be to add both, as was done with Yamaha; the shield version as Scuderia Ferrari and the version below (extracted from their website source) as Ferrari.

Wait, don't you mean the shield with the horse as _Scuderia Ferrari_, and _Ferrari_ with the wordmark?
Otherwise it's just the horse two times... 🐴

PeterShaggyNoble:
I'm suggesting the horse on the shield with the `SF` Scuderia Ferrari and the horse on its own, as above, for Ferrari.

runxel:
Okay, I see.
Nevertheless I think that we don't need the horse twice. (doesn't matter with or without shield)
I would at least expect the wordmark for once.
But lets listen to someone else from the @simple-icons/maintainers! :)

ericcornelissen:
I would say the horse (without shield) is recognizable for Ferrari, personally I think it is more recognizable than the wordmark. The horse is used in various forms (in a shield, in a square, in isolation) and is used to represent their brand on all of their social media (based on which we could argue it should be inside rectangle). Also, it is more appropriate for our purposes, I think. 

I might be confused about the difference between Ferrari and Ferrari Corporate but for **Ferrari** I think we should definitely use the horse (without shield). The horse with shield could be added, on request, for Scuderia Ferrari. Potentially we could add the Ferrari wordmark for Ferrari Corporate?

But correct me if I got the "Ferrari" brand incorrectly here 😅

phatbhoy67:
Just the horse without a-neigh accoutrements for me 👀

runxel:
Okay great! 💯 
To sum it up:
- `Ferrari` → horse, nothing else.
- `Scuderia Ferrari` → horse in shield (later PR)
- `Ferrari Corp.` (_this PR_) → the wordmark.

---

Maybe you want to make a Ferrari PR, @PeterShaggyNoble, since you already have it all together? :)
That would be sick.

> Ferrari

PeterShaggyNoble:
Yup, will do so Monday. Can you open an issue and assign me as a reminder?

ericcornelissen:
Just wondering, why `Ferrari Corp` rather than `Ferrari Corperate` (or the native `Ferrari N.V.`, see [Wikipedia](https://en.wikipedia.org/wiki/Ferrari)) 🤔

runxel:
> 
> 
> Just wondering, why `Ferrari Corp` rather than `Ferrari Corperate` (or the native `Ferrari N.V.`, see [Wikipedia](https://en.wikipedia.org/wiki/Ferrari)) 🤔

Good point, like always!
Will change that to `Ferrari N.V.`
Does the file be named `ferrarin-dot-v-dot` then? 😅

ericcornelissen:
> Does the file be named ferrarin-dot-v-dot then?

I think so 🤔😅 


> Will change that to `Ferrari N.V.`

I'm equally fine with `Ferrari Corporate` which would improve the filename 😄 Any other opinions on which name we should use @simple-icons/maintainers

PeterShaggyNoble:
All looks good now, @runxel but, before merging, [where did we land on the colour](https://github.com/simple-icons/simple-icons/pull/2360#issuecomment-573762506)?

runxel:
You already got the rosso corsa, so I will take the red from the stylesheet :)

PeterShaggyNoble:
Thanks for all your work on this one, @runxel

PR #2365, opened 2020-01-14
Add icon count to website & README

PeterShaggyNoble:
We could do with touting the fact that, come the next release, we'll have over 1000 icons in our library ([Wear OS was the 1000th](https://github.com/simple-icons/simple-icons/pull/2305#issuecomment-573322864)).

Do we want the count on the website to be live or not (assuming it's possible)? Personally, I always prefer the "over `X`" phrasing to the exact figure - dunno why that is!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 </a>
 <h3 align="center">Simple Icons</h3>
 <p align="center">
-Free SVG icons for popular brands. See them all on one page at <a href="https://simpleicons.org">SimpleIcons.org</a>. Contributions, corrections & requests can be made on GitHub. Started by <a href="https://twitter.com/bathtype">Dan Leech</a>.</p>
+Over 1000 Free SVG icons for popular brands. See them all on one page at <a href="https://simpleicons.org">SimpleIcons.org</a>. Contributions, corrections & requests can be made on GitHub. Started by <a href="https://twitter.com/bathtype">Dan Leech</a>.</p>
 </p>
 
 ## Usage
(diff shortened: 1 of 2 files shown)


ericcornelissen:
I'm personally not a fan of saying "over a 1000 icons" rather than just "free icons", but that is just personal preference.

That said, if we decide to go ahead with this we might as well make it so the website programmatically uses the exact number of icons.

PeterShaggyNoble:
> That said, if we decide to go ahead with this we might as well make it so the website programmatically uses the exact number of icons.

Can that be done in the templating, @ericcornelissen? If so, I'll leave that to you. If not then I'll knock together a quick, wee bit of JS to do it.

ericcornelissen:
> Can that be done in the templating, @ericcornelissen? If so, I'll leave that to you. If not then I'll knock together a quick, wee bit of JS to do it.

That can be done with templating, yes, and we would ideally do so as then it would have to happen only once when GitHub puts together the page for GitHub pages (rather than with every page load).

I'll wait for further input before working on this 👍

runxel:
Well I think that would be great!
I actually wondered how many icons we got now just yesterday. So yeah, would be neat! 🤷‍♂ 
:)

PeterShaggyNoble:
@runxel, Tip: Github truncates lists of files to 1000 items so, now that we're over 1000 icons, you can see how many we have at https://github.com/simple-icons/simple-icons/tree/develop/icons - 1027 as of this comment.

PeterShaggyNoble:
Abandoning this in favour of Eric's proposed implementation.

PR #2375, opened 2020-01-15
Improve Formula for Display Colours

### LINKED ISSUE #2315: Improve Formula for Display Colours
### Kind of issue
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [ ] The Packagist Package
  - [x] The website
  - [ ] Other, namely: 

### Description
The current formula used to determine whether an icon & its related should be displayed on the website in black or white doesn't work to well with Apache Cordova, for exmaple, being a particularly bad offender.

Apache Cordova

A formula I've been using for a good few years now, which I think I originally got from the WCAG, is `red*.299+green*.587+blue*.144<128` - if that returns `true` then use white, otherwise use black. I know there is a website redesign being planned (ref: #980) that may render this moot but, in the interests of accessibility, I think this is something we should fix on the current site in the interim.

You can see the changes to each brand on [this temporary, page](https://petershaggynoble.github.io/sic/), which uses the following JavaScript version of the formula to determine the colour:
```
parseInt(hex.substr(0,2),16)*.299+parseInt(hex.substr(2,2),16)*.587+parseInt(hex.substr(4,2),16)*.114<128?`#fff`:`#000`
```

ericcornelissen:
**Issue:** Closes  #2315

### Checklist

n/a

### Description

As per the discussion in #2315 I updated the logic that determines if an icon is displayed in black or white depending on the brand colour. Rather than a threshold of `128` (which is better for a11y) I put a threshold of `210` (which is more in line with what we already have, and, I think, looks better).

I put a screenshot of the whole site with both thresholds below or you can play with the value yourself by changing it on line 105 of index.html.

Feedback is welcome, I'm fine with changing the threshold value as this one is kind of arbitrary 😄

PS. I still use the term "luminance" in the code even though I'm not sure if the current computation is still about luminance... 

| 128 | 210 |
|---|---|
|  |  |
| <p align="center"><b>160</b></p> | |
| Screenshot_2020-01-17 Simple Icons | |

### DIFF
--- a/index.html
+++ b/index.html
@@ -98,26 +98,14 @@
         {% assign rgbMin = rgbBlue %}
     {% endif %}
 
-    {% if rgbRed <= 0.03928 %}
-        {% assign lRed = rgbRed | divided_by: 12.92 | times: 0.2126 %}
-    {% else %}
-      {% assign lRed = rgbRed | times: 0.3053 | plus: 0.6822 | times: rgbRed | plus: 0.0125 | times: rgbRed | times: 0.2126 %}
-    {% endif %}
-    {% if rgbGreen <= 0.03928 %}
-        {% assign lGreen = rgbGreen | divided_by: 12.92 | times: 0.7152 %}
-    {% else %}
-        {% assign lGreen = rgbGreen | times: 0.3053 | plus: 0.6822 | times: rgbGreen | plus: 0.0125 | times: rgbGreen | times: 0.7152 %}
-    {% endif %}
-    {% if rgbBlue <= 0.03928 %}
-        {% assign lBlue = rgbBlue | divided_by: 12.92 | times: 0.0722 %}
-    {% else %}
-        {% assign lBlue = rgbBlue | times: 0.3053 | plus: 0.6822 | times: rgbBlue | plus: 0.0125 | times: rgbBlue | times: 0.0722 %}
-    {% endif %}
-    {% assign luminance = lRed | plus: lGreen | plus: lBlue %}
-    {% if luminance >= 0.83 %}
-        {% assign class = "grid-item--dark" %}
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble:
I'm struggling to see a few if the icons in the `210` screenshot, most notably in the lighter yellows, blues & pinks so I think we'll need to go down closer to, if not all the way to, `128`. Sometimes, unfortunately, aesthetics have to be sacrificed in the interests of accessibility.

ericcornelissen:
I'm perfectly fine with `128` for accessibility. I added a preview for a compromise at `160` and would like to have some more input from others (including @simple-icons/maintainers) 🙂 

In case many prefer aesthetics over accessibility, we could add a toggle to make it configurable (e.g. switch between `128` and `160` or `210`)?

PeterShaggyNoble:
I *was* going to suggest adding a slider tat defaults to `128` but thought that might be a bit much! 😆

runxel:
I like the 160 one! 🌮  That's as good as it can get.
128 is maybe an overkill, I'd say.

ericcornelissen:
> I was going to suggest adding a slider that defaults to 128 but thought that might be a bit much! 

A slider is definitely overkill and not very hulpfull/user friendly I think. But I really don't think havin a toggle/a few radio buttons is a bad idea, though it might be too much work for now given that most design in #980 get rid of the problem entirely.

> I like the 160 one! 🌮 That's as good as it can get. 128 is maybe an overkill, I'd say.

Updated it to `160` now, Let's see if we can get one more opinion 😄

runxel:
Everyone good now? :)
Great, let's merge! 💯

PR #2376, opened 2020-01-15
Cinema4d icon

massamino:
**Issue:**

### Checklist

### Description

Manually vectorized


foo-dogsquared:
Alexa ranking: [~20k](https://www.alexa.com/siteinfo/maxon.net) for the parent company (Maxon). 

I am a bit familiar with the tool and it is considered as one of the common choices for 3D modelling. 
Also, it's not likely to be rejected since it is one of their popular products, after all. 

For clarification, where you did get the icon? 
I found [a press kit page](https://www.maxon.net/de/header-meta-navigation/ueber-maxon/pressematerial/) that contains the EPS of the logo (although no monochrome version of the logomark so far).

PeterShaggyNoble:
@foo-dogsquared, should the name not be just "Cinema 4D"?

foo-dogsquared:
I don't think there would be any naming conflicts with this so we'll go with "Cinema 4D". 

Please update the name in the JSON file instead @massamino as @PeterShaggyNoble suggested.

PeterShaggyNoble:
You'll also need to update the `title` tag in the SVG to match.

And the source URL in the JSON should be where you sourced the original artwork.

runxel:
Hey @massamino!
Welcome to simple-icons and thank you very much for your PR! 🎉 

The others already gave some good advice to improve it so we can merge this bad boy in! 😎 
Thanks! 👋

massamino:
Tnx, I'll fix it. I couldn't open the kit on my mac to check for logo files, so I made it from scratch.

massamino:
@PeterShaggyNoble should the source point to the presskit page then?

massamino:
What am I doing wrong. The build failed. I can't see a typo and all is renamed to "Cinema 4D" ...

PeterShaggyNoble:
Thanks for updating, @massamino. I'd agree with that choice for the source URL.

The build is failing because Cinema 4D is not in the correct spot in the JSON, which needs to be sorted alphabetically by title.

massamino:
OK, sorry, didn't know that. I'll fix that as well :)

PeterShaggyNoble:
It's a brand spanking new addition to our linting :wink: We have plans in place to automate the sorting so the build won't fail in future.

massamino:
Fixed. Thanks for your help.

PeterShaggyNoble:
Thanks for sorting all that out, @massamino :+1:

Now we just have to decide if we're happy with this monochrome treatment of the icon. I'll defer to @foo-dogsquared on the final call as he's familiar with the software. I really like [this version](https://www.deviantart.com/physxpsp/art/Cinema-4D-Metro-logo-344402024) of it, though, so maybe we could do something similar?

Cinema 4D by PhysXPSP

massamino:
I saw that version as well when I was searching for a svg icon for the software. For me this version has to little contrast. That's why I choose to make the dark side of the 'c' transparent to reflect that. But if the final decision is to use the full white one, I'll fix the errors in the screenshot and use your version.

runxel:
I would follow your argumentation. The monochromization is well done and I like the look more then the too dense version Peter found.

> I saw that version as well when I was searching for a svg icon for the software. For me this version has to little contrast. That's why I choose to make the dark side of the 'c' transparent to reflect that. But if the final decision is to use the full white one, I'll fix the errors in the screenshot and use your version.


However I can't agree on the color being black.
_I_ would have suspected it to be `#011a6a` matching the blue orb.

massamino:
@runxel I would agree if I could find that color as a main color on their website. I was looking for a different match than black to, but couldn't find it. I can see a blue color on the website, but it doesn't match the blue orb. I'll check again :)

runxel:
I wouldn't give so much attention to the website in this case.
The page is rather bland, nothing to gain there.

But yes, I understand that the decision is not so easy, since there is also no flat icon version.

massamino:
I'll change the color to the suggested orb color.

PeterShaggyNoble:
If we're all agreed that the current version is better then that works for me too 👍 Nice work, @massamino

PR #2379, opened 2020-01-16
Add AngularJS

### LINKED ISSUE #2358: AngularJS
**Name:** AngularJS
**Website:** https://angularjs.org/
**Alexa Rank:** [~18.5k](https://www.alexa.com/siteinfo/angularjs.org)
**Colour:** `#b52e31` (or `#e23237`?)
**Official resources for icon and color:** https://angularjs.org/img/angularjs-for-header-only.svg

First, we'll need to decide *if* we want to include AngularJS in order to distinguish it from Angular. Then we'll need to decide how we make the icon different from Angular's. In MDI, we use the `A` and the frame of the shield, like so:

[angularjs](https://materialdesignicons.com/icon/angularjs)

PeterShaggyNoble:
AngularJS

**Issue:** Closes #2358

### Checklist

### Description
SVG found in website header. Other colour options are `#e23237` from the other half of the shield or `#dd0031`, the same as Angular.


PeterShaggyNoble:
Well, that's weird! Looking at it originally, I thought there *was* a line running down the middle but, with the icon zoomed in to fill my large monitor, it was gone when I knocked the 2 red panels out so figured it was an optical illusion. It must be *hairline* thin that I missed it!

Whether it's there by design or by accident, I don't think it makes sense to keep it in the monochrome version. I'll try to get it removed tomorrow.

PeterShaggyNoble:
@runxel, that should be fixed now.

runxel:
Looking good! Thanks :)

PR #2392, opened 2020-01-17
Add LXDE

mhf-ir:
**Issue:**

### Checklist

### Description
Alexa rank about 510K but considering Most users using [Lubuntu](https://www.alexa.com/siteinfo/lubuntu.net) is about 120K and other desktop environment like gnome is already there good to have LXDE.


PeterShaggyNoble:
Thanks for the contribution, @mhf-ir :+1:

To help us decide whether or not this is within scope, could you expand a bit on the relationship between Lubuntu & LXDE? At the rate LXDE's rank is climbing, though (over 400k in the past 3 months) it looks like it could well break 500k soon so it may be a moot issue.

PeterShaggyNoble:
Using the icon on their website homepage (in grey below), there are some discrepancies in your version (in red) that will need to be cleaned up before we can merge this in.

Also, where did you source the colour from? I don't see it used anywhere on their site. Which colour to pick is going to be tricky, though, as they use a few with no indication of which should be considered the primary one.

LXDE

mhf-ir:
You're right color not specified seems the default color is `000000` for mono logo.

https://wiki.lxde.org/en/index.php?curid=37864

PeterShaggyNoble:
Despite being on their Wiki, I personally wouldn't consider that a realiable, official source as it's not used on any pages on their Wiki, was uploaded by an unregistered third-party and is over 3 years old. Given that, could you update this to use the SVG from their homepage.

The colour is still up in the air - I'd go with `#2e3842` from their website header or `#ed4933` from the "button" links but let's see what the other @simple-icons/maintainers think before deciding.

mhf-ir:
I think there no colour from any resource it's already mono color logo.

Lubuntu use blue 
https://en.wikipedia.org/wiki/Lubuntu#/media/File:Lubuntu.svg

But official website on github release logo as white:
https://github.com/lxde/lxde.github.io/blob/master/images/lxde-logo.svg

PR Updated

PeterShaggyNoble:
@mhf-ir, can you update this to match the source SVG, which isn't quite square. Yours being square accounts for the differences seen above.

I'd also suggest using either the website homepage or the `images` directory in their repo as the source URL, rather then the SVG file.

PeterShaggyNoble:
Thanks for updating, @mhf-ir - SVG looks spot on to me now. I'm still not convinced on the colour but don't have a better alternative to offer so let's see if one of the other @simple-icons/maintainers can figure it out.

ericcornelissen:
@mhf-ir may I ask why you choose to add LXDE rather than lubuntu? The lubuntu logo is essentially the LXDE logo in a circle (also, the colour, as you pointed out, is much clearer).

That said, for the colour I agree that `#2e3842` would be a good choose, seen as they use a similar colour in their favicon 👍

PeterShaggyNoble:
@ericcornelissen, we could add both.

ericcornelissen:
> we could add both

I'm not opposed to that, I was just interested in the reasoning for adding both. Seen as their icons are nearly identical and, as an outsider, I would say lubuntu is more like a brand than LXDE 🤔

runxel:
> I would say lubuntu is more like a brand than LXDE 🤔

True, but we also have GNOME already ...

PeterShaggyNoble:
Leaving this open (for now) despite [this comment](https://github.com/simple-icons/simple-icons/pull/2402#issuecomment-579852495) as, bar the colour and discussion on whether we should include it or its parent brand or both, it's pretty much ready to go.

ericcornelissen:
Again, I see no problem with adding both. The primary reason I suggested adding the lubuntu logo is that picking the colour would be easy and the logo is essentially the same. But if there is demand for both version we can add them if someone contributes them 🙂 So: if we can find a suitable colour for LXDE this can be merged 👍

PeterShaggyNoble:
Colour choices as I see 'em are:
1. `#2e3842`, from the website header & footer (current colour)
2. `#ef5e4a` from the "button" links on their site
3. `#21b2a6` from their homepage
4. `#505393` also from their homepage

Not sold on `3` & `4` at all - I think `2` is probably the best option but I'm fine with `1` as well, if that's the consensus.

ericcornelissen:
The [Alexa rank](https://www.alexa.com/siteinfo/lxde.org) is now down to ~630k, I wonder what happened in early January what got them to almost 500k :thinking: 

Anyway, regarding the colour. `#21b2a6` is best based purely on the website but I think `#2e3842` is better in general.

PeterShaggyNoble:
> I wonder what happened in early January what got them to almost 500k 🤔

Probably all of us checking out the colour! 😆

Seeing as their rank is on the decline, though, and given that we didn't get a reply to [this question](https://github.com/simple-icons/simple-icons/pull/2392#issuecomment-575697309), I'd be OK with closing this as out of scope.

ericcornelissen:
In that case, I'm closing this and marking it as `out of scope` for now. If you want to continue working on this @mhf-ir feel free to continue the disucssion and/or your work right here! :slightly_smiling_face: 

For anyone else who would like to see the LXDE icon added, feel free to leave a comment here or open an issue for it :+1:

Issue #2380, opened 2020-01-16
Automated fixes in Pull Requests


ericcornelissen:
Opening this issue to discuss the possibility of implementing automated fixes in Pull Requests first(?) discussed in #2334. Some things that we could automatically fix by pushing commits are:

- [ ] Uppercase hex value
- [ ] Add `<title>` to SVG when it is missing based on the title in the JSON
- [ ] Run `npm run svgo -- [icon]` for the new SVG (see also #2061)
- [ ] Order JSON file alphabetically
- [ ] Pretiffy JSON file
- [ ] The treatment of newlines
- [ ] Remove `#` from the hex field if it is accidentally included
- [ ] Unify implicit/explicit closepath commands (see also #5070)

More suggestions for things we could automate are welcome 🙂

PeterShaggyNoble:
In case they're not covered by "SVG optimisation" above:

- Strip `doctype`s
- Strip newlines & tabs
- Add missing attributes to `svg` tag and strip unnecessary attributes
- Replace `<path d="..." />" and "<path d="..."></path>` with `<path d="..."/>`
- Strip unnecessary attributes from `path` tag

ericcornelissen:
By "SVG optimization" I basically meant "Run `npm run svgo -- [icon]` for the new SVG" 😄 (updated original for clarity)

runxel:
- The treatment of newlines

ericcornelissen:
Good one @runxel 👍 updated the issue description

phatbhoy67:
I have included the # in the hex value more than once.............. 😆

ericcornelissen:
I'm currently in the process of creating a GitHub action to "Run `npm run svgo -- [icon]` for the new SVG" 🙂 see: https://github.com/ericcornelissen/svgo-action

ericcornelissen:
@mondeja [brought up](https://github.com/simple-icons/simple-icons/issues/5070#issuecomment-782213396) that doing automated fixes in Pull Requests might be 1) perceived as being invasive and 2) confusing for newcomers (_"huh, I can't push because my branch is out of sync?"_). And that we should instead consider pushing automated fixes AFTER the pull request has been merged, especially since some of these things are quite trivial.

I'm personally still of the opinion that we should do the fixes within the context of a Pull Request so as to avoid merging in "invalid" (*) Pull Requests for the following reasons:

[...]

PR #2397, opened 2020-01-17
Update .svgo.yml

runxel:
### Description
Illustrator likes to put some extra attributes in SVGs. This feature will remove those.
It's just a small fix and shouldn't break anything.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -33,6 +33,11 @@ plugins:
   - removeUnknownsAndDefaults:
       keepRoleAttr: true
 
+  # remove any unnecessary attributes
+  - removeAttrs:
+      attrs:
+        - baseProfile
+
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble:
See also #2380

.svgo.yml
```diff
@@ -33,6 +33,11 @@ plugins:
   - removeUnknownsAndDefaults:
       keepRoleAttr: true
 
+  # remove any unnecessary attributes
```
ericcornelissen: For clarity, do you mind reordering this so that the `removeUnknownsAndDefaults` configuration comes after this comment as well?

runxel: Good idea! I've done so. Thanks for the valuable comments :)

PR #2401, opened 2020-01-18
Add LG

mhf-ir:
**Issue:**

### Checklist

### Description
Alexa rank is about [1500](https://www.alexa.com/siteinfo/lg.com)


runxel:
Hey @mhf-ir,
thanks for the PR.

First of all I'm not sure about the color, it could be the dark blue as well. But that's not the big point.
What's interesting here is, that MariaDB themself use another version, which is monochromized. See:

db

Maybe you can find out if they use it somewhere else too. I'd slightly prefer this version over yours.

PeterShaggyNoble:
Wrong PR, @runxel!

PeterShaggyNoble:
Brand guidelines can be found at https://www.lg.com/global/about-lg/brand-identity

Unfortunately, that page does state that:

> The Symbol Mark may not be used alone.

So, sadly, I think we'll have to include the LG here.

Also, from the downloads available on that page, I get a colour of `#a50034`.

ericcornelissen:
Thanks for your work on this @mhf-ir 🎉 It is much appreciated

PR #2402, opened 2020-01-18
Add MariaDB Foundation

mhf-ir:
**Issue:**

### Checklist

### Description
Logo from official website is svg, colour picked from body of seal.
Alexa is about [54K](https://www.alexa.com/siteinfo/mariadb.org)


mhf-ir:
@runxel About the minimal icon I like your verstion too but, Hard rules about must be exact same and url even colour lead me to official ones without doubt. Send me the good one url vector i will change it.

PeterShaggyNoble:
https://mariadb.org/wp-content/themes/twentynineteen-child/sponsor-logos/mariadb_logo.svg

runxel:
Sorry for posting at the wrong [PR](https://github.com/simple-icons/simple-icons/pull/2401#issuecomment-576266062) 🙈 😅 

@PeterShaggyNoble best detective as always!

runxel:
I personally would omit the "MariaDB" (if allowed).
What do you think @PeterShaggyNoble ?

PeterShaggyNoble:
Tricky one, @runxel. On the one hand, they don't use it without the wordmark (that I can see) anywhere on [this site](https://mariadb.org/) - they even include it in the [favicon](https://mariadb.org/wp-content/uploads/2019/02/cropped-mariadb_org_rgb_r_512-1-32x32.png). On the other hand, it does pop up in a couple of places on [this site](https://mariadb.com/) (e.g., header & favicon) without the wordmark. I haven't found anything that outright says it can't be used without the text - the only *suggestion* that that might be the case is that all the files available on [this page](https://mariadb.org/about/#logos-and-badges) include it.

When in doubt, I like to let the fact that we're an icon library sway my decision so, after all that, I'd say let's go without!

phatbhoy67:
Hey @mhf-ir thanks for your work on this one 👍 

It isn't clear to me (maybe @PeterShaggyNoble and @PeterShaggyNoble can pitch in here) what this icon is intended to represent, is it [MariaDB Foundation](https://mariadb.org/), [MariaDB](https://mariadb.com/) the business or just the technology in general? 🤔 

I think this might be a similar situation to the one discussed in #2446 

It seems that MariaDB the business has many planned or existing [products](https://mariadb.com/products/) which are different to the open source ones, they also are one of the main sponsors of MariaDB Foundation so can't (in my view) be the same thing. 🤔 

The seal/sealion and wordmark seem to represent the business, the foundation has a separate and different identity and logo.

Any suggestions to avoid an future conflicts?

PeterShaggyNoble:
@phatbhoy67, it looks to *me* like MariaDB Foundation is the parent organisation and uses the outlined seal as its icon (complete with wordmark) and MariaDB is the product, using the filled seal currently in this PR as its icon but I don't know if I'm right - it's quote confusing!

phatbhoy67:
Apologies, no idea how I managed to double tag you there, it was meant to be one @PeterShaggyNoble and one @runxel 🤣 

The alexa rank ☝️ is for (https://mariadb.org/), as is the source.

It looks like we have based the icon on the "business" version .svg displayed on the "foundation" site as a sponsor? (https://mariadb.org/wp-content/themes/twentynineteen-child/sponsor-logos/mariadb_logo.svg)

Given the logo in the header 👇  (https://mariadb.org/wp-content/themes/twentynineteen-child/icons/mariadb_org_rgb_h.svg)
icon

I would expect it to look like this 👇, or this with the full wordmark
rect937

Lss, I think we need a MariaDB Foundation icon and a MariaDB icon.

This PR seems to be a mixture of both 🤔

PeterShaggyNoble:
Ah, yes, we do seem to have become mixed up somewhere along the way, don't we?! 😕

OK, so, let's keep this PR to the foundation and revert to @mhf-ir original version only without the wordmark as it's not an *explicit* requirement. And then, on top of that, add the version currently in this PR separately as just MariaDB.

Sorry for all the messing, @mhf-ir!

mhf-ir:
PR is in your hand, you can change it. after work with you guys i'v been thinking about alternative repo that no so strict just add simple icons, you can change it and revert. Thanks for you're work anyway.

PeterShaggyNoble:
I've updated this as the MariaDB Foundation and reverted the original, unfilled seal. I've opted not to include the wordmark as it's not explicitly required anywhere and I also think the blue is the more appropriate choice of colour as they make more use of it on their website than the brown.

PeterShaggyNoble:
D'oh! Forgot to fix a conflict *again*! Thanks, @phatbhoy67 - sorted now.

PR #2405, opened 2020-01-18
Add Fila

mhf-ir:
**Issue:**

### Checklist

### Description
Alexa rank is about [74k](https://www.alexa.com/siteinfo/fila.com)


PeterShaggyNoble:
Thanks for your work on this one, @mhf-ir; SVG lines up nicely with [this one](https://statics.glamit.com.ar/skin/frontend/default/fila16/images/filanew.svg) I found on [their Argentinian site](https://statics.glamit.com.ar/skin/frontend/default/fila16/images/filanew.svg). I would much prefer if we could just use the `F` but, other than their favicons, they don't use it anywhere that I can see.

They seem to be very inconsistent with their colours but the one you've chosen does appear a good bit across their sites so I think we can go with that.

PR #2406, opened 2020-01-18
Add Premier League

mhf-ir:
**Issue:**

### Checklist

### Description
Alexa is about [~840](https://www.alexa.com/siteinfo/premierleague.com)


PeterShaggyNoble:
I was thinking about sports¹ icons last week and whether or not to add some as you could easily get an entire, dedicated icon library out of them on their own. My line of thought was that we'd need to draw a line somewhere to stop us from becoming inundated with requests, similar to what happened to [BrandColors](https://brandcolors.net/) - they added a couple of educational institutions which opened the floodgates to a deluge of requests for more of them. The consclusions I came to were:

1. Sporting organisations would be fine to add
2. Sports clubs and teams would be a bit too much

However, I couldn't make a decision on where sporting competitions fall, which isn't much help here! 😆 But all that's just my own, personal thinking on it. We'd need a few other @simple-icons/maintainers to share their thoughts on the topic.

That aside, if we are to add this then, unless there are some brand guidelines disallowing it, I'd just use the lion's head here, without the wordmark.

-----
<sup>1. Loath, as I am, to call soccer a sport! 😆</sup>

runxel:
I'd say its okay to add the lion of the Premier League; they are fitting into (1) of yours somehow.
And I second the thought, that single clubs are a bit too much.

ericcornelissen:
> That aside, if we are to add this then, unless there are some brand guidelines disallowing it, I'd just use the lion's head here, without the wordmark.

Definitely agreed.


>  My line of thought was that we'd need to draw a line somewhere to stop us from becoming inundated with requests, similar to what happened to BrandColors - they added a couple of educational institutions which opened the floodgates to a deluge of requests for more of them. The consclusions I came to were:
> 
>     Sporting organisations would be fine to add
>     Sports clubs and teams would be a bit too much
> 
> However, I couldn't make a decision on where sporting competitions fall, which isn't much help here! 😆 But all that's just my own, personal thinking on it. We'd need a few other @simple-icons/maintainers to share their thoughts on the topic.
> 
> That aside, if we are to add this then, unless there are some brand guidelines disallowing it, I'd just use the lion's head here, without the wordmark.

For sport clubs, I think we can draw a clear line. For university, perhaps as well. For sporting competitions we could say the same, but arguably there aren't as many **big** sporting competitions, though we might consider using stricter requirements for them?


> I was thinking about sports icons last week and whether or not to add some as you could easily get an entire, dedicated icon library out of them on their own.

If there is enough interest, we could use our tools to set that up 🤔 Though I don't think we currently have the resources to support that project 😞

PeterShaggyNoble:
Closing as abandoned due to [this comment](https://github.com/simple-icons/simple-icons/pull/2402#issuecomment-579852495). If anyone else would like to take this on, feel free to open a new PR or issue but but please see the continuation of the discussion above in #2513 first.

PR #2408, opened 2020-01-19
Add Pokémon icon

KovuTheHusky:
### Checklist

### Description
As a companion PR to #2340.


ericcornelissen:
Not a full review but just wanted to say I'm okay (and I think I prefer) the full Pokémon wordmark. That said, the Pokéball is a good alternative (as per [this comment](https://github.com/simple-icons/simple-icons/pull/2340#issuecomment-575706794)) 👍

PeterShaggyNoble:
I'd vote for the Pokeball over the full wordmark as it suits our purposes as an icon library better and is iconic enough on its own.

ericcornelissen:
> I'd vote for the Pokeball over the full wordmark as it suits our purposes as an icon library better and is iconic enough on its own.

Although the Pokéball is a fictional brand, I personally feel like it is separate :thinking: Any input from other @simple-icons/maintainers?

runxel:
I'd prefer the wordmark actually...

ericcornelissen:
> One thing, why did you choose red for the hex value, I would expect it to be blue or yellow?

Good point, I guess red **is** based on the Pokéball :laughing: but indeed yellow or blue are better options if we go with the wordmark. I'm personally in favour of yellow.

PeterShaggyNoble:
As a Pokémon "outsider" (far too old for it!), I'd associate yellow with it, as a brand. That, however, may well be influenced by Pikachu being the most well known character.

ericcornelissen:
It seems there is consensus on yellow so I updated it 👍 

> That, however, may well be influenced by Pikachu being the most well known character.

I don't even think that is a bad reason to choose yellow 🙃

phatbhoy67:
Thanks for your work on this one @KovuTheHusky and thanks to @ericcornelissen for updating the colour 👍

PR #2411, opened 2020-01-19
Add Olympic Games

mhf-ir:
**Issue:**

### Checklist

### Description
Alexa rank about [26k](https://www.alexa.com/siteinfo/olympic.org)


foo-dogsquared:
Even though it passes the initial requirements, I don't think we can add this since the org (IOC) behind this logo is (understandably) strict rivaling Disney's aggressiveness when it comes to protecting its copyright. 
What do the other @simple-icons/maintainers think about this? 

On a side note, [some info about usage of the logo](https://www.olympic.org/olympic-rings).

mhf-ir:
Logo get from Wikipedia page and license is Public Domain:
https://en.wikipedia.org/wiki/Olympic_Games#/media/File:Olympic_rings_without_rims.svg
https://en.wikipedia.org/wiki/Public-domain-equivalent_license
I don't think that's would be problem. 

> Use and rights
> 
> All rights to the Olympic properties, as well as all rights to the use thereof, belong exclusively to the IOC, including but not limited to the use for any profit-making, commercial or advertising purposes.

**not limited to the use for any profit-making, commercial or advertising purposes**

foo-dogsquared:
> not limited to the use for any profit-making, commercial or advertising purposes

Ah, I missed that specific part for some reason. >_>
~~I'll review this later when I get some sleep.~~

PeterShaggyNoble:
I'd be wary of adding this for the reasons @foo-dogsquared mentioned in his initial comment. Even though *we're* not directly profiting from its inclusion in our library, if someone using our library were to use the rings in a manner the IOC don't like, I could well see them throwing a cease-and-desist our way. We'll need some feedback from some of the other @simple-icons/maintainers before we can make a final decision on this, though.

All that aside, I'd suggest changing the name to "Olympic Games", rather than just "Olympic".

ericcornelissen:
> Even though it passes the initial requirements, I don't think we can add this since the org (IOC) behind this logo is (understandably) strict rivaling Disney's aggressiveness when it comes to protecting its copyright.

If this is true I agree that we might be better of not including it, at least for now...

ericcornelissen:
Marked this as `out of scope` for now to keep the list of open Pull Requests a little bit clean. The reason I closed it rather then merging it is the concern originally voiced in [this comment](https://github.com/simple-icons/simple-icons/pull/2411#issuecomment-576030073).

Continued discussion regarding whether or not we should add this can happen in this thread! Any feedback on whether or not our concerns are valid is welcome :slightly_smiling_face:

PeterShaggyNoble:
[This](https://github.com/simple-icons/simple-icons/issues/2575#issuecomment-585791641) reminded me that the IOC once issued WWE with a cease-and-desist to stop (Olympic gold medalist) Kurt Angle from calling one of his moves "The Olympic Slam".

Issue #2442, opened 2020-01-21
Khronos, OpenGL, WebGL, and Vulkan (and maybe other standards from Khronos)


foo-dogsquared:
I could've made this into its individual issue but I figured it would be easier to keep it all in one issue especially if the other standards are considered to be added. 

For now, I'll only list the ones I'm (superficially) familiar with. 
If you think the other items are suitable to be added, feel free to do so. 

**Name:** Khronos 
**Website:**  https://www.khronos.org/
**Official resources for icon and color:**

*  https://www.khronos.org/legal/trademarks/
* Brand color and the official logos are in the linked page. 
In fact, all of the standards to be requested are available there. 
  

Alexa ranking for Khronos: [~45k](https://www.alexa.com/siteinfo/khronos.org). 

The ones I'm sure to consider are WebGL, OpenGL, and Vulkan. 
They're kind of a universal thing at this point. 

* WebGL is supported throughout the major browsers (i.e., Firefox, Chrome). 
* OpenGL is a popular way to get into graphics programming. 
* Vulkan is a modern and low-level API for graphics computing. 

-----
- [x] Khronos
- [x] OpenGL
- [x] Vulkan
- [x] WebGL

runxel:
Closed by delivering all icons in:
#2448, #2450, #2451, #2542

PR #2448, opened 2020-01-22
Khronos Group (#2442)

runxel:
**Issue:** #2442

### Checklist

### Description
On the basis of `Khronos_RGB_June18.svg` from official download. `#CC3333` comes out of this svg.


PeterShaggyNoble:
I haven't looked into these ones yet but could we get away with just using the first, stylised `O`, I wonder? Or, if not, at least ditch the `Group`?

runxel:
Yeah, that would be nice.
But I think that's [disallowed](https://www.khronos.org/legal/logo-usage-guidelines).

phatbhoy67:
Nice work @runxel 👍

Issue #594, opened 2017-10-09
Fix: Remove Trademark symbols from MySQL & StubHub.


davidklebanoff:
The Intel, MySQL, and StubHub icons all have either a ™ or ® mark on them that seems inappropriate for the icons. They are the only icons with such marks.

birjolaxew:
I don't think we should change icons more than necesarry; if they are provided in monochrome, we should avoid removing elements from it.

Logo usage guidelines also usually require that the ™/® is always included.

davidklebanoff:
They are not requirements, but rather "requests". For instance, Oracle, which owns MySQL, "would _appreciate_ you attributing ownership of Oracle trademarks to Oracle Corporation by using trademark symbols."

Further reading on the matter from other sources seems to indicate that use of such trademark symbols is not required, but helps clearly indicate that the mark is a trademark to others. Doing quick Google Image searches, you can see that within companies and their very own products they do not always use the trademark symbols. For instance many of Intel's product packaging lacks the trademark symbol.

If you look at StubHub's press page, the press page's logos that are available for download do not have the TM mark on them. Unfortunately, those are PNG files, so we had to source an SVG from the site's header which happened to contain a TM mark.

Many of the brands in simple icons are trademarks (some registered, some not). The majority of them, all but the three indicated, do not have such trademark symbols attached to them. This was an effort to increase consistency across all our icons.

So, I don't know if all that may sway your opinion, but I thought I'd provide some background to my initial thoughts.

Source:
https://www.oracle.com/legal/trademarks.html

ericcornelissen:
I agree with @davidklebanoff that the icons would look better without the trademark symbol, but I also agree with @birjolaxew that the icon should stay true to the actual logo. If the designers of the logo think the trademark symbol should be included (even if we think it shouldn't) I think we should stick with that.

Just my two cents 🙂

davidklebanoff:
I would argue that the trademarks are not part of the actual logo. See my previous comment.

birjolaxew:
@davidklebanoff In cases were the trademark is explicitly optional (e.g. in Oracles case), I can agree to removing it. In cases where the official source doesn't include the mark, there's is no arguing that our version shouldn't either.

However, I'd like to keep the mark on Intel - I have been unable to find an official guideline on the mark usage, nor an official source that doesn't include it.

davidklebanoff:
[...]

PR #2453, opened 2020-01-22
Add Microsoft SQL Server

### LINKED ISSUE #2445: Need icon for SQL Server
**Name:** SQL Server
**Website:** https://www.microsoft.com/es-es/sql-server/sql-server-2019
**Official resources for icon and color:** 🤷‍♂
  
microsoft-sql-server

PeterShaggyNoble:
Microsoft SQL Server
**Issue:** Closes #2445

### Checklist

### Description
Colour chosen is the [calculated midpoint](https://meyerweb.com/eric/tools/color-blend/#EE352C:A91D22:1:hex) of the red gradient.


ericcornelissen:
Wanted to point out two things about the addition in this PR:

- The [source you linked](https://en.wikipedia.org/wiki/Microsoft_SQL_Server) does not actually have the logo on it 😅 [The SVG that was linked in the issue](https://de.wikipedia.org/wiki/Datei:Microsoft_SQL_Server_Logo.svg) is on [the German version of that page](https://de.wikipedia.org/wiki/Microsoft_SQL_Server)
- I can't find an official source for this, but judging by how the logo typically appears I think we **should** include the TradeMark symbol in the SVG 😞

PeterShaggyNoble:
Thanks, @ericcornelissen. My bad on the source link; could've sworn I use the German page there! 🙄 I'm OK just commiting that change straight to `develop`, right?

We do have precedent for removing (registered) trademark symbols (ref: #2177 & #594) when they're not explicitly required but we should probably have a broader discussion on when they should and shouldn't be included.

ericcornelissen:
> Thanks, @ericcornelissen. My bad on the source link; could've sworn I use the German page there! 🙄 I'm OK just commiting that change straight to develop, right?

Go right ahead!


> We do have precedent for removing (registered) trademark symbols (ref: #2177 & #594) when they're not explicitly required but we should probably have a broader discussion on when they should and shouldn't be included.

If it is not explicitly stated I'm fine if it is not there, I mainly pointed it out to see if you had considered it 😄 That said, having a (new, given #594) "formal" discussion about it in a separate issue might be useful.

runxel:
The TradeMark symbol ™ is a misleading symbol and has no legal meaning whatsoever (opposed to the _registered_ trademark).
I would always exclude it.

In the end both are a tautological boilerplate. 😃

ericcornelissen:
> The TradeMark symbol ™ is a misleading symbol and has no legal meaning whatsoever (opposed to the registered trademark).
> I would always exclude it.
> 
> In the end both are a tautological boilerplate. 😃

I agree with that, however, if a brand insist on ~~making their logo ugly~~ including the ™ in the icon then have to include it as well (regardless of it whether it has legal meaning).

PR #2420, opened 2020-01-20
Use GitHub Actions to automatically label Pull Requests

ericcornelissen:
**Issue:** _None_

### Checklist

n/a

### Description

We got access to [GitHub Actions](https://github.com/features/actions) a while back when it was still beta-access only but unfortunately haven't used it yet (we initially wanted to use it for #982) and I want to start using it to implement some stuff, among others #2380 and https://github.com/simple-icons/simple-icons-font/issues/1. However, I decided to start with something that is a bit simpler, namely:

**Feature**: Adding the labels `new icon` and `updated icon` to Pull Requests automatically based on the files added or changed in the Pull Request.<sup>1</sup>

For those unfamiliar with GitHub Actions the file `.github/workflows/pull_requests.yml` defines a "workflow", which is analogous to a build defined in `.travis.yml`. In it I say that `on: [pull_requests]` I want to do some stuff, run one or more "actions", namely I want to `run(s-)on: ubuntu-latest` a step as defined by `ericcornelissen/labeler@master`. In the case of the "action" is configured by `.github/labeler.yml` (which is a hard requirement from the "action" itself). The (currently incomplete) documentation for the configuration can be found in the [project's README](https://github.com/ericcornelissen/labeler/blob/master/README.md).

I tested the workflow in my fork of this repository, you can see the final results in [this Pull Request](https://github.com/ericcornelissen/simple-icons/pull/36) and [this Pull Request](https://github.com/ericcornelissen/simple-icons/pull/37) (and a bunch more attempts in the [list of closed Pull Requests](https://github.com/ericcornelissen/simple-icons/pulls?q=is%3Apr+is%3Aclosed)). ~~Also, the action should run on this Pull Request, but it does nothing as the files changed don't match the globs in the configuration.~~ Based on [this article](https://github.community/t5/GitHub-Actions/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/td-p/15426) it looks like the actions won't start running until this is merged because this PR is coming from a fork (rather than this repo) so I hope it will start working 🤞 

Why do this with GitHub Action rather than Travis CI? To try it out and learn, but also because I think it is easier to use and work with especially when you want to do stuff like in #2380. We don't need to move our Travis CI builds to GitHub Actions, but we could do that as well. Also, I like it very much that all the configuration is in the `.github/` directory 🙃 

[...]

PR #2459, opened 2020-01-22
Manual patching of icons (#2061)

runxel:
**Issue:** 
https://github.com/simple-icons/simple-icons-font/issues/29
and #2061

### Checklist
[x] yes.

### Description

What I am doing here is basically cleaning my own mess:
The icon for ABB robotstudio was one of my first contributions here. It was minimized using SVGOMG, which has problems – see issues linked above for more info.

I will continue to update all icons which has these problems here. This first PR should be merged by one of you @simple-icons/maintainers, so you know whats going on. The next ones I should be able to merge alone.
The correction is actually done quite easily (if you want to help):
1. See what icon has a problem (the Travis build for the font will tell you).
2. Optimize the logo again with svgo (locally!!!).
  It will correct the problems without changing the overall structure.
3. merge back in.
4. profit :)


ericcornelissen:
I'm not against doing this in theory, though I would prefer one giant PR for all the icons at once 😄

If you're doing it for https://github.com/simple-icons/simple-icons-font/issues/29 then I would like to inform you that was planning to try and include the transformation you describe above in the build process of the simple-icons-font. So, in that regard this is not necessary. (I have been thinking this for a while, but seen as the font isn't particularly popular it just doesn't have high priority)

Lastly, until we can fix this for new contributions I think this will be too time consuming to do... I actually think an automated fix for this can be implemented with GitHub Actions (as opposed to #2420) If you are very concerned about this issue I will move creating it to the top of my priority list for Simple Icons 😃

runxel:
> I'm not against doing this in theory, though I would prefer one giant PR for all the icons at once 😄

Oh yeah, sure, that would be better.
I just wasn't smart enough how to get a list all the shady icons at once. (which then could be automatically put thru svgo.)

> If you're doing it for [simple-icons/simple-icons-font#29](https://github.com/simple-icons/simple-icons-font/issues/29) then I would like to inform you that was planning to try and include the transformation you describe above in the build process of the simple-icons-font. 

You mean there will be no need for people to optimize their icons by hand, but instead our CI will do that and push the optimized svg back?
Well, would be neat! Even easier for people to contribute then.

> Lastly, until we can fix this for new contributions I think this will be too time consuming to do... I actually think an automated fix for this can be implemented with GitHub Actions 

Do you want to run all existing icons thru a GH action?

> If you are very concerned about this issue I will move creating it to the top of my priority list for Simple Icons 😃

Haha 🙈 no no, its okay :D

ericcornelissen:
> > If you're doing it for simple-icons/simple-icons-font#29 then I would like to inform you that was planning to try and include the transformation you describe above in the build process of the simple-icons-font.
>
> You mean there will be no need for people to optimize their icons by hand, but instead our CI will do that and push the optimized svg back?
> Well, would be neat! Even easier for people to contribute then.

Sorry for the confusion, that particular comment was meant to apply solely to **simple-icons/simple-icons-font**. My idea was to update [it's build pipeline](https://github.com/simple-icons/simple-icons-font/blob/36000da53a092b8bf36edee589e1d12ae5940445/scripts/build.js) to re-optimize the icons at build-time only to make the font.


However...

> > Lastly, until we can fix this for new contributions I think this will be too time consuming to do... I actually think an automated fix for this can be implemented with GitHub Actions
>
> Do you want to run all existing icons thru a GH action?

Yes! At some point I want that "there will be no need for people to optimize their icons by hand, but instead our CI will do that and push the optimized svg back" (see #2380). Of course this would still require some work from contributors, e.g. fitting it in the 24x24 viewbox, but the SVGO(MG) step will be eliminated. 😄

runxel:
Okay great!
Fixing the icons at the build pipeline helps with those icons, which are currently "corrupted".
Having our CI doing all the heavy lifting will help in the future. Both's fabulous!

However, what's with the existing icons here? They would still be bad, since the iconfont only does the fixing at build time and doesn't give the fixed icons back to this repo.
Maybe I can find a solution to still find & fix those.... 🤔

ericcornelissen:
> Fixing the icons at the build pipeline helps with those icons, which are currently "corrupted".
Having our CI doing all the heavy lifting will help in the future. Both's fabulous!

👍 


> However, what's with the existing icons here? They would still be bad, since the iconfont only does the fixing at build time and doesn't give the fixed icons back to this repo.

Those do indeed still need to be fixed, however, if you start doing that before we have fixed the source of problem you will need to keep updating. If we do it once after we implemented a fix for the problem we can patch it and send out a patch-release immediately. Especially since, as far as I can tell, there is no gain for doing it now I don't really see a reason to do it.

Then again, as I said before, do feel free to do update them now 🙂 

One thing you could try to investigate is see if you can automate the detection of "corrupted" icons so that fixing them can be automated 🤔

ericcornelissen:
I will close this now due to inactivity (both here and w.r.t. #2061)

PR #2466, opened 2020-01-23
Aircall

runxel:
**Issue:** –

### Checklist

### Description
Svg with color taken from website header.
[Alexa rank ~28k](https://www.alexa.com/siteinfo/aircall.io)


PeterShaggyNoble:
@runxel, it looks like the path in this one is slightly wider than 24px.

runxel:
Once again! Glad to have you here!
It was even worse than expected: The dimensions are not the same on each side...
But now it's fixed with correct height and centered on the viewbox.

PeterShaggyNoble:
I'm seeing some differences in the "feet" of the `A` now, @runxel (yours in red, original in green) - is that as a result of your fixes?

**LEFT**
image

**RIGHT**
image

runxel:
Not sure what happened there. Rebuilt the entire thing. Should be good now!

Thanks for spotting! 👍

PeterShaggyNoble:
Looks good now, @runxel. Nice one 👍

PR #2499, opened 2020-01-29
Update Babel

### LINKED ISSUE #2485: Update Babel?
**Name:** Babel
**Official resources for icon and color:** https://github.com/babel/website/blob/master/website/static/img/babel-black.svg

Given the lack of guidelines saying otherwise, should we update our Babel icon to be just the `B`, as they do in their favicon?

Babel

PeterShaggyNoble:
Babel

**Issue:** Closes #2485

### Checklist


PeterShaggyNoble:
I think the former is the better option, @ericcornelissen. As you say, it's easy to search for `svg` in the repo to find the latest icon whereas, with a direct link to the file, it won't be immediately obvious to people visiting that URL in the future that a brand might have changed its icon.

ericcornelissen:
> I think the former is the better option, @ericcornelissen. As you say, it's easy to search for svg in the repo to find the latest icon whereas, with a direct link to the file, it won't be immediately obvious to people visiting that URL in the future that a brand might have changed its icon.

I now actually think the latter is a better option 🤔 As I stated earlier, I don't think many people actually go to the source URL to check if it should be updated. Rather, someone notices the brand is using a new icon and from there they go look for a new source. The source in our metadata would then be a good jumping of point to start looking for the new icon. If it concerns a GitHub URL for a specific file in a specific commit, I think most of us are well aware that we can click on "Code"  and go search for new `.svg`s in the repository.

But again, we should really set a good and consistant standard for this, It not only affects GitHub URLs but also e.g. Wikipedia/Wikimedia URLs. In the case of Wikipedia/Wikimedia: having the Wikipedia URL will allow you to quickly see if anything changed (assuming it has been updated), the Wikimedia URL will shows you exactly what the icon was supposed to look like and that a good place to look for a new source is the Wikipedia page of the source.

PeterShaggyNoble:
Source updated, @ericcornelissen. With source URLs, I had been trying to answer both where the current version was found and where someone might be likely to easily find a new version but I see what you're saying: that we should err more on the former.

ericcornelissen:
Thanks for the contribution @PeterShaggyNoble 😄 I'll see if I can draft a clarification for this in the [Contributing Guidelines](https://github.com/babel/website/blob/93330158b6ecca1ab88d3be8dbf661f5c2da6c76/website/static/img/babel-black.svg)

Issue #2460, opened 2020-01-23
On (Registered) Trademark Symbols


PeterShaggyNoble:
Spinning off a discussion in #2453 and following up on an old discussion in #594, let's try to create a definitive reference here for when to and when not to include (registered) trademark symbols in an icon. The situations, as I see it, that need to be covered are:

1. There is no explicit requirement that the symbol be included.
2. It is stated that the symbol *should* be included but *may* be removed under certain conditions (usually when the icon is to be displayed at small sizes).
3. There is an explicit requirement that the symbol be included but there is ambiguity about under which conditions it may be removed (e.g., #1516).
4. Whether it's inclusion is an explicit requirement or not, the symbol is incorporated into the design of the icon (e.g., Intel).
5. There is an explicit requirement that the symbol be included.

My own, personal opinion is that they should *never* be included in an icon (except, maybe, in scenario 4) as they simply don't make sense for the size people typically use icons at. However, as a member of the SI team, I recognise that we do have to do everything we can to make sure we adhere to the wishes & requirements of each brand in order to cover ourselves so my suggestions for each scenario above are:

1. Don't include it
2. Don't include it
3. If it is a registered trademark symbol then include it. If it is a trademark symbol, don't include it, Either way we should reach out to the brand for clarification, updating as necessary if/when they respond to us
4. Include it
5. Include it

runxel:
>     1. Don't include it
>     2. Don't include it
>     3. Don't include it, proceed with adding it and reach out to the brand for clarification, updating as necessary if/when they respond to us
>     4. Include it
>     5. Include it

Nailed it, imho.
(Even tho I would choose to never include it 😈 😛)

PeterShaggyNoble:
> Even tho I would choose to never include it

As would I!

ericcornelissen:
List seems good to me, nice work @PeterShaggyNoble 👍 Even though the 4th point is definitely going to lead to some ambiguities 😅 e.g., as far as I can tell all the logos [in the Google images search for Microsoft SQL server](https://www.google.com/search?q=microsoft+sql+server) do include ™

For point 3, I would like to propose changing it slightly, based on [this comment by @runxel](https://github.com/simple-icons/simple-icons/pull/2453#issuecomment-577629869): if it concerns a _registered trademark symbol_, include it, if it concerns a _trademark symbol_, don't include it. Thoughts?

PeterShaggyNoble:
[...]

PR #2501, opened 2020-01-29
Add NFC

### LINKED ISSUE #2491: NFC
**Name:** NFC (or Near-Field Communication?)
**Website:** https://nfc-forum.org/
**Alexa Rank:** [~1.2m](https://www.alexa.com/siteinfo/nfc-forum.org), which shouldn't matter as it *is* an important standard
**Colour:** `#002e5f`
**Official resources for icon and color:** https://nfc-forum.org/wp-content/uploads/2016/10/NFC_NMark_Guidebook_9.16.pdf (found at https://nfc-forum.org/our-work/nfc-branding/n-mark/guidelines-and-brand-assets/)

[image](https://github.com/simple-icons/simple-icons/files/4123044/nfc.zip)

PeterShaggyNoble:
NFC

**Issue:** Closes #2491

### Checklist


ericcornelissen:
With #2460 in mind, from the PDF in the source you linked:

> The N-Mark symbol is a trademark. By adhering to the following regulations, guidelines, and recommendations, you will maintain the integrity of the N-Mark. In jurisdictions where the N-Mark is registered, the N-Mark must always be  accompanied by the registered trademark ® symbol. Elsewhere, it must  appear with the trademark ™ symbol. If you employ the N-Mark in both  registered and unregistered jurisdictions, use of the ™ symbol is permitted,  unless otherwise advised by the NFC Forum.

I'm not entirely sure, but, assuming the SVG will be used "in both  registered and unregistered jurisdictions", I think we should include the trademark symbol (™) here...

PeterShaggyNoble:
I went by the following line, found in the paragraph after the one you quoted:

> If the N-Mark is smaller than 5mm in height, or **if the means of display makes it infeasible to render** the ™ or ® symbol legibly, the ™ or ® symbol may be omitted. 

Another reason I opted to omit them is that I wasn't sure:
1. Of the legalities of claiming something is a *registered* trademark in a jurisdiction where it's not.
2. Whether showing the trademark symbol in jurisdictions where it is a registered trademark offers the necessary legal protections for that trademark.

The final reason was purely aesthetic, though; including it with the boxed version, which is their preference, would mean that the icon wouldn't fill the canvas in either direction in order to retain its proportions. A way around that, of course, would be to use the unboxed version, which could fill the entire vertical space and still be centred horizontally even with the inclusion of the symbol.

ericcornelissen:
> > If the N-Mark is smaller than 5mm in height, or if the means of display makes it infeasible to render the ™ or ® symbol legibly, the ™ or ® symbol may be omitted.

Assuming [a 96 dpi screen, a 24x24 icon is actually ~6x6mm](https://www.pixelto.net/px-to-mm-converter). But I do agree that it might still be too small. In all honestly, in this regard, I would like to see it as displayed on our site before making a decision.


> Of the legalities of claiming something is a registered trademark in a jurisdiction where it's not.

That is a good point which I hadn't considered 🤔 


> Whether showing the trademark symbol in jurisdictions where it is a registered trademark offers the necessary legal protections for that trademark.

I don't know if it does and honestly I don't think it matters: if the guidelines say you should include it, you should include it.

PeterShaggyNoble:
Here's what they looked like with the `TM` in place. Their illegibility at native size was why I took them up on the option of

> if the means of display makes it infeasible to render the ™ or ® symbol legibly, the ™ or ® symbol may be omitted

PeterShaggyNoble:
I'll take another pass at the optimisation on Monday, see if it needs to be that high. Can't remember now why I *did* go that high but, usually, I zoom in as far as I can in SVGOMG so the icon fills my largest monitor, start at one end of the precision scale or the other, keep moving up or down until I no longer notice a visual difference and then do a quadrant-by-quadrant overlay comparison with the original in Inkscape in case my eye missed any hairline difference(s). Of course, it's equally likely to just be a mistake/copy-paste error! :laughing:

Also, as there was some uncertainty on the `™` symbol, we should get one of the other @simple-icons/maintainers to weigh in and confirm, *or* dispute, our final decision.

PeterShaggyNoble:
Yeah, doesn't look like the precision _did_ need to be that high - clearly I was just seeing things! 😆

Updated.

ericcornelissen:
Alright then, thanks for the contribution @PeterShaggyNoble 😃

PR #2506, opened 2020-01-30
Add MariaDB

PeterShaggyNoble:
MariaDB

**Issue:** Follow-up to #2402

### Checklist

### Description
Could do with some help with the colour choice here; I lifted the current one from the logo found in the header of [this site](https://mariadb.com/) when you scroll down. Also, even though the SVG was found at mariadb.org, would mariadb.com be a better option?


phatbhoy67:
Hey @PeterShaggyNoble I think it would be better to update the source to be (https://mariadb.com/) .

There is a .svg version of the [sealion](https://mariadb.com/wp-content/themes/mariadb/assets/images/favicon/safari-pinned-tab.svg) favicon in the source.

This version is also slightly different to the PR version, the lines around the flipper are cleaner in the new source.

Looking at the favicon for (https://mariadb.com/) 👇 

favicon

I picked `#003545`.

No mention of any official colour here (https://mariadb.com/trademarks/).

PeterShaggyNoble:
Thanks, @phatbhoy67. I'd seen the SVG favicon but dismissed it because Potrace! Comparing the 2 though, it does look like the version on mariadb.org was shifted down a bit in the conversion, clipping the bottom of the tail so I've updated to use the favicon instead.

In the absence of any brand guidelines or a clear brand colour - yours is as good as any, and not a million miles from the one I grabbed, so let's go with that.

PR #2515, opened 2020-02-01
Add Peugeot

runxel:
**Issue:**
#2059

### Checklist

### Description
Color was taken from wordmark in official SVG.


phatbhoy67:
Nice work @runxel 👍

PeterShaggyNoble:
Should the eye not be knocked out of this? And I wonder would an outline be a better treatment? 🤔

runxel:
> Should the eye not be knocked out of this? And I wonder would an outline be a better treatment? 

No and no.
Brief explanation:
- The eye isn't even knocked out in the original version, it's just a small black dot.
- Also at small sizes nobody can actually see it.
- An outline would look weird here, because the icon is not uniform enough (think of the tail...!).

ericcornelissen:
> The eye isn't even knocked out in the original version, it's just a small black dot.
> Also at small sizes nobody can actually see it.

That, to me, doesn't sound like we shouldn't cut it out. Though it is an artistic interpretation, I do think it would make the logo more recognizable 🤔 


> An outline would look weird here, because the icon is not uniform enough (think of the tail...!).

Although I agree a filled version is more accurate, I don't think a outlined version would look weird. (in fact, an older version of the logo is an outline!)

runxel:
The outline works, if it's the same size as the tail, otherwise the tail would either look to big or to small.
Anyway, if I have time I will make a version with eyes for you in the next days 👀

PR #2523, opened 2020-02-04
Add Dynamics 365

### LINKED ISSUE #2512: Dynamics 365
**Name:** Dynamics 365
**Website:** https://dynamics.microsoft.com/
**Alexa Rank:** n/a (Microsoft subdomain)
**Colour:** `#002050`
**Official resources for icon and color:** It was a bugger to track down but I finally found it at http://thepartnerchannel.com/wp-content/uploads/Dynamics365_styleguide_092816.pdf

[Dynamics 365](https://github.com/simple-icons/simple-icons/files/4139455/dynamics365.zip)

PeterShaggyNoble:
Dynamics 365

**Issue:** Closes #2512

### Checklist

### Description
... Or _Microsoft_ Dynamics 365?

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1161,6 +1161,11 @@
             "source": "https://www.duolingo.com/"
         },
         {
+            "title": "Dynamics 365",
+            "hex": "002050",
+            "source": "http://thepartnerchannel.com/wp-content/uploads/Dynamics365_styleguide_092816.pdf"
+        },
+        {
             "title": "Dynatrace",
             "hex": "1496FF",
             "source": "https://www.dynatrace.com/company/press-kit/"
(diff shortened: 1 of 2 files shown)


phatbhoy67:
Hey @PeterShaggyNoble thanks for your work on this one and nice find on the .pdf btw 👏 

When reviewing this I went to page 3 of the .pdf where the different size and colour icons are laid out.

From the "64 PX DYNAMICS 365 SYMBOL" section, I tried to match the PR version against the (almost) black version. It didn't match. I tried a few of the other versions on that page of different sizes to make sure and none of them matched. 🤔 

I then isolated the icon from the layout image 👇 on page 2 and it matches the PR version, so now I am wondering which is the correct version of the icon we should be using? 😕 
dyn365

PeterShaggyNoble:
I used the right-hand one from the page in your screenshot, @phatbhoy67. I'll try again with the 32px version on the next page, though, as it's closer to our native size. Or, alternatively, we could go with the [Fabric Icons](https://uifabricicons.azurewebsites.net/) version (`Dynamics365Logo`)?

phatbhoy67:
Both versions in the layout page match your PR.

I might get my protractor out 📐  and measure that quoted 60 degree angle on the layout page to see if its an "artist's impression" of the icon. 🤣 

AFAIK all the icons on page three are the same.

PR #2534, opened 2020-02-05
Add DC

### LINKED ISSUE #2528: DC
**Name:** DC
**Website:** https://www.dccomics.com/
**Alexa Rank:** [~47k](https://www.alexa.com/siteinfo/dccomics.com)
**Colour:** `#0078f0`
**Official resources for icon and color:** [SVG](https://m.media-amazon.com/images/G/01/Comics/web/i/images/dc/DC-logo.svg) in header of [this site](https://www.readdc.com/)

[DC](https://github.com/simple-icons/simple-icons/files/4154717/dc.zip)

PeterShaggyNoble:
DC

**Issue:** Closes #2528

### Checklist

### Description
Possible alternative name: DC Entertainment


PeterShaggyNoble:
My reason for suggesting the alternative name, by the way, was to avoid potential clashes in the future.

PeterShaggyNoble:
DC Entertainment is the parent company and uses the same logo, which is why I chose it over DC Comics.

(Also because, like people who say "PIN number", "DC Comics" has always annoyed my brain as the "C" in "DC" stands for "Comics"! 😆)

ericcornelissen:
> DC Entertainment is the parent company and uses the same logo, which is why I chose it over DC Comics.

Hmm, didn't read careful enough. I think you're right and DC Entertainment is the better option :+1: Now here's to hoping that the logo's don't diverge :crossed_fingers: 

Once you change the title, @PeterShaggyNoble, this can be merged :+1: 

---

> Also because, like people who say "PIN number", "DC Comics" has always annoyed my brain as the "C" in "DC" stands for "Comics"! laughing

Well, in this case they do it themselves :rofl: Based on [the Wikipedia page](https://en.wikipedia.org/wiki/DC_Comics), I would guess the company was, in fact, incorporated as "DC Comics"!

PeterShaggyNoble:
Pushing an update now ...

> Now here's to hoping that the logo's don't diverge

If they ever do then we can add DC Comic separately.

> Based on the Wikipedia page, I would guess the company was, in fact, incorporated as "DC Comics"!

Yeah, it's a complicated and messy story of how they arrived at that name but it's covered quite succinctly in the "Origins" & "Golden Age" portions of the history section on their Wikipedia page.

Issue #1585, opened 2019-08-11
Updated Microsoft Office icons


adriaanzon:
Microsoft updated the logos of their office suite a while ago.  https://news.microsoft.com/europe/2018/11/29/say-hello-to-microsofts-new-office-icons/

**Name:** microsoftoutlook, microsoftonedrive, microsoftword, microsoftexcel, microsoftpowerpoint, microsoftonenote, microsoftaccess
**Official resources for icon and color:** https://developer.microsoft.com/en-us/fabric#/styles/web/office-brand-icons

Their brand page above also includes monochromatic icons, but they are outdated. This [PDF document](https://static2.sharepointonline.com/files/fabric/assets/microsoft_fabric_assets_license_agreement_may_2019.pdf), also mentioned on their brand page, includes a monochromatic Word logo on the second page. Here is a screenshot of that section in the PDF:
<img width="573" alt="image" src="https://user-images.githubusercontent.com/4326420/62834303-7949eb00-bc4a-11e9-8bbe-68d5823969df.png">
Though, for the other office applications, I haven't found icons in this particular style. 

Related: monochromatic icon suggestions for Microsoft Teams https://github.com/simple-icons/simple-icons/issues/1261#issuecomment-495383567

PeterShaggyNoble:
Good news: I have found monochrome versions of all the Microsoft Product icons buried in a font file on one of MS's sub-sites. I'll try to get some work done on them before the end of the week.

adriaanzon:
In the meantime, new icons have also been added here: https://developer.microsoft.com/en-us/fabric#/styles/web/icons#available-icons. Are they the same as the icons you found?

PeterShaggyNoble:
Nice one, @adriaanzon - it looks like the fonts I've found are indeed (horribly out of sync) subsets of Fabric Icons.

PeterShaggyNoble:
PRs have been opened for all of these now, linked from the checklist above.

PeterShaggyNoble:
Closing this out as they have all been addressed.

PR #2541, opened 2020-02-06
Update Microsoft Access

PeterShaggyNoble:
Microsoft Access

**Issue:** #1585

### Checklist

### Description
<s>Holding off on the rest as I need some help deciding on the source URL for them all:
1. The icons were sourced at https://uifabricicons.azurewebsites.net/
2. There is a subset of that font used at https://developer.microsoft.com/en-us/office, and all the icons are visible there
3. The individual product pages (in this case https://developer.microsoft.com/en-us/access) use that same font subset and, although a product's page doesn't always use its icon anywhere, they do all include the product's colour.

As both the icon and the colour can be retrieved from the individual product pages, I've gone with option 3.</s>
Never mind, just found https://developer.microsoft.com/en-us/fabric#/styles/web/colors/products which includes the exact colour references and, as both icon and colour are available, I think that's a better choice for the source URL too.


PeterShaggyNoble:
... Or would one of the colours from the full colour brand icon be a better choice?

https://static2.sharepointonline.com/files/fabric/assets/brand-icons/product/svg/access_48x1.svg in this case.

PR #2553, opened 2020-02-07
Update Microsoft Outlook

PeterShaggyNoble:
Microsoft Outlook

**Issue:** #1585

### Checklist

### Description
Had to go with a different source URL from the others (see #2541) here again as [this page](https://developer.microsoft.com/en-us/fabric#/styles/web/colors/products) doesn't include Outlook at all.


ericcornelissen:
@PeterShaggyNoble, I don't see a version of the logo you're using in the source you provided. I'd guess you took the logo from the font (as you can find on [this page](https://developer.microsoft.com/en-us/fabric#/styles/web/icons))? If so, did you use `OutlookLogo` or `OutlookLogoInverse`?

PeterShaggyNoble:
I'm seeing it on the right-hand side of the page but I guess not everyone else will. Not sure what to do for the source - I'll have another quick hunt 'round, see if I can find another page or site that features both the icon and the colour.

image

PeterShaggyNoble:
Found one! Updating now.

PR #2558, opened 2020-02-07
Update Yammer

PeterShaggyNoble:
Yammer

**Issue:** #1585

### Checklist

### Description
See #2541 for notes on source URL


ericcornelissen:
The "Y" has some weird angles to it but it seems the source material does too :thinking: I'm fine with adding it this way but I'll request a second pair of :eyes: from another @simple-icons/maintainers just in case :+1:

runxel:
It looks weird, yes, but not _toooo_ shoddy.
I think we can add this.

PeterShaggyNoble:
No that it's been pointed out, it's annoying me! 😆 But beyond my limited skills to try to fix! 🙁

runxel:
Hey @PeterShaggyNoble!
I can totally understand the pain :D 

What do you think of this?

[yammer](https://github.com/simple-icons/simple-icons/files/4207392/yammer.zip)

----

It's totally redrawn, has much less control points, and straight lines are actually straight. (And of course you would see differences if you do an overlay – but that's clearly expected and a mudged origin can never yield a totally equal looking _constructed_ svg)

PeterShaggyNoble:
Appreciate that, thanks @runxel 😀

Your version was just a hair over 24px wide and scaling down to size required increasing the precision to 4 when running it through SVGOMG.

runxel:
> Your version was just a hair over 24px wide and scaling down to size required increasing the precision to 4 when running it through SVGOMG.

Ach, that reminds me to check my svgs more careful after I svgo them. Precision 3 is indeed not always sufficient...

Issue #2527, opened 2020-02-04
Contactless Payment


PeterShaggyNoble:
**Name:** Contactless Payment
**Website:** https://www.emvco.com/emv-technologies/contactless/
**Alexa Rank:** [~467k](https://www.alexa.com/siteinfo/emvco.com)
**Colour:** `#000000`
**Official resources for icon and color:** https://upload.wikimedia.org/wikipedia/en/8/83/Universal_Contactless_Card_Symbol.svg (Brand guidelines available [here](https://www.emvco.com/about/trademark-centre/))

And, so we finally come to the last icon on my list from MDI! 🎉 And one I don't know that we should include as, to me, it pushes the boundaries of standards & protocols a step to far. Thoughts?

[![Contactless Payment](https://user-images.githubusercontent.com/15157491/73759621-1f8f2d00-4764-11ea-86a0-46d21c9bcbfc.png)](https://github.com/simple-icons/simple-icons/files/4154337/contactlesspayment.zip)

ericcornelissen:
I personally think this is a common enough standard/protocol to be included, at least from a European perspective. I believe it is becoming more common in the U.S./Canada as well, but I don't know for sure (no idea about other countries/continents).

I do agree that it stretching the boundaries a bit, but I guess this is comparable to [RSS](https://simpleicons.org/?q=rss). I think this is brand-like enough to be included. Any opinions from other @simple-icons/maintainers (@birjolaxew)?

PR #2567, opened 2020-02-10
Update guidelines for source URLs

ericcornelissen:
**Issue:** Following https://github.com/simple-icons/simple-icons/pull/2499#issuecomment-580798030

### Description

Updated the guidelines for the `source` URL,: slightly changing what we use them for, and providing more detailed guidelines for common sources.

This is more like draft, would appreciate feedback on the wording etc. 🙂

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,7 +152,7 @@ Make sure the icon is added in alphabetical order. If you're in doubt, you can a
 
 #### Source Guidelines
 
-We use the source URL as a reference for the current SVG in our repository and as a jumping-off point to find updates if the logo changes. If you used one of the sources listed below, make sure to follow these guidelines.
+We use the source URL as a reference for the current SVG in our repository and as a jumping-off point to find updates if the logo changes. If you used one of the sources listed below, make sure to follow these guidelines. If you're unsure about the source URL you can open a Pull Request and ask for help from others.
 
 If the new SVG is sourced from:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,7 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Website headers (you can use [svg-grabber](https://chrome.google.com/webstore/detail/svg-grabber-get-all-the-s/ndakggdliegnegeclmfgodmgemdokdmg) for Chrome)
 1. Favicons
 1. Wikimedia (which should provide a source)
+1. GitHub repositories
 
 Working with an SVG version of the logo is best. In the absence of an SVG version, other vector filetypes may work as well (e.g. EPS, AI, PDF). In the absence of vector logos, a vector can be created from a high quality rasterized image, however this is much more labor intensive.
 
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble:
There are some files that aren't available on Wikimedia, only on WIkipedia ([this one](https://upload.wikimedia.org/wikipedia/en/8/83/Universal_Contactless_Card_Symbol.svg) I used for #2527, for exmaple) so I'd suggest including either the brand's or logo's Wikipedia page as an alternative source in those cases.

We should also list GitHub repos as a potential source in [this section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors).

ericcornelissen:
> There are some files that aren't available on Wikimedia, only on WIkipedia ([this one](https://upload.wikimedia.org/wikipedia/en/8/83/Universal_Contactless_Card_Symbol.svg) I used for #2527, for exmaple) so I'd suggest including either the brand's or logo's Wikipedia page as an alternative source in those cases.

Interesting 🤔 I updated the description for Wikipedia/Wikimedia, let me know what you think @PeterShaggyNoble 


> We should also list GitHub repos as a potential source in [this section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors).

Good point, I will update that list as well 👍

ericcornelissen:
> Just a minor tweak;
> 
> > For an SVG from Wikipedia/Wikimedia the source URL should link to the ~~page containing file information~~ **logo's page on the relevant site, and not the brand's Wikipedia page**

I changed it slightly to say "[...] logo **file's** page [...]" :+1:

PeterShaggyNoble:
Even better 😀 This looks good to me now, Eric but, again, as it relates to our documentation, I'll leave it to someone else to throw another eye over it before merging.

runxel:
One question (because I'm not sure if I got that right):

> In either case the commit hash should be part of the URL.

Can one just append the commit hash to the base url of a repo?
Does this then point to the repo being at this point of time of the repo?

ericcornelissen:
> Does this then point to the repo being at this point of time of the repo?

That is correct and exactly why I want it. This prevents issues when e.g. the file is removed from the project later on.


> Can one just append the commit hash to the base url of a repo?

Not really, for GitHub it should be somewhere in the middle of the URL (replacing the branch name, to be exact). Unfortunately, as far as a I know, GitHub doesn't have an easy way of getting this kind of _permalink_ :confused:

Do you think it would be a good idea to add a step-by-step tutorial on how to get the URL we want @runxel?

runxel:
> > Can one just append the commit hash to the base url of a repo?
> 
> Not really, for GitHub it should be somewhere in the middle of the URL (replacing the branch name, to be exact). Unfortunately, as far as a I know, GitHub doesn't have an easy way of getting this kind of _permalink_ 😕

Ach, that's what I thought!
(I was just confused for a moment thinking there might be something I've always missed until now on Github 😆)

> Do you think it would be a good idea to add a step-by-step tutorial on how to get the URL we want @runxel?

Yeah, that would be much appreciated, I'd say, @ericcornelissen!  👍

ericcornelissen:
Updated the guidelines @runxel, turns out there is an "easy" way to get the commit hash in the URL, but it is hidden behind a keybinding... Pressing <kbd>y</kbd> on any(?) GitHub page will include the commit hash (as per [the docs](https://help.github.com/en/github/managing-files-in-a-repository/getting-permanent-links-to-files)) :upside_down_face: Glad I looked that up

runxel:
Woah, that's interesting!
Cool you've found that!

I think this PR is good to go 😎

PeterShaggyNoble:
😮 *Great* find, Eric - that's gonna save a bit of time in the future!

Looks like this is good to go now,

PR #2582, opened 2020-02-11
Add Windows 95

### LINKED ISSUE #2535: Windows 95
**Name:** Windows 95 (or 98?)
**Website:** n/a
**Alexa Rank:** n/a
**Colour:** `#02a5ff` from the SVG. Only chosen 'cause of the sky background associated with those versions of Windows - other suggestions welcome.
**Official resources for icon and color:** https://upload.wikimedia.org/wikipedia/commons/1/1e/Windows_Logo_1995.svg

OK, _one_ last icon from MDI then!

Under discussion here is, not only if we want to add this particular icon, but also whether  we want to start entertaining requests for legacy or retro brands, which I think we would need to do if we added this one. It should be noted that we do already have icons for the likes of Playstation, Playstation 3 & Wii.

[image](https://github.com/simple-icons/simple-icons/files/4161087/windows95.zip)

PeterShaggyNoble:
Windows 95

**Issue:** Closes #2535

### Checklist


runxel:
Duuuuuude, the color should obv. be `#008080`....
Everything else would be so misleading ;P

But seriously. Why the blue? Nobody would ever associate that color with Win95, would they? 😅

PeterShaggyNoble:
My defining memory of Windows 9x is [this image](https://en.wikipedia.org/wiki/File:Windows98shutdown-en.png) so I associate the sky blue with it.

The teal I definitely associate with the OS but not so much with the brand. What do the other @simple-icons/maintainers think?

runxel:
Okay. I would still argue, the OS _is_ the brand. :D 
Even if you said, you associate it with the sky blue, then why an aggresive blue like your `#02A5FF`?
Wouldn't it be more like `#a5c6de`?

Still: for _me_ it must be teal. `#008080`

PeterShaggyNoble:
OK, you've convinced me! 😆 PR updated.

I think, for these older brands, where brand guidelines can no longer be found - if they ever existed - we're OK to go with our best judgement of the colour people most associate with it.

I'd still like to get some feedback from the other @simple-icons/maintainers before we merge this in, though.

runxel:
No one complained for 8 days :)
I will merge this in for now, we can always update the color, if somebody disagrees later.
Thanks for the work, @PeterShaggyNoble!

Issue #2587, opened 2020-02-12
[COLLECT] Airlines


runxel:
- [x] Aer Lingus (#2607)
- [x] Aeroflot (#4881)
- [x] Aeroméxico (#4760)
- [x] Air Asia (#4892)
- [x] Air Canada (#2833)
- [x] Air China (#4799)
- [x] Air France (#2891)
- [x] Alitalia (#4808)
- [x] All Nippon Airways (ANA) (#4821)
- [x] American Airlines (#2827)
- [x] British Airways (#4908)
- [x] China Eastern (#4827)
- [x] China Southern (#4837)
- [x] Delta (#4844)
- [x] easyJet (#2598)
- [x] Emirates (#4862)
- [x] Ethiopian Airlines (#4909)
- [x] Etihad (#4957)
- [x] KLM (#2588)
- [x] Lion Air (#4959)
- [x] LOT (#5221)
- [x] Lufthansa (#2113)
- [x] Norwegian Air (#5145)
- [x] Pegasus (#5144)
- [x] Qantas (#2839)
- [x] Qatar Airways (#4869)
- [x] Ryanair (#2606)
- [x] S7 (#4888)
- [x] Southwest (#2896)
- [x] TUI (#2851)
- [x] Turkish Airlines (#1527)
- [x] United (#2803)
- [x] Wizz Air (#2847)

others:
- [ ] PanAm ? :D – just for the chuckles, and as a "retro" brand

---

See [wikipedia](https://en.wikipedia.org/wiki/Largest_airlines_in_the_world).

PeterShaggyNoble:
Dibs on Ryanair & Aer Lingus! :laughing:

PeterShaggyNoble:
Thoughts on this treatment of the American Airlines icon? (ref: #1874)

![American Airlines](https://user-images.githubusercontent.com/15157491/77050368-92293480-69c1-11ea-88c5-95e37912154e.png)

runxel:
Looks great! I like that :)

PeterShaggyNoble:
Morbid as it may sound, we have made the decision to put the remaining brands in this issue on hold until the current crisis is over and we see which airlines and travel operators survive - we don't want to be adding a bunch of icons that we may need to remove again in a few weeks or months.

PeterShaggyNoble:
So, 9 months later, do we want to continue holding off on these or do we take it that any airlines, etc. that have survived this long are going to be OK?

adamrusted:
[...]

PR #2598, opened 2020-02-14
Add easyJet

runxel:
**Issue:** #2587

### Checklist

### Description
There was sadly no svg to be found, neither on their website nor Wikipedia.
However it was easy to recreate, since the wordmark is set in [Cooper Black](https://fontsinuse.com/uses/16586/easyjet). The kerning was adjusted to fit to a higher resolution picture of the logo.
The color `#FF751A` was lifted from the website's css.


PeterShaggyNoble:
I extracted the wordmark from page 14 of [this PDF](https://easy.com/wp/wp-content/uploads/2018/10/easyGroup_Brand_Manual-1.pdf) and converted it to a path and your version (in read) doesn't _quite_ line up. From page 18 ofhat PDF, as well, there are a couple more choices for the colour:
- `#fe5000` from converting the Pantone
- `#ff5700` from converting the CMYK
- `#ff6600`

easyJet

runxel:
Hey @PeterShaggyNoble. thanks for joining in here!
The link to the brand guide is great, I saw it referenced mutliple times but the links were all dead; nice to see you've found it!

However, please do your comparison with the _actual_ logo from the website – you will see that indeed the kerning in the PDF is wrong (it was just set in Cooper Black with no additional kerning done on it).

I'm fine with either of those colors, we just need to agree on one. :D

PeterShaggyNoble:
Ah, I _completely_ misread your initial comment, don't know what I think you meant now! Sorry 'bout that.

From past experience, you're much better at comparisons between vector and raster than I so I'll defer to you there, pending a review from one of the other @simple-icons/maintainers, of course.

On the colour: I'd vote `#ff6600` as it's provided twice in the brand guidelines. But I also think that your initial choice is an acceptable alternative given that it's also sources from easyJet, if that's the consensus.

runxel:
`#ff6600` it is, then!

We could link to `https://easy.com/` ? 
I mean, that's the parent page. I am not sure if we should link against the document itself. Pretty sure @PeterShaggyNoble would be concerned of link rot, and actually I am as well in this case.
(Given that this just a basic wordpress link it doesn't really look persistent enough)

ericcornelissen:
@runxel I think the easyJet website is better than the easy website. I agree that link rot is a potential problem, and linking the PDF itself is not ideal in any case. The main reason I asked for suggestions is because you said earlier:

> The link to the brand guide is great, I saw it referenced multiple times but the links were all dead; nice to see you've found it!

So I was wondering if any of the pages where you saw it referenced could be used as a source. If not, then I think this is fine as is.

<sup>at least we have a reference to the brand guidelines in this discussion...</sup>

runxel:
> > The link to the brand guide is great, I saw it referenced multiple times but the links were all dead; nice to see you've found it!
> 
> So I was wondering if any of the pages where you saw it referenced could be used as a source. If not, then I think this is fine as is.

Ah, got it, sorry!
Funnily enough, the links were all on _unrelated_ websites (like from `fontsinuse.com` and therelike).
So yes, best is to stay on easyjet.com, like it was already.

PeterShaggyNoble:
All looks good to me 😃

PR #2624, opened 2020-02-19
Add micro:bit

PeterShaggyNoble:
micro:bit

**Issue:** [MDI#4009](https://github.com/Templarian/MaterialDesign/issues/4009)

### Checklist

### Description
Icon taken from SVG in website header. Light green seems to be their primary brand colour so I grabbed the green from [this file](https://github.com/microbit-foundation/microbit-svg/blob/master/microbit-drawing.svg) but there's also `#00ed00` which is the starting point of the background gradient of their site and a case could also be made for `#001700` seen in the header and footer, which is obtained by layering 90% black over `#00ed00`.

**Alexa rank:** [31891](https://www.alexa.com/siteinfo/microbit.org)

**Note:** The build has failed due to the `:` in the brand name, which we don't currently handle. It has been addressed in #2625.


PeterShaggyNoble:
Hmm ... anyone know why the build is still failing after merging #2625? 🤔

ericcornelissen:
Ow, I always forget this :sweat_smile:  We need to update `index.html` separately as it doesn't (and cannot?) use the JavaScript scripts... One more reason to start working on #770 and #898 for me :upside_down_face:

I'll update the code in `index.html` for you @PeterShaggyNoble  :+1:

ericcornelissen:
The build errors for this PR should be resolved once https://github.com/simple-icons/simple-icons/pull/2654 is merged. As proven by c36d4774e84505fdf3065786a7a99338546dbb53 (specifically [this build](https://travis-ci.com/simple-icons/simple-icons/jobs/289731159)) :smile:

PeterShaggyNoble:
> We need to update index.html separately as it doesn't ... use the JavaScript scripts

Will try to remember that for the future. Merged #2654 so trying the build on this again ...

PeterShaggyNoble:
:+1: Success!

_data/simple-icons.json
```diff
@@ -3060,6 +3060,11 @@
             "hex": "DE4F4F",
             "source": "http://logo.meteorapp.com/"
         },
+        {
+            "title": "micro:bit",
+            "hex": "0191DC",
```
ericcornelissen:
```suggestion
            "hex": "00ED00",
```

PeterShaggyNoble: Ah, feck! Somehow managed to copy & paste the wrong colour from that SVG file. The green in it is actually `#10d95e`. But, looking at the social media & apps, I'd agree with the `#00ed00` choice. Thanks, Eric.

ericcornelissen: Would've been okay with both if it is sourced from the SVG, thanks for updating :+1:

Issue #1531, opened 2019-07-15
Resolve non-alphanumeric characters in brand slugs


ericcornelissen:
### Kind of issue 
  - [ ] Improvement
  - [ ] Bug
  - [x] Other, namely: _future proofing_

### This issue concerns 
  - [x] The NPM Package
  - [ ] The website
  - [ ] Other, namely: 

### Description

> Ye, that definitely should be removed in slugs too. We might want to consider a catch-all stripping of unknown characters (e.g. `.replace(/\W/g, "")`) at the end of our `titleToFilename`.

_Originally posted by @birjolaxew in https://github.com/simple-icons/simple-icons/pull/1519#issuecomment-511227997_

Not sure if it should be all unknown characters (`/\W/g`), all non-ascii characters, or all non-alphanumeric characters. Anyway, the idea arose from removing at least single quotes (`'`) from the filename/brand slug.

For reference, currently this is our algorithm to convert the brand name to the filename/brand slug:

https://github.com/simple-icons/simple-icons/blob/7672bfb2dbd0fa89bfa9d45c1749d288a1c48093/scripts/utils.js#L6-L24

Pinging @birjolaxew, @davidklebanoff, and @paulmelnikow for feedback.

paulmelnikow:
Hi! Thanks for the ping. 😀

This is the formula @RedSparr0w developed in https://github.com/badges/shields/pull/3641#issuecomment-509394351:

> Leave `-`, `.`, `a-z` and `0-9` characters:

Input | Output
:-- | :--
`let’s-encrypt` | `lets-encrypt`
`.NET` | `.net`
`Not a Re@l Service` | `not-a-rel-service`
`Yahoo!` | `yahoo`

It seemed to work for everything except Google and Google+.

I'm curious whether or not that's an improvement over what you're using now. If looks like what you have will change `google+` to `googleplus` which seems better, though I guess moot since #1513 is merged.

Are there any with @ signs in them?

Also pinging @chris48s who, like @RedSparr0w has done a lot of work on the logos in Shields, and may have input.

ericcornelissen:
> Leave -, ., a-z and 0-9 characters:

Leaving dashes (`-`) in seems reasonable to me, unfortunately introducing dots (`.`) is a bit of a problem for us at this point in time (and also, in my opinion, not ideal for filenames).

[...]

PR #2625, opened 2020-02-19
Replace Colons & Multiple Accented Characters in Filenames

PeterShaggyNoble:
As seen in #2624, we don't currently handle colons in brand names when creating the filename. The change in this PR simply strips them out but, alternatively, we could replace them with hyphens instead.


PeterShaggyNoble:
Hadn't seen that issue, Eric. Would the better solution, I wonder, be to remove the line I modified in this PR and instead add a new `replace` to the end of the chain like so: `.replace(/[^\-\w]/g,"")`.

I also now notice that the replacement of accented accented letters isn't using a global RegEx; we should update that, too.

ericcornelissen:
I'd say we merge this in as is* and proceed the discussion in #1531. Using `/[^\-\w]/g` is probably fine, but we might want to use similar solutions to `.Net` -> `dot-net` for other symbols as well...


>  also now notice that the replacement of ~~accented~~ accented letters isn't using a global RegEx; we should update that, too.

That, we should :+1: *Do you want to include that in this Pull Request?

PeterShaggyNoble:
> Do you want to include that in this Pull Request?

Just pushed that update.

ericcornelissen:
> Just pushed that update.

Okay, that is a simple change and can be merged without a problem.

I would like to have an extra pair of :eyes: from the @simple-icons/maintainers look at the change of **removing ":" (colons) from brand names**. Feedback from the community is also welcome! (@paulmelnikow perhaps?)

PeterShaggyNoble:
Added a couple more replacements:

- Stripping out regular `'`s
- Replacing `š` with `s`

runxel:
Added more common diacritics in 0fd4cdb
`č, ć, ň, ń, ś, ž, ź`

paulmelnikow:
Pinging @chris48s, do you have any input on behalf of Shields?

chris48s:
I don't see any issue with this for our usage :+1:

PeterShaggyNoble:
Thanks, @paulmelnikow & @chris48s :+1: Merging this in, so.

Issue #1984, opened 2019-11-24
[COLLECT] Add icons of CAD brands


runxel:
Some I have in mind: (Please comment, if you have more suggestions)

- [x] ANSYS (#3211)
- [x] Archicad (#2633)
- [x] Autodesk Autocad (#9920)
- ~CATIA~ _(invalid, see comment below)_
- [ ] fritzing
- [ ] Autodesk Inventor
- [x] KiCad #7517
- [x] Bentley Microstation (#8934)
- [ ] Nemetschek Allplan
- [ ] PTC
- [x] Autodesk Revit (#8956)
- [x] Rhino & Grasshopper (#2197)
- [x] Sketchup (#3587 & #4311)
- [ ] SolidEdge
- ~SolidWorks~ _(invalid, see comment below)_
- [x] Vectorworks (~#3586~ #4764)

fatihmeh:
FYI, #1549 added Dassault Systemes logo which has similarities to Solidworks and Catia logos.

runxel:
> FYI, #1549 added Dassault Systemes logo which has similarities to Solidworks and Catia logos.

Oh yeah, you're right. I still remembered when they had their own distinctive branding.

sjstark:
Was about to make a request for the following before I found this existing issue:
[Vectorworks](https://www.vectorworks.net/en-US): [Alexa Rank 137,530](https://www.alexa.com/siteinfo/vectorworks.net)
[Full White SVG Found](https://www.vectorworks.net/cached/themes/vectorworks_redesign_2018/assets/images/logos/vw-logo-full.svg)
Trailing text would need to be clipped

[Solidworks](https://www.solidworks.com/): [Alexa Rank 11,345](https://www.alexa.com/siteinfo/solidworks.com)
[Full White SVG Found](https://www.solidworks.com/sites/default/files/2018-02/SWlogo33.svg)
Trailing text would need to be clipped

[Sketchup](https://www.sketchup.com/): [Alexa Rank 1,896](https://www.alexa.com/siteinfo/sketchup.com)
[Best SVG Found](https://commons.wikimedia.org/wiki/File:SketchUp_logo.svg)
Official site has logo saved as PNG :/

jack-mil:
Kicad added in #7517

uncenter:
[...]

PR #2633, opened 2020-02-19
Add Archicad (#1984)

runxel:
**Issue:** #1984

### Checklist

### Description
Archicad (or how the producer wants "ARCHICAD") is one of the leading CAD/BIM programs.
You can see the logo (which ist rather sparsely to be found on the homepage) [here](https://www.graphisoft.com/archicad/), if you scroll down a bit.
It also acts as the program's icon on Windows and Mac. 

I found a [svg on one of their sites](https://wildcads.graphisoft.de/wp-content/uploads/2019/10/AC-Logo_2.svg).
I made it monochrome like others did already in [a similiar fashion](https://www.bimarium.com/). (See the first logo there.)

The color is kinda tricky, since the homepage has not so much of it.
They like it blue, anyhow.

See:
acad

The blue `#313D6B` I took is from the homepage's CSS (one of the few blues) (and is also in the current icon).

[Alexa rank ~ 40k.](https://www.alexa.com/siteinfo/graphisoft.com)


PeterShaggyNoble:
Nice one, @runxel :+1:

I'm seeing a minor difference in the left inner curve (original in black) that you might want to double check but I'd say it could be down to differences in how we both cropped and scaled the original.

runxel:
> I'm seeing a minor difference in the left inner curve (original in black) that you might want to double check but I'd say it could be down to differences in how we both cropped and scaled the original.


Yeah that was being expected. In the original are _a lot_ unnecessary control points, most of them malformed (sharp kinks and there like). I rebuilt this curve to have a better control point structure, so it might be a tiny bit off. (which should be unnoticable)

PeterShaggyNoble:
In that case, I'd say this is good to go.

PR #2647, opened 2020-02-20
Add Vauxhall

runxel:
**Issue:** #2059

### Checklist

### Description
The icon was taken from a PDF @PeterShaggyNoble posted [here](https://github.com/simple-icons/simple-icons/issues/2059#issuecomment-589177828).
Color is from there as well.

Just don't try to look and match it with the original. Those pesky auto manufacturer svgs are horrible and working on them will certainly cut a few years of your lifespan. Also nothing lines up.


PeterShaggyNoble:
Wow, great work, @runxel 😀 Trying to figure that one out yesterday made my brain itch! As the original was such a mess, we'll need one of the other @simple-icons/maintainers to throw another eye over it before merging.

Note that the red they use on their main website is `#c90c0f` and on the website I sourced the PDF its `#ce181e` but, I think your choice is a valid one, too.

Speaking of the source, I'd suggest [this page](https://www.vauxhall.co.uk/microapps/vxr/corsa-vxr/), where the PDF can be found.

runxel:
Updated the source

ericcornelissen:
Thanks for the work @runxel, looks like it was pretty rough :100:

PR #2692, opened 2020-02-27
Add Note on Previews to PR Template

PeterShaggyNoble:
Adds a link to [my preview image generator](https://petershaggynoble.github.io/MDI-Sandbox/simpleicons/preview/) to our pull request template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,13 +2,11 @@
 Before opening your pull request, have a quick look at our contribution guidelines:
 https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md
 
-Add a preview image of your submission using:
-
+Consider adding a preview image of your submission using:
 https://petershaggynoble.github.io/MDI-Sandbox/simpleicons/preview/
 -->
 
 **Issue:**
-
 
 ### Checklist
   - [ ] I updated the JSON data in `_data/simple-icons.json`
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,5 +1,11 @@
 <!--
-Before opening your pull request, have a quick look at our contribution guidelines: https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md
+Before opening your pull request, have a quick look at our contribution guidelines:
+https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md
+-->
+
+<!--
+Add a preview image of your submission using
```
ericcornelissen: I would:

- Just include it in the same comment.
- Add a colon at the end, like the existing line.
- Either have an extra empty line in both or in neither.

```suggestion

Add a preview image of your submission using:

```

ericcornelissen: I would reword this using "consider" so  contributors don't feel obliged to use it. For now I would like to keep this as an optional thing so we can get some feedback on the system. Perhaps even ask for feedback here? What do you think @PeterShaggyNoble?

```suggestion
-->

<!--
Consider adding a preview image of your submission using
```

PeterShaggyNoble:
Thanks for the detailed feedback, Eric. Agree with all those changes, particularly the wording - the less work there is for contributors putting PRs together, the better - definitely don't want people feeling obliged to add them.

> Also, what would you say about adding a note about this in the Contributing Guidelines?

Good idea. Although, I'd hold off until we've had a bit of feedback and I've worked the last few kinks out. Or maybe even until we get the generator integrated into the new site or set up in its own SI repo, as we discussed.

PR #2706, opened 2020-02-28
Add McDonald's

PeterShaggyNoble:
McDonald's

**Issue:** [MDI#4930](https://github.com/Templarian/MaterialDesign/issues/4930) (assuming the issue title was a typo!)

### Checklist

### Description
Icon & colour sourced from [this PDF](https://www.mcdonalds.com/content/dam/uk/nfl/pdf/reports/260619-McDonalds45.pdf), linked from the source URL. Personally, I'd consider red to be their primary brand colour but I think most other people will be expecting the yellow.

**Alexa rank:** [6066](https://www.alexa.com/siteinfo/mcdonalds.com)


PeterShaggyNoble:
It looks to me like the build is failing [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/build-package.js#L34) and/or [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/templates/icon-object.js#L2) , because of the `'` in the icon title.

ericcornelissen:
> 
> 
> It looks to me like the build is failing [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/build-package.js#L34) and/or [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/templates/icon-object.js#L2) , because of the `'` in the icon title.

Actually it only fails in the first instance, the second instance [is escaped here](https://github.com/simple-icons/simple-icons/blob/21b23da612aa9dcde607c634a13aa11694401fba/scripts/build-package.js#L38). I'll open a Pull Request with a fix for the build 👍

PR #2710, opened 2020-03-01
Escape icon title for object key

ericcornelissen:
**Issue:** #2706

### Description

Updates the build script to escape `'` in object keys.


PeterShaggyNoble:
Looks good to me, Eric - thanks for the quick fix 🙂

PR #2746, opened 2020-03-06
Add Toyota

PeterShaggyNoble:
Toyota

**Issue:** #2059 (Could've sworn one of us added this in the first round)

### Checklist

At _last_ a car manaufacturer with _proper_ brand guidelines!


PeterShaggyNoble:
Forgot to include this alternative treatment, which is what I would have expected this to be but I didn't spot it in use or available anywhere in their brand guidelines.

Toyota

ericcornelissen:
I am very surprised the Visual Identity System shows the logo in a square :confused: I would totally expect your "alternative" treatment to be their logo... but given the guidelines I think the version currently in the PR is the way to go.

Any other opinions from the community or @simple-icons/maintainers?

ericcornelissen:
That seems to be some kind of consensus :smile: Thanks for the addition @PeterShaggyNoble 


> My last resort is the argument they didn't put rectangle logos on their cars...

As much as I love that argument, I'm afraid we're going with the rectangle version for now :upside_down_face:

PR #2766, opened 2020-03-10
Rename TikTok

PeterShaggyNoble:
**Issue:** n/a

### Checklist

### Description
There shouldn't be a space in the name of this one. ~I don't _think_ this constitutes a breaking change - the filename remains `tiktok.svg` - but didn't want to commit directly, just in case.~

It was indeed a breaking change (see comments below) so **do not** merge this until `v3`.


ericcornelissen:
I'm confident this **is** a breaking change for our NPM and Packagist(?) packages. In Node TikTok can currently be accessed as:

```javascript
simpleIcons.get("Tik Tok");
```

And after the change, I'm fairly certain, this would no longer work. However I'm currently on my phone so I am not 100% confident and can't test...

PeterShaggyNoble:
Ah. Feck! Good thing I *didn't* commit directly, so!

Adding this to the `v3` miletsone.

ericcornelissen:
Tested it now, it does indeed break backwards compatibility so we can make this change in `v3` 👍

PeterShaggyNoble:
In the interests of keeping all these together, this will be addressed in #2942.

PR #2777, opened 2020-03-12
Add Oshkosh (#2059)

runxel:
**Issue:** #2059

### Checklist

### Description
EPS can be found on the [Press/Media page](https://oshkoshdefense.com/media/photos). Color is black or orange.
oshkosh

oshkosh


PeterShaggyNoble:
Given it's inclusion in #2059, I was expecting this to be the parent [Oshkosh Corporation](https://oshkoshcorp.com/), rather than their military vehicle division. Their Alexa rank is only [876647](https://www.alexa.com/siteinfo/oshkoshdefense.com) but Alexa only has data from the ast 2 months (rather than the usual 3), suggesting this might be a relatively new site.

The PR itself looks good to me but I would suggest, for the purposes of future disambiguation, renaming it as Oshkosh Defense.

If this is somehow in-scope despite it's Alexa rank, I will be leaving this to one of the other @simple-icons/maintainers to merge in as I have a moral issue with companies that profit from war and, personally, don't think they're something we should be including - this would set that precedent. That said, I'm always happy to abide by the consensus of the team, if it's decided that we _do_ want to include companies like this.

runxel:
> The PR itself looks good to me but I would suggest, for the purposes of future disambiguation, renaming it as Oshkosh Defense.
 
> If this is somehow in-scope despite it's Alexa rank, I will be leaving it to one of the other @simple-icons/maintainers to merge in as I have a moral issue with companies that profit from war.

Total fair point and I'm on your site.
When I was a kid we had a quartet game (those cards, maybe you can recall that – it's a lost thing nowadays, I assume) with big trucks. An 'Oshkosh' was one of them. They make airport fire trucks and "normal" trucks as well.
To me was just a memory of childhood. But we can totally skip that.

However, _if_ we do we should maybe update the contribution guidelines. (On the other hand its quite hard to decide – most corporations deal with the military. We would need to take down Amazon, Microsoft, Google, et al. as well...)

PeterShaggyNoble:
I _thought_ I hadn't heard of this company before, which was the main reason I assumed your request was for the Oshkosh _Corporation_, but, when you mentioned those cards from your childhood, it got me thinking of my own and I remembered that [Matchbox](http://www.matchbox.com/) had a couple of Oshkosh trucks (way) back in the day. I also seemed to _vaguely_ recall the branding being an orange or a yellow so I went off to do a bit more research on them.

First thing I discovered is that they do, as you said, have other subsidiaries ([Oshkosh Airport](https://www.oshkoshairport.com/) for example) that use the same logo, which somewhat assuaged my reservations about them _solely_ being in the "business" of war. If possible, though, I'd like to see if we can find an alternative source URL.

I also found [this SVG](https://commons.wikimedia.org/wiki/File:Oshkosh_Corporation_logo.svg) in use on their [German](https://de.wikipedia.org/wiki/Oshkosh_Corporation) and [Norwegian](https://no.wikipedia.org/wiki/Oshkosh_Corporation) Wikipedia pages featuring the orangey colour I mentioned above. The same logo can be seen on [this page](https://fs.oshkoshglobal.com/adfs/ls/) linked from the footer of their Airport site. I think black is also a valid choice, though, as my preference for the orange is probably being influenced by unreliable memories of my childhood.

runxel:
> I think black is also a valid choice, though, as my preference for the orange is probably being influenced by unreliable memories of my childhood.

Oh yes! this was also how I remembered the logo! I just didn't found the color on their site, but you are – again – a _battle_ proven (sorry for the pun) detective.
I will update the color! :)

Any idea for a better link? The svg I have there is already monochrome, but with the colored ones there is no cut out.

PeterShaggyNoble:
> Any idea for a better link?

Not that I've been able to find, unfortunately. So, let's not get hung up on my personal issues - we can update the source URL at a later date if we do find a better one.

Once the colour is updated then the only outstanding issue is that of the Alexa rank but, now that I remember it and have done all that research, I think we can say it's a well-established brand and therefore within scope. We'll still need the opinion of at least one of the other @simple-icons/maintainers on that, though.

PeterShaggyNoble:
Don't forget to update the colour, @runxel 😉

Or are you waiting for further feedback on whether it's the correct choice of colour?

ericcornelissen:
Based on a quick glance at [the Wikipedia page for the corporation](https://en.wikipedia.org/wiki/Oshkosh_Corporation) it seems to me that Oshkosh as seen in this PR is for _Oshkosh defense_ (hence the URL), which is a subsidiary (or w/e) of the the _Oshkosh Corporation_, which is in the business of vehicle manufacturing in general.

To avoid adding a business that is solely in the business of war I would suggest switching to the _Oshkosh Corporation_.

I don't have a problem necessarily with adding _Oshkosh defense_ in addition to that, but perhaps we wait until it is requested by the community? If we do decide to add it now, I'd say we need to rename this one to _Oshkosh defense_.

PeterShaggyNoble:
@ericcornelissen, note that they use the same logo for other subsidiaries such as [Oshkosh Airport](https://www.oshkoshairport.com/). The reason for the source URL being what it is is that it's the only one, so far, that either @runxel or myself can find which features a vector version of their monochrome logo.

That's not to say, of course, that we can't add an icon for the parent corporation as well.

runxel:
Yeah, the parent company is just that, the parent. 
The logo in this PR is for me more essential; it's whats on the trucks.
They bought alot of other companies, that's why they have the parent company.

Maybe let's rename it to `Oshkosh Airport & Defense`? :)

ericcornelissen:
> Yeah, the parent company is just that, the parent.
> The logo in this PR is for me more essential; it's whats on the trucks.

That's fair, wasn't sure which of the two it was but if this is the logo you want then that's fine :+1: 


> Maybe let's rename it to Oshkosh Airport & Defense? :)

That sounds good to me, given @PeterShaggyNoble comment I'm not sure if that covers it all? Either way, I personally feel some renaming would be appropriate, though I am also okay with adding this as just _Oshkosh_.

PeterShaggyNoble:
I think we should treat it the same as Ferrari; add this as just "Oshkosh" and the parent company as "Oshkosh Corporation".

ericcornelissen:
> I think we should treat it the same as Ferrari; add this as just "Oshkosh" and the parent company as "Oshkosh Corporation".

Alright, then it's settled :+1: Thanks for the addition @runxel :smile:

PeterShaggyNoble:
Weren't we going to go with the `#e6830f` colour?

ericcornelissen:
> Weren't we going to go with the `#e6830f` colour?

Apologies :sweat_smile:  I'm okay with both `#000000` and `#e6830f` so feel free to update

PeterShaggyNoble:
@runxel, do you agree with the `#e6830f`, too? If so then let's update :+1:

PR #2819, opened 2020-03-20
Add Logitech

PeterShaggyNoble:
Logitech

**Issue:** Templarian/MaterialDesign#941

### Checklist

### Description
Icon & colour from [this SVG](https://www.logitech.com/content/dam/logitech/common/header/logitechg-overlay.svg) in website header. I've gone with the same treatment that they use for [their favicon](https://www.logitech.com/favicon.ico) and that can be seen on some of their products ([example](https://www.logitech.com/assets/65762/streamcam.jpg)). They seem to have a palette of colours but the blue appears a few times throughout their site so I went with that.

**Alexa rank:** [4158](https://www.alexa.com/siteinfo/logitech.com)


runxel:
The dot of the `i` looks really weird, it's not round at all.
I mean, it's in the source as well, but I can't think that this is right... 🤔 
Looks a bit like being potraced.

PeterShaggyNoble:
Well  spotted, @runxel :+1: It also wasn't centred over the `i` so I fixed that, too.

Issue #2799, opened 2020-03-17
New Relic


skyzyx:
**Name:** New Relic
**Website:** https://newrelic.com
**Official resources for icon and color:** https://newrelic.com/about/media-assets

phatbhoy67:
[Alexa rank: 13,367](https://www.alexa.com/siteinfo/newrelic.com)

Suggest source used is this [official svg](https://newrelic.com/assets/newrelic/source/NewRelic-logo-bug-clr-w.svg) available to download from [media-assets](https://newrelic.com/about/media-assets) page, removing the wordmark to match the favicon configuration of the logo 👇  

![favicon](https://newrelic.com/assets/apple-touch-icon.png?v=WGoP4WGmYG)

Possible hex values include `#008c99` and `#70ccd3` taken from [colour svg](https://newrelic.com/assets/newrelic/source/NewRelic-logo-bug.svg) available from [media-assets](https://newrelic.com/about/media-assets) page.

PR #2835, opened 2020-03-23
Added the New Relic icon

skyzyx:
**Issue:**

### Checklist

### Description

Fixes https://github.com/simple-icons/simple-icons/issues/2799.


_data/simple-icons.json
```diff
@@ -3680,6 +3680,11 @@
             "hex": "00C7B7",
             "source": "https://www.netlify.com/press/"
         },
+        {
+            "title": "New Relic",
+            "hex": "008c99",
```
phatbhoy67:
```suggestion
            "hex": "008C99",
```

PeterShaggyNoble:
Thanks for taking this on, @skyzyx :+1:

I was just about to comment the same as @phatbhoy67. However, from the background colour used for the white logos at the source URL, I would suggest using the lighter of the 2 colours.

@phatbhoy, I would have expected [this treatment](https://newrelic.com/assets/newrelic/source/SVG/Vert_1col_Rev.svg) rather than the solid version you suggested. And, although I too would prefer it without the wordmark, that it isn't an option on their media assets page might mean we must include it.

skyzyx:
> And, although I too would prefer it without the wordmark, that it isn't an option on their media assets page might mean we must include it.

Looking at their source for the _Safari pinned tab_, I see this:

```html
<link rel="mask-icon" href="/etc/designs/newrelic/products-icon/safari-pinned-tab.svg" color="#5bbad5"/>
```

Perhaps there's some wiggle room to duplicate what they themselves are doing? I'll wait for confirmation from you about whether or not you're OK going down this path before updating the PR.

phatbhoy67:
> @phatbhoy, I would have expected [this treatment](https://newrelic.com/assets/newrelic/source/SVG/Vert_1col_Rev.svg) rather than the solid version you suggested. And, although I too would prefer it without the wordmark, that it isn't an option on their media assets page might mean we must include it.

I suggested a version that would look like this 👇 😕 

rect4774

Which maintains the format of the original favicon. The version you included @PeterShaggyNoble, the ellipses are different

PeterShaggyNoble:
Sorry, @phatbhoy67, mixed up the links in my comment above, meant to link [this one](https://newrelic.com/assets/newrelic/source/NewRelic-logo-bug-clr-w.svg).

phatbhoy67:
> Sorry, @phatbhoy67, mixed up the links in my comment above, meant to link [this one](https://newrelic.com/assets/newrelic/source/NewRelic-logo-bug-clr-w.svg).

Hey @PeterShaggyNoble, I'm confused, that is the exact same source I suggested should be used in https://github.com/simple-icons/simple-icons/issues/2799#issuecomment-599985516 as the basis for the icon? 😕 

rect4774

With the wordmark removed and fill unset it would look like this ☝️ 

So, I guess the question is, do we go with the version above or follow the treatment as seen in the [`safari-pinned-tab.svg`](https://newrelic.com/etc/designs/newrelic/products-icon/safari-pinned-tab.svg) as suggested by @skyzyx of filling/merging the two outer coloured ellipses.

PeterShaggyNoble:
> Hey @PeterShaggyNoble, I'm confused

No, looks like I'm the one that's confused; I misinterpreted your original comment on the issue! 😆

My vote goes to the completely outlined version - I don't think the pinned tab SVG is an accurate representation of the logo - and the lighter of the 2 colours.

PeterShaggyNoble:
Closing this out due to lack of activity. I'll get a fresh PR together for it shortly.

Issue #2822, opened 2020-03-20
[COLLECT] Metro Networks


samuelpolat:
- [ ] Beijing Subway
- [ ] Shanghai Metro
- [ ] Guangzhou Metro
- [ ] Seoul Subway
- [ ] Tokyo Metro
- [ ] Moscow Metro
- [ ] Shenzhen Metro
- [ ] Hong Kong Mass Transit Railway
- [ ] New York City Subway (#2820)
- [ ] Mexico City Metro
- [ ] Paris Métro
- [x] London Underground (#2742)
- [ ] São Paulo Metro
- [ ] Cairo Metro
- [ ] Singapore Mass Rapid Transit 

Based on [List of metro systems](https://en.wikipedia.org/wiki/List_of_metro_systems)

samuelpolat:
![seoulmetro](https://user-images.githubusercontent.com/10519501/77224941-5083bf80-6b62-11ea-9c38-34e3aadd991f.png)

I extracted the icon from the [Seoul Metro Wikipedia page](https://en.wikipedia.org/wiki/Seoul_Metro) and chose `#293896` as color as it is referenced in their [brand guidelines](http://www.seoulmetro.co.kr/kr/page.do?menuIdx=507).

**Alexa rank:** [~174k](https://www.alexa.com/siteinfo/seoulmetro.co.kr)

samuelpolat:
> ![seoulmetro](https://user-images.githubusercontent.com/10519501/77224941-5083bf80-6b62-11ea-9c38-34e3aadd991f.png)
> 
> 
> 
> I extracted the icon from the [Seoul Metro Wikipedia page](https://en.wikipedia.org/wiki/Seoul_Metro) and chose `#293896` as color as it is referenced in their [brand guidelines](http://www.seoulmetro.co.kr/kr/page.do?menuIdx=507).
> 
> 
> 
> **Alexa rank:** [~174k](https://www.alexa.com/siteinfo/seoulmetro.co.kr)

Another popularity metric would be their [annual ridership of ~1.9 billion](https://en.m.wikipedia.org/wiki/Seoul_Metropolitan_Subway) (2017).

samuelpolat:
> > ![seoulmetro](https://user-images.githubusercontent.com/10519501/77224941-5083bf80-6b62-11ea-9c38-34e3aadd991f.png)
> > 
> > 
> > 
[...]

PR #2845, opened 2020-03-25
Add Cairo Metro

PeterShaggyNoble:
Cairo Metro

**Issue:** #2822

### Checklist

### Description
Colour from [website](https://cairometro.gov.eg/) stylesheet.

A [low Alexa rank](https://www.alexa.com/siteinfo/cairometro.gov.eg) but a good suggestion was made by @samuelpolat in [#2822](https://github.com/simple-icons/simple-icons/issues/2822#issuecomment-602503457) to use the annual ridership to ascertain the popularity of transportation networks and the 1.3b figure for Cairo Metro would put it on a par with London's Underground at 1.4b.


phatbhoy67:
Hey @PeterShaggyNoble thanks for taking this one on.

The .svg is great 👍 

I can find the hex value chosen in the stylesheet but can't seem to see it used on the web page, or am I just being a bit dim? 🤔 

The red linear gradient I can see used most on the page has stops `#ef533c` and `#a00b11` with a [calculated midpoint](https://meyerweb.com/eric/tools/color-blend/#A00B11:EF533C:1:hex) of `#C82F27`, perhaps use that instead?

PeterShaggyNoble:
Sorry, @phatbhoy67, should have given a bit more detail as to where I grabbed that colour from - it's from the cubes in the preloader which looked to be close enough to the red used in the [PNG](https://cairometro.gov.eg/images/metrologo.png) from the same preloader. If you have a look in the HTML source, you can find those cubes nested in the last `div` before the `script` tags at the end - they have a class of `.sk-cube`. It can also be found in use for the FA icons in the footer and is pretty close to the red used in the source SVG, which is `#c01414`

Compared to the colour in their logo, the midpoint of the gradient looks a bit "washed" out though so if we were to change it `#af1015` might be a better choice instead instead. They use it for, among other things, the links in their menu.

PR #2851, opened 2020-03-26
Add TUI

PeterShaggyNoble:
TUI

**Issue:** #2587

### Checklist

### Description
Icon from [this SVG](https://www.tuiholidays.ie/static-images/_ui/mobile/hotfix_11a_20-202003140731/framework/tui/TUI-Logo.svg), colour from website header.

**Edit:** ~Came across [a styleguide](https://tui-nordic.github.io/common-ux/) after putting this PR together which confirms the [hex reference](https://tui-nordic.github.io/common-ux/pages/atoms/colors.html) for "TUI Blue". Will do a comparison in a while between the icon in this PR and the one available [here](https://tui-nordic.github.io/common-ux/pages/atoms/icons.html).~ There is one minor _hairline_ difference in the curve of the smile, not enough to worry about seeing as the version in the PR is from an official source so this is good to go as-is unless one of the other @simple-icons/maintainers thinks otherwise.


runxel:
Wait, _blue_? 😨

ericcornelissen:
I agree that red would be good choice as well, but judging from their branding I would say they "want" blue to be their brand colour, It's just that the logo itself is red...

runxel:
Where exactly? I seem to have missed the point where it says blue is the main color.
Even if there is this part: The link goes to "TUI nordic", which seems to differ a bit from "global TUI"

> Additionaly to the colors defined in the global profile the Nordic market also use its own ‘Red Nordic’ and ‘Blue Nordic’.

PeterShaggyNoble:
> I agree that red would be good choice as well, but judging from their branding I would say they "want" blue to be their brand colour, It's just that the logo itself is red...

Yup, exactly this. From their website to [their printed material](https://www.tui.co.uk/brochures/) to their TV ads, blue is the dominant colour with the red only being used for the logo and as an accent colour.

runxel:
Okay, it might be their favorite brand color.... :shipit: 
But the logo is still in red? And I mean, we're collecting the logos here.
Leave the other colors to brandcolors.net. 😅

PeterShaggyNoble:
So, I guess the choice then depends on whether we're collecting _brand_ colours or _logo_ colours here. I've always thought it was the former.

runxel:
For most companies those two are at leat closely related.
I was just thinking: Somebody uses our package to display the logo. Then they also rely on the color for it. The TUI logo is now blue. Pretty sure people would think that they encountered a bug... 🤔

PeterShaggyNoble:
Hmm ... 🤔 ... I can see your point. I'd still make the case for the blue, though. But let's see if we can get a consensus from the other @simple-icons/maintainers. So far, that's one vote for each.

ericcornelissen:
Historically we have been leaning towards the _brand colour_ rather than the _logo colour_, therefore I am in favour of blue as well. That said, I'd be willing to make an exception here because of how consistently red is used for the logo.

phatbhoy67:
Found this [.pdf](http://media.tuicontent.nl/banners/tui/brandguide/TUI_Logo_Manual.pdf) - on page three, under heading translated as "corporate identity" TUI-Blue, `#70cbf4` is stated as primary colour, TUI-Red listed as secondary colour?

runxel:
Okay okay, I admit defeat 😅

ericcornelissen:
In that case, I merged this in :smile: Thanks for the contribution @PeterShaggyNoble!

PR #2869, opened 2020-03-30
Add Fastify

PeterShaggyNoble:
Fastify

**Issue:** #2363

### Checklist

### Description
There are a number of apparent imperfections in this but, as they are present in [all the versions of their logo](https://github.com/fastify/graphics) that I checked, I'm not sure whether they're intentional or if we should try to fix them.

**Alexa rank:** [~255k](https://www.alexa.com/siteinfo/fastify.io)


PeterShaggyNoble:
@runxel, do you mean the second tooth from the front on the top jaw?

runxel:
Exactly! :)
I think you can just delete that point, since the neighbour sits on the exact line to the next tooth.

PeterShaggyNoble:
:+1: Updated

PR #2877, opened 2020-03-30
Add Loom icon

### LINKED ISSUE #2837: Icon request: Loom
**Name:** Loom
**Website:** loom.com
https://www.loom.com/press
#fd5e60

LuckyF:
loom

**Issue:** Closes #2837 

### Checklist


PeterShaggyNoble:
Thanks for the contribution, @LuckyF. Which version of their logo did you base your version on? It seems to be very inconsistent between files and even, at times, _within_ individual files.

Also, looking at their press page in more detail, I notice this line:
> When you are limited on space, use only the logotype.

Doesn't make sense to me why they'd choose the wordmark over the icon but, if that's what they want, we may have to go with it. What do the other @simple-icons/maintainers think?

ericcornelissen:
> > When you are limited on space, use only the logotype.
> 
> Doesn't make sense to me why they'd choose the wordmark over the icon but, if that's what they want, we may have to go with it. What do the other @simple-icons/maintainers think?

That... is weird. I'm fairly confident that by "limited space" they are thinking about something other than what simple-icons is... So I would stick with (something that resembles) the current SVG.

Other than that this looks good to me, besides @PeterShaggyNoble's other comment.

LuckyF:
Hi @PeterShaggyNoble,
I used the _"loom_logo_symbol_black.svg"_ from their [official press kit](https://www.dropbox.com/sh/n9cjqz3uf025e8d/AAAc-XYD__WzbJ43IzkdIMjXa/Logo?dl=0&subfolder_nav_tracking=1).

PeterShaggyNoble:
A-ha! I hadn't seen that, @LuckyF; I was using the SVGs on [this page](https://www.loom.com/press) and was getting some differences so wanted to make sure I was comparing like with like. Thanks for the link.

I think you'll need to up the precision on this by one as there are a couple of small differences, such as in the top right corner in the first screenshot below (your version in red). The original also has some imperfections around the outer edge of the inner "circle" which we should probably clean up for them, as seen in the second screenshot below.

LuckyF:
Unfortunately I have messed up my git repo and had to create a new pull request (#2951).

PR #2891, opened 2020-04-02
Add Air France

PeterShaggyNoble:
Air France

**Issue:** #2587

### Checklist

### Description
Icon & colour from [this SVG](https://objects.airfrance.com/FR/common/common/img/tab1st/header/logo_air_france.svg). Do we need to go with the full wordmark here? :thinking:


runxel:
Good question! I don't know if the stylized wing is enough already.
Technically everything looks good! We could merge.

They once (long ago) had a really [funny logo](https://i.pinimg.com/originals/10/6a/a9/106aa944e50db7b5b79cb390d4be85ca.jpg) btw!

PeterShaggyNoble:
The more I look at it, the less I'm convinced that it's the right choice either. What do the other @simple-icons/maintainers think?

ericcornelissen:
I don't see any reason why we would change it to the wordmark. The shape is quite distinct and they use it like this as well. If you worry that it is not recognizable enough, @PeterShaggyNoble, I'd say that that is not our worry (on top of that, at least for me, I'm just not at all familiar with the brand and logo)

PeterShaggyNoble:
My primary concern was that, other than their favicon, I hadn't spotted in use on its own anywher on their site. However, I now see that they use it for [their app](https://play.google.com/store/apps/details?id=com.airfrance.android.dinamoprd) icon and on [social media](https://www.facebook.com/airfrance/) (although backed up by the full logo underneath) so I think we're OK to do the same.

ericcornelissen:
Ah I didn't spent a long time on their website, but indeed the app and social media profiles use it in conjunction with the wordmark.

Still, I think this is good to go so I will merge it now. If there are any objections we can change it later :slightly_smiling_face:

ericcornelissen:
Thanks for the work @PeterShaggyNoble :tada:

PR #2914, opened 2020-04-08
Add Luos

Emanuel-Allely:
luos

**Issue:** N/A

### Checklist

### Description

Luos is building a new, open and universel framework for embedded development. Their rank on Alexa in not very high because they recently change their domain name to luos.io but a lot of projects are using this technology. On Crunchbase their rank is 19k.

I pick the hex value because it's the official color coming with the logo.
I didn't manually vectorize the logo, I took the vector source file.


PeterShaggyNoble:
Welcome to Simple Icons, @Emanuel-Allely and thanks for the contribution 🙂

Could you provide us with a link to the vector file you based your icon on? I don't see one on their website nor can I find one in any of their GitHub repos.

Also, to help us gauge the popularity of this brand, can you let us know what their old domain was? Currently both their Alexa rank and the activity in their GitHub repos would put them outside our scope.

Emanuel-Allely:
Thanks @PeterShaggyNoble for your answer! I'm one of the founders of this company, that's why I directly get an .svg file. Before our domain was luos-robotics.com. Luos is not really famous yet (we just got out of INRIA research lab) but more and more projects are using our technology and libraries (such as https://tcrn.ch/2s6l8gd). We would like to propose a badge (like shields.io proposes) for each project embedding our libraries. I would be really great if Luos logo could in Simple Icons! I hope you'll enjoy this project and allow it to be on SI :)

PeterShaggyNoble:
Thanks for the additional info, @Emanuel-Allely

It looks like our old domain had a rank of ~4.5m before the change over, which, unfortunately, still puts it beyond our 500k cut-off point. Coupling this with the current low rank of your new site and the numbers on your GitHub repos & social media accounts, I'm afraid I'm going to have to say that this is out of scope for now. But we'll be more than happy to revisit it once you're more established.

Thanks again for your work on the PR and all the info you provided.

Emanuel-Allely:
Hi @PeterShaggyNoble, thanks for your answer.

I don't really understand why you're limiting the number of companies on Simple Icons. I understand that you can't bring all logos of all brands on your own and you need to focus on brands with a lot of traction. But what about brands like Luos which try to do the job bringing you nice PR 🙂. Maybe I'm wrong but I think that the more brands you have on SI, the more traction and visits you'll get.

Beyond that it could be so helpful for Luos to be on Simple Icons! We are bringing a new tool for developers working on embedded systems. We want to change the way Electronics is made worldwide. Bringing Luos on SI could real contribution to this project! Please, could you reconsider your position, you will not regret it! 🙂

ericcornelissen:
Hi @Emanuel-Allely, chiming in here for a second. The main reason we don't want to add just any icons, even if they are contributed, is twofold: first is that every icon requires some level of maintaining as it needs to stay up to date, and second (and more importantly) it is not that easy for us to just remove the icon once the brand no longer exists. This is because removing an icon breaks backwards compatibility.

However, again, if you can give us an indication of how "popular" Luos is, that might change or decision. Keep in mind that the @simple-icons/maintainers are perhaps not very familiar with robotics :slightly_smiling_face: And don't forget, even if Luos is not "popular enough" now, we may still add it in the future!

Last but not least, simple icons is not the only way to get your logo on badges/shields! For example, with [Shields.io](https://shields.io/), you can use the `?logo` URL parameter to manually set to logo to anything you want (PNGs and SVGs work, and more). You, as the project owner of Luos, can simply give the correct URL to anyone that wants to include it in their README.

PR #2930, opened 2020-04-13
Update CDN URLs in the README

ericcornelissen:
**Issue:** -

### Checklist
n/a 

#### Update example URLs

Before, the example links included `@latest`. This is a problem when we make a major release. If anyone is using an icon we removed with an @latest URL, then the icon no longer works. Therefore, I changed `@latest` to `@v2`, which resolves to the latest `2.x.x.` version available.

This behaviour is illustrated by the URLs for `@v1`:

- https://cdn.jsdelivr.net/npm/simple-icons@v1/
- https://unpkg.com/browse/simple-icons@v1/

#### The Unpkg URL

The Unpkg URL in the text linked to the Unpkg homepage but the JSDelivr URL in the text linked to simple-icons' page on JSDelivr. Now, both of them link to the simple-icons page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Icons can be downloaded as SVGs directly from [our website](https://simpleicons.
 
 ### CDN Usage
 
-Icons can be served from a CDN such as [JSDelivr](https://www.jsdelivr.com/package/npm/simple-icons) or [Unpkg](https://unpkg.com). Simply use the `simple-icons` npm package and specify a version in the URL like the following:
+Icons can be served from a CDN such as [JSDelivr](https://www.jsdelivr.com/package/npm/simple-icons) or [Unpkg](https://unpkg.com/browse/simple-icons/). Simply use the `simple-icons` npm package and specify a version in the URL like the following:
 
 ```html
 <img height="32" width="32" src="https://cdn.jsdelivr.net/npm/simple-icons@v2/icons/[ICON NAME].svg" />
--- a/README.md
+++ b/README.md
@@ -18,15 +18,15 @@ Icons can be downloaded as SVGs directly from [our website](https://simpleicons.
 Icons can be served from a CDN such as [JSDelivr](https://www.jsdelivr.com/package/npm/simple-icons) or [Unpkg](https://unpkg.com). Simply use the `simple-icons` npm package and specify a version in the URL like the following:
 
 ```html
-<img height="32" width="32" src="https://cdn.jsdelivr.net/npm/simple-icons@latest/icons/[ICON NAME].svg" />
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble:
Could we make a case for showing both URLs (`@latest` _and_ `@vX`) so people can choose their own, preferred option?

We'd also need to remember to update the version number with each major release.

ericcornelissen:
> Could we make a case for showing both URLs (@latest and @vX) so people can choose their own, preferred option?

I suppose, but: for anyone that already knows the difference, they can easily change it if they want; whereas for anyone that doesn't know the difference it is quite meaningless and possibly confusing. 

The reason I prefer `@vX` is that it is _safer_. Of course, someone using `@vX` that doesn't understand it might want to use an icon release in `v(X+1)` and be confused that it doesn't work...

In either case (showing both `@latest` and `@vX` or not) it might be good idea to link to some webpage that does a good and quick job at explaining the topic.

What do you think @PeterShaggyNoble?


> We'd also need to remember to update the version number with each major release.

Yes, of course, that is perhaps a bit unfortunate... But, even if we forget to update it we'll probably get a report about it soon enough.

PeterShaggyNoble:
> The reason I prefer @vX is that it is _safer_

Agreed.

> In either case (showing both @latest and @vX or not) it might be good idea to link to some webpage that does a good and quick job at explaining the topic.

Yep, a bit more detail on the differences and pros & cons between using an exact version number, the major version number only and `@latest` definitely wouldn't hurt and would help people not savvy on these things make a more informed decision as to the best option for them.

ericcornelissen:
Alright, any ideas for what we should link to? Here are some possibilities:

- [The SemVer specifications](https://semver.org/)
- [Documentation from JSDelivr related to this topic](https://github.com/jsdelivr/data.jsdelivr.com#resolve-a-version-range)
- [NPM's documentation on SemVer](https://docs.npmjs.com/misc/semver)
- [This blog post that ranks high on Google](https://nodesource.com/blog/semver-a-primer/)

Or do you think we should write something simple ourselves? (in addition to a link?)

PeterShaggyNoble:
Yeah, I think add our own simple description with links out to something more detailed.

ericcornelissen:
I added the following line about it :arrow_down: Currently without a link as I felt this actually captures quite well what we want to say. And I think non of the links in my previous comment add something useful.

> These examples use the latest major version. This means you won't receive any updates following the next major release. You can use `@latest` instead to receive updates indefinitely. However, this will result in a `404` error if the icon is removed.

PeterShaggyNoble:
That works for me, sums it up quite nicely :+1:

As this involves a change to our documentation, we'll need another eye thrown over it before merging it in.

PeterShaggyNoble:
In that case, we can go ahead and merge this in.

PR #2961, opened 2020-04-21
Add Jamboard

PeterShaggyNoble:
Jamboard

**Issue:** Templarian/MaterialDesign#5056

### Checklist

### Description
Unfortunately, the icons in the source PDF are all raster so I traced over the one on the top right of page 32. My vector image "skills" suck so if someone can do a better job then please feel free to update this PR. Of course, if you can find a vector version we can use instead, which would also give us a better source URL, even better! Another option might be to base it off the `jam_drive` icon found in Google's [Material Icons Extended font](https://fonts.googleapis.com/icon?family=Material+Icons+Extended) (available to view [here](https://petershaggynoble.github.io/MDI-Sandbox/google/)). As seen below, though, the gaps between elements are much bigger.

Colour is from page 50 of source PDF.

Jamboard


PeterShaggyNoble:
> Do you happen to remember where you found the `source` pdf file?

Only the PDF itself came up in my search results, @ericcornelissen, I wasn't able to find any pages linking to it.

> The only thing left is judge popularity. Jamboard does not seem to have it's own domain separate from Google so we can't use its Alexa rank.

They have jamboard.com which doesn't register on Alexa but as it just redirects to Jamboard's GSuite page, I don't think we can read too much into that.

> Did you find any other sources for the popularity of Jamboard @PeterShaggyNoble?

I think that, other than the simply being a Google product, the other metric I went by was it having over 500k downloads on [Google Play](https://play.google.com/store/apps/details?id=com.google.android.apps.jam) which seems reasonably significant, although we don't yet have a cut-off for that as a metric.

ericcornelissen:
> I think that, other than the simply being a Google product, the other metric I went by was it having over 500k downloads on [Google Play](https://play.google.com/store/apps/details?id=com.google.android.apps.jam) which seems reasonably significant, although we don't yet have a cut-off for that as a metric.

I'm not sure if 500k+ downloads over 3 years (first release was May of 2017) is super popular, but I'd suppose we need to investigate this further before we actually start using it (are updates included in this count? Does 500k+ mean between 500k and 500k or between 500k and 1m? How many downloads do brands already included in the collection have? Etc.). But as a Google service I'm fine with adding this for now :+1:

PR #2965, opened 2020-04-21
Clarify we don't want overlapping paths

ericcornelissen:
**Issue:** n/a

### Description

Following the discussion on 69ae2c829b1c33708e77c02998224d038802d980 this updates the Contributing Guidelines _"Extract the Icon from the Logo"_ section based on @runxel's suggestion.

An alternative (almost equal) solution to consider is:

```diff
  1. Isolate the icon from any text or extraneous items.
+ 1. Merge any overlapping paths and compound all paths into one.
  1. Change the icon's viewbox/canvas/page size to 24x24.
```

---

<sup>Went with a PR rather than a direct commit because I wasn't 100% sure if the discussion was finished. Also, maybe, some input from the contributing community</sup>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,7 +65,9 @@ There are many different tools for editing SVG files, some options include:
 
 Using your preferred tool you should:
 
-1. Isolate the icon from any text or extraneous items. Please also make sure that there are no overlapping paths by merging them.
+1. Isolate the icon from any text or extraneous items.
+1. Merge any overlapping paths.
+1. Compound all paths into one.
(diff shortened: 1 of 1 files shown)


runxel:
I think both solutions here a valid. :)
Thanks for taking care of this!

PeterShaggyNoble:
The version the PR currently uses is best, I think.

But should it be "combine" rather than "compound" in step 3?

runxel:
`Compound` is at least the right terminus technically. @PeterShaggyNoble 
Combine is rather ambigous, while if you don't know what compound means, Google will point you in the right direction.

PR #2979, opened 2020-04-25
Update Fur Affinity icon

KovuTheHusky:
**Issue:** Fur Affinity [updated their logo](https://www.furaffinity.net/journal/9364332) at the end of last year.

### Checklist

### Description

The [raster icon](https://www.furaffinity.net/themes/beta/img/banners/fa_logo_20191231.png) was vectorized using Adobe Illustrator's image trace.


PeterShaggyNoble:
Thanks for the update, @KovuTheHusky. I'm not too good with comparing raster to vector so I'll leave that to one of the other @simple-icons/maintainers.

However, should the colour not also be updated to one of the greys used on their site?

Also, their favicon is quite different to their logo and would be better suited to our purposes. Having said that, on a quick run through their site, I don't see it in use anywhere else which is usually a case _against_ going that route. If we do decide to then an SVG of that icon can be found at https://github.com/FurAffinity/node-web/blob/2b9e53c334496af7300e74fae3c450a7c74de6dc/public/images/favicon.svg

KovuTheHusky:
> Thanks for the update, @KovuTheHusky. I'm not too good with comparing raster to vector so I'll leave that to one of the other @simple-icons/maintainers.

Understood!

> However, should the colour not also be updated to one of the greys used on their site?

I didn't feel confident enough to pick one color myself, though my recommendation would possibly be the deep blue from the scarf in the logo as the website has multiple themes that change the UI colors entirely.

> Also, their favicon is quite different to their logo and would be better suited to our purposes. Having said that, on a quick run through their site, I don't see it in use anywhere else which is usually a case _against_ going that route. If we do decide to then an SVG of that icon can be found at https://github.com/FurAffinity/node-web/blob/2b9e53c334496af7300e74fae3c450a7c74de6dc/public/images/favicon.svg

That favicon has been used for years but has never been used elsewhere. The previous icon in simple-icons was also from the header/nav though I understand if there's a preference for the old favicon since it's a simpler shape. It is fairly generic though.

PeterShaggyNoble:
> the website has multiple themes that change the UI colors entirely.

Would you like to try reaching out to them, see if they have a preference for the colour?

> That favicon has been used for years but has never been used elsewhere.

In that case, I'd agree that it wouldn't be the right choice for us - a favicon can occasionally not be an accurate reflection of the branding.

PeterShaggyNoble:
Did you have any luck getting feedback from them, @KovuTheHusky? If not then I'd suggest in the absence of anything else to go by then I think we should use one of the colours from their default theme.

KovuTheHusky:
My ticket is still open with no response. ☹️

PeterShaggyNoble:
In that case, let's pick a colour ourselves and update, if necessary, when/if they get back to you. The options that have been presented so far are:

1. Leave the current colour as-is,
2. Use one of the greys from the default theme (my vote), and,
3. The blue from the scarf in the source image (@KovuTheHusky's vote).

What do the other @simple-icons/maintainers think?

ericcornelissen:
Leaving the color as is definitely incorrect. I would be inclined to agree with @PeterShaggyNoble but given that @KovuTheHusky is most familiar with the brand as well as the remark that:

> [...] the website has multiple themes that change the UI colors entirely.

makes me think the suggested blue (`#36566F`) is perhaps a better choice for now.

PeterShaggyNoble:
> given that @KovuTheHusky is most familiar with the brand

A fair point :+1:

PeterShaggyNoble:
As far as my abilities go to compare vector to raster, your path looks pretty spot on to me, @KovuTheHusky so, after the colour is updated, if there are no other objections, I think this can be merged in.

ericcornelissen:
Thanks for the update @KovuTheHusky 🎉 Glad it's finally merged.

PR #3018, opened 2020-05-01
Add audio-technica 

### LINKED ISSUE #3015: audio-technica logo request
**Name: audio-technica**
**Website: https://audio-technica.com/**
**Official resources for icon and color: https://de.wikipedia.org/wiki/Datei:Audio-technica.svg**

Orchomenos:
audio-technica (2)

**Issue:** Closes #3015

### Checklist

### Description

Source for logo and colour is: https://upload.wikimedia.org/wikipedia/commons/b/b5/Audio-technica.svg, but I edited the logo so it is only contains the icon and fits the requirements.


Orchomenos:
> Glad to see you figured out how to create a Pull Requests @Orchomenos 👍
> 
> Your work looks pretty good already. There is just one thing that you will need to change, as you can see from your preview and the image below the logo does not fit the 24x24 viewbox exactly.
> 
> [The extract the icon section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#2-extract-the-icon-from-the-logo) of our guidelines should explain how to fix this. Essentially, you will need to make sure that you have a 24x24 pixels canvas in you editing software and then fit the logo in there. (if you need more help, feel free to ask us here and tell us what software you're using)
> 
> audio

I thin I habe found my mistake. I am using Inkscape and there is a setting which is called something like scale for svg-user units per pixel, which I had on 2,6(standard setting) and now I put changed it to one.

Orchomenos:
> Nice job @Orchomenos, the SVG looks good now 👍
> 
> I'm just wondering if the title should be `"Audio-Technica"` instead of `"audio-technica"` (with capital letters). From my point of view, though I'm not familiar with the brand at all, the wordmark seems to use lowercase letters but in text the name is with capital letters. Look, for example, at the name in the website title or their social media pages.
> 
> So, could you updated the title in both the JSON file and the SVG 🙂 (or make a case for staying with all lowercase)


audio-technica packaging

While I see where you are comming from, I used their packaging as help, because there it is written with smaller case letter...
What do you think is more representative of a company - their website or packaging?

ericcornelissen:
Well, it is not just there website but also their social media. As I pointed out, it seems to me that their wordmark logo uses lowercase letters and the brand name is with capitals.

If you could, can you verify that the box does not contain a capitalized version of the name at all (e.g. on the back).

Either way, I will ask for feedback from the other @simple-icons/maintainers on the issue to get some more input.

Orchomenos:
IMG_20200503_202235

On my box they used it with capital letters when using their offical company name and with smaller case letters when referring to their website or brand as a whole, so now I am even more confused.

runxel:
I think the crucial part is that they write themselves "Audio-Technica" in the copyright part. So my vote goes there! :)

PeterShaggyNoble:
I'd say it should be title case too.

ericcornelissen:
In that case, could you update the title in the JSON and SVG files @Orchomenos, thanks!

ericcornelissen:
Thanks for the update and thanks for the contribution @Orchomenos :tada: It is very much appreciated!

PR #3022, opened 2020-05-04
Add Razer

EinsEli:
razer
 
### Checklist
  - [x] I optimized the icon with ~~SVGO or~~ SVGOMG

### Description

The logo is from [wikipedia ](https://de.wikipedia.org/wiki/Razer#/media/Datei:Razer_Schlangen_logo.svg)and the colors are from the same image.


PeterShaggyNoble:
Thanks for the PR, @EinsEli.

I had a quick look around their site for an official and better quality vector version of that logo but couldn't find one. However, I _would_ suggest using [this version](https://en.wikipedia.org/wiki/File:Razer_snake_logo.svg) instead as it is more recent but still with its own little imperfections which we could do with cleaning up.

I agree with your choice of colour as it's used on their site, too. Although, in the absence of any official guidelines, I _could_ see a case being made for one of the greys they use.

ericcornelissen:
> I agree with your choice of colour as it's used on their site, too. Although, in the absence of any official guidelines, I could see a case being made for one of the greys they use.

Chiming in here just to say that green is definitely the brand colour of Razer.

EinsEli:
I've changed the icon to the one on [wikipedia ](https://en.wikipedia.org/wiki/File:Razer_snake_logo.svg)

PeterShaggyNoble:
Thanks for the update, @EinsEli.

The new path however, is coming in a `24.001` in height, rather than `24`. There are also a number of imperfections in the source file (examples below) which we should consider fixing.

EinsEli:
> Thanks for the update, @EinsEli.
> 
> The new path however, is coming in a `24.001` in height, rather than `24`. There are also a number of imperfections in the source file (examples below) which we should consider fixing.
> 
> 

Should be fixed now :)

Orchomenos:
> Thanks for the PR, @EinsEli.
> 
> I had a quick look around their site for an official and better quality vector version of that logo but couldn't find one. However, I _would_ suggest using [this version](https://en.wikipedia.org/wiki/File:Razer_snake_logo.svg) instead as it is more recent but still with its own little imperfections which we could do with cleaning up.
> 
> I agree with your choice of colour as it's used on their site, too. Although, in the absence of any official guidelines, I _could_ see a case being made for one of the greys they use.
> 
> **Alexa rank:** [~2.8k](https://www.alexa.com/siteinfo/razer.com)

I tried finding their logo on their website too and found this version as an .svg: https://d4kkpd69xt9l7.cloudfront.net/sys-master/images/h6c/he1/8858452525086/wordmark.svg

ericcornelissen:
@Orchomenos I don't think using the wordmark is appropriate for Razer, but thanks for the help.

PeterShaggyNoble:
Thanks for your work on this one, @EinsEli :+1: I'd be happy to merge this in as-is. I'm going to hold off for a day or 2, though, just to give the other @simple-icons/maintainers an opportunity to double check that we didn't miss any other imperfections in the source that should be fixed.

My first instinct upon seeing their website was that we should be using the wordmark, too, @Orchomenos, or maybe just the `E`. However, from their favicon and [this page](https://www.razer.com/eu-en/legal/trademark-use-guidelines), I think the snakes icon is the correct choice.

PeterShaggyNoble:
As there have been no objections, I'm going to go ahead and merge this in. Thanks again for your work on it, @EinsEli :+1:

Issue #2734, opened 2020-03-04
Adobe Icon Review


PeterShaggyNoble:
In light of #2733, and following on from #2364, we should do a full review of all of our Adobe icons. Most all of their brands have their own product pages (e.g., [Phostoshop](https://www.adobe.com/products/photoshop.html)) with SVG logos.

- [x] Adobe (#2769)
- [x] Adobe Acrobat Reader (#2769)
- [x] Adobe After Effects (#2769)
- [x] Adobe Audition (#2769)
- [x] Adobe Creative Cloud (#2769)
- [x] Adobe Dreamweaver (#2769)
- [x] Adobe Illustrator (#2769)
- [x] Adobe InDesign (#2769)
- [x] Adobe Lightroom CC
- [x] Adobe Lightroom Classic
- [x] Adobe PhoneGap
- [x] Adobe Photoshop (#2733)
- [x] Adobe Premiere (#2770)
- [x] Adobe Typekit (#2771, #2776)
- [x] Adobe XD

PeterShaggyNoble:
Closing this out as I have the rest of these ready to go once #2769 is merged (I stupidly worked on the same branch! 🙄)

PR #3032, opened 2020-05-06
Update Adobe Creative Cloud

PeterShaggyNoble:
Adobe Creative Cloud

**Issue:** #2734 and separated out of #2769

### Checklist

### Description
Icon & colour from [this SVG](https://www.adobe.com/content/dam/shared/images/product-icons/svg/creative-cloud-logo.svg) found in source URL header.

Possible alternative treatment based on the icon in the header of [fonts.adobe.com](https://fonts.adobe.com/) and in use in a few other places:

Adobe Creative Cloud

### Comparison
With our current version in red.
Adobe Creative Cloud


PeterShaggyNoble:
Well spotted, @phatbhoy67, thanks. There was a `clip-path` in the source that I'd missed which was throwing things off slightly; should be OK now.

I also think that, as it's the one they use on product pages, that the version in the PR is the one we should go with. It's also clearer and more easily identifiable at `24*24`.

PR #3035, opened 2020-05-06
Add Rancher (#3034)

### LINKED ISSUE #3034: Request: Rancher icon
**Name:** Rancher
**Website:** www.rancher.com
**Official resources for icon and color:** https://rancher.com/brand-guidelines/

Happy to take this up.

kshitishp:
Rancher

**Issue:** Closes #3034

### Checklist

### Description
All hex values, vectors and other brand requirements taken from https://rancher.com/brand-guidelines/


kshitishp:
> 1. You can remove the registered trademark symbol. Unless a company explicitly & irrevocably requires its inclusion or its an integral part of the design, we generally omit them.

I checked up #2460 and felt this icon falls most appropriately under point 5 (There is an explicit requirement that the symbol be included) - though there is a guideline that the ® should be included if/when the copy contains certain labels (https://rancher.com/brand-guidelines/). Do I take it as this creating ambiguity and making this fall under point 3 (There is an explicit requirement that the symbol be included but there is ambiguity about under which conditions it may be removed)?

> 2. You'll need to remove the CloudStack JSON entry and SVG from your branch.

Oops :) Rookie mistake!

PeterShaggyNoble:
Ah! I was looking for guidance on the trademark symbol in the wrong place; under what _not_ to do with the logo. I see it now under what _to_ do.
> Use the cow (with ®)

That certainly suggests that it _should_ be included but not necessarily that it _must_ be included. I was looking at the "Don't"s, looking for something that explicitly stated _not_ to remove it. So, yeah, there is a wee bit of ambiguity there, which from #2460 would mean we should include it. Although that issue is still in discussion.

My other concern now is the rest of that point:
> ... if the Rancher Labs, RancherOS, Rancher is visible in copy

That suggests that the cow should _only_ be used on its own if the company name is mentioned in the copy of the page it's being displayed on, which would mean we'd need to include the wordmark (but could omit the symbol). But, again, it's ambiguous so I'd say we can get away without including it.

Let's see how the other @simple-icons/maintainers feel.

kshitishp:
End of the day, it's an icon, right? So it shouldn't be an issue as such. The ®, at those small sizes, only makes stuff rather unreadable. I don't see a lot of brands covering glyphs in their brand guidelines - so it's ambiguous, to say the least, and it should be ok to not include the mark.

I'll wait for the maintainers before pushing another commit.

runxel:
Chiming in on that topic:
I'm in favor of using the cow alone, but, and I sadly have to agree, we are in need to include the ®️ then...
Even tho #2460 is not finally settled: If a company is stupid enough to brace their demand of including a tautological boilerplate and smudging the icon, we can't deter it. 😃

ericcornelissen:
I'd agree as well that it is better to include the (R) symbol in this case

kshitishp:
> 1. You can remove the registered trademark symbol. Unless a company explicitly & irrevocably requires its inclusion or its an integral part of the design, we generally omit them.

I've left the symbol as it is based on the comments here.

> 2. You'll need to remove the CloudStack JSON entry and SVG from your branch.

I had created the new branch from the wrong base. I've rebased and updated the PR. I hope I haven't messed it up further 🙈

PeterShaggyNoble:
Dang! Overruled again! 😆

@kshitishp, can you update the SVG to centre the path vertically on the canvas? Once that's done, this should be good to go.

kshitishp:
> Dang! Overruled again! 😆
> 
> @kshitishp, can you update the SVG to centre the path vertically on the canvas? Once that's done, this should be good to go.

Done. It was again a matter of second decimals :|

PeterShaggyNoble:
Thanks for your efforts on this one, @kshitishp - everything looks perfect now :+1: (Except, of course, for the inclusion of the trademark symbol at all 😆)

PR #3064, opened 2020-05-13
Add SSRN

davidjb:
ssrn

**Issue:**

### Checklist

### Description

Color determined from the SVG logo at https://www.ssrn.com/index.cfm/en/. Logo was extracted and cleaned up, removing extraneous paths and, oddly, an embedded JPEG.

SSRN ranks at 6,753 on Alexa (https://www.alexa.com/siteinfo/ssrn.com) -- very common in the research sector.


PeterShaggyNoble:
Welcome to Simple Icons, @davidjb and thanks for the contribution(s).

The SVG looks spot on to me :+1: Although, there are a few minor imperfections in the original which we may want to consider fixing - the (non-exhaustive) examples below show the top of the right "leg" of the `R` and the top-left of the `N`. Of course, they could well be intentional 🤷‍♂️

They use a couple of different blues on their site and, without any guidance as to which the primary is, I think the blue from the logo is as good a choice as any.

davidjb:
@PeterShaggyNoble I think the imperfections mostly stem from the "stroke" effect applied to define an "edge" to each of the subsequent letters overlapping (eg S above S above R above N) -- the chunks missing and shapes of the underlying letters follow the shape and serifs as if a stroke were there.

That said, SSRN's had a few variants of its logo and they might have been (auto-)traced at some point creating unusual paths.  For instance, https://static.ssrn.com/cfincludes/img/favicon/safari-pinned-tab.svg?v=2 is notably different than the header logo -- for example, the serifs are shaper in this latter version, where the S and R meet is much flatter and the R is convex rather than concave.

In any case, the icon in this PR is the one directly from the current header, only modified through a run through `svgo`.  Let me know if you think I should tweak anything.

PeterShaggyNoble:
Unless they're glaringly obvious, I'm generally OK with adding SVGs with imperfections stemming from the source but other members of the @simple-icons/maintainers team prefer that we fix them.

Some of them I did figure were a result of the strokes and cut-outs but others, such as the 2 I mentioned above, are not.

ericcornelissen:
I agree with this @PeterShaggyNoble on this one, given that the source is the brand itself I'm okay with leaving the imperfections. So I'll be merging this in as is, so thanks for the work @davidjb :tada: 

If we want to remove the imperfections, the SVG path can be updated easily the future.

PR #3066, opened 2020-05-13
Add Microsoft Academic

davidjb:
microsoftacademic

**Issue:**

### Checklist

### Description

Logo hand traced from the touch icon https://academic.microsoft.com/apple-touch-icon.png. Hex value colour was determined from the same PNG touch icon.


PeterShaggyNoble:
Thanks for the contribution, @davidjb :+1:

However, there is a problem with the source SVG in that it has been converted from a raster file using [Potrace](http://potrace.sourceforge.net/) and, as is often the case, it's not quite perfect when compared to the original raster as you can see in the comparison with [this PNG](https://academic.microsoft.com/apple-touch-icon.png) below. Although it's far from the _worst_ conversion we've seen, we will need to get it a bit closer to the original before we can add it in.

On the choice of colour, they use `#5bbad5` in the `link` tag for the `mask-icon`. It doesn't look as vibrant as the blue they use in the PNG above, though, so I'd be happy to run with your choice.

davidjb:
Cool thanks @PeterShaggyNoble, good catch about Potrace.  I'll do some more hunting and may just tighten up the tracing manually. Stay tuned.

davidjb:
Updated the SVG, replacing it with a hand-traced one created from scatch. Far simpler in terms of path content and aligned much better than theirs:

microsoftacademic2

The black is mine.  There's a tiniest bit of of feathering around the edges; see what you think.

PR #3069, opened 2020-05-13
Add Trove

davidjb:
trove

**Issue:**

### Checklist

### Description

Color determined from the fill colour of the icon in the SVG logo at https://trove.nla.gov.au/. `T` shaped part of the logo was extracted, which is the icon used as the site's Favicon.


PeterShaggyNoble:
Thanks for this one, @davidjb 

First thing we need to do is to determine whether this is within scope for us. As it has no Alexa rank of its own for us to go on, we need to look at some secondary metrics, i.e. their social media. They have 22.1k followers on [Twitter](https://twitter.com/TroveAustralia) which, to me makes it significant enough for inclusion but, as we're still trying to determine our cut-off points for things like this, we'll need some feedback from one of the other @simple-icons/maintainers too.

If it's decided that it _is_ within our scope then there are a couple of imperfections and excess points from the source file that will need to be cleaned up before we merge it in - I see you caught a couple of them already.

davidjb:
Thanks @PeterShaggyNoble -- there's some other icons in this research-related space I could contribute as well, depending on if there's a decision that this fits.

ericcornelissen:
I think this is sufficiently popular and I would be happy to add it.

> If it's decided that it is within our scope then there are a couple of imperfections and excess points from the source file that will need to be cleaned up before we merge it in - I see you caught a couple of them already.

Could you point them out @PeterShaggyNoble?

PeterShaggyNoble:
In the original file, nearly all of the square corners have an unneeded point, the four outer, rounded corners have a few where they meet the straight edges, and the diagonals at the bottom have a few too. And then there's this imperfection on both of the innermost diagonals:

davidjb:
The icon was _quite_ a mess -- I'll check things more carefully in future.  For this one, here's what I changed:

* Removed the extra points from the curves
* Cleaned the jagged edges at the bottom of the central diagonals and at the right edge
* Farthest nodes left edge of the icon didn't line up - the top of segment didn't align with the bottom in terms of X coordinates; they do now
* Cleaned another jagged edge at the top-right curve - cluster of points reduced to a single node
* Nodes positioned at the same conceptual Y coordinates (eg curves and levels of the T shape, etc) have been aligned. They were a few thousandths of the unit out on the left of the T versus the right.

davidjb:
Any more changes needed on this one before merging?

davidjb:
No worries, thanks @ericcornelissen. I’m glad the linting passed, that could have been embarrassing if it failed the size test 😅

PR #3079, opened 2020-05-13
Fix 'wslint' command that isn't linting the simple-icons.json file.

mondeja:
This is only a small change for `.editorconfig` file to avoid the insertion of an useless final newline in `_data/simple-icons.json` file.


ericcornelissen:
Thanks for the suggestion @mondeja, however I disagree with the decision. There are some reasons why final newlines are preferred (read e.g. [this thread](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file)), and whether you agree with them or not there is not really a reason to not include a final newline in the JSON file. The only thing it does is save you a single byte of storage.

If you added the rule to enforce consistency (and clean git diffs), the first rule in the `.editorconfig` file (should) already ensure there **is** always a final newline, in all files (except SVGs).

Still, I'll leave this open for a bit to hear what others thinks (@simple-icons/maintainers?)

runxel:
Yeah, I'd also argue it should be the other way around (i.e. ensuring that final new line). 🤔

mondeja:
I've submitted this pull because sometimes, when I add a icon, I've noticed that a final newline have been added to the file. So, the actual behaviour is not consistent and includes a bit of noise in git diffs.

mondeja:
And what about add an [`editorconfig-checker` step on linting process](https://github.com/mondeja/simple-icons/blob/eclint/package.json#L33-L34)?

ericcornelissen:
> I've submitted this pull because sometimes, when I add a icon, I've noticed that a final newline have been added to the file. So, the actual behaviour is not consistent and includes a bit of noise in git diffs. The master currently doesn't have a final newline, but if you are not agree I will close this. It doesn't matters to add a final newline or not, but the consistence would be great checking the diffs.

That is probably because not every contributor to this project is familiar with Editorconfig and uses it in their editor. Still, for the JSON file I think we can ask contributors to include the final newline. I haven't investigated why this is not being reported by our `wslint` command. If you could investigate that, that would be very helpful.


> And what about add an editorconfig-checker step on linting process? I can make the pull cleaning all the files that doesn't match the `.editorconfig` spec

Actually, if you look [a few lines below the lines you linked to](https://github.com/mondeja/simple-icons/blob/2f4be974d0a1d50b7b625e7f25995660df5ef745/package.json#L37) you see that we do already lint for whitespace issues. However, we are excluding SVGs because of earlier feedback (see #1682) from supporters who found it annoying because some editors (particularly on Unix systems) actually make it quite difficult to not add the final newline (and many EditerConfig plugins do not configure the editor to remove it if it is there).

mondeja:
> If you could investigate that, that would be very helpful.

I've added a commit that fix this. Seems that `editorconfig-checker` accepts regular expressions instead of globs. Also, I've revert the change on `.editorconfig` file and added a final newline to the JSON. That will resolve the inconsistencies.

PeterShaggyNoble:
This seems to have broken something in our linting process:
- [This build](https://github.com/simple-icons/simple-icons/runs/709781449) is failing because there's a trailing newline in `simple-icons.json`.
- While [this one](https://github.com/simple-icons/simple-icons/runs/709800760) is failing because there _isn't_.

ericcornelissen:
Seems to be related to #3112? Note that the two build fails because of different reasons, though I agree that the error message is not very clear (but that is a problem that should be solved at the linter's  end)

It's quite unfortunate that this change is not being reflected in a lot of older Pull Requests, causing some annoying failing builds in newer Pull Requests and even on the `develop` branch itself 😞

PR #3083, opened 2020-05-15
Avoid to run svglint for icons/*.js files.

mondeja:
Only run `svglint` for SVG files inside `icons/` folder, so you can build the library and lint SVG files after. It's just for comfort.


ericcornelissen:
Thanks for the fix @mondeja 😃

PR #3107, opened 2020-05-26
Add linter to check svg path dimensions

davidjb:
**Issue:** https://github.com/simple-icons/simple-icons/pull/3065#issuecomment-632729673

### Checklist

### Description

This adds a custom linter to `svglint` to pull out the path, run it through a bounding box checker and test that either the width or height is 24.  The math checks to 3 decimal places, to avoid rounding errors in floating point comparisons (eg `24` vs `24.000000000001`), which follows how svgo [does things](https://github.com/svg/svgo/blob/master/plugins/cleanupNumericValues.js) and is configured in `.svgo.yml`.

The library (https://github.com/icons8/svg-path-bounding-box) is admittedly faily old and may warrant review but it works well and is simple enough.  Others I tried (eg `svg-path-bounds` choked on icon paths, erroring on certain icons' curves). There are definitely [others on NPM](https://www.npmjs.com/search?q=bounding%20svg&ranking=popularity), including one `svg-path-bbox` under active development which I didn't try, and there is the source library where `svg-path-bounding-box` gets its algorithm from being `canvg`, but the overhead to set up an object [seemed excessive](https://canvg.github.io/canvg/index.html), adding even more dependencies.  At any rate, I manually compared a dozen or so icons in the now-failing lint output to the path's dimensions visible in Inkscape and the bounds were exact, to the 3 decimal places Inkscape shows at least.

Example lint output:

```
------------------------------------ Files ------------------------------------
x icons/500px.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 24.001 x 5.980
x icons/abbrobotstudio.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 23.997 x 20.128
x icons/accusoft.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 23.999 x 15.591
x icons/airbnb.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 22.330 x 23.990
x icons/airbus.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 23.999 x 4.445
x icons/airplayaudio.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 24.001 x 23.635
x icons/amazon.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 23.990 x 21.797
x icons/amazonlumberyard.svg
  x custom-2 Size of <path> must be exactly 24 in one dimension; currently 23.999 x 6.331

[...snip...]

----------------------------------- Summary -----------------------------------
✓ 1044 valid files.
x 291 invalid files.
```

291 icons to fix 😢 but at least linting can catch it 😎


mondeja:
It seems that if we want this check working we need to update all failed icons in some automated way before include it.

PeterShaggyNoble:
Thanks for putting this together, @davidjb, looks like it would be a good addition to our linting process :+1:

I think we'd need to work on the wording of the error message thrown, though, as, the vast majority of time, these issues creep in at the optimisation stage of icon creation process which can lead to confusion if someone has taken care to ensure that the dimensions are _exactly_ right. I know, the first time it happened to me, I was confused as to what had happened until one of the other team members pointed it out to me.

I also think we'd need to up the precision checked to 6 as, while 3 is our _ideal_ maximum precision, there are cases where a higher precision is needed in order to exactly match the source. We currently, by my quick count, have:
- 1 icon with at least one path command that has 17(!) decimal points in it,
- 1 with 10 decimal places,
- 4 with 8, 6 & 5, and,
- 61 with 4

(Although, separate to this, we should review all icons with a precision higher than 5, as well as the outliers with a lower number of decimal places - 23 have a precision of just 1 and 5 have no decimal places in any of their path commands.)

I don't know that automatically scaling the path if it fails the linting is a viable solution here, @mondeja. As a I mentioned above, this issue usually arises at the optimisation stage and there may be other imperfections or inconsistencies that were created at the same time. Or they may creep in when automatically resizing the path.

davidjb:
@PeterShaggyNoble Regarding the wording, I agree -- my first PR was a carefully-sized path, and was surprised to see `svgo` had changed it to `23.998` or thereabouts.  Perhaps better wording + the error could link to a section of the contributor docs to further explain the situation.  Some tightening up of the error reporting when/if an incorrect path is present might be needed as well; unlikely to happen but still.

With the PR, please use whatever makes sense and modify accordingly, it's essentially a PoC to discuss and cherry-pick.

mondeja:
We can extract the maximum number of floating points of paths coordinates to set the accuracy accordinglly, so if an icon has 10 decimal points in their coordinates, the error couldn't be higher than  0.0000000001.
Also, can be established a minimum and maximum of error. Setting 9 decimal places as the maximum and 6 as minimum, the icon size error can be between 0.000001 and 0.000000001.

The real problem here is to update manually all the icons that doesn't fit this lint rule.

.svglintrc.js
```diff
@@ -1,7 +1,10 @@
 const data = require("./_data/simple-icons.json");
 const { htmlFriendlyToTitle } = require("./scripts/utils.js");
+const getBounds = require('svg-path-bounding-box');
```
ericcornelissen, line 3: Just for consistency, since a few changes are required anyways: 

```suggestion
const getBounds = require("svg-path-bounding-box");
```

package.json
```diff
@@ -22,6 +22,7 @@
     "jest": "26.0.1",
     "jsonlint2": "1.7.1",
     "npm-run-all": "4.1.5",
+    "svg-path-bounding-box": "^1.0.4",
```
ericcornelissen, line 25: To avoid a PR by [Renovate](https://docs.renovatebot.com/) that pins this dependency, please update this to:

```suggestion
    "svg-path-bounding-box": "1.0.4",
```

.svglintrc.js
```diff
@@ -46,6 +49,18 @@ module.exports = {
...
             }
           },
+          function(reporter, $, ast) {
+            const iconPath = $.find("path").attr("d");
+            const bounds = getBounds(iconPath);
+            const width = +bounds.width.toFixed(floatPrecision);
+            const height = +bounds.height.toFixed(floatPrecision);
+
+            if (width === 0 && height === 0) {
+              reporter.error(`Path bounds reported as 0 x 0; check the path is valid`);
```
ericcornelissen, line 59: Update the error message so it reads as an English sentence (and I'd suggest using `"` since this string is not actually a template literal)

```suggestion
              reporter.error("Path bounds reported as 0 x 0; check if the path is valid");
```

davidjb: I adjusted this a tiny bit further to make it more readable as an English sentence

davidjb:
@ericcornelissen Changed made, thanks for the suggestions. Next step might be to also eslint or similar config for linting the linter js file 😄 

> Depending on the difficulty of implementing this I like @mondeja suggestion of just taking whatever the precision of the icon is.

This will be hard to determine given that when a path gets loaded into Javascript, you're subject to floating point calculations and thus rounding errors. As an example, loading the path from `icons/x-pack.svg` with `svgpath` (the lib `svg-path-bounding-box` uses), points recorded in the SVG as `1.07` are seen in JS as `1.0700000000000003` -- I'm not sure where this is creeping in; other libs might handle this type of rounding issue better.

A text analysis of the path is an option but that doesn't necessarily equate to what the bounding box's precision would or should be.  A curve or arc between points with one precision will produce path of a higher precision (eg particularly so with angle calculations).  This might or might not matter given the icon; I don't know how 

Maybe a solution is to set a common maximum precision (precision of 5 sees the lint errors jump to 532 😓) or certain icons could include a metadata field for the desired precision if their icons are that precise.

Either way, apologies for the work creation scheme -- hopefully this helps more than it hinders.

ericcornelissen:
> Next step might be to also eslint or similar config for linting the linter js file smile

Given how uncommon it is for the JS of this project to be changed, I don't want to bother with it. I think it's quite nice to keep as few dependencies as necessary, it makes `npm install` quicker :smile: 

> points recorded in the SVG as 1.07 are seen in JS as 1.0700000000000003 -- I'm not sure where this is creeping in;

That is just the effect of floating point numbers, I think. But it's good that you bring it up as it tells us that we should definitely put a limit on the maximum precision we **can** possibly consider.

> This will be hard to determine given that when a path gets loaded into Javascript, [...] A text analysis of the path is an option but that doesn't necessarily equate to what the bounding box's precision would or should be. A curve or arc between points with one precision will produce path of a higher precision (eg particularly so with angle calculations). This might or might not matter given the icon; I don't know how.

Alright, that sounds to me like we shouldn't do it dynamically, at least for now.


What about the following (@davidjb @PeterShaggyNoble @mondeja): keep the precision of this analysis at 3 (maybe 4?) for now and find a solution for the problems with existing icons. I think, currently, it's not such a big deal if an icon exceeds the 24x24 box by less than 0.001.

mondeja:
> it's not such a big deal if an icon exceeds the 24x24 box by less than 0.001

Yes, I agree, we can start from that point, but it's not difficult to achieve the extraction of numbers inside the path with a function like [this](https://github.com/mondeja/svg-path-bbox/blob/e2037834897c66ab6dacfa04cedefd5d596b3cfe/src/util.js#L3) (using regular expressions), and extract the maximum number of floating points with something like [this](https://github.com/mondeja/svg-path-bbox/blob/e2037834897c66ab6dacfa04cedefd5d596b3cfe/src/util.js#L42). So, the problem of excesive floating point numbers would be not a problem if we can control the bounding box path processor. This is one of the problems for which I started some weeks ago in the development of [svg-path-bbox](https://github.com/mondeja/svg-path-bbox), but it's not mature yet.

However, I think that it's better starts from 0.001 and go further when first failing icons would be scaled.

davidjb:
Just had a interesting discovery when editing an SVG in Inkscape -- the UI reports a node at `19.000,17.000` but in the source file it is stored as `18.999984,17.000015`.  Manually typing `19.000` over the original value sees the node move ever so slightly, so Inkscape is using/displaying higher precision but rounding the display.  Anyway, this clearly Inkscape's issue (and that level of precision will get optimised by the default `svgo` config) but I thought I'd mention it as someone might get a surprise to find, as I did, what looks like `24.000` is actually something else entirely.

ericcornelissen:
@PeterShaggyNoble do you agree with [this comment](https://github.com/simple-icons/simple-icons/pull/3107#issuecomment-634524250)? If so I think we should start looking into a solution to fix/ignore the issues with existing icons :smile:

PeterShaggyNoble:
Yup, I think a fixed precision of 3 or 4 should be OK.

ericcornelissen:
What are the opinions with regards to hardcoding the icons that currently generate an error and ignoring any errors they produce. This is the least work, we should just not forget to remove them from the list of hardcoded icons when they're updated :sweat_smile: Perhaps we can do this based on the exact path value (or full SVG) so that an update to one of the icons is certainly subject to this linting step?

If someone is willing to manually resize over 290 icons they are of course welcome to :upside_down_face:

PeterShaggyNoble:
> If someone is willing to manually resize over 290 icons they are of course welcome to

He says, knowing I'm just about crazy enough to take something like that on! 😆 I _have_ been thinking of a _full_ icon review lately! 🙈

ericcornelissen:
> He says, knowing I'm just about crazy enough to take something like that on! 😆 I have been thinking of a full icon review lately! 🙈

Hahaha 🤣 It would be awesome if you could do it! (I would say this list takes priority over a "generic"/"general" icon review.) Still, unless you plan to do it in less than 1 week I don't think it hurts to implement a temporary solution to save ourselves some time reviewing new and existing PRs 😉

PeterShaggyNoble:
I'll create a new "tracking" issue for them in a bit so we can get some help with them, work through them as we can.

ericcornelissen:
Okay 👍 @davidjb, are you interested in implementing some filtering mechanism for the existing icons, or would you prefer us to implement that?

PeterShaggyNoble:
> Perhaps we can do this based on the exact path value (or full SVG) so that an update to one of the icons is certainly subject to this linting step?

I think, for better future proofing and so that we don't _have_ to remember to remove an icon from the list as we work through updating them, that this is the best option - if the path/SVG exactly matches one in the list then ignore it. If you let me know what format you need that list in, I can very quickly generate it from the list of names I've already done up.

davidjb:
@ericcornelissen For the filtering, it sounds like you've got ideas in mind as to how you'd like to manage the errant icons so it seems simpler if you add a method which works best for you as maintainers.  I can't think of anything better than @PeterShaggyNoble's comparison strategy, sort of perhaps using SHA hashes to simplify storage rather than the full string path. The simplest option is a JS object in the .svglint file with that data, annotated in some way (comments etc) of identifying which hash/path corresponds to which icon so you can know when to remove them.

PeterShaggyNoble:
Would this format do the trick, @davidjb?

```
{
	"500px": "M7.439 9.01A2.994 2.994 0 0 0 4.449 12a2.993 2.993 0 0 0 2.99 2.99 2.994 2.994 0 0 0 2.99-2.99 2.993 2.993 0 0 0-2.99-2.99m0 5.343c-1.297 0-2.354-1.056-2.354-2.353s1.057-2.353 2.354-2.353S9.792 10.703 9.792 12s-1.056 2.353-2.353 2.353m6.472-5.343a2.994 2.994 0 0 0-2.99 2.99 2.993 2.993 0 0 0 2.99 2.99 2.994 2.994 0 0 0 2.99-2.99 2.994 2.994 0 0 0-2.99-2.99m0 5.343c-1.298 0-2.353-1.056-2.353-2.353s1.055-2.353 2.353-2.353c1.297 0 2.353 1.056 2.353 2.353s-1.056 2.353-2.353 2.353m-11.612-3.55a2.1 2.1 0 0 0-1.596.423V9.641H3.39c.093 0 .16-.017.16-.292 0-.269-.108-.28-.18-.28H.396c-.174 0-.265.14-.265.294v2.602c0 .136.087.183.247.214.141.028.223.012.285-.057l.006-.01c.283-.408.9-.804 1.486-.732.699.086 1.262.644 1.34 1.327a1.512 1.512 0 0 1-1.501 1.685c-.635 0-1.19-.408-1.421-1.001-.035-.088-.092-.152-.343-.062-.229.083-.243.181-.212.268a2.11 2.11 0 0 0 1.976 1.386 2.102 2.102 0 0 0 .305-4.18m16.646-1.764c-.805.062-1.434.769-1.434 1.61v2.661c0 .154.117.186.293.186s.293-.031.293-.186v-2.668c0-.524.382-.974.868-1.024a.972.972 0 0 1 .758.247.984.984 0 0 1 .322.729c0 .08-.039.34-.217.581-.135.182-.391.399-.844.399h-.009c-.115 0-.215.005-.234.28-.013.186-.012.269.148.29.286.04.576-.016.865-.166.492-.256.822-.741.861-1.267a1.562 1.562 0 0 0-.452-1.222 1.56 1.56 0 0 0-1.218-.45m3.919 1.559l1.085-1.085c.04-.039.132-.132-.055-.324-.08-.083-.153-.125-.217-.125h-.001a.163.163 0 0 0-.121.058l-1.088 1.088-1.086-1.093c-.088-.088-.19-.067-.322.065-.135.136-.157.24-.069.328l1.086 1.092-1.064 1.064-.007.007c-.026.025-.065.063-.065.125-.001.063.042.139.126.223.07.071.138.107.2.107.069 0 .114-.045.139-.07l1.068-1.067 1.091 1.092a.162.162 0 0 0 .114.045h.002c.069 0 .142-.04.217-.118.122-.129.143-.236.061-.319l-1.094-1.093z",
	"abbrobotstudio": "M24 12.46a9.6 9.6 0 01-19.2 0h1.07a8.53 8.53 0 108.53-8.53V2.86a9.6 9.6 0 019.6 9.6zm-9.6-3.2a3.2 3.2 0 103.2 3.2 3.2 3.2 0 00-3.2-3.2zm-2 0l-.6-6.67-2.46 1.92-1.46-1.44a4.67 4.67 0 00-5.62-.37L.24 4a.54.54 0 00-.15.74.54.54 0 00.74.15l2-1.31a3.64 3.64 0 014.29.22l1.37 1.38L6.2 7z",
	"accusoft": "M18.078,16.345c-0.209-0.261-8.709-11.13-9.005-11.496 c-0.279-0.366-0.209-0.47-0.157-0.523c0.105-0.122,0.261-0.105,0.871-0.105c0.366,0,3.989-0.017,4.372-0.017 c0.784,0,0.906,0.07,0.993,0.087c0.087,0.035,0.296,0.209,0.453,0.383c0.087,0.105,7.699,9.214,7.768,9.301 c0.087,0.105,0.209,0.279,0.314,0.435c0.087,0.157,0.105,0.366-0.035,0.453c-0.087,0.052-4.302,1.794-4.424,1.829 c-0.122,0.052-0.348,0.139-0.523,0.105C18.618,16.798,18.357,16.676,18.078,16.345 M22.38,16.136l0.314,0.052 c0,0,1.01,0.192,1.115,0.226C23.913,16.432,24,16.519,24,16.554c0,0.087-0.052,0.122-0.139,0.174 c-0.07,0.052-4.633,2.856-4.72,2.909c-0.087,0.052-0.192,0.105-0.435,0.139c-0.453,0.087-1.306-0.157-1.585-0.209 c-0.261-0.052-11.461-2.543-11.548-2.578c-0.105-0.035-0.174-0.035-0.174-0.139c-0.017-0.157,0.226-0.209,0.418-0.279 c0.192-0.07,5.452-1.968,5.643-2.055c0.192-0.087,0.418-0.105,0.54-0.105s0.853,0.105,1.184,0.157s1.271,0.192,1.271,0.192 l2.142,2.769c0.366,0.435,0.61,0.61,0.923,0.627c0.157,0.017,0.331-0.035,0.453-0.087C18.061,18.035,22.38,16.136,22.38,16.136 M10.153,9.343c0,0,1.846,2.369,1.864,2.386c0.017,0.035,0.035,0.07,0.07,0.087v0.035c-0.07,0.052-3.362,3.1-3.379,3.118 l-3.466,1.271c0,0-0.105,0.035-0.157,0.07c-0.052,0.035-0.122,0.105-0.105,0.261c0,0.052,0.017,0.853,0.035,1.045 c-0.017,0.017,0,0-0.017,0.017c0,0-4.267,1.359-4.302,1.359c-0.209,0.07-0.61,0.209-0.662,0.174 c-0.087-0.07,0.017-0.174,0.07-0.244c0.052-0.07,8.796-8.674,9.127-9.022C9.7,9.395,10.153,9.343,10.153,9.343",
	"airbnb": "M11.998 18.267c-1.352-1.696-2.147-3.182-2.412-4.455-.263-1.026-.159-1.847.291-2.464.477-.71 1.187-1.055 2.12-1.055s1.642.345 2.119 1.062c.446.61.558 1.432.286 2.464-.291 1.298-1.085 2.784-2.411 4.456zm9.597 1.14c-.185 1.245-1.034 2.278-2.2 2.782-2.251.98-4.48-.583-6.388-2.703 3.155-3.95 3.738-7.025 2.384-9.014-.795-1.14-1.933-1.695-3.393-1.695-2.943 0-4.561 2.49-3.925 5.38.37 1.564 1.351 3.342 2.915 5.33-.98 1.084-1.909 1.855-2.73 2.332-.636.344-1.245.557-1.828.608-2.677.399-4.776-2.198-3.823-4.877.132-.345.395-.98.845-1.961l.025-.053C4.94 12.36 6.717 8.75 8.759 4.746l.053-.132.58-1.115c.45-.822.635-1.19 1.351-1.643.345-.209.769-.314 1.245-.314.954 0 1.697.557 2.015 1.006.158.239.345.557.582.953l.558 1.088.08.159c2.04 4.002 3.819 7.605 5.276 10.789l.026.025.533 1.22.318.764c.243.612.294 1.221.213 1.857zm1.219-2.389c-.186-.583-.504-1.271-.9-2.093v-.03c-1.887-4.005-3.64-7.605-5.304-10.84l-.111-.162C15.313 1.461 14.464 0 11.998 0 9.56 0 8.524 1.694 7.465 3.897l-.081.16c-1.668 3.234-3.42 6.839-5.301 10.842v.053l-.558 1.219c-.21.504-.317.768-.345.847-1.35 3.712 1.432 6.972 4.8 6.972.027 0 .132 0 .264-.027h.372c1.75-.213 3.553-1.325 5.382-3.316 1.828 1.988 3.633 3.103 5.38 3.316h.372c.132.027.238.027.264.027 3.368.003 6.15-3.26 4.8-6.972z",
	"airbus": "M11.062 11.296c0 .74-.389 1.161-.993 1.304-.007 0 .967 1.531.967 1.531h-1.18l-1.427-2.277h1.006c.435 0 .597-.24.597-.532 0-.285-.156-.532-.59-.532H8.266v3.341H7.228V9.869h2.206c1.096 0 1.628.616 1.628 1.427m-5.573 2.835h1.038V9.869H5.489v4.262zM2.174 9.869L0 14.131h1.168l.352-.714h1.75l-.435-.895h-.873l.646-1.311h.013l1.453 2.92h1.194L3.095 9.869h-.921zm12.678 2.05c.409.143.688.519.688 1.019 0 .72-.577 1.194-1.46 1.194h-2.524V9.869h2.427c.863 0 1.376.461 1.376 1.148-.001.428-.176.72-.507.902m-2.258-.396h1.382a.368.368 0 0 0 .376-.376.367.367 0 0 0-.37-.376h-1.388v.752zm1.414 1.713a.435.435 0 0 0 .448-.441c0-.247-.195-.428-.448-.428h-1.414v.869h1.414m4.808-.986c0 .647-.298 1.006-.889 1.006-.584 0-.882-.359-.882-1.006V9.869h-1.064v2.303c0 1.317.694 2.05 1.946 2.05s1.953-.733 1.953-2.05V9.869h-1.064v2.381zm3.834-.688c-.985-.24-1.2-.263-1.2-.545 0-.218.246-.324.662-.324.551 0 1.139.138 1.473.344l.331-.869c-.428-.227-1.058-.389-1.791-.389-1.097 0-1.713.545-1.713 1.278 0 .789.46 1.109 1.518 1.337.824.182.999.295.999.526 0 .251-.227.363-.675.363a3.565 3.565 0 0 1-1.706-.415l-.318.908c.513.273 1.278.448 2.05.448 1.077 0 1.719-.5 1.719-1.337.001-.673-.433-1.104-1.349-1.325"
}
```

ericcornelissen:
> [U]sing SHA hashes to simplify storage rather than the full string path

That's certainly something worth considering :thinking: (using any hashing mechanism for that matter: SHA, MD5 or simpler)


> The simplest option is a JS object in the .svglint file with that data, annotated in some way (comments etc) of identifying which hash/path corresponds to which icon so you can know when to remove them.

I would propose putting them in a separate JSON file and loading it, but that's a small detail. I think commenting them would be unnecessary, when a particular SVG is updated no SVG should match the hash any more, and once #3169 is finished we can simply remove the entire JSON file (and the loading logic).


> Would this format do the trick [...]?

It definitely could. It's been a while since I worked with the SVG linter but if it provides easy access to the `<path>`'s `d` value it is. Otherwise I would prefer just having the full SVG in the JSON, that makes the comparison a lot easier.

If we decide to go with SHA (or MD5, since I'm not worried about collisions in this dataset) hashes the values should be hashes though. Don't see why we would compute the hash every time the CI is ran.

PeterShaggyNoble:
> It definitely could. It's been a while since I worked with the SVG linter but if it provides easy access to the `<path>`'s `d` value it is.

Are the additions in this PR already accessing the path's data?

> Otherwise I would prefer just having the full SVG in the JSON, that makes the comparison a lot easier.

Yup, it would definitely be easier but my (admittedly minor) concern would be that the #3169 review could take a good while and, in the meantime, we could in the future, for some reason we can't foresee now, decide to change the requirements for or formatting of the markup of our files - using the path alone gives us a little bit of extra future proofing.

> the values should be hashes though. Don't see why we would compute the hash every time the CI is ran.

Agreed. Should have said in that in my last comment: let me know which hashing function is easiest to work with that we want to use and I'll generate the complete file.

ericcornelissen:
> Are the additions in this PR already accessing the path's data?

Jup :sweat_smile: Nevermind my earlier comment


> [I]n the meantime, we could in the future, for some reason we can't foresee now, decide to change the requirements for or formatting of the markup of our files - using the path alone gives us a little bit of extra future proofing.

That is also a good point


> Agreed. Should have said in that in my last comment: let me know which hashing function is easiest to work with that we want to use and I'll generate the complete file.

I would prefer something that is super quick to compute, as we are going to be doing it at least 390 times every time `npm run svglint` is executed...

PeterShaggyNoble:
Hmm ... which is the lesser of 2 evils here? The larger overhead involved in loading in a file with unhashed data or having to unhash all 290 paths for each build?

davidjb:
Svglint executes `.svglintrc.js` once for all images (wasn't sure, just tested it), so if you import / construct a lookup object at the top of the file, it'll only get processed once for all the hundreds of icons. 

The fastest lookup will (probably, depending on the interpreter) be a string property check on an object (eg like @PeterShaggyNoble's structure, just reversed so the path is the key which we're looking up; truncated for readability):

```
const ignoredIcons = {
	"M7.439 9.01A2.994. [...]": "500px",
        "M24 12.46a9.6 [...]": "abbrobotstudio",
        "M18.078,16.345c-0.209 [...]: "accusoft",
        [...]
}
```

and then checking for the presence of the path as a property via `igoredIcons.hasOwnProperty(pathString)` before running the tests (some anecdotal reports say `in` is a faster lookup than `hasOwnProperty` but for me on Node 14.3 it's about 5x faster).  The data could be loaded in from a JSON file or otherwise; there'll be a bit of data in that file given the number of icons but it'll only be loaded into svglint's memory once.

Hashing would make the data file smaller but the computation of a hash on all icons (not just the 290 ignored...) will be a magnitude slower than an interpreter-optimised string comparison / property lookup.

ericcornelissen:
Good thinking @davidjb, I had realized some (but not all) of those points as well since our discussion yesterday. I agree that the string-property check on an object is probably the most efficient thing we can do. Given that you already spend some time trying things out, could you implemented a proof of concept (perhaps with just a handful of SVGs ignored)?

davidjb:
I've just pushed a new PR.  With this updated PR, I've implemented the complete implementation of ignoring the files that are non-conforming at this moment in time.  A precision of 3 causes 290 errors and a precision of 4 raises that to 469 so I've left the precision at 3 given the workload of adjusting icons; the precision can be raised over time, if desired, to clean up the others.

For future ignore files, this is the code I used in `.svglinrc.js` to dump out the failing paths, added when the reporter would normally error on incorrect path sizes:

```js
const fs = require('fs')
const title = $.find("title").text().split(' ')[0];
const iconName = htmlFriendlyToTitle(title);
fs.writeFileSync(
  '.svglint-ignored.json',
  '    ' + JSON.stringify({[iconPath]: iconName}).replace('{', '').replace('}', '')+',\n',
  {flag: 'as'},
  () => {}
)
```

and then I manually added the `{}` and stripped off the trailing `,` (svglint's custom runners have no knowledge of their run order, hence it was necessary just to use sync output and tweak the output into valid JSON).

I also tweaked the reporter name so it's nicer than `custom-2`.  The same might want to be applied to the other custom check for titles as well.

davidjb:
Pushed the change as suggested by @ericcornelissen for the existing reporter, and also manually linted the .js file to add semicolons and use double quotes (muscle memory from other projects I'm working on).  That should be it - let me know if there's any other changes required.

ericcornelissen:
> Pushed the change as suggested by ericcornelissen for the existing reporter, and also manually linted the .js file to add semicolons and use double quotes (muscle memory from other projects I'm working on). That should be it - let me know if there's any other changes required.

Seems like something went wrong @davidjb? At least I'm not seeing a name for the existing reporter :thinking:

davidjb:
Once more with feeling!  That should do it @ericcornelissen; looks like I'd re-pushed the same content without the updated commit.

PeterShaggyNoble:
This all looks good to me - great work, @davidjb 👏

Thinking on it some more over the weekend, I think we should leave the precision at `3`. Inkscape, and I'm sure other programmes too, round to 3 decimal places in its toolbars and, even if there is some way of changing that (I haven't found it yet), I don't think it's something we should expect people to be doing to work on icons for us.

The only addition I can see that hasn't been made yet is this suggestion from @ericcornelissen:
> I think a better compromise, as @davidjb suggested, is to put it in the [Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/47530c59a552fca8bb3d7c6f40b61b7bad2e256e/CONTRIBUTING.md) (possibly linking that in the error message)

ericcornelissen:
> Thinking on it some more over the weekend, I think we should leave the precision at 3. Inkscape, and I'm sure other programmes too, round to 3 decimal places in its toolbars and, even if there is some way of changing that (I haven't found it yet), I don't think it's something we should expect people to be doing to work on icons for us.

I think I agree, but we can always review that at a later point in time :+1: 


> The only addition I can see that hasn't been made yet is this suggestion [...]

Good call, forgot about that :sweat_smile: Do you have any ideas regarding this @PeterShaggyNoble?

How about adding a sub-bullet to th [Check the icon section](https://github.com/simple-icons/simple-icons/blob/47530c59a552fca8bb3d7c6f40b61b7bad2e256e/CONTRIBUTING.md#5-check-the-icon)'s _"Be scaled to fit the viewbox, while preserving the icon's original proportions."_ bullet stating something along the lines of "We check this up to 3 decimal points of precisions; as a result the optimization may affect this."

PeterShaggyNoble:
I think that line is best placed in the ["Optimize the Icon"](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#3-optimize-the-icon) section as that's the stage where this usually crops up. Maybe something along the lines of:

> After optimising the icon, double check it against your original version to ensure no visual imperfections have crept in and that the dimensions of the path have not been changed so that the icon no longer fits exactly within the canvas, up to 3 decimal points of precision.

ericcornelissen:
That's a good idea @PeterShaggyNoble, I took the liberty of adding it to this Pull Request (as well as rewriting it just slightly) :ok_hand:

CONTRIBUTING.md
```diff
@@ -91,6 +91,8 @@ All icons in Simple Icons have been optimized with the [SVGO tool](https://githu
   * Leave the remaining settings untouched (or reset them with the button at the bottom of the settings).
   * Click the download button.
 
+After optimizing the icon, double-check it against your original version to ensure no visual imperfections have crept in. Also make sure that the dimensions of the path have not been changed so that the icon no longer fits exactly within the canvas, up to 3 decimal points of precision.
```
PeterShaggyNoble, line 94: A small change to the end to (maybe) make it a bit clearer.

```suggestion
After optimizing the icon, double check it against your original version to ensure no visual imperfections have crept in. Also make sure that the dimensions of the path have not been changed so that the icon no longer fits exactly within the canvas. We currently check the dimensions up to a precision of 3 decimal points.
```

ericcornelissen: Regarding "double check" or "double-check", the internet suggests to me to use the latter. Then again, you are (I believe) a native speaker 🤔 

Other than that I agree 👍

PeterShaggyNoble: Oop, dunno how that happened. You're correct, it should be hyphenated. I've corrected it above.

ericcornelissen: Alright, done @PeterShaggyNoble 😃

ericcornelissen:
And with that I merge this Pull Request. Thanks so very much for your work on this @davidjb, this is a truly amazing contribution! 🎊 

We will be forcing Pull Requests to be up-to-date with the base branch (i.e. `develop`) for a while to make sure existing PRs comply with this linting step.

PeterShaggyNoble:
Great work, @davidjb - many thanks for this addition.

Tested it on #3126 & #3192 and it looks like it's working as intended :+1:
- https://travis-ci.com/github/simple-icons/simple-icons/jobs/347043563
- https://travis-ci.com/github/simple-icons/simple-icons/jobs/347045288

PeterShaggyNoble:
Coming back around to this one: would we be better off checking on [line 69](https://github.com/simple-icons/simple-icons/blob/develop/.svglintrc.js#L69) whether the width **or** the height is `0`?

I also wonder if it would be possible to build on what was done here to check whether the path is centred on the canvas? Maybe within a precision of `.001` to allow for rounding errors?

davidjb:
If I recall correctly, the 0x0 dimensions were the library reporting an invalid path - a quick glance at their source (https://github.com/icons8/svg-path-bounding-box/blob/master/lib/Path/BoundingBoxView.js#L10-L11) indicates that the dimensions start as NaN and get logical OR'd with 0.  So both dimensions would be 0 in the case of an invalid path.

As for checking if the path was centred, sounds like an idea.  Calculate the centre of the path (e.g half way between min/max points in X & Y), round it off and compare that to the centre of the canvas (12x12).  Sounded pretty quick since that data's already in the library...ahem.. https://github.com/simple-icons/simple-icons/pull/3250

PR #3133, opened 2020-05-29
Add N26

vigonotion:
N26

**Issue:** -

### Checklist

### Description
Adds the N26 logo (app-based bank from Germany)

Alexa Rank: ~[11,834](https://www.alexa.com/siteinfo/n26.com)

Logo has been extracted from the [website](https://n26.com/) header (top left).

Color `#5FAD97` has been extracted from their app logo: https://n26.com/en-de/logo-assets


PeterShaggyNoble:
Nice work, @vigonotion - SVG looks perfect to me :+1:

Not entirely sold on the choice of colour, but I can see why you went with it. My first instinct would have been to go with the `#266678` shade as it's the most dominant on their site. If others agree that the lighter shade is the correct choice then I'd suggest `#48ac98` from their stylesheet, as eyedropping from an image isn't always 100% reliable.

vigonotion:
`#266678` is only dominant on their website, which is not really used that much as N26 is mainly an app based bank.  I can confirm the color `#48ac98`, the Android app icon actually also uses that color.

Also look at the first video in the app description, especially the splash screen in the beginning, to get an idea of their branding:
https://play.google.com/store/apps/details?id=de.number26.android

PeterShaggyNoble:
OK, I'm happy to run with that colour, so. Thanks for your work on this one, @vigonotion.

PR #3156, opened 2020-06-02
Add Metro de Ciudad de México

mondeja:
metrodeciudaddemexico

**Issue:** #2822

### Checklist

### Description
- Logo extracted from [2019 Lance Wyman design](https://es.wikipedia.org/wiki/Metro_de_la_Ciudad_de_M%C3%A9xico#/media/Archivo:Metro_de_la_Ciudad_de_M%C3%A9xico_(logo)_version_2019.svg).
- Hex value from [original 1968 logo](https://es.wikipedia.org/wiki/Archivo:Metro_de_la_Ciudad_de_M%C3%A9xico_logo.svg). The new 2019 design is filled by `#ff5000` colour, but I'm not sure which one to use. Any suggestions?


PeterShaggyNoble:
Hmm ... the Lance Wyman design, would _appear_ to be an officially commissioned one and the SVG of the other one is an original work based on a JPG, which could account for the fact that the 2 versions don't line up. So, despite the fact that they don't use it on their website (that I can see, at least), I'm OK with the source you've chosen.

Your SVG looks spot on when compared to the source and, of the two colours, the one you've chosen is closest to what they use on their site but the shade they use, as seen on [this page](https://www.metro.cdmx.gob.mx/organismo/acerca-de) is actually `#f77e1c` so I think that's what we should go with.

Other than that, you're just missing the "la" from the name - Metro de _la_ Ciudad de México

PeterShaggyNoble:
Thanks for your work on this one, @mondeja :+1:

PR #3161, opened 2020-06-04
Update Mixcloud Logo

NovaGL:
simpleicons(2)

### Checklist

Logo embedded in website, branding site not up yet.
PS: Sorry for all the commits, I'm trying to keep my fork upto-date with your base.


PeterShaggyNoble:
Thanks for the PR, @NovaGL, and for bringing this to our attention.

I think we should hold off on merging this in, though, until they've added the new brand guidelines and clarified what we should use as the icon, whether it be "Mix" as per their app icon or just the "M" as per their favicon. The new brand guidelines would also confirm the choice of colour.

NovaGL:
No problem. Just noticed as it changed to this logo in the Sonos app

PeterShaggyNoble:
Anyone any objections to giving this another week to see if Mixcloud release their updated brand guidelines and, if not, making a judgement call to run with the choices of icon & colour, @NovaGL has already made? We can always come back to update again based on their brand guidelines when they _do_ release them?

ericcornelissen:
> Anyone any objections to giving this another week to see if Mixcloud release their updated brand guidelines and, if not, making a judgement call to run with the choices of icon & colour NovaGL has already made? We can always come back to update again based on their brand guidelines when they _do_ release them.

I think it is safe to do that as, indeed, we can always update at some later point.

PeterShaggyNoble:
OK, still no official guidelines from them so let's run with this as-is.

Using the first SVG (the one with the 170*20 `viewBox`) found in the HTML of the homepage's header, I'm seeing some minor differences (see screenshot of the middle of the "M" below for example, with your version in red). If you used the same version then it's probably just a precision issue. IF you used a different version, though, let me know and I'll compare to that instead.

I'd also suggest removing the `#brand` anchor from the source URL, too.

PeterShaggyNoble:
Don't forget about this one, @NovaGL 😉 Still no sign of their brand guidelines, all these weeks later.

NovaGL:
Hi Peter,

I have changed the brand colors as it seems that the logo is mainly used in this color. It can be changed back if required.
Happy to do whatever others want.

image

fbernhart (changes requested):
Blue still seems to be their main brand color, so I would suggest changing to to `#5000FF` (which can be found in their official brandguidelines, a PDF file in their zipped brand kit at https://www.mixcloud.com/about/#brand). `#5000FF` is as well the first color mentioned in their brand kit. The "Color Misuse" section in their brand guidelines as well mentions:

> Do not use large amount of 'High Pink' in brand contexts. We should still remain a predominately blue brand.

As well a precision of 4 doesn't seem to be necessary. I've optimized it with a precision of 3 and it still looks perfect. Feel free to use this path with a precision of 3 instead:

```
<path d="M2.462 8.596l1.372 6.49h.319l1.372-6.49h2.462v6.808H6.742v-5.68l.232-.81h-.402l-1.43 6.49H2.854l-1.44-6.49h-.391l.222.81v5.68H0V8.596zM24 8.63v1.429L21.257 12 24 13.941v1.43l-3.235-2.329h-.348l-3.226 2.329v-1.43l2.734-1.94-2.733-1.942V8.63l3.225 2.338h.348zm-7.869 2.75v1.24H9.304v-1.24z"/>
```

mondeja:
I agree with `#5000FF` color and the path provided by @fbernhart, so I think that we can go ahead. I'm merging this coauthoring @NovaGL and @fbernhart. Thank you all for the reviews!

PeterShaggyNoble:
@NovaGL, was this checked against the brand kit that they finally released at https://www.mixcloud.com/about/?

PR #3180, opened 2020-06-05
Normalise accented characters in website search

PeterShaggyNoble:
Went looking for ŠKODA on our site earlier but using "skoda" didn't produce any results as we're not currently normalising accented characters in the search term or the brand name. This PR fixes that by adding the [replacements](https://github.com/simple-icons/simple-icons/blob/develop/scripts/utils.js#L16-L25) used in `scripts/utils.js` to  the [`normalizeSearchTerm`](https://github.com/simple-icons/simple-icons/blob/develop/index.html#L196-L198) function in `index.html`. I don't have a copy of the site running locally so, despite being a straightforward change, I'd suggest someone test this change before merging it.

We also need to figure out how to handle symbols. For example: searching for "dot" should return .NET in the results and searching for "plus" should return C++. One way to do that could be to search by file name as well as brand name.


ericcornelissen:
Haven't tested it yet, just some quick thoughts: first, nice find!

Second, it's a bit unfortunate that we need to do this for all >1000 icons every time the website is loaded :confused:  Ideally would do this when the website is build, but until #898 I don't think that is very easy to do :thinking: On the other hand, [we do already have Liquid logic that does what you implemented](https://github.com/simple-icons/simple-icons/blob/0a301e1ae95ab221dcb2331e265d92f3356d4198/index.html#L35), but off the top of my head I don't know how we could leverage that for this problem...

---

> We also need to figure out how to handle symbols. For example: searching for "dot" should return .NET in the results and searching for "plus" should return C++. One way to do that could be to search by file name as well as brand name.

Should it? I'm not against it but it doesn't seem necessary to me.

PeterShaggyNoble:
I did, initially, look into trying to do this within Liquid but couldn't figure it out and, with no way of testing it, I went the JS route instead.

I wonder, though, if replacing [lines 149-153](https://github.com/simple-icons/simple-icons/blob/develop/index.html#L149-L153) with something along the lines of the following would do the trick:

```
{% assign allIconNames = "" %}
{% assign normalisedIconNames = "" %}
{% for icon in iconsArray %}
    {% assign iconArray = icon | split: "," %}
    {% capture allIconNames %}{{ allIconNames }}"{{ iconArray[5] }}",{% endcapture %}
    {% assign normalised = iconArray[5] | downcase %}
    {% assign normalised = normalised | replace: "à", "a" | replace: "á", "a" | replace: "â", "a" | replace: "ã", "a" | replace: "ä", "a" %}
    {% assign normalised = normalised | replace: "ç", "c" | replace: "č", "c" | replace: "ć", "c" %}
    {% assign normalised = normalised | replace: "è", "e" | replace: "é", "e" | replace: "ê", "e" | replace: "ë", "e" %}
    {% assign normalised = normalised | replace: "ì", "i" | replace: "í", "i" | replace: "î", "i" | replace: "ï", "i" %}
    {% assign normalised = normalised | replace: "ñ", "n" | replace: "ň", "n" | replace: "ń", "n" %}
    {% assign normalised = normalised | replace: "ò", "o" | replace: "ó", "o" | replace: "ô", "o" | replace: "õ", "o" | replace: "ö", "o" %}
    {% assign normalised = normalised | replace: "š", "s" | replace: "ś", "s" %}
    {% assign normalised = normalised | replace: "ù", "u" | replace: "ú", "u" | replace: "û", "u" | replace: "ü", "u" %}
    {% assign normalised = normalised | replace: "ý", "y" | replace: "ÿ", "y" %}
    {% assign normalised = normalised | replace: "ž", "z" | replace: "ź", "z" %}
    {% capture normalisedIconNames %}{{ normalisedIconNames }}"{{ normalised }}",{% endcapture %}
{% endfor %}
```

And then, the assignment to the JS `icons` variable on (the current) [line 199](https://github.com/simple-icons/simple-icons/blob/develop/index.html#L199) would just become:

```
      var icons = [{{ normalisedIconNames }}];
```

> Should it? I'm not against it but it doesn't seem necessary to me.

If it's just me, then no, maybe not! :laughing: I don't use our JSON file in my projects so I do, on rare occasion, find myself searching our site for them by their file name.

PeterShaggyNoble:
Gave the Liquid version a try on my fork and something's gone wrong that I can't quite figure out: https://petershaggynoble.github.io/simple-icons/?q=%C5%A0KODA

PeterShaggyNoble:
Figured it out: it was me overlooking `greyscaleIconsArray` that was causing the problem, although I'm not sure what that does. The changes required to achieve this in Liquid are now as follows:

1. Replace [lines 149-147](https://github.com/simple-icons/simple-icons/blob/develop/index.html#L149-L157) with the below,
2. Replace the assignment of the JS `icons` varioable as above, and,
3. Optionally, move `normalizeSearchTerm` to `site_script.js`.

The big problem here is increasing the number of places times the same `replace`ments are being run and, therefore, the number of places new `replace`ments need to be added:
1. Our linting script
2. The `assignment` of `filename` beginning at line 31 of `index.html`
3. The two new ones below
4. The `normalizeSearchTerm` function

Of course, if Liquid has support for functions then that could help eliminate 2 of them.

----
```
{% assign allIconNames = "" %}
{% assign normalisedIconNames = "" %}
{% for icon in iconsArray %}
    {% assign iconArray = icon | split: "," %}
    {% capture allIconNames %}{{ allIconNames }}"{{ iconArray[5] }}",{% endcapture %}
    {% assign normalised = iconArray[5] | downcase %}
    {% assign normalised = normalised | replace: "à", "a" | replace: "á", "a" | replace: "â", "a" | replace: "ã", "a" | replace: "ä", "a" %}
    {% assign normalised = normalised | replace: "ç", "c" | replace: "č", "c" | replace: "ć", "c" %}
    {% assign normalised = normalised | replace: "è", "e" | replace: "é", "e" | replace: "ê", "e" | replace: "ë", "e" %}
    {% assign normalised = normalised | replace: "ì", "i" | replace: "í", "i" | replace: "î", "i" | replace: "ï", "i" %}
    {% assign normalised = normalised | replace: "ñ", "n" | replace: "ň", "n" | replace: "ń", "n" %}
    {% assign normalised = normalised | replace: "ò", "o" | replace: "ó", "o" | replace: "ô", "o" | replace: "õ", "o" | replace: "ö", "o" %}
    {% assign normalised = normalised | replace: "š", "s" | replace: "ś", "s" %}
    {% assign normalised = normalised | replace: "ù", "u" | replace: "ú", "u" | replace: "û", "u" | replace: "ü", "u" %}
    {% assign normalised = normalised | replace: "ý", "y" | replace: "ÿ", "y" %}
    {% assign normalised = normalised | replace: "ž", "z" | replace: "ź", "z" %}
    {% capture normalisedIconNames %}{{ normalisedIconNames }}"{{ normalised }}",{% endcapture %}
{% endfor %}
{% for icon in greyscaleIconsArray %}
    {% assign iconArray = icon | split: "," %}
    {% capture allIconNames %}{{ allIconNames }}"{{ iconArray[5] }}"{% unless forloop.last %},{% endunless %}{% endcapture %}
    {% assign normalised = iconArray[5] | downcase %}
    {% assign normalised = normalised | replace: "à", "a" | replace: "á", "a" | replace: "â", "a" | replace: "ã", "a" | replace: "ä", "a" %}
    {% assign normalised = normalised | replace: "ç", "c" | replace: "č", "c" | replace: "ć", "c" %}
    {% assign normalised = normalised | replace: "è", "e" | replace: "é", "e" | replace: "ê", "e" | replace: "ë", "e" %}
    {% assign normalised = normalised | replace: "ì", "i" | replace: "í", "i" | replace: "î", "i" | replace: "ï", "i" %}
    {% assign normalised = normalised | replace: "ñ", "n" | replace: "ň", "n" | replace: "ń", "n" %}
    {% assign normalised = normalised | replace: "ò", "o" | replace: "ó", "o" | replace: "ô", "o" | replace: "õ", "o" | replace: "ö", "o" %}
    {% assign normalised = normalised | replace: "š", "s" | replace: "ś", "s" %}
    {% assign normalised = normalised | replace: "ù", "u" | replace: "ú", "u" | replace: "û", "u" | replace: "ü", "u" %}
    {% assign normalised = normalised | replace: "ý", "y" | replace: "ÿ", "y" %}
    {% assign normalised = normalised | replace: "ž", "z" | replace: "ź", "z" %}
    {% capture normalisedIconNames %}{{ normalisedIconNames }}"{{ normalised }}"{% unless forloop.last %},{% endunless %}{% endcapture %}
{% endfor %}
```

ericcornelissen:
Thanks for the investigation and sorry for not getting back to this earlier. We can actually piggy back on [the filename transformation](https://github.com/simple-icons/simple-icons/blob/463450cec89ffc37785e68bf32b267fedbbeeb03/index.html#L31-L45) already present in `index.html` and base the list of search terms directly on that, i.e. we don't need to do any mapping. You can check out [a working prototype here](https://github.com/ericcornelissen/simple-icons/blob/931fd857e6f967dea4b4018b9e8bb6ff8d10c8cd/index.html) and [a diff with this PR here](https://github.com/ericcornelissen/simple-icons/commit/931fd857e6f967dea4b4018b9e8bb6ff8d10c8cd)

Feel free to update this PR according to those changes. Note that I excluded some modifications made for the `filename` from the `searchTerm`. Do also be aware that the `normalizeSearchTerm` function should still be updated as seen currently in this PR to ensure that someone searching for "ŠKODA" (with "Š") also gets the expected search result.

What do you think?

PeterShaggyNoble:
Ah, that definitely works a lot better, Eric as it eliminates the 2 additional replacements I mentioned in point 3 above. Nice one :+1:

I'll get this tested and the PR updated this morning.

PeterShaggyNoble:
Yup, that all seems to be working fine - you can test it out [here](https://petershaggynoble.github.io/simple-icons/) but you may need to force refresh [`site_script.js`](https://petershaggynoble.github.io/simple-icons/site_script.js).

I've updated the PR with the suggested changes, plus the additional change of moving `normalizeSearchTerm` from `index.html` to `site_script.js`.

ericcornelissen:
Nice 👍 Glad we can finally push this update to the live website, this is quite an important fix 😃

Issue #2887, opened 2020-04-01
Apache Icon Review


PeterShaggyNoble:
Following on from #2364 & #2734, we should do a full review of all our Apache icons using https://apache.org/logos/ as the source. As with the Google review, I think that if a project maintains its own design guidelines with vector artwork (e.g., [Flink](https://flink.apache.org/material.html)) then they should take precedence.

If I've missed any on the list, please let me know.

- [x] Apache (#4294)
- [x] Apache Airflow (#4501)
- [x] Apache Ant (#4406)
- [x] Apache CloudStack (#3023)
- [x] Apache Cordova (~~#4422~~) - LGTM (@adamrusted)
- [x] Apache Druid (#2874)
- [X] Apache ECharts (#3166)
- [x] Apache Flink (#4343)
- [x] Apache Kafka (#4344)
- [x] Apache Maven (#4500)
- [x] Apache NetBeans IDE (#4421)
- [x] Apache OpenOffice (#3969)
- [x] Apache Pulsar (#3181)
- [x] Apache RocketMQ (#4345)
- [x] Apache Solr (#4375)
- [x] Apache Spark (#4502)
- [x] Cassandra (#2890, #2942)

Why do I keep doing these things to myself?! 🙈

runxel:
Whuuaaaaat? 😵
Why didn't we know of this site earlier? .... 🙈

creepertron95:
The [Groovy](https://simpleicons.org/icons/groovy.svg) icon should be renamed to [Apache Groovy](https://apache.org/logos/#groovy).

QAInsights:
@PeterShaggyNoble Could you please add [Apache JMeter](https://apache.org/logos/#jmeter) to the list? Thanks!

mondeja:
It seems that the main Apache icon needs to be corrected (note the gap of the center of the feather), so I've added to the list:

![apache-main-update](https://user-images.githubusercontent.com/23049315/101537866-7f65ee00-399c-11eb-8262-b78508670349.png)

adamrusted:
Have reviewed Apache Ant - is the correct version, and better than my exported one with 3 decimals!

mondeja:
> Have reviewed Apache Ant - is the correct version, and better than my exported one with 3 decimals!

The sources must be updated also.

adamrusted:
The Apache Review is now finished 🎉
Just waiting on PRs being merged now.

PeterShaggyNoble:
[...]

PR #3181, opened 2020-06-05
Update Apache Pulsar

mondeja:
apachepulsar

**Issue:** #2887

### Checklist

### Description
Path extracted from [this file](https://svn.apache.org/repos/asf/comdev/project-logos/originals/pulsar.svg).


PeterShaggyNoble:
Thanks, @mondeja.

Am I missing something, though, or is the source URL the only thing that's been changed here. Not the I'm averse to doing this for all Apache brands that don't have their own design guidelines, I also don't think it's strictly necessary if there's no update needed to the SVG. Also, if we _are_ going to update the source URLs for those icons that don't need to be updated and don't have their own guidelines,I'd rather we do them all in a single PR. Any thoughts from the other @simple-icons/maintainers?

mondeja:
Seems that the path has, at least, one minor change, note the `v1.574` new command instead of `v1.573`.

ericcornelissen:
> Seems that the path has, at least, one minor change: note the v1.574 new command instead of v1.573.

That seems correct, but it doesn't seem like a notable difference :thinking: Though I'm not against updating it, I'm also not sure if it's really necessary.

>  I also don't think it's strictly necessary if there's no update needed to the SVG. Also, if we are going to update the source URLs for those icons that don't need to be updated and don't have their own guidelines,I'd rather we do them all in a single PR.

I agree with both of these points.

PeterShaggyNoble:
> Seems that the path has, at least, one minor change: note the v1.574 new command instead of v1.573.

And that's why I shouldn't try to "eyeball" the differences between two paths! :laughing:

PeterShaggyNoble:
> Seems that the path has, at least, one minor change: note the `v1.574` new command instead of `v1.573`.

And that's why I shouldn't try to "eyeball" the differences between two paths! :laughing:

It looks like the difference between the 2 is simply one of rounding during the optimisation process as, looking at the path data for the [SVG](https://pulsar.apache.org/img/pulsar.svg) I used vs. the one you used, there are only a very few minor differences with a couple of absolute commands being replaced by relative commands:

```
## https://pulsar.apache.org/img/pulsar.svg
M61,13.73H46.09C42.06,6,37.8,0,32,0c-7.55,0-11.1,6.59-14.53,13l-.3.56H0v3.81H10c3.53,0,4.25,1.19,4.25,1.19-2.77,4.3-6.1,8.55-11.65,8.55H0v4H2.62c10.76,0,15.46-8.73,19.23-15.74C25.31,8.91,27.72,5,32,5c4.44,0,9.11,9.66,13.22,18.19.6,1.24,1.21,2.51,1.83,3.77-15.76.5-23.27,4.19-29.38,7.2-4.47,2.2-8,4.93-13.81,4.93H0v4H3.83c7,0,11.36-2.16,16-4.45,6.43-3.16,13.72-6.74,30.9-6.74h9.93v-4H55.05a3,3,0,0,1-2.47-1C51.6,25,50.64,23,49.7,21L48.4,18.34s.93-1,3.51-1H61Z

## https://svn.apache.org/repos/asf/comdev/project-logos/originals/pulsar.svg
M61 13.73H46.09C42.06 6 37.8 0 32 0c-7.55 0-11.1 6.59-14.53 13l-.3.56H0v3.81h10c3.53 0 4.25 1.19 4.25 1.19-2.77 4.3-6.1 8.55-11.65 8.55H0v4h2.62c10.76 0 15.46-8.73 19.23-15.74C25.31 8.91 27.72 5 32 5c4.44 0 9.11 9.66 13.22 18.19.6 1.24 1.21 2.51 1.83 3.77-15.76.5-23.27 4.19-29.38 7.2-4.47 2.2-8 4.93-13.81 4.93H0v4h3.83c7 0 11.36-2.16 16-4.45 6.43-3.16 13.72-6.74 30.9-6.74h9.93v-4h-5.61a3 3 0 0 1-2.47-1C51.6 25 50.64 23 49.7 21l-1.3-2.66s.93-1 3.51-1H61z
```

So I'd agree with Eric that, while not strictly necessary, updating does no harm and, at the very least, updates to the more correct source URL so I'm going to merge it in. Going forward, though, we should focus initially on those icons where there is a significant visual difference.

PR #3201, opened 2020-06-11
Add InAccel Coral icon

eliaskoromilas:
### Checklist

### Description
- Filled https://inaccel.com/coral

inaccelcoral


PeterShaggyNoble:
Welcome to Simple Icons, @eliaskoromilas and thanks for the contribution.

Going by the domain of the parent brand, its Alexa rank is [~5.1m](https://www.alexa.com/siteinfo/inaccel.com) which would put it beyond the scope of this project (our cut-off is 500k). Is there another metric of popularity we can use to assess the popularity of the Coral brand?

Also, could you elaborate on where you sourced the filled version of the icon? I could only find [this partially filled version](https://www.alexa.com/siteinfo/inaccel.com) at the source URL, which, if in scope, is what I think we should use.

eliaskoromilas:
### Description
- Partially filled https://inaccel.com/coral

inaccelcoral

---

That's unfortunate! I guess we have to postpone this PR.

Regardless, I believe that the filled (no-stroke) version would look better on badges, what do you think?

PeterShaggyNoble:
> That's unfortunate! I guess we have to postpone this PR.

In that case, we'll have to close this out as "out of scope", unfortunately. Thanks for you efforts on it nonetheless and, if the popularity ever increases to bring it within our scope, please do reopen this PR.

> I believe that the filled (no-stroke) version would look better on badges, what do you think?

I definitely agree, but we can only go by what is officially available from the company themselves.

PR #3239, opened 2020-06-21
Add linting step for extraneous elements in SVGs

### LINKED ISSUE #3223: Add linting for extraneous elements in SVGs
To prevent issues like #3220. The simplest solution would be to run a regular expression on all SVGs to guarentee they start with `<svg` and end in `</svg>` (e.g. `^<svg.*</svg>$`). But ideally we include this linting step as a part of [svglint](https://github.com/simple-icons/simple-icons/blob/develop/.svglintrc.js), and I'm not sure if that allows us to do such a regular expression.

ericcornelissen:
**Issue:** Closes #3223

### Description

This adds a new custom linter to the SVG linter (`npm run svglint`) that checks if SVGs start with `<svg` and end in `</svg>` (allowing for a single newline).

It works as expected, as illustrated by [this build failure](https://travis-ci.com/github/simple-icons/simple-icons/jobs/351847196#L231), which was fixed in 498561c1b2a4c92f0790df50f7810baf1191818d (the icon had one trailing whitespace).

Suggested regarding the error message and reporter name are welcome 😄


.svglintrc.js
```diff
@@ -70,6 +72,14 @@ module.exports = {
               reporter.error(`Size of <path> must be exactly ${iconSize} in one dimension; the size is currently ${width} x ${height}`);
             }
           },
+          function(reporter, $, ast) {
+            reporter.name = "extraneous";
+
+            const rawSVG = $.html();
+            if (!svgRegexp.test(rawSVG)) {
+              reporter.error("Detected unexpected characters outside the <svg>");
```
PeterShaggyNoble, line 80: Perhaps for the error:
```suggestion
              reporter.error("Unexpected character(s) detected outside the opening and/or closing <svg> tags");
```

PeterShaggyNoble:
Other than the change above, this looks good to go to me. When I get around to reviewing the complete markup for all our icons, we may need to expand on this to include other checks.

Although, if we could somehow automate fixing niggly little things like this, that, I think, would be a better solution.

.svglintrc.js
ericcornelissen: Thanks for the suggestion @PeterShaggyNoble, that's much better

ericcornelissen:
> Other than the change above, this looks good to go to me.

Changed it 👍 Thanks for the suggestion


> When I get around to reviewing the complete markup for all our icons, we may need to expand on this to include other checks.

I'd suggest defining a separate custom check for each of those.


> Although, if we could somehow automate fixing niggly little things like this, that, I think, would be a better solution.

Even if we do that, it's still useful to have this 1) just in case and 2) for anyone to check their contributions locally.

PR #3246, opened 2020-06-22
Add Pimcore

### LINKED ISSUE #3243: Pimcore
**Name:**
Pimcore

**Website:**
https://pimcore.com/en

[press-kit.zip](https://github.com/simple-icons/simple-icons/files/4813177/press-kit.zip)

aarongerig:
pimcore

**Issue:** Closes #3243 
**Alexa rank:** [153,358](https://www.alexa.com/siteinfo/pimcore.com)

### Checklist

### Description
Icon is taken from the issue #3243. The color can be found in their guidelines available at the source URL in the "Typeface & Colors" tab.


PeterShaggyNoble:
Thanks for the contribution, @aarongerig :+1:

I am noticing some _hairline_ differences between the AI file supplied in the issue and your version (in red in example below) as well as a difference of `.004` in the heights of the 2. Most likely, this is simply down to requiring a slightly higher precision when optimising your SVG. Or it may be that you based your version on a different source file in which case, let us know and we'll do a comparison to that instead.

Of course, as @ckemptner has already contributed to the process of getting this added then, if he's happy to run with it as-is then so are we 🙂

aarongerig:
@PeterShaggyNoble Thanks for reviewing! How did you compare the two files? I'd like to see the differences (I couldn't find the example with the red notes).

PeterShaggyNoble:
When reviewing, I open to the source file and resize & centre it. Then I create a new path from the data in the contributed SVG, ensuring that both paths use different fill colours with a high contrast. Then I just eyeball it in there, zooming in as far as Inkscape allows where necessary. I then switch the layering of the 2 paths and do the same again.

aarongerig:
@PeterShaggyNoble I just did the same and I can see those minor differences between the two versions. I tried to eliminate the difference by exporting the SVG again (I'm using Adobe Illustrator), but after passing it through [SVGOMG](https://jakearchibald.github.io/svgomg/) it always has the mentioned differences. Do I have to crank up the precision slider to max? In your docs it says you have to use the default settings of SVGOMG and that's what I'm doing. :)

PeterShaggyNoble:
We recommend starting at a precision of 3 - although, if you _can_ get it down to 2, fantastic! - and then, if there's a discrepancy, increase that precision by 1 until it's rectified.

PeterShaggyNoble:
@aarongerig, could you go ahead and update the PR with the changes mentioned in [this comment](https://github.com/simple-icons/simple-icons/pull/3246#issuecomment-649417975), please?

aarongerig:
@PeterShaggyNoble Of course, I just did update the icon with precision level 4. Sorry for the long wait.

Issue #3169, opened 2020-06-04
The Great 2020 Icon Review


PeterShaggyNoble:
As brought to light recently by @davidjb (mostly) and myself, we have quite a number of icons in our library with one or more imperfections from the following list:
- They do not fit the canvas
- Their level of precision may be too high
- They are not perfectly centred on the canvas

It's going to be a fairly hefty job getting _all_ these fixed so we'd like to appeal to the members of our community to help us out if they can by taking on one or more of these icons for review on all **3** points above. (We know, example, that there are 76 icons on the first list that are also not perfectly centred on the canvas.)

To avoid having people working at cross-purposes on the same icons, if you are reviewing one, please either check it off the list or leave a comment. Also, so that this issue doesn't become cluttered with comments, if an icon warrants further discussion, please keep that discussion to the icon's PR, or create an issue for it.

After you fixed an icon, please run `SI_UPDATE_IGNORE=true npm run svglint` to update the list of icons that is ignored by the SVG linter so that the icon you fixed won't be excluded in the future.

## Icons that don't fit the canvas (#3107)

- [x] 500px - #3312
- [x] ABB RobotStudio - #3961
- [x] Accusoft - #3751
- [x] Airbnb - #3874
- [x] Airbus - #3962
- [x] AirPlay Audio - #3963 
- [x] Amazon - #3619
- [x] Amazon Lumberyard - #3964 
- [x] Anaconda - #3965 
- [x] Android Studio - #3802
- [x] AngelList - #3966
- [x] Angular - #3814
- [x] Apache - #3967 
- [x] Apache Cassandra - #3968
- [x] Apache OpenOffice - #3969 
- [x] Apple - #3970 
- [x] Apple Music - #3971 
- [x] Apple Pay - #3972 
- [x] ARCHICAD - #3327
- ~~arXiv - #3973~~ 
- [x] AT&T - #3980
- [x] Atlassian - #3872
- [x] Audiomack - #3981
- [x] Automatic - #3982
- [x] Awesome Lists - #3783 
- [x] Azure Functions - #3983
- [x] Baidu - #3984
- [x] Bamboo - #3992
- [x] Bath ASU - #3993
- [x] Beatport - #3994
- [x] Beats - #3995
- [x] Beats by Dre - #3996
- [x] Behance - #3997
- [x] Bitrise - #3998
- [x] BMC Software - #3999
- [x] Boeing - #4000
- [x] Boost - #4001
- [x] Brand.ai - #4002
[...]

Issue #3120, opened 2020-05-27
Review of icons with outlying precision levels


PeterShaggyNoble:
Following on from a discussion in #3107, I propose we review all icons with a precision higher than 5 or lower than 2 and update them if required. The 38 icons I have found are as follows, with the maximum number of decimal places in parentheses:

- [x] Jupyter (17) - #3170
- [ ] Highly (15)
- [ ] Audio-Technica (8)
- [ ] GitKraken (8)
- [ ] Ruby on Rails (8)
- [ ] TUI (8)
- [ ] Arch Linux (6)
- [ ] Autodesk (6)
- [ ] Code Climate (6)
- [ ] YouTube Music (6)
- [ ] Bitrise (1)
- [ ] CEVO (1)
- [ ] Craft CMS (1)
- [ ] Eventbrite (1)
- [ ] FACEIT (1)
- [ ] Flipboard (1)
- [ ] Gitpod (1)
- [ ] Gradle (1)
- [ ] HP (1)
- [ ] JetBrains (1)
- [ ] Kentico (1)
- [ ] Known (1)
- [ ] Koding (1)
- [ ] Kotlin (1)
- [ ] Linux Foundation (1)
- [ ] Logstash (1)
- [ ] Material Design Icons (1)
- [ ] Micro.blog (1)
- [ ] Microsoft (1)
- [ ] Roundcube (1)
- [ ] TensorFlow (1)
- [ ] Unity (1)
- [ ] Xcode (1)
- [ ] Delicious (0)
- [ ] Discover (0)
- [ ] Figma (0)
- [ ] Framer (0)
- [ ] PeerTube (0)

PeterShaggyNoble:
Closing this as it has been merged with #3169

PR #3250, opened 2020-06-23
Add svg linter to check center (with tolerance)

davidjb:
**Issue:** https://github.com/simple-icons/simple-icons/pull/3107#issuecomment-648089119

### Description

This adds a custom linter to svglint in the same vein as the last one #3107 to pull out the path, run it through a bounding box checker, calculate the centre X/Y of the path and test that the result is 12,12 (being the centre of 24,24). The math precision is to 3 decimal places, configured from the same config variable as last time, to avoid rounding errors in floating point comparisons. Whether some additional tolerance should be added to say accept values between 11.999 and 12.001 is up for discussion...

In this case, adding this linter with this precision fails on 461 files..so I suppose another ignore list is in order given the additional fixes required. 😅 Hence the WIP title to open this to discussion.  I can add an ignore in the exact same way as before, just with a different filename so you can track which lint failures are being ignored, or simply add them into the existing ignore file and any future change in path (eg fixing the icon size) will knock it off the ignore list anyway...

Example lint output (ignore list bypassed):
```
[...]
x icons/webpack.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.971 x 12.357
x icons/radiopublic.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.002
x icons/wikipedia.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.001
x icons/wikimediacommons.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12.001 x 12
x icons/windows95.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12.001 x 12
x icons/wolframlanguage.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.001
x icons/write-dot-as.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12.003 x 12
x icons/x-dot-org.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 11.999
x icons/xamarin.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.001
x icons/xcode.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.933 x 12.031
x icons/xdadevelopers.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.999 x 12
x icons/xmpp.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.001
x icons/youtube.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 11.999
x icons/zalando.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.509 x 12
x icons/zdf.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 11.998
x icons/zillow.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.001
x icons/zoom.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.024

------------ Summary ------------
✓ 909 valid files.
x 461 invalid files.
```

Some are very close and some are quite a ways off (relatively speaking when considering 3 decimal points..!), in either dimension.

461 more icons to fix 😰  - don't shoot the messenger 😎.  With the existing ignore list in place, it's showing as 244 invalids at time of writing.


PeterShaggyNoble:
Whoa! Nice, quick PR, @davidjb 🙂 And there I thought it might require some further discussion in the original PR first! 😆

Before we get to reviewing this properly, out of interest, how many SVGs fail the test if you _do_ allow for a variance of `.001`?

ericcornelissen:
Nice work once again @davidjb 👌 

Regarding ignoring these errors, I would suggest keeping it separate from the other list, but somehow include it in the same file (e.g. an array with two objects)


> Before we get to reviewing this properly, out of interest, how many SVGs fail the test if you do allow for a variance of .001?

Seems like that would be 51 with the existing list of ignored SVGs and 127 in total.

PeterShaggyNoble:
> Seems like that would be 51 with the existing list of ignored SVGs and 127 in total.

I think we should allow for that variance, so as it could well be caused by rounding in icons with a precision greater than 3. And, of course, creates less work for us in fixing it! 😆

> Regarding ignoring these errors, I would suggest keeping it separate from the other list, but somehow include it in the same file

I'd agree with that.

I'd also suggest generating a list of _just_ the 51 icons not also caught by #3107 and adding them to #3169 for fixing, work through it all as one.

davidjb:
I can add the above suggestions in.  Before I revise the PR, following the discussion at #3253, is there a desire to add the same or similar level of tolerance to the size linter as well? e.g to accept dimensions ranging from 23.999 to 24.001. 

My feeling is that if people are editing or updating icons then getting it exactly right isn't much more effort, at least no more so than my experience grappling with rounding caused by `svgo`.  That said, I'm not the one doing the editing.  I suppose certain icons may be more difficult than others to ensure they're correct so it's up to the project to decide what tolerance is acceptable and in what way.

ericcornelissen:
> I'd also suggest generating a list of just the 51 icons not also caught by #3107 and adding them to #3169 for fixing, work through it all as one.
>
> EDIT: Should we merge #3120 with #3169 while we're at it?

I would personally keep them separate, though that might get us in some confusing situations when we don't cross an icon off both lists :thinking: In that regard I'm fine either way, and I would merge all three of the lists into one.


> I can add the above suggestions in. Before I revise the PR, following the discussion at #3253, is there a desire to add the same or similar level of tolerance to the size linter as well? e.g to accept dimensions ranging from 23.999 to 24.001.

That, to me, seems like a separate discussion that can be handled in a separate PR. Could you update this PR with the above suggestions @davidjb?

For now, I haven't considered it very well but it seems sensible to include a similar level of tolerance in the size linter but @PeterShaggyNoble [disagrees](https://github.com/simple-icons/simple-icons/pull/3253#issuecomment-651074277).

davidjb:
Okay so just to confirm the changes I need to make are:
* adding a tolerance of +/- 0.001 to the centre-checking linter, and
* adding the ignore list for the ~51 icons that fail this new test (which will involve restructuring the ignore file in the process)

Just checking before I get started.

ericcornelissen:
Correct @davidjb :+1:

davidjb:
All done and pushed.  The PR now has tolerance for at most +/- 0.001 in any direction.  The ignore file was updated and restructured (I used the code at https://github.com/simple-icons/simple-icons/pull/3107#issuecomment-640132743 to dump again) -- 81 additional icons were ignored because of the centring linter.  The size linter only checks against the `size` ignore list but the `center` linter checks against both.

I thought of this a while ago with the ignore list - potentially adding an ignore reporter like this into the svglintrc, maybe as a disabled-by-default, separate linter?  That way, project admins or enthusiastic contributors can easily discover which icons were ignored and why.  I would say enable the warnings by default since warnings won't break a CI build - but the 369-odd files on the ignore list generate a lot of output that'd probably spook a contributor.  Anyway, here's the code as an example:

```js
            if (iconIgnored.size.hasOwnProperty(iconPath)) {
              reporter.warn(`Icon was ignored due to size linter`);
              return
            } else if (iconIgnored.center.hasOwnProperty(iconPath)) {
              reporter.warn(`Icon was ignored due to center linter`);
              return
            }
```

.svglintrc.js
```diff
@@ -80,6 +81,33 @@ module.exports = {
...
+            const centerY = +((bounds.minY + bounds.maxY) / 2).toFixed(iconFloatPrecision);
+            const devianceY = centerY - targetCenter;
+
+            if (
+              devianceX < -iconTolerance ||
+              devianceX > iconTolerance ||
+              devianceY < -iconTolerance ||
+              devianceY > iconTolerance
+            ) {
+              reporter.error(`<path> must be centered on the canvas; the center is currently ${centerX} x ${centerY}`);
```
runxel, line 108: No real opinion on that, but maybe write how much it differs (`devianceX/Y`) instead of the current center?
But again, I'm not even sure what I would prefer.

ericcornelissen: I'm not against this, but I'm not sure how useful that would be in practice. I'd say if we find it would be useful we can always add it later?

runxel: Yeah, sure! :)
Like I said, it's not really a must. I think both values say basically the same, but with the deviance you can see on one glance where something has gone wrong, and in which direction. The center value just needs a tiiiiiny bit of more brain time to process, I guess.

.svglintrc.js
```diff
@@ -80,6 +81,33 @@ module.exports = {
...
+            const centerX = +((bounds.minX + bounds.maxX) / 2).toFixed(iconFloatPrecision);
+            const devianceX = centerX - targetCenter;
+            const centerY = +((bounds.minY + bounds.maxY) / 2).toFixed(iconFloatPrecision);
+            const devianceY = centerY - targetCenter;
+
+            if (
+              devianceX < -iconTolerance ||
+              devianceX > iconTolerance ||
+              devianceY < -iconTolerance ||
+              devianceY > iconTolerance
```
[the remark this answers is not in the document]
davidjb: Great catch - I was going to use abs but then must have gotten distracted. I've pushed this into the PR.

.svglintrc.js
davidjb: On a slight tangent, it could be hard to ascertain that the centre should be 12,12 (e.g. one either needs to know the size should be 24 x 24 and figure it out or already know it should be 12,12 -- details of which are in the contributor guide but still...). When I was writing the PR, it's reasonably clear to see from the lint output of 203 erroring icons what the value should probably be:

```sh
[...]
x icons/zalando.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.509 x 12
x icons/zdf.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 11.998
x icons/zoom.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 12 x 12.024
```

but you lose this point of reference as these icons are being ignored and eventually, when icons are fixed, a contributor _should_ only be seeing one error at a time.  If the centre is a long way out, there's some effort needed to parse something like:

```sh
x icons/fakeicon.svg
  x icon-centered <path> must be centered on the canvas; the center is currently 11.444 x 12.638
```

Anyway, I agree with @runxel that knowing where and in which dimension something's gone wrong is useful. At the same time I feel like seeing the expected value + problematic value together explains it in absolute terms so
I've pushed a change to the output to indicate what the centre should be, plus a formatting change to use coordinate format versus `w x h` dimensions.  I don't know how this output goes for countries that use the comma as a decimal separator; I added a space to hopefully help in that case + for readability.    See what you think 👍   

```sh
x icons/zoom.svg
  x icon-centered <path> must be centered at (12, 12); the center is currently (12, 12.024)
```

davidjb:
As above, I've added @runxel's suggestion of `Math.abs` (thanks!) and modified the reporter output to explain what the centre should be and what it currently is.

.svglintrc.js
ericcornelissen: I'm completely fine with the current error message @davidjb, nice work.

Regarding "the comma as a decimal separator": the space does make that clear but as far as I know programming languages (and by extension programmers) always use a dot as a decimal separator so it shouldn't be a problem to begin with.

davidjb: Cool, thanks @ericcornelissen.  I was aiming for as explicit as possible, and as a plus it seems more readable/scannable with spaces.

ericcornelissen:
Thanks again for this very useful contribution @davidjb, we all appreciate it very much! 🎉🎉 

@PeterShaggyNoble will you look into [updating (and merging?) the list(s) of icons that are being ignored](https://github.com/simple-icons/simple-icons/pull/3250#issuecomment-649427340)?

davidjb:
You’re welcome! I feel bad as I keep making more work for people. Is there anything else that needs linting? 😅

PeterShaggyNoble:
Thanks for another valuable contributions, @davidjb :+1:
> I feel bad as I keep making more work for people.

Feel free to pitch in on cleaning it all up! 😆
> will you look into updating (and merging?) the list(s) of icons that are being ignored?

On it!

PeterShaggyNoble:
@davidjb, looks like something went wrong when generating the values for the list of icons to ignore; only the first word in each brand name was included.

davidjb:
Hmm very curious. I see what happened, my title extraction code was incredible naïve, assuming no spaces in the brand name. Shall I get on it and send an update PR with regenerated/updated ignore lists?

PeterShaggyNoble:
:+1: Yes please, @davidjb

davidjb:
All done @PeterShaggyNoble, see the new PR over at #3433. To allow automated updating of the ignore file much easier, the code's now embedded in the svglintrc; that'll help with any future linters that might come along too.

PR #3266, opened 2020-06-28
Rename Firefox and add parent brand

### LINKED ISSUE #2462: Firefox
**Name:** Firefox
**Website:** https://www.mozilla.org/en-US/firefox/
**Alexa Rank:** n/a (Mozilla subdirectory)
**Colour:** `#321c64`
**Official resources for icon and color:** Logo from https://mozilla.design/firefox/logos-usage/ Colour from https://mozilla.design/firefox/color/ - there's a load of them to select from, with no clear indication if any one colour is preferred for any one of their brands. I chose one that seemed to make sense to me but it may not to others.

This is the Firefox parent brand (as opposed to the browser) - I've added this to the `v3` milestone with the "pending" label as, if we do add it, it would require renaming our existing Firefox icon as "Firefox Browser".

[Firefox](https://github.com/simple-icons/simple-icons/files/4104019/firefox.zip)

PeterShaggyNoble:
Firefox

**Issue:** Closes #2462
**Alexa rank:** n/a

### Checklist

### Description
I've gone with the same orange for the parent brand. Also, it's not 100% clear to me whether or not we need to keep the "Mozilla" prefix. I've left it out for now but I'm happy with either. Colour from [this page](https://mozilla.design/firefox/color/).

Firefox Browser


ericcornelissen:
I will leave this open for a bit to allow for discussion regarding whether or not "Mozilla" should be included in the name for Firefox. To me, however, it is pretty clear from the [Logos & Usage page](https://mozilla.design/firefox/logos-usage/) that it is not definitely not necessary and should probably not be there in any case.

Relatedly, though, this also brings to the discussion of whether or not the "Mozilla" prefix should be omitted from [Thunderbird](https://github.com/simple-icons/simple-icons/blob/ccd927ec186103efab35543414dc838c0b58751b/_data/simple-icons.json#L3974) :thinking:

PR #3283, opened 2020-07-03
Rename Line as LINE

PeterShaggyNoble:
**Issue:** n/a
**Alexa rank:** [~227](https://www.alexa.com/siteinfo/line.me) & [~1.4k](https://www.alexa.com/siteinfo/webtoons.com)

### Checklist

### Description
Line should be stylised as "LINE" ~and it looks like WEBTOON dropped the "LINE" from its name some time last year~.


PeterShaggyNoble:
I'd missed that, @ericcornelissen. Given that, leaving it as it currently is is definitely the safest option as it will, as you say, show up no matter which version of the name people search for. I'll roll back that commit now.

And I'd counter your argument that people should be looking for the English version with the argument that we should be aiming to serve an international, multilingual audience :wink: It's one of the reasons, for example, I've started adding, and encouraged others to add, public transport systems in their native language rather than English.

ericcornelissen:
> Given that, leaving it as it currently is is definitely the safest option as it will, as you say, show up no matter which version of the name people search for. I'll roll back that commit now.

👍 In that case I will merge this in as is. If it turns out that it is actually reasonable to rename (Line) Webtoon as well that can be handled in a separate PR.


> And I'd counter your argument that people should be looking for the English version with the argument that we should be aiming to serve an international, multilingual audience 😉 It's one of the reasons, for example, I've started adding, and encouraged others to add, public transport systems in their native language rather than English.

I intentionally said "[One] **could** argue ...", I don't really agree with it myself 🙃

PR #3328, opened 2020-07-15
Add Rakuten Icon

### LINKED ISSUE #3294: Rakuten
**Name:** Rakuten
**Website:** rakuten.co.jp
**Alexa rank:** 93
  

davidkryzaniak:
rakuten

**Issue:** Closes #3294 
**Alexa rank:** 93

  

### Checklist


davidkryzaniak:
That's odd. Yes, I did use the referenced image. I just attempted to recreate the logo again and got the same output. Maybe I'm doing something wrong? (too aggressive on the optimization?)

ericcornelissen:
> 
> 
> That's odd. Yes, I did use the referenced image. I just attempted to recreate the logo again and got the same output. Maybe I'm doing something wrong? (too aggressive on the optimization?)

Yes, it could very well be the optimization. Could you try with less aggressive optimization (i.e. increasing the `precision` value)? Alternatively you can share the SVG produced by your editing software here and we can work together from that.

PeterShaggyNoble:
I can confirm that when running this through SVGOMG with a precision of 3, there is a very definite visual shift. However, in my version it was only apparent in the inner curve of the "R". Unfortuntately, to completely resolve it, I had to up the precision to 6! 😮 SVGO may fare better than that, though.

ericcornelissen:
SVGOMG just runs SVGO for you but with a UI so it shouldn't make a difference (other than the fact that there are more options available)

PeterShaggyNoble:
> it shouldn't make a difference

It shouldn't, no, but on _very_ rare occasions in the past both @runxel and I have noticed differences in the output between the two 🤷‍♂️

adamrusted:
Found a new SVG version of the underlined 'R' [here](https://global.rakuten.com/corp/assets/img/site-icons/rakuten-black.svg) - and rebuilt from source. Seems to be working with a precision of 3, but could another @simple-icons/maintainers look at this?

rakuten

mondeja (approved):
Perfect @adamrusted, great work! :100:
Thank you for rescuing this addition and @davidkryzaniak for the previous effort :pray:

PR #3367, opened 2020-07-28
Add PyCharm

### LINKED ISSUE #3353: PyCharm
**Name:** PyCharm (in #625 but it was closed)
**Website:** https://www.jetbrains.com/pycharm/
**Alexa rank:** 1,747
  

https://commons.wikimedia.org/wiki/File:PyCharm_Logo.svg 
color: black (000000)

I don't know how to process a svg image so I need someone's help. Thanks a lot.

MetLee:
image

**Issue:** Closes #3353 
**Alexa rank:** 1,747
  

### Checklist

### Description

I downloaded the svg from here: https://www.jetbrains.com/company/brand/logos/ .
I deleted the characters and resized the logo to 24 * 24.


MetLee:
Should I add more icons from https://www.jetbrains.com/company/brand/logos/ ?

PeterShaggyNoble:
So, after posting my comment on the original issue yesterday, I came across https://www.jetbrains.com/company/brand/ which states:
> **Do Not Separate the Black Box and Beams**
> Never use the Black Box without the beams behind. This will make the logo disappear in the background and lose the energy created.

Initially I thought that that just applied when being used on a dark background but, further down the page, they reinforce that with:
> You may not alter Logos in any way except pixel dimensions

I don't ever remember seeing that restriction before so it may be new. Either way, we're unfortunately going to have to update this and all our existing JetBrains icons to include the "beams".

PeterShaggyNoble:
@MetLee, which file did you base your SVG off? The letters don't line up with `icon-pycharm.svg` but I'm guessing that's might be because you used `logo-pycharm.svg` - as it's intended to be used as an icon, I think we should base our version off the former.

Also, could you update the source URL to the logos page, please, rather than directly to the zip?

MetLee:
> @MetLee, which file did you base your SVG off? The letters don't line up with `icon-pycharm.svg` but I'm guessing that's might be because you used `logo-pycharm.svg` - as it's intended to be used as an icon, I think we should base our version off the former.
> 
> Also, could you update the source URL to the logos page, please, rather than directly to the zip?

There is a problem confusing me: 
If we use `icon-pycharm.svg`, we need to change the color of its beam (according to the `CONTRIBUTING.md`: `7. Remove all colors. The icon should be monochromatic`).
But on JB's official guideline, we can't change this color.

MetLee:
BTW, the [official website](https://www.jetbrains.com/company/brand/) uses only black box on the top left corner without the beam, which conflicts with its guideline.
image

It's a bit strange.

MetLee:
image

PeterShaggyNoble:
> Should I add more icons from https://www.jetbrains.com/company/brand/logos/ ?

Yup, provided they meet our popularity criteria, work away - that'd be great. To help with reviewing, though, please submit a separate PR for each.

> You meant that I should use the black box of icon-pycharm.svg, right?

Indeed

> BTW, the official website uses only black box on the top left corner without the beam, which conflicts with its guideline.

Often, a company's internal brand guidelines will differ from those they make publicly available, usually because, when working with first-party implementations of their branding, they're working in an environment where the brand is already firmly established.

PR #3372, opened 2020-07-30
Add PhpStorm

MetLee:
image

**Issue:** #625 
**Alexa rank:** 1,862
  

### Checklist


PeterShaggyNoble:
Thanks for the PR, @MetLee :+1: Note, though, that the Alexa rank you've quoted is for jetbrains.com and doesn't give an accurate reflection of the popularity of PhpStorm so the first thing we'll need to do is determine the popularity of both this and GoLand. After Alexa ranks, the next thing we usually look at is GitHub stars (our cut-off is 5k) but, unfortunately, neither has it's own dedicated repo. Could you suggest another metric we could use to assess the suitability of these brands?

MetLee:
image
I found this table on https://www.alexa.com/siteinfo/jetbrains.com

PeterShaggyNoble:
Interesting idea, @MetLee :+1: But comparing them with their parent company probably isn't going to give us the most accurate data. If we could find some sort of comparison between each brand and a direct competitor that we can definitively say would be in-scope for us, that might give us a better idea.

MetLee:
> Interesting idea, @MetLee 👍 But comparing them with their parent company probably isn't going to give us the most accurate data. If we could find some sort of comparison between each brand and a direct competitor that we can definitively say would be in-scope for us, that might give us a better idea.

I am not good at finding the rank of a specific brand; may you give me some advice?

cevin:
@PeterShaggyNoble 

Another product of the same company has been included. It is also a very popular commercial IDE tool. There is no reason why PhpStorm cannot be included.

PeterShaggyNoble:
> Another product of the same company has been included

Thanks for the feedback, @cevin. However, another product from the same company already being available isn't reason alone for us to add a brand, no matter how big that company is. For example, we regularly discuss the suitability of various Google products for inclusion and have just recently [rejected a product from Adobe](https://github.com/simple-icons/simple-icons/pull/2792).

> There is no reason why PhpStorm cannot be included

What we're trying to do here is come up with some way to assess the popularity of _all_ JetBrains products that fall within their own domain so we can use that for any future requests to add more of their products (such as [GoLand](https://github.com/simple-icons/simple-icons/pull/3373)).

MetLee:
> > Another product of the same company has been included
> 
> Thanks for the feedback, @cevin. However, another product from the same company already being available isn't reason alone for us to add a brand, no matter how big that company is. For example, we regularly discuss the suitability of various Google products for inclusion and have just recently [rejected a product from Adobe](https://github.com/simple-icons/simple-icons/pull/2792).
> 
> > There is no reason why PhpStorm cannot be included
> 
> What we're trying to do here is come up with some way to assess the popularity of _all_ JetBrains products that fall within their own domain so we can use that for any future requests to add more of their products (such as [GoLand](https://github.com/simple-icons/simple-icons/pull/3373)).

Agree. So we need to find a method to show its popularity.

cevin:
how about Google Trends
https://trends.google.com/trends/explore?q=phpstorm

MetLee:
https://trends.google.com/trends/explore?q=phpstorm,jetbrains
It shows that PhpStorm is of the same popularity as JB.
@PeterShaggyNoble

cevin:
> https://trends.google.com/trends/explore?q=phpstorm,jetbrains
> It shows that PhpStorm is of the same popularity as JB.
> @PeterShaggyNoble

https://trends.google.com/trends/explore?q=phpstorm,jetbrains,webstorm
Even higher than the already included WebStorm

mondeja:
> https://trends.google.com/trends/explore?q=phpstorm,jetbrains,webstorm
> Even higher than the already included WebStorm

Definitively this comparison reflects that WebStorm can be within the scope of simple-icons. I agree with the inclusion of PhpStorm. @PeterShaggyNoble are you against this?

service-paradis:
I also agree with the inclusion. We already have some other JetBrains products (JetBrains, PyCharm, IntelliJ IDEA and potentially TeamCity). I think the popularity of PHPStorm is comparable to those and within the scope.

mondeja (approved):
Looks good to me, so I'm merging this. Thank you for adding this one @MetLee :sparkles:

PR #3373, opened 2020-07-30
Add GoLand

MetLee:
image

**Issue:** None
**Alexa rank:** 1,862
  

### Checklist


mondeja:
Sorry, but I think that Goland is not in the scope of simple-icons for now. Comparing with already included Webstorm using Google Trends https://trends.google.com/trends/explore?q=gloand,webstorm just as it was done in #3372  we can see that the usage of Goland is far enough. If nobody have any arguments against this, I will close this pull.

Issue #3369, opened 2020-07-29
Update Adobe Icons


PeterShaggyNoble:
Adobe have recently updated the majority of their product logos so that they all now have filled, square backgrounds with rounded corners so we'll need to update to match. In all cases an SVG of the logo can be found on the individual product's page.

We'll need to decide, too, what to do for the colours and be consistent; either we keep our current colours, use the colour of the background square of the new logo or use the colour of the icon/text from the new logo.

Icons we need to update are:
- [x] Adobe After Effects  (#3378)
- [x] Adobe Audition (#3379)
- [x] Adobe Dreamweaver (#3380)
- [x] Adobe Illustrator (#3377)
- [x] Adobe InDesign (#3381)
- [x] Adobe Lightroom CC (#3397, #3396)
- [x] Adobe Lightroom Classic (#3382)
- [x] Adobe Photoshop (#3376)
- [x] Adobe Premiere Pro (#3384)
- [x] Adobe XD (#3386)

MetLee:
I will update the other icons.

PeterShaggyNoble:
Thanks for taking these on, @MetLee, it's a big help.

MetLee:
According to the websites (https://www.adobe.com/products/photoshop-lightroom.html and https://www.adobe.com/creativecloud/catalog/desktop.html), I suggest that `Adobe Lightroom CC` should be renamed `Adobe Lightroom`.

PeterShaggyNoble:
Yes, I'd agree with that, @MetLee. As it's a breaking change, though, it will need to wait until [v4](https://github.com/simple-icons/simple-icons/milestone/3). Can you create a separate issue so we can add it to that milestone, please?

PeterShaggyNoble:
To clarify: It's just the name change that needs to wait until v4; you can go ahead with the updates to the icon & colour.

MetLee:
> Yes, I'd agree with that, @MetLee. As it's a breaking change, though, it will need to wait until [v4](https://github.com/simple-icons/simple-icons/milestone/3). Can you create a separate issue so we can add it to that milestone, please?

Sure!

MetLee:
#3396

PeterShaggyNoble:
Going to close this out as everything in this issue has been taken care of, pending a decision on the colour choices.

Thanks once again for all your work here, @MetLee :+1:

PR #3376, opened 2020-07-30
Update Adobe Photoshop

MetLee:
image

**Issue:** #3369 
**Alexa rank:**
  

### Checklist

### Description

I use the color of the text 'PS'.


MetLee:
@PeterShaggyNoble  Should I continue to update the other 9 icons?

ericcornelissen:
> [...] SVG looks perfect to me. I agree with you that:
>
>    1. We should update our colours, and,
>    2. We should use the foreground colour of each icon,

I complete agree with that 👍

runxel:
I'll give my support for this as well!
(By "foreground" we mean the text color, which part is now knocked out, right?)

PeterShaggyNoble:
Excellent, looks like we have a consensus, so 🙂

> By "foreground" we mean the text color, which part is now knocked out, right?

That's it, ay, @runxel.

PR #3380, opened 2020-07-30
Update Adobe Dreamweaver

MetLee:
image

**Issue:** #3369 
**Alexa rank:**
  

### Checklist


MetLee:
> There are a few excess points in the original version of this, as there are in all these new icons, if you'd like to take a stab at cleaning them up while we decide on the colours. Otherwise, this is ready to go as-is.

I don't know how to clean them up correctly. I attempted to use SVGOMG to compress the svg file but it didn't work... I mean that the size of this svg file will be changed (to 24*23.99) and it can't pass the automated check.

PR #3433, opened 2020-08-08
Update svglint ignore list, add ignore list updater

davidjb:
**Issue:** #3250 
**Alexa rank:** N/A
  

### Checklist

### Description

This updates the ignore list as per #3250 and embeds code for updating
the ignore file when the environment variable `SI_UPDATE_IGNORE=true`
is set.

This also:

* Normalises the top-level object key in the ignore file to match each linter name
* Sorts the ignore output on top-level key and icon name value to make output consistent
* Speeds up icon center checking if ignored
* Ignores icons in size linter if icon is otherwise ignored from another linter (by centralising the ignore code)


davidjb:
Updated the PR to ensure the ignore output is consistently sorted; the svglint processing is async so object key and valuing ordering varied between runs.  This is handled by sorting on top-level key (linter name) and then on icon name (actually an Object _value_, which is a little interesting.

davidjb:
With all the various updates coming down the line (ala #3169), getting this merged would help with maintaining the ignore lists.  Any changes required on this?

ericcornelissen:
Sorry for the delay @davidjb, so if I understand correctly this would allow us to run `npm run svglint` after an icon has been (supposedly) fixed and it will update the `.svglint-ignored.json` to the new list of SVGs that are not yet fixed (and should therefore be ignored). That is quite convenient as I'd imagine it might get pretty repetitive to remove fixed icons manually.

I have one small remark/questions. Currently `isIgnored` seems to check if the icon should be ignored by any linter (i.e. it uses `Object.values(iconIgnored)`. Wouldn't it be better to change this to a specific linter by including the `linterName` (something along the lines of `iconIgnored[linterName].map(...).some(...)`)?

davidjb:
Yep, that's right @ericcornelissen - so run as `SI_UPDATE_IGNORE=true npm run svglint` and that'll trigger an update to the ignore file.

As for the `isIgnored` - the current implementation (e.g. the code as it is before this PR) has a bit of a weird mix of ignores - the size linter only ignores from the size list but the centring linter ignores from both size and centre lists (https://github.com/simple-icons/simple-icons/pull/3433/files#diff-6a7a61d95c2af287ca7aae4861a6720e553f5aaebcfab1e90b1189bd6639f14cL90-L91).  The size linter came first and that was implemented as-is and then for the centring linter, there was a request from someone (can't recall who) to make it like that to avoid either extra errors (e.g. if ignored once, ignore for all) and/or avoiding a duplication of entries in the ignore file.

Since the ignore file is automated and will be shrinking, your suggestion of just checking the specific linter's ignore list makes sense.  Let me know and I'll update the PR plus also update the ignore file as well (since a bunch of PRs have been merged since and it's conflicted).

ericcornelissen:
> Yep, that's right - so run as SI_UPDATE_IGNORE=true npm run svglint and that'll trigger an update to the ignore file.

Alright, sounds good to me :+1: I think we should add a note about this to the original post of #3169 once this gets merged.


> The size linter came first and that was implemented as-is and then for the centring linter, there was a request from someone (can't recall who) to make it like that to avoid either extra errors (e.g. if ignored once, ignore for all) and/or avoiding a duplication of entries in the ignore file.

I see, seems that was a decision made in #3250.


> Since the ignore file is automated and will be shrinking, your suggestion of just checking the specific linter's ignore list makes sense. Let me know and I'll update the PR plus also update the ignore file as well (since a bunch of PRs have been merged since and it's conflicted).

Yes, I think I like to keep them separate from now on. Any thoughts on this @runxel or @NovaGL ?

runxel:
Also a sorry from me, @davidjb, for having this one lying aroung so long!
It was a bit over my head and I didn't fully understood what's going on, so I didn't merge.

Back to topic: I agree with @ericcornelissen, a seperation makes sense to me.

Again, thanks for this work!

davidjb:
No worries @runxel!  All done and pushed - ignore file has updated against the latest `develop` and code modified so that each linter only checks against its own ignore lists. 

Now it's down 350 icons ignored -- it was at least 461 a while back.  Some great work going on over in #3169 👍

runxel:
Wonderful! ❤👌

PR #3473, opened 2020-08-20
Adds link to vue-simple-icons

noahlitvin:


runxel:
Hi @noahlitvin 👋 
Thanks so for much for creating the Vue package and this PR. Would you mind sorting it in alphabetically?
So Svelte would come before Vue. Thanks!

noahlitvin:
@runxel 👍 All set.

PR #3478, opened 2020-08-21
Add Osano

pgoforth:
osano (1)

**Alexa rank:**
Global: 41,269
US: 15,261

### Checklist

### Description
I am a lead software engineer for this company.
Hex value is primary style guide color and logo is from original vector.


runxel:
Hi @pgoforth! 👋 
Thank you for this PR!
However I'm a bit confused: Your version and the full logo don't match, as you can see in this screenshot. Your version is a perfect round "doughnut" shape, while the Osana "o" is not.... 🤔 

comp

pgoforth:
@runxel 
I am using the solo `O` logo from our company's shared assets folder. I will take your question directly to the founder...he may even reply inline here. Regardless, we will make sure this is 100% correct, and thank you for pointing this out.

pgoforth:
@runxel 
Per my conversation with the founder, the version submitted is the correct version and the version on the website is an old version in need of updating. I've included a screenshot of his response in our Slack:
Screen Shot 2020-08-31 at 10 15 13 AM

runxel:
Okay, @pgoforth, great to know! 👍 
Then we are pretty good to go, just one simple wish: Since the linked "source" version is not correct I would just use the base url instead.

pgoforth:
@runxel 
All changed and ready to go! Thank you so much for your diligence.

runxel:
Thanks for your patience @pgoforth! We're good to go!

PR #3507, opened 2020-08-27
Add logo p5.js

OOQQ:
**Name:** p5.js
**Website:** https://p5js.org/
**Alexa rank:** 38.638

**Checklist**
- I optimized the icon with SVGO or SVGOMG
- The SVG viewbox is 0 0 24 24

**Description**
The famous Processing vector library ported to javascript and the HTML5 <canvas> element for embedding.
[p5js.svg.zip](https://github.com/simple-icons/simple-icons/files/5136061/p5js.svg.zip)


runxel:
Congratulations @OOQQ on your first PR!

As you can see the build is currently failing, since the icon is not exactly 24px in one dimension, but 23.95.
Seems like a minor difference but we just try to get it right. 😃 
Would you please check the size again? (Using a bigger precision in SVGO helps, when the error was introduced there) 
Thanks! 💯

adamrusted:
Thanks again to @OOQQ for your work on this icon! I was able to pull this into Inkscape and fix the sizing, and it's now passing the linter with no problems. Could another @simple-icons/maintainers check this over before merge?

service-paradis:
LGTM 🎉 Thanks @OOQQ and @adamrusted!

PeterShaggyNoble:
Could we get away with using just the asterisk here, as they do in their favicon?

Issue #3515, opened 2020-08-30
Icon Request: Ren'Py


creepertron95:
**Name:** Ren'Py
**Website:** https://www.renpy.org/
**Alexa rank:**[104,196](https://www.alexa.com/siteinfo/renpy.org)
  

**Official resources for icon and color:** 
https://www.renpy.org/static/index-logo.png 
https://github.com/renpy/renpy/blob/master/sphinx/source/logo.png

NovaGL:
I think this logo would be very difficult to convert to monochrome. If you want to have a go feel free.

PR #3523, opened 2020-08-31
Add Ren'Py Icon

creepertron95:
renpy(2)

**Issue:** #3515
**Alexa rank:** 104,196
  

### Checklist

### Description

I have attempted to manually vectorize the logo from the source image the best as I could due to the logo being .png only. I picked the hex value from the mascot's hair color.


creepertron95:
cc @renpytom

runxel:
@creepertron95 Thank you so much for this PR!
Manual vectorizing is always difficult. We appreciate your effort!

As you can see, the build is currently failing, because it seems the path is not perfectly centered. Would be nice to check that again. You _might_ need to bump up the precision of svgo to get a pass in the linter.

creepertron95:
@runxel Should I go ahead and invert the snake as well as the founder suggested?

runxel:
@creepertron95 We try to respect the wishes from the brand owners, so normally we look at things like brand manuals. But the founder replying in a thread has even more importance. So yes, that would be great.

Also have a look at the svg again in general. The build fails because the structure isn't right. But thats easy to fix. :)

creepertron95:
@runxel How can I go about fixing the structure?

runxel:
Thanks for your work!
The lint is still failing however 🙈 
`<path> must be centered at (12, 12); the center is currently (12.012, 12)`

I don't know what tool you are using – sometimes with Illustrator there are problems to align it correctly. Inkscape seems to do a better job at this. 

Otherwise it looks great! 💯  I think @renpytom will agree :)
If not: now's the best time to demand changes!

creepertron95:
@runxel I am using Inkscape 1.0, How can I center the path correctly to fix the lint?

runxel:
I don't know Inkscape well, tbh, but I do think you can position it correctly with the numerical position input (there should be some boxes).

creepertron95:
@runxel I have tried to recenter the logo by adding and subtracting '0.012' to the X position input and the build still fails.

runxel:
No! @creepertron95 it looks all good now! 💯  You can click on the Travis link btw, to see why the build failed.
The report now says, that the location of the new item inside the JSON is wrong... Pretty sure the `'` is problematic. I'm not sure tho what we actually expect (luckily most brands use only alphanumeric values 😅). Gonna tag @PeterShaggyNoble here who will probably have a clue! :)

creepertron95:
@runxel Done! I found the problem and fixed it.

runxel:
Great! We're good to go!
Big thanks to anyone involved! 💯

Issue #3529, opened 2020-09-02
Request: NumPy Icon


rajrohanyadav:
**Name:**NumPy
**Website:**https://numpy.org/
**Alexa rank:**14,027
  

**Official resources for icon and color:**
  
https://numpy.org/images/logos/numpy.svg

NovaGL:
Color:  `#4d77cf` or `#4dabcf`

Thanks for the request. Do you think you would be up to trying to do a pull request?

Please see the [contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) if you are upto it.

rajrohanyadav:
Yes @NovaGL I want to give it a shot.

rajrohanyadav:
Hey @NovaGL,
I gave it a shot, thanks for asking.
I'm a newbie, & need a little advice: Do I write this sort of message after creating a PR, or is it not needed/unprofessional.
Thanks anyways.

NovaGL:
There's no right or wrong way. Just create a PR and I'll let you know if it needs changes. As you can see it's a bit slow around here so might take awhile to get approved

rajrohanyadav:
Thankyou,
I figured what was wrong, and made necessary changes. Probably all is fine now.

PR #3531, opened 2020-09-03
add NumPy icon

rajrohanyadav:
**Issue:**#3529
**Alexa rank:**14,027
  

### Checklist

### Description

I chose Website accent color as the hex value. SVG logo was taken from the official website.


ericcornelissen (changes requested):
Thanks for taking the time to contribute @rajrohanyadav, and nice job on figuring out why the build failed 👍 

I have a question w.r.t. the SVG. [The SVG you linked in the original issue](https://numpy.org/images/logos/numpy.svg) **is not** square, but the SVG you contributed **is** square (see the comparison below, red is the original). Could you correct this so that the original dimensions of the logo are maintained (within a 24x24 canvas)?

Also, where on the website exactly can I find the color you specified (`#013243`)? Also, @NovaGL, why did you suggest `#4d77cf` or `#4dabcf` as brand colors?

comparison

NovaGL:
Hi @ericcornelissen 
When suggesting colors I just provide the colors used in the SVG and nothing else, if the contributor feels there are more suitable colors its up to their discretion.

I agree that `#013243` is a better choice in this case as it seems to be the main color used on the website. For example its the background color used for the footer.

ericcornelissen:
@rajrohanyadav I see you updated the contribution. However, now it looks like the icon is not scaled to fit the viewbox, as stated in [the Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#2-extract-the-icon-from-the-logo):

> Scale the icon to fit the viewbox, while preserving the icon's original proportions. This means the icon should be touching at least two sides of the viewbox.

rajrohanyadav:
Hey @ericcornelissen Original ratio of the icon was 18.59 * 19.87. If I scale it, it does not fit the box exactly. Also, centering is also failing. I'm using figma. I'd really like to do it, can you tell if figma is ok, or I should use something else.
Also, you said while touching atleast two sides, if I try to accomplish that should it be at the cost of centering.
Thankyou.

rajrohanyadav:
hey @ericcornelissen Thankyou for your help, I fixed the dimension and scaling.
The only issue is:
For maintaining the logo ratio, I resized it to 22.46 * 24
Now I don't find a way to center it perfectly unless I resize it to 22 * 24.
Can you please help. Thankyou

rajrohanyadav:
Hey @ericcornelissen I'm getting center at (11.998, 12), which is causing fail. I am using Figma which has only 2 decimal places accuracy. Should I use some other program?
Thankyou

ericcornelissen:
@rajrohanyadav you could try changing the location entirely and manually setting it so that the center is at 12x12, or you could maybe try [Inkscape](https://inkscape.org/) (with which I don't have experience) instead. Do also note that the SVGO(MG) optimization may be causing the SVG to be off-center. You can see if this is the case by increasing the precision that SVGO uses. Hope that helps :slightly_smiling_face:

rajrohanyadav:
Hey @ericcornelissen and @NovaGL many thanks for your help. I think I've done it right this time. 
Maybe next time I won't need to ask so many questions.
I chose `#013243` because the brand name, the primary action button as well as the footer are of this color. It seems to be the accent color of the brand.

ericcornelissen (approved):
> many thanks for your help. I think I've done it right this time.
> Maybe next time I won't need to ask so many questions.

Yes, awesome :tada: It's not a problem that you ask questions :smile: Many thanks for the contribution!

> I chose `#013243` because the brand name, the primary action button as well as the footer are of this color. It seems to be the accent color of the brand.

Okay, thanks for the clarification. Given that @NovaGL agrees as well I complete fine merging this as is :+1:

PR #3551, opened 2020-09-11
Add TurboSquid

creepertron95:
turbosquid

**Name:** TurboSquid
**Website:** https://www.turbosquid.com/
**Alexa rank:** [4,818](https://www.alexa.com/siteinfo/turbosquid.com)
  

### Checklist

### Description
 There is actually several versions of this logo. The favicon is encourged to be used if size is below 50 pixels and a detailed squid version. The color is TS Orange from the Colors section from the branding page. This PR is using the favicon, I can use the detailed squid logo if needed.


creepertron95:
@runxel I wanted your opinion of if I should the current favicon logo or should I use the squid present in their logo?

runxel:
@creepertron95 yeah good question!
I'm not familiar with the brand _at all_, so I can't tell it for sure. (Like, how well is brand recognition).

However, the brand guideline is quite clear on that on the other hand:

> Should be used with our URL in browsers or a substitute for the squid icon when its below 50 pixels.

We always try to respect a brand guideline as well as possible. And since our "default size" is indeed below 50 pixels _we should_ follow the guide. Due to this I'm personally totally okay with having the simpler icon to be included.

runxel (commented):
Everything looks spot on. Thanks for the work, @creepertron95! 💯

PR #3560, opened 2020-09-13
Add Lospec

creepertron95:
lospec

**Name:** Lospec
**Website:** https://lospec.com/
**Alexa rank:** [318,564](https://www.alexa.com/siteinfo/lospec.com)
  

### Checklist

### Description
 The hex value is the logo color from the branding page. The background color is (332F35). The logo was from the branding page as svg format. The usage section of the branding page says,

> The Logo and logomark should be placed on a dark background, preferably our background color, but can be reversed if needed. As the logo is essentially pixel art, it's best if it's resized to an exact multiple of 100% using a nearest-neighbor algorithm to prevent blurring.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -3606,6 +3606,11 @@
             "source": "https://loop.frontiersin.org/"
         },
         {
+            "title": "Lospec",
+            "hex": "EAEAEA",
+            "source": "https://lospec.com/brand"
+        },
+        {
             "title": "Lua",
             "hex": "2C2D72",
             "source": "https://www.lua.org/docs.html"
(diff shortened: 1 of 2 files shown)


runxel (changes requested):
Hi, @creepertron95!
Your icon is sooooo slightly off 😅 
lospec

creepertron95:
@runxel Well then. How can that be fixed? I used the original svgs from their branding page and rescaled them and kept the exact dimensions when scaling. Also, what should I do about the usage section of their branding page?

runxel:
Mhm, maybe one of us made an error while rescaling...?
OR, most possible it's a problem with the precision of SVGO. Try to up it first :)

runxel:
Regarding the brand guide:
Well... The usage is not in our hands, so the only other way I see would to not include it at all. Don't worry too much.

runxel (commented):
Thanks for your work on this one, @creepertron95! 💯

PR #3572, opened 2020-09-17
Focus the search input on load

### LINKED ISSUE #1982: [website] Input focus
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [ ] The Packagist Package
  - [x] The website
  - [ ] Other, namely: 

### Description

The input focus should be automatically set to the search bar after load.
Also it should be designed better. Maybe put it in the middle and make it bigger? :)

ericcornelissen:
This updates the website source so that the search input is focused on load. (this cannot be achieved with [the HTML autofocus attribute](https://www.w3schools.com/tags/att_input_autofocus.asp) because the field is disabled by default in case JavaScript is disabled in the browser).

Closes  #1982

### DIFF
--- a/site_script.js
+++ b/site_script.js
@@ -17,6 +17,7 @@
   // Remove the "disabled" attribute from the search input
   $searchInput.setAttribute('title', 'Search Simple Icons');
   $searchInput.removeAttribute('disabled');
+  $searchInput.focus();
 
   // include a modified debounce underscorejs helper function.
   // see


runxel:
Ah, that was an easy fix. Good to know that the HTML attribute wouldn't work! 
Thanks for the PR, @ericcornelissen!

PeterShaggyNoble:
Ah, so that's who did that! Autofocus is evil! 😝

PR #561, opened 2017-09-21
Updated Netflix icon.

davidklebanoff:
Issue #369.

PR #3669, opened 2020-10-07
Correct Netflix Icon

service-paradis:
image

**Issue:** #3169
**Alexa rank:**
  

### Checklist

### Description


mondeja:
Hello @service-paradis. I can see that you've updated the icon to fit with latest Netflix icon version, maybe manually (I'm not sure). I can see next difference between previous Netflix icon in simple-icons and your one (your version in red and my version correcting size from previous Netflix icon in simple-icons in black) at the bottom-right corner of the left rectangle of the "N" symbol:

netflix-diff

Checking [Netflix brand guidelines](https://brand.netflix.com/en/assets/brand-symbol/) I can see that the symbol icon has been updated in brand guidelines after this icon was updated in #561, 3 years ago. So my question is: what file are you using as source in this update? If you are using the symbol shared by Netflix brand guidelines, you should update the source in `_data/simple-icons.json`, otherwise I think that we should go ahead with official brand symbol in this update.

service-paradis:
Hum... from memory I took the simple-icon original for this one. Looks like something went wrong, sorry!
When I compared my logo with the previous version, I though it looks pretty similar
image
I will give it another try starting from the original again!

Maybe another issue could be open to update the icon using the new guideline. 
Unfortunately, it looks like it will be difficult to respect this new guideline!

service-paradis:
@mondeja 
I gave it another try! Is it better? :)
image

mondeja (approved):
Now seems correct to me :+1: Thanks for the quick reply and the fix @service-paradis. I will open an issue for update the Netflix icon using current brand guidelines.

PR #3688, opened 2020-10-08
Correct Next.js icon

service-paradis:
image

**Issue:** #3169
**Alexa rank:**
  

### Checklist

### Description
- I took the original SVG.


ericcornelissen (approved):
Looking good @service-paradis, there is a small difference between the previous version and the correct version (see ⬇️) but I'm going to make the judgement call that this is an improvement. If anyone disagrees, feel free to comment here 🙂 

comparison

PR #3750, opened 2020-10-20
Proposal: Check outlying precision while running linter

service-paradis:
**Issue:** #3120, #3169

### Description

_**Only a proposal and some adjustments might be needed.**_

I added a custom lint check to verify the average precision level. I report an error and update the ignored list if the average precision level is below or above a defined treshold. I used the values that were indicated in #3120. The error reported looks like this:
`Precision level should be between ${iconMinFloatPrecision} and ${iconMaxFloatPrecision}; the average is currently ${precisionAverage}`

Ex.
image


.svglintrc.js
```diff
@@ -131,6 +133,23 @@ module.exports = {
               }
             }
           },
+          function(reporter, $, ast) {
+            reporter.name = "icon-precision";
+
+            const iconPath = $.find("path").attr("d");
+            const decimalRegex = /\d+\.(\d+)/g;
```
mondeja, line 140: What about [exponential notation](https://www.w3.org/TR/SVG11/paths.html#PathData)? The icon `uploaded.svg` has `5e-3` as last number in his path.

service-paradis: You're right, it is not relly optimized. I should have make this a draft instead of add proposal to the title.
Wanted to open the discussion about it before, and kind of create a POC.

mondeja: I personally agree with this lint rule.

service-paradis: Thanks for the feedback :)
I have been looking a little bit more at svgpath package. It is surely a great lib to apply transformations, but I dont know if it can be used to retrieve decimal information. Would need to take a look.
I also check the current icons for the exponential notation. I think, `uploaded.svg` is the only one right now (does not mean we should not take it into account though) and coincidally, `5e-3` respect the precision (`0.005`).
Maybe we could detect this with another regex.

mondeja: Take into account that exponential notation can be defined with uppercase `E` following [SVG 1.1 spec](https://www.w3.org/TR/SVG11/paths.html#PathData). The ideal regex would fit with next BNF grammar rules:

```
nonnegative-number:
    integer-constant
    | floating-point-constant
number:
    sign? integer-constant
    | sign? floating-point-constant
integer-constant:
    digit-sequence
floating-point-constant:
    fractional-constant exponent?
    | digit-sequence exponent
fractional-constant:
    digit-sequence? "." digit-sequence
    | digit-sequence "."
exponent:
    ( "e" | "E" ) sign? digit-sequence
sign:
    "+" | "-"
digit-sequence:
    digit
    | digit digit-sequence
digit:
    "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
```

mondeja: Hello again @service-paradis. I have take another look on svgpath library and can be used to parse correctly all the points. Check this code:

```javascript
"use strict"

const parsePath = require("svgpath/lib/path_parse");

const path = "m14.651 13.862c0 2.5477-2.0653 4.613-4.613 4.613-2.5477 0-4.613-2.0653-4.613-4.613 0-6.1507 9.226-6.1507 9.226 0zm3.497 6.04c-1.33 1.334-3.109 3.957-8.119 4.095-5.451 0.148-9.959-4.484-9.978-10.282-0.01-2.945 0.774-4.659 2.82-6.792 1.224-1.278 7.004-6.923 7.004-6.923 2.052 2.052 0.96 4.873-0.108 5.93l-3.732 3.705c-0.47 0.465-1.968 1.925-1.931 4.444 0.045 3.202 2.51 5.85 5.799 5.8 2.632-0.042 3.756-1.356 4.376-1.98l3.841-3.828c1.103-1.093 3.77-2.027 5.829 5e-3z";
console.log(parsePath(path))
```

<details>
<summary>Output</summary>

```
{
  err: '',
  segments: [
    [ 'M', 14.651, 13.862 ],
    [
      'c',    0,
      2.5477, -2.0653,
      4.613,  -4.613,
      4.613
    ],
    [
      'c',     -2.5477,
      0,       -4.613,
      -2.0653, -4.613,
      -4.613
    ],
    [ 'c', 0, -6.1507, 9.226, -6.1507, 9.226, 0 ],
    [ 'z' ],
    [ 'm', 3.497, 6.04 ],
    [
      'c',   -1.33,
      1.334, -3.109,
      3.957, -8.119,
      4.095
    ],
    [
      'c',     -5.451,
      0.148,   -9.959,
      -4.484,  -9.978,
      -10.282
    ],
    [
      'c',    -0.01,
      -2.945, 0.774,
      -4.659, 2.82,
      -6.792
    ],
    [
      'c',    1.224,
      -1.278, 7.004,
      -6.923, 7.004,
      -6.923
    ],
    [
      'c',   2.052,
      2.052, 0.96,
      4.873, -0.108,
      5.93
    ],
    [ 'l', -3.732, 3.705 ],
    [
      'c',   -0.47,
      0.465, -1.968,
      1.925, -1.931,
      4.444
    ],
    [
      'c',   0.045,
      3.202, 2.51,
      5.85,  5.799,
      5.8
    ],
    [
      'c',    2.632,
      -0.042, 3.756,
      -1.356, 4.376,
      -1.98
    ],
    [ 'l', 3.841, -3.828 ],
    [
      'c',    1.103,
      -1.093, 3.77,
      -2.027, 5.829,
      0.005
    ],
    [ 'z' ]
  ]
}
```

</details>

I've used the path of `uploaded.svg` icon and you can see that the latest point in output is `0.005`. So I think that you can use svgpath library to parse the points and that it is the correct library to use because the [parser module](https://github.com/fontello/svgpath/blob/3b938bd9d423aaf377d76752d0df96edac28399c/lib/path_parse.js) is written at low level.

service-paradis: You're absolutely right! Thanks for the feedback! I will definitely take a look at svg path to improve this rule. I will try to push changes soon :)

.svglintrc.js
```diff
@@ -129,6 +132,36 @@ module.exports = {
...
+            const countDecimals = (num) => {
+              let text = num.toString();
+              if (text.indexOf('e-') > -1) {
+                let [base, trail] = text.split('e-');
+                let elen = parseInt(trail, 10);
+                let idx = base.indexOf('.');
+                return idx == -1 ? 0 + elen : (base.length - idx - 1) + elen;
+              }
+              let index = text.indexOf('.');
+              return index == -1 ? 0 : (text.length - index - 1);
```
mondeja, line 151: Looks good, but doesn't work for exponential positives numbers right? I possible that someone could add a exponential positive in a path? I don't know. I have been experimenting with this function for supporting exponential positive numbers:

```javascript
if (num && num % 1) {
    let [base, op, trail] = num.toExponential().split(/e([+-])/);
    let elen = parseInt(trail, 10);
    let idx = base.indexOf('.');
    return idx == -1 ? elen : base.length - idx - 1 + (op === '+' ? -elen : elen);
  }
  return 0;
```

.svglintrc.js
```diff
@@ -3,12 +3,15 @@ const fs = require('fs');
...
 const getBounds = require("svg-path-bounding-box");
+const parsePath = require("svgpath/lib/path_parse");
 
 const titleRegexp = /(.+) icon$/;
 const svgRegexp = /^<svg( [^\s]*=".*"){3}><title>.*<\/title><path d=".*"\/><\/svg>\r?\n?$/;
 
 const iconSize = 24;
 const iconFloatPrecision = 3;
+const iconMinFloatPrecision = 2;
+const iconMaxFloatPrecision = 5;
```
mondeja, line 14: Maybe this variables can be called `iconMinFloatPrecisionAverage` and `iconMaxFloatPrecisionAverage`?

.svglintrc.js
service-paradis: LGTM! Surely more flexible this way!

.svglintrc.js
```diff
@@ -129,6 +132,34 @@ module.exports = {
...
+                let elen = parseInt(trail, 10);
+                let idx = base.indexOf('.');
+                return idx == -1 ? elen : base.length - idx - 1 + (op === '+' ? -elen : elen);
+              }
+              return 0;
+            };
+            const precisionArray = segmentParts.map(countDecimals);
+            const precisionAverage = precisionArray && precisionArray.length > 0 ?
+              Math.round(precisionArray.reduce((prev, curr) => prev + curr) / precisionArray.length) :
+              0;
```
mondeja, line 154: Why do you need this? :thinking:

```suggestion
            const precisionAverage = Math.round(precisionArray.reduce((prev, curr) => prev + curr) / precisionArray.length);
```

service-paradis: It surely is a edge case, but it is to prevent errors if the path is invalid. With the proposal, it gives an error like the following:
image
The invalid path is already catch by another reporter and the error coming from the precision is not really useful for the end user.

If I keep the code as-is, this is what I get with an invalid path:
image
Not exactly the truth, but better than a exception I think!

What do you think?

mondeja: I see, perfect! :+1:

package.json
```diff
@@ -25,6 +25,7 @@
     "svg-path-bounding-box": "1.0.4",
     "svglint": "1.0.6",
     "svgo": "1.3.2",
+    "svgpath": "^2.3.0",
```
ericcornelissen, line 28: Can you pin this new (direct) dependency to `2.3.0`?

```suggestion
    "svgpath": "2.3.0",
```

ericcornelissen (commented):
Thanks for the detailed review so far @mondeja and thanks for the suggestion @service-paradis. I definitely like the upper bound to ensure small file sizes (whether 5 is the correct value is debatable but that can be changed if needed). I'm not so font of the lower bound, if an icon can work with a precision of `0` or `1` that's great! Unfortunately, I think we will need to leave icons with a precision that is too low to manual reviews... Thoughts?

Two question

- Is "average precision" the right choice here? Shouldn't we be looking at the maximum precision?
- I'm surprised by the number of ignored cases in `.svglint-ignored.json`, in  #3120 (and  #3169) there are only 38 icons listed. I'm not sure how we got the lists there, maybe that was done manually... Anyway, any idea why there is a discrepancy?

mondeja:
> I think we will need to leave icons with a precision that is too low to manual reviews... Thoughts?

I agree with you, @ericcornelissen. Discrepances in 0.01 pixel ranges and lower are easy to check reviewing manually and can be uploaded an icon with path fitting entirely in integer points.

> - Is "average precision" the right choice here? Shouldn't we be looking at the maximum precision?
> - I'm surprised by the number of ignored cases...

- I'm also inclined to try the maximum number of decimal places check instead of the mean.
- I was confusing about this discrepance because of #3120 from #3107 discussion. Perhaps the list, as you have pointed, was manually created.

package.json
service-paradis: Sure :)

service-paradis:
Thanks for the feedback!
- I pinned svgpath version
- I use only the maximum precision level instead of the average
- I validate only if it is greater than the defined maximum

The number of ignored icon is way less this way!

.svglintrc.js
```diff
@@ -129,6 +131,34 @@ module.exports = {
               }
             }
           },
+          function(reporter, $, ast) {
+            reporter.name = "icon-precision";
+
+            const iconPath = $.find("path").attr("d");
```
mondeja, line 137: Hey @service-paradis, the icons ignorance is missing.

```suggestion
            const iconPath = $.find("path").attr("d");
            if (!updateIgnoreFile && isIgnored(reporter.name, iconPath)) {
              return;
            }
```

.svglintrc.js
```diff
@@ -129,6 +131,34 @@ module.exports = {
...
+            const { segments } = parsePath(iconPath);
+            const segmentParts = segments.flat().filter((num) => (typeof num === 'number'));
```
ericcornelissen, line 139: It seems that this custom check may fail here on NodeJS v10 or lower because `.flat()` is not a function. [The documentation for `.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) suggests that we should be safe with NodeJS v11. So we need to at least update [`.travis.yml`](https://github.com/simple-icons/simple-icons/blob/develop/.travis.yml) to use `node_js: 11`. Can you update the `.travis.yml` file and push a change that should make Travis fail on this new check to verify that it does?

I'm also thinking of adding this detail to the Contributing Guidelines, thoughts on this (@mondeja as well)?

.svglintrc.js
```diff
@@ -129,6 +131,34 @@ module.exports = {
...
+              }
+              return 0;
+            };
+            const precisionArray = segmentParts.map(countDecimals);
+            const precisionMax = precisionArray && precisionArray.length > 0 ?
+              Math.max(...precisionArray) :
+              0;
+
+            if (precisionMax > iconMaxFloatPrecision) {
+              reporter.error(`Precision level should not be greater than ${iconMaxFloatPrecision}; the maximum is currently ${precisionMax}`);
```
ericcornelissen, line 156: I think this should be reworded for clarity, what about:

```suggestion
              reporter.error(`Precision level should not be greater than ${iconMaxFloatPrecision}; the precision is currently ${precisionMax}`);
```

ericcornelissen (changes requested):

.svglintrc.js
ericcornelissen: Good catch! I was a little surprised this problem didn't show up with the Travis build but I figured out why. See my comments below.

.svglintrc.js
mondeja: I suggest:

```suggestion
              reporter.error(`Maximum precision level should not be greater than ${iconMaxFloatPrecision}; the precision is currently ${precisionMax}`);
```

.svglintrc.js
mondeja: All arrays returned by `parsePath` have 1 array in depth, so why not use this?

```suggestion
            const segmentParts = segments.reduce(
              (acc, val) => acc.concat(val), []
            ).filter((num) => (typeof num === 'number'));
```

ericcornelissen: We can definitely do it manually, but I don't see a problem with updating the NodeJS version as it does not affect our NPM package.

I'm fine with upgrading to either 12 or 14 as well, definitely if 11 is not an LTS release.

.svglintrc.js
ericcornelissen: I personally see "maximum precision" and "precision" (in this context) as the same thing, but I don't mind being explicit about it. The only thing I worry about is that the message is getting pretty long... How about:

- Removing "level" from the first sentence?
- Changing the second sentence to `it is currently ${precisionMax}`?

```suggestion
              reporter.error(`Maximum precision should not be greater than ${iconMaxFloatPrecision}; it is currently ${precisionMax}`);
```

.svglintrc.js
mondeja: > I'm also thinking of adding this detail to the Contributing Guidelines...

What detail, the minor version of NodeJS required to run the linting process?

.svglintrc.js
mondeja: > - Removing "level" from the first sentence?
> - Changing the second sentence to it is currently ${precisionMax}?

Totally agree.

.svglintrc.js
ericcornelissen: > What detail, the minor version of NodeJS required to run the linting process?

Exactly, and maybe something similar for Jekyll and potentially Docker? Maybe other things as well? Something along the lines of a "Prerequisites" section... I'm unsure how to approach this as none of these are "hard" prerequisites.

mondeja: I don't think if this will be necessary if we use the stable versions of Jekyll and NodeJS, maybe only will add noise. But if we add it, simply pointing in "Building locally" section the list item "Make sure you have NodeJS v12 at least installed." at first should be enough.

- Building the Docker image locally you will see that `3.12` installs nodejs (12.18.4-r0) so is not necessary to add nothing to the Docker part.
- Jekyll, installed locally without specify a version will fetch the latest stable, which is what are we using for simple-icons.

ericcornelissen: I do think it's useful to point out that NodeJS v12 is required, especially since svglint fails silently when a custom lint function errors (perhaps we should open an issue for that).

Regarding Docker, I don't imagine we use anything that is only supported in newer versions, but if so it would be useful to specify a minimum version as well. (for Docker I'm interested in whether you can build our docker image by running `docker build .`). As for Jekyll, I'm not too familiar with how people use it, but if someone still has an old version installed (for whatever reason), then they run into trouble. Although someone might figure this out from the output of Jekyll, having the Contributing guidelines just say that you need at least Jekyll v3.5 (following #3824) makes sense to me. For the same reason one would do it for NodeJS.

That said, this discussion (apart from including a minimum NodeJS version) should not block this PR from being merged!

mondeja: I've open #3959 to track this improvements separately. As for now, we can add at the start of the file this check to prevent silent failures:

```javascript
if (Array.prototype.flat === undefined) {
  console.error(`Minimum NodeJS v11.15.0 is required, but you are running ${process.version}.`);
  process.exit(1);
}
```

service-paradis: My proposal for this would be to use the `engines` property of package.json and enforcing it using a `.npmrc`. This way, people wont be able to run `npm i` if the required version is not used.
I just commited the proposal.. Let me know what you think!

.svglintrc.js
service-paradis: English is not my native language as you can see 😛

.svglintrc.js
ericcornelissen: > As for now, we can add at the start of the file this check to prevent silent failures [...]

Though not ideal, I'm not against it. In the meantime I opened https://github.com/birjolaxew/svglint/issues/20.

ericcornelissen: > My proposal for this would be to use the engines property of package.json and enforcing it using a .npmrc. This way, people wont be able to run npm i if the required version is not used.
I just commited the proposal.. Let me know what you think!

I'm not 100% of this, but I think this also affects projects that install us as a dependency and the minimum NodeJS version for that is lower than v12 (not sure exactly what it is, I think the newest thing we use in the package is [`Object.defineProperty`](https://github.com/simple-icons/simple-icons/blob/2f5b6ee307479252c8b00b66999abc14c009ebf8/scripts/templates/index.js#L3)).

package.json
```diff
@@ -7,6 +7,11 @@
     "svg",
     "icons"
   ],
+  "engines": {
+    "node": ">=12.0.0",
+    "npm": ">= 6.0.0",
+    "yarn": ">= 1.15.0"
```
mondeja, line 13: Why these two engines?

.svglintrc.js
mondeja: > I'm not 100% of this, but I think this also affects projects that install us as a dependency and the minimum NodeJS version for that is lower than...

Yes, I've created a package as a test for `node: ">=16.0.0"` engine with `strict-engine=true` in `.npmrc` and this is the error trying to install it as dependency:

```
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: a-project-with-rare-name@1.0.0
npm ERR! notsup Not compatible with your version of node/npm: a-project-with-rare-name@1.0.0
npm ERR! notsup Required: {"node":">=16.0.0"}
npm ERR! notsup Actual:   {"npm":"7.0.8","node":"v15.2.1"}
```

> not sure exactly what it is...

- `Object.defineProperty` is [available since 0.12.18](https://node.green/#ES2015-built-ins-Symbol-Object-defineProperty-support).

service-paradis: Oh, no problem! I will remove it and use @mondeja proposal for now.

ericcornelissen: Alright :+1:  Now [Travis is failing because the NodeJS version is too low](https://travis-ci.com/github/simple-icons/simple-icons/jobs/443701618#L230), as expected. Can you update `.travis.yml` @service-paradis?


> Object.defineProperty is available since 0.12.18.

Hmm... :thinking: Then my memory is failing me :sweat_smile: I recall using a NodeJS feature that was fairly new at the time (like one or two LTSes before the latest at the time)... Anyway, this shouldn't block this PR. Would be nice if we _can_ specify a minimum NodeJS version in `package.json` at some point though.

.svglintrc.js
```diff
@@ -1,14 +1,21 @@
+if (Array.prototype.flat === undefined) {
```
ericcornelissen, line 1: Minor thing, but could you move this right before the `if (updateIgnoreFile) {` line?

.svglintrc.js
mondeja: > Would be nice if we can specify a minimum NodeJS version in package.json at some point though.

I've included it in the TODO list of #3959 :+1:

.svglintrc.js
service-paradis: Sure!

.svglintrc.js
service-paradis: > Alright 👍 Now [Travis is failing because the NodeJS version is too low](https://travis-ci.com/github/simple-icons/simple-icons/jobs/443701618#L230), as expected. Can you update `.travis.yml`

Done :)

ericcornelissen: Awesome, now [this build](https://travis-ci.com/github/simple-icons/simple-icons/jobs/443847121#L234) also shows that the check is able to fail on Travis. Thanks 👍

ericcornelissen (approved):

mondeja (approved):

.svglintrc.js
service-paradis: Was pushing commits step by step to be sure :D

PR #3818, opened 2020-11-06
Correct curl icon

service-paradis:
image

**Issue:** #3169
**Alexa rank:**
  

### Checklist

### Description


ericcornelissen:
Hmm, it looks like your update slightly changed the shape of the logo, as you can see in the comparisons below. Specifically, the bottom right circle seems to have a slightly different shape than before. Any ideas why this is the case? Did you compare your version to the source SVG?

| Full | Bottom right |
| --- | --- |
| image | image |

service-paradis:
> Hmm, it looks like your update slightly changed the shape of the logo, as you can see in the comparisons below. Specifically, the bottom right circle seems to have a slightly different shape than before. Any ideas why this is the case? Did you compare your version to the source SVG?

Hum... you're right! Sorry about this. I will take another look!

service-paradis:
I gave it another try. The SVGO tools is messing with circles when I use a precision of 3. Looking fine with a precision of 4. Do I keep it this way?
image

ericcornelissen (approved):
A precision of 4 is fine (as per #3120), thanks for the quick update @service-paradis :+1:

PR #3820, opened 2020-11-06
Correct Rust icon

service-paradis:
image

**Issue:** #3169
**Alexa rank:**
  

### Checklist

### Description


mondeja (changes requested):
I'm seeing differences in circles between the original version without optimize after size and align corrected and your version:

rust-circle-diffs


Might be an optimization problem with SVGO. Could you increase the precision optimizing to 4 @service-paradis?

service-paradis:
> Might be an optimization problem with SVGO. Could you increase the precision optimizing to 4 @service-paradis?

Hum... You're right! Looks like the precision of 3 messes with Rust circles. Just gave it a try with a precision of 4 and it looks good (I think)!
image

mondeja (approved):
It looks much better now. Thank you again @service-paradis! :+1: :tada:

PR #3823, opened 2020-11-07
Exclude unnecessary files and directories from website build

mondeja:
After build `_site` locally, if I run tests, next errors are raised:

<details>
<summary>Jest errors</summary>

```
FAIL  _site/tests/icons.test.js
 ● Test suite failed to run

   Cannot find module '../_data/simple-icons.json' from '_site/tests/icons.test.js'

   > 1 | const { icons } = require('../_data/simple-icons.json');
       |                   ^
     2 | const { titleToFilename } = require('../scripts/utils.js');
     3 | 
     4 | icons.forEach(icon => {

     at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:306:11)
     at Object.<anonymous> (_site/tests/icons.test.js:1:19)

FAIL  _site/tests/index.test.js
 ● Test suite failed to run

   Cannot find module '../_data/simple-icons.json' from '_site/tests/index.test.js'

   > 1 | const { icons } = require('../_data/simple-icons.json');
       |                   ^
     2 | const simpleIcons = require('../index.js');
     3 | const { titleToFilename } = require("../scripts/utils.js");
     4 | 

     at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:306:11)
     at Object.<anonymous> (_site/tests/index.test.js:1:19)
```

</details>

The problem is that jest discover the directory `_site/tests` and run it along with `tests`. But, we don't need the next files inside website `_site` directory:

- `tests/` -> http://simpleicons.org/tests/icons.test.js
- `scripts/` -> http://simpleicons.org/scripts/build-package.js
- `composer.json` -> http://simpleicons.org/composer.json
- `CONTRIBUTING.md` -> http://simpleicons.org/CONTRIBUTING.md
- `Dockerfile` -> http://simpleicons.org/Dockerfile
- `index.js`
- `LICENSE.md` -> http://simpleicons.org/LICENSE.md
- `package-lock.json` -> http://simpleicons.org/package-lock.json
- `package.json` -> http://simpleicons.org/package.json
- `README.md` -> http://simpleicons.org/README.md

So the solution is not to include them from website build using a [Jekyll configuration file](https://jekyllrb.com/docs/configuration/). The new `_site` tree would be:

```
_site
├── CNAME
├── icons
├── images
├── index.html
├── site_script.js
└── stylesheet.css
```

## Steps to reproduce

```bash
gem install jekyll bundler
jekyll build
npm install
npm run test
```

PR #3824, opened 2020-11-07
Remove 'index.html' empty lines to reduce built file size.

mondeja:
We can use [Liquid whitespace control](https://shopify.github.io/liquid/basics/whitespace) feature to remove empty lines created by `{% ... %}` statements in `index.html`. This reduce the output file `_site/index.html` size:

- Before: 3061507 bytes (~2.9197 MB)
- After: 2859632 bytes (~2.7272 MB)

It's not a massive amount of optimization, but will be bigger for each new icon.

> [Minimum Jekyll version required v3.5](https://stackoverflow.com/a/44663902/9167585) (we currently use v4.1.1).

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
 {%- assign iconCount = site.data.simple-icons.icons.size -%}
 {%- assign iconsUnsortedString = "" -%}
 {%- assign greyscaleIconsUnsortedString = "" -%}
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Thank you very much for the improvements here and in #3823 :smiley:

PR #3843, opened 2020-11-09
Allow exponential notation with E uppercase inside icon paths.

mondeja:
I just realized that [SVG 1.1 Specification](https://www.w3.org/TR/SVG11/paths.html#PathData) allows exponential notation using `E` uppercase inside paths, so we can allow it extending #2946.


ericcornelissen (approved):

ericcornelissen:
Good catch, thanks for updating this @mondeja :smiley:

PR #3847, opened 2020-11-10
Improve TravisCI configuration

mondeja:
This pull optimize all the TravisCI jobs, especially the "Build website":

1. Removed root `language` because is explicitly declared in jobs.
2. Unify stage names, both Capitalized.
3. Add `git.depth: 1` to reduce `git clone` command time for all jobs ~~except "Git tag" (I'm not sure if there could be risky)~~.
4. Add ~~`deploy.skip_cleanup: true`~~ `deploy.cleanup: false` for "Git tag" job because there is nothing to clean (or I'm missing something? :thinking:). I couldn't tested this. EDIT: Added it also for "NPM package".
5. Upgrade Ruby version with preinstalled TravisCI `2.5.3` in "Build website" job.
6. Cache Ruby installation in "Build website" job as [recommended by TravisCI](https://docs.travis-ci.com/user/caching/#cache-rvm-ruby-version-for-non-ruby-projects). The cache [will expire in 45 days](https://docs.travis-ci.com/user/caching/#caches-expiration). Since we are not freezing gem versions in a Gemfile, this is perfect for our Jekyll use case. This is the greatest optimization in this pull, as you can see in [this job](https://travis-ci.com/github/mondeja/simple-icons/jobs/432363573), taking about the same time as the other jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,18 @@ jobs:
       name: "Lint"
       language: node_js
       node_js: 10
+      git:
+        depth: 1
       script:
         - npm run jsonlint
         - npm run svglint
         - npm run wslint
         - npm run our-lint
-      git:
-        depth: 1
     - name: "Build website"
       language: ruby
       rvm: 2.5.3
+      git:
+        depth: 1
       cache:
         directories:
           - /home/travis/.rvm/
@@ -23,15 +25,13 @@ jobs:
         - gem install jekyll
       script:
         - jekyll build
-      git:
-        depth: 1
     - name: "Test package"
       language: node_js
       node_js: 10
-      script:
-        - npm run test
       git:
         depth: 1
+      script:
+        - npm run test
 
     - stage: "Deploy"
       name: "Git tag"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 os: linux
-language: node_js
 
 jobs:
   include:
@@ -12,20 +11,29 @@ jobs:
         - npm run svglint
         - npm run wslint
         - npm run our-lint
+      git:
+        depth: 1
     - name: "Build website"
       language: ruby
-      rvm: 2.4.1
+      rvm: 2.5.3
+      cache:
+        directories:
+          - /home/travis/.rvm/
       install:
         - gem install jekyll
       script:
         - jekyll build
+      git:
+        depth: 1
     - name: "Test package"
       language: node_js
       node_js: 10
       script:
         - npm run test
+      git:
+        depth: 1
 
-    - stage: deploy
+    - stage: "Deploy"
       name: "Git tag"
       language: shell
       if: branch = master
@@ -38,9 +46,12 @@ jobs:
       deploy:
         provider: releases
         token: "$GITHUB_TOKEN"
+        skip_cleanup: true
     - name: "NPM Package"
       language: node_js
       node_js: 10
+      git:
+        depth: 1
       if: branch = master
 
       deploy:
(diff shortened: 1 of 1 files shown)


ericcornelissen:
> Add `git.depth: 1` to reduce `git clone` command time for all jobs except "Git tag" (I'm not sure if there could be risky).

What do you mean by "I'm not sure if there could be risky"? That it could be risky to add `git.depth: 1` to the "Git tag" job? I'm not sure why that would be risky 🤔 


> Add `deploy.skip_cleanup: true` for "Git tag" job because there is nothing to clean (or I'm missing something? 🤔). I couldn't tested this.

I removed this in #2536 because Travis says that the `skip_cleanup` key is deprecated. You can see this if you check the warnings under the "config" tab on the build for this PR. But it seems the documentation for deploying still says you can use it (the alternative `cleanup` in under the "config" tab didn't work back when I did #2536).

Soo yea, I'm not sure what to do with this. If I remember correctly, setting `skip_cleanup: true` will/used to prevent something along the line of `git reset --hard` from being run before the deploy. This won't have any effect in either of the deploy jobs. Since Travis says its deprecated, and cleaning shouldn't be too expensive anyway I would say we don't use `skip_cleanup`. What do you think?


> Cache Ruby installation in "Build website" job [...]

Awesome, good thinking 👍

mondeja:
> What do you mean by "I'm not sure if there could be risky"? That it could be risky to add git.depth: 1 to the "Git tag" job? I'm not sure why that would be risky

I did not dare to add it because I'm not sure how the releases deploy provider works. Has something in their behaviour related with previous commits of the branch?

> Since Travis says its deprecated, and cleaning shouldn't be too expensive anyway I would say we don't use skip_cleanup. What do you think?

In TravisCI deployment v2 (still in beta) [is replaced by `cleanup`](https://docs.travis-ci.com/user/deployment-v2#cleaning-up-the-git-working-directory) and that says the warnings of [the job](https://travis-ci.com/github/simple-icons/simple-icons/builds/147690404/config) that you have posted at point number 1 in #2536. I will try with `cleanup: false` and checks if Travis generates any warnings.

mondeja:
Now [it doesn't raises any warning](https://travis-ci.com/github/simple-icons/simple-icons/builds/200317156/config). I've added to "NPM Package" job also because I've seen that you added it in previous TravisCI configuration versions.

ericcornelissen:
> I did not dare to add it because I'm not sure how the releases deploy provider works. Has something in their behaviour related with previous commits of the branch? EDIT: Definitively you are right, it's not "risky" at all. I will add to that job also.

Alright, glad to have that resolved :slightly_smiling_face: 

> In TravisCI deployment v2 (still in beta) is replaced by cleanup and that says the warnings of the job that you have posted at point number 1 in #2536. I will try with cleanup: false and checks if Travis generates any warnings.
>
> Now it doesn't raises any warning. I've added to "NPM Package" job also because I've seen that you added it in previous TravisCI configuration versions.

Thanks for the investigation and trying this out. Even though [the documentation](https://docs.travis-ci.com/user/deployment-v2#cleaning-up-the-git-working-directory) says the default is to do no cleanup, having it explicit is nice I think :+1:

ericcornelissen (changes requested):
One minor nitpick, could you change this so that the "git" key always comes after the `node_js`/`rvm` key (just like in the "NPM Package" job) (the "Git tag" job is good already)

mondeja:
It's done

ericcornelissen (approved):

ericcornelissen:
Thank you very much for this change @mondeja 🎉

PR #3866, opened 2020-11-16
Correct GNU icon

mondeja:
gnu

**Issue:** #3169

### Checklist


fbernhart:
@mondeja Perfect! 🚀 💯 

In case we'll ever be in need of an official SVG version to compare against, I've found it here: https://www.gnu.org/graphics/empowered-by-gnu.svg

PR #3867, opened 2020-11-16
Update labeler.yml

ericcornelissen:
**Issue:** n/a
**Alexa rank:** n/a

### Checklist

n/a

### Description

Run [the labeler workflow](https://github.com/simple-icons/simple-icons/blob/develop/.github/workflows/labeler.yml) only when a Pull Request is opened.


mondeja (approved):
Definitively, this should solve the problem. If an user changes another icon adding one or adds another icon updating one we configure the labels manually :+1:

PR #3894, opened 2020-11-18
Updated CRYENGINE

fbernhart:
cryengine

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja:
Hey @fbernhart, the width of your version has been reduced to 10,835, but this icon only needs to be centered. Can you give it another shot?

fbernhart:
@mondeja That has been a problem of SVGOMG. I centered it, optimized it (precision of 3) and then saw in Inkscape, that through the optimization the icon has changed from 24.000 to 24.016 in width. So I had to resize it a bit. With a precision of 4 the same effect still occured.

 I've now set the precision to 5 and after the optimization the icon now keeps its width of 24.000.

mondeja (approved):
Now it's perfect! Thank you @fbernhart :muscle:

PR #3905, opened 2020-11-18
Updated HappyCow

fbernhart:
happycow

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (commented):
Seems that the logo it's outdated in the source (your path is the red in comparison). You want to update it in this pull or I will review as is and we open another for the update?

image

fbernhart:
happycow


@mondeja I've rebuilt it now from the source, using the new logo. And gave it a kind of different approach. 😁 

Most of the icon is filled now. And I've separated the color changes with gaps (width 0,35px). What do you think?

mondeja:
> I've rebuilt it now from the source, using the new logo. And gave it a kind of different approach.

Surely, but seems that you forgot to push... or I'm going crazy? :smile:

fbernhart:
@mondeja  haha, I'm sorry. I've pushed now ;)

mondeja:
I think that, although your treatment about what geometries must be filled is good, we must respect as far as we can the original source. Because of that, maybe choose a gap width based on our fancy could not be the best choice if the original icon for the brand has stroke geometries, which is the case, and we could incur a deformation of the brand symbol.

respect-width-happycow

Have you checked if the original stroke width looks fine in small sizes?

fbernhart:
@mondeja Thanks for your feedback I'll try it again with the original width and see how it looks like. 👍🏼

fbernhart:
happycow

@mondeja How about this version? I've now used the original strokes to create the gaps. Unfortunatelly a precision of 3 didn't make it through the linter.

mondeja (approved):
It looks perfect now :1st_place_medal:
Nice work @fbernhart!

PR #3909, opened 2020-11-18
Update IconJar

fbernhart:
iconjar

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (commented):
The press-kit of the source (https://geticonjar.com/press-kit/) is down and the website is out of scope for simple icons according to Alexa Rank ([~600k](https://www.alexa.com/siteinfo/geticonjar.com)). So I think that this icon should be removed and mark this pull as a breaking change for the next major release. What do you think?

fbernhart:
@mondeja Done.

PeterShaggyNoble:
Hmm ... surprised to see this one with such a low Alexa rank; perhaps being Mac-only hampers it. However, we should try to see if we can find another metric to judge it by in the hopes of keeping it because, up until now, I would have said it was a pretty well-known piece of software. Maybe that's just my own bias from working with and contributing to icon libraries for years or maybe people have simply gravitated towards other, newer software.

Removing this also sets a precedent for retroactively rejecting icons that no longer meet our requirements. On the whole, I'm OK with that; brands have their day in the sun and then fall out of favour all the time. But I think we should set some parameters to firstly _consider_ an icon for removal on these grounds and then to actually remove it as there may be extenuating contributing factors we're not aware of accounting for a temporary slump in its rank - we don't want to be removing a brand only to see it bounce back a couple of releases later and having to re-add it.

PeterShaggyNoble:
Looks like this has had a resurgence; it's rank has been back up over 500k for about a month now.

PeterShaggyNoble:
Took the liberty of reverting this PR, @fbernhart, hope you don't mind. Can you double check everything is as it should be and then take it out of draft?

fbernhart:
@PeterShaggyNoble Thanks. 👍🏼  I've now solved the merge conflict and it's ready to be reviewed.

But as the press kit is still down, there isn't any current version of the logo to compare against.

PeterShaggyNoble:
> But as the press kit is still down, there isn't any current version of the logo to compare against

Well, as we were merely centring it and, other than that, it matches up with our existing icon I think we can say its OK. As for the source: with the press kit gone maybe use the homepage instead? It at least includes the treatment we use as a favicon.

PeterShaggyNoble:
Updated the source so merging this in.

PR #3910, opened 2020-11-18
Updated Kaspersky

fbernhart:
kaspersky

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (commented):
I'm seeing a very small difference repect to [this localized source (es_ES)](https://www.kaspersky.es/content/es-es/images/homepage/kaspersky-logo.svg). Also, the color is different in this source (`#006D5C`). I don't know what was the exact source used. Can you check this @fbernhart?

fbernhart:
@mondeja Yeah, there is a minimal difference.

If I'm visiting https://www.kaspersky.com it's redirecting me to https://www.kaspersky.de (the German website), which has the following SVG as header:

https://service.maxymiser.net/cm/images-eu/kaspersky-com/A386D9BFC296DBD6F507D510782B63B2D90178927B0784DEE02C9F16DC3DB5DA.svg

The SVGs do look like the same. But the one in the header of the German website uses `#00A88E` as color. 


Acessing kaspersky.com through a webproxy from the US, I'm seeing the same color as on the Spanish kaspersky.es site. This version even seem to be hosted on the German site, as https://www.kaspersky.de/content/de-de/images/homepage/kaspersky-logo.svg returns the same SVG file. But they don't use it (yet?).

I guess the .es and .com version are already using the new color while the .de version is still stuck with the old one. So I'd be in favour of changing it to `#006D5C`.

fbernhart:
Feel free to merge it, if you agree with the color.

kaspersky

mondeja (approved):
Thanks for the investigation @fbernhart! :+1:

PR #3927, opened 2020-11-18
Updated Qi

fbernhart:
qi

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (changes requested):
I'm seeing slight differences between your version (red) and current source (black) using both versions (`.ai` and `.eps`), so it seems that the icon has been updated. Could you rebuild from source @fbernhart?

image

fbernhart:
@mondeja I couldn't get it down to a precision of 3. Only 4 worked. But the icon itself should be ok now. 😊

mondeja (approved):
We can go ahead with this. Thank you @fbernhart :+1:

PR #3940, opened 2020-11-18
Updated Sourcegraph

fbernhart:
sourcegraph

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis:
Thanks again for your work 😄 
It looks a little bit different than the original from https://sourcegraph.com/.assets/img/sourcegraph-mark.svg (or https://github.com/sourcegraph/about/blob/main/_resources/assets/sourcegraph-mark.svg)
image

fbernhart:
I'll rebuild it from scratch by tomorrow. 👍🏼

Funny how some of the icons are a tiny bit different from the original ones.

As I've only done some alignment and scaling (with the same scale factor for height and width of course) to all those icons, I can't really explain where the differences come from.

1) Have the icons been a bit offset ever since they were added to simple-icons?

2) Have the brand owners very slightly changed their icons? (Which would be weird)

3) Or does this happen while optimizing the icons and setting precision to 3 to 5. (Which would be bad)

I'm hoping it's 1) 😄

service-paradis:
It is probably a mix of those 😛 

Some of the icons were added 2 years ago (and maybe more). The review process was probably different, and maybe those icons have been merged this way.
Some of them have been updated. For example, this one (Sourcegraph) have been slightly updated in July this year according to https://github.com/sourcegraph/sourcegraph/issues/12008.
Could also be, from time to time, the optimisation process.

fbernhart:
Ok, good to know.

At least it seems like it wasn't me who messed it up. 😂👍🏼

fbernhart:
@service-paradis Done. I've rebuilt it from https://sourcegraph.com/.assets/img/sourcegraph-mark.svg.

I've as well changed the source to https://sourcegraph.com/.assets/img. I guess we can always find the newest logo there, even if they rename the file.

service-paradis (approved):
Looking good, thanks 💯

PR #3942, opened 2020-11-18
Updated TuneIn

fbernhart:
tunein

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis:
Hello @fbernhart 
I think they might have updated their logo since I'm not able to find this one on their website. Looks like they now uses full words `TUNE IN`. Maybe I'm missing something. What do you think?

fbernhart:
@service-paradis 

I would go for the logo with the full word mark then.

Seems like it's used more often than the old one in the favicon.

If I visit [their website](https://tunein.com/) on my phone, I can see it in the header, the side menu and as advertisement for their app.



I would as well go for that kind of turquoise (in the logo) as color.

Do you agree?

service-paradis:
I think I would use the full wordmark logo as well. There are 2 primary color used in their website `#14D8CC` and `#1C203C`. I agree with you for the turquoise color. Maybe we can ask another @simple-icons/maintainers their opinion?

mondeja:
I also agree with `#14D8CC`

NovaGL:
Hmm. I think there is already a PR for this

service-paradis:
> Hmm. I think there is already a PR for this.
> How is this different to this #3115

According to #3169, we're updating every icons that do not satisfy the guidelines. For example, the current icon for tunein is not centered
image

While reviewing, we also check if we need to update those icons.

NovaGL:
Thanks for that. Maybe we should add a tag or something if it's just a positioning change not logo change

service-paradis:
> The logo and colour is the same so why are you changing it?

Look like they always use full wordmark as their logo. The only place where I see the initials is in the favicon.
Same for `14D8CC`. This color does not seem to be widely used on their website.
Maybe I'm missing something? Is there some official sources I'm not aware of?

NovaGL:
The favicion and sonos app uses it. I don't feel we should change the logo for the sake of it. This change seems over zealous.

ericcornelissen:
I agree with @NovaGL that this PR **can** be merged as is since it fixes the centering of the existing icon.

That said, I have no strong opinions on whether we should switch to the full wordmark. Them using it for favicons is an indication that they prefer it for small sizes (and 12x12 is pretty small) (not sure about the size of icons in the Sonos app). But on the other hand, if many users don't instantly recognize this version then arguably we should switch to the more recognizable full wordmark.

So, we can also update to the full wordmark here if @fbernhart is open to doing that and if most @simple-icons/maintainers agree, currently it seems 2 are in favor of updating, 1 is indifferent, and 1 is against.

fbernhart:
I wouldn't mind working on a new logo (the full wordmark). Just let me know what the majority wants. 😊

NovaGL:
I use this logo for projects that require a small logo eg L for logitech and N for Netflix. I would suggest we have two varients of these logos if we want to set a precedent with this logo

ericcornelissen:
> I use this logo for projects that require a small logo eg L for logitech and N for Netflix. I would suggest we have two varients of these logos if we want to set a precedent with this logo

Simple icons provides "Free SVG icons for popular brands", we are not aiming for good UI elements. If the brand forces us to use something that doesn't look good at small sizes that is the brands problem not ours.

That said, as long as TuneIn is still using the "LN" version of the logo, it seems to me we should use that since it works better at small sizes (where they use it as well).

mondeja:
I am not very familiar with this brand and do not really understand the discussion raised here, although from what I can understand I am in favor of @ericcornelissen in using the "LN" version.

service-paradis:
They dont use the small logo a lot, but they still uses it. So lets keep this simple for now!

fbernhart:
What's the "LN version"? Are we talking about the letters "TI" like in the preview of my initial post?

I'm divided on this one.

They use the full wordmark on every part of their website (apart from the favicon) as well on [Google's](https://play.google.com/store/apps/details?id=tunein.player&hl=de&gl=US) and [Apple's](https://apps.apple.com/de/app/tunein-radio-musik-sport/id418987775) app stores.

**But** they seem to use just the "TI" in all of their social media presence. [Facebook](https://www.facebook.com/TuneIn/), [Instagram ](https://www.instagram.com/tunein/) and [Twitter](https://twitter.com/tunein). With around 770k followers in total, I would assume that people would as well recognise the brand with the "TI" logo.

So both versions seem to be frequently used by TuneIn.

ericcornelissen:
> What's the "LN version"? Are we talking about the letters "TI" like in the preview of my initial post?

Ow wow, yes I meant the "TI" version :upside_down_face: 

* * *

That sounds like some kind of a consensus to me, at least for now. We can always revisit this later and update the SVG if needed :slightly_smiling_face: At least now we don't need to ignore it in `npm run svglint` anymore. Thanks a bunch @fbernhart! :tada:

PR #3966, opened 2020-11-19
Updated AngelList

fbernhart:
angellist

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted:
Just looked at this in Figma - there's a tiny difference on the lower right side. @simple-icons/maintainers - would this class this as 'close enough'?

Capture

mondeja (changes requested):
I'm seeing significative differences rebuilding from source (red) against your version @fbernhart (black). Could you rebuild from source?

image

mondeja:
> Just looked at this in Figma - there's a tiny difference on the lower right side

Have you rebuild from source @adamrusted?

adamrusted:
> Have you rebuilt from source @adamrusted?

My bad - I was comparing the changes between the previous version, and PR version. Didn't even look at source! 😅

adamrusted:
Think it's purely to do with the optimization process. I just rebuilt from source, with Illustrator working to 3 decimals and got a different result to @mondeja (see below)

Capture

mondeja:
> Just looked at this in Figma - there's a tiny difference on the lower right side. @simple-icons/maintainers - would this class this as 'close enough'?

In theory must fit visually 100% with your version, but I would do some rare exceptions. For example, if the source is not a vector, looks without differences in design and we are updating only the size and centering (like in this case), I'm not even comparing it with the source (here yes because the source it's not a raster image). Also, could be slight differences of processing between different editors, which I would pass if they are very minimal, but i haven't faced that case yet.

> Think it's purely to do with the optimization process. I just rebuilt from source, with Illustrator working to 3 decimals and got a different result to @mondeja (see below)

:thinking: Could be, let's see what is the result of @fbernhart rebuilding from the source in this case...

fbernhart:
@mondeja @adamrusted Thanks for your input. 👍🏼 

I've now rebuilt it from the source (https://angel.co/logo --> the "hand alone" version, using the file `AngelList_Black_Victory_Hand.eps`) and it now perfectly matches, even after optimization.

Let me know it you still get a different result.

adamrusted (approved):
Yep! Spot on! 🎉

mondeja (approved):
All looks spot on now :v:
Thank you both!

PR #3969, opened 2020-11-19
Updated Apache OpenOffice

fbernhart:
apacheopenoffice

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I've rebuilt the icon from https://apache.org/logos/


fbernhart:
As the existing icon didn't pass the linter (even at a precision of 5), I've rebuilt it using the new source https://apache.org/logos/, as suggested in #2887.

If this get's merged, we can remove it from the list in #2887.

mondeja (changes requested):
I'm seeing a lot of more coordinates than necessary. See the next comparison between your version (red) and mine (blue):

aoo-diff

fbernhart:
@mondeja No, I used "Version 3", which is the logo alone.

Seems like they used different SVGs to build the different versions.

mondeja:
I think that the source I've used it's more optimized, because I don't see differences in the geometries. Would you try again using that version?

fbernhart:
@mondeja I've found the reason. The SVGs are actually the same.

I've made a mistake in Inkscape. The "birds" are actually already cut out of the blue circle in the original SVG. But they still exist as extra path. So when I created the difference between the circle (with the birds already cut out) and the birds themselves, some more points were created.

I've solved it now by just deleting the extra birds path. 😄

mondeja (approved):
It's perfect now :sparkles: Thank you @fbernhart!

Issue #3959, opened 2020-11-19
Update NodeJS version and document requisites


mondeja:
Following the discussion in #3750, we must make some improvements, both in documentation and in the testing workflow.

- [x] Upgrade NodeJS to v12 or v14 because of the maintenance of v10 [will end in May 2021](https://nodejs.org/en/about/releases/). Added in #3750
- [x] Document that minimal NodeJS version is 11.15.0 due to [missing of `Array.prototype.flat`](https://node.green/#ES2019-features-Array-prototype--flat--flatMap--Array-prototype-flat) in previous NodeJS versions. Added in #4151
- [x] Document that minimum Jekyll version required is v3.5. This is because Liquid Whitespace Control feature used in `index.html` since #3824 is supported by Liquid from v4, included as Jekyll dependency since v3.5. Added in #4116
- [x] Add `engine.node: ">=0.12.18"` to `package.json` to specify minimum version supported by simple-icons installing it as dependency. This is because `Object.property`, used in the generated `index.js`, is [available since that version](https://node.green/#ES2015-built-ins-Symbol-Object-defineProperty-support). Added in #3979

ericcornelissen:
Your changes in a970353558a190f6625f53d89a6eb44a15db3ea5 bring up a small question. It seems you're using a version of NodeJS/NPM that switched to `lockfileVersion` 2, whereas the existing `lockfileVersion` is 1. I'm having trouble figuring out when version 2 was introduced... It seems to be around v7. However (using `nvm` to test this) it seems even Node v14 comes bundled with an NPM version less than v7, and running `npm install -g npm` doesn't upgrade me to v7 either (only explicitly running `npm install -g npm@7`).

@mondeja, what are the Node and NPM versions you used in #3985? Any idea regarding the NPM version required for `lockVersion` 2? Any opinion on whether we should stick with `lockVersion` 1 for now or whether we should upgrade to `lockVersion` 2?

For the record, even when installing with an NPM version that doesn't "support" `lockfileVersion` 2, everything still seems to work fine.

mondeja:
I've used Node v15.2.1 and NPM v7.0.8.

This new `package-lock.json` version seems to be [backwards compatible](https://github.com/npm/cli/blob/latest/CHANGELOG.md#files-on-disk) with NPM CLIs using previous version of lockfiles, but older clients print a warning about the version mismatch.

[...]

PR #3979, opened 2020-11-20
Add minimum NodeJS version requirement to package.json

mondeja:
**Issue:** #3959

Minimum version of NodeJS required to install simple-icons will be 0.12.18 after this pull.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,6 @@
     "get-filename": "node scripts/get-filename.js"
   },
   "engine": {
-    "node": ">=>=0.12.18"
+    "node": ">=0.12.18"
   }
 }
--- a/package.json
+++ b/package.json
@@ -42,5 +42,8 @@
     "posttest": "npm run postpublish",
     "svgo": "svgo --config=./.svgo.yml",
     "get-filename": "node scripts/get-filename.js"
+  },
+  "engine": {
+    "node": ">=>=0.12.18"
   }
 }


package.json
```diff
@@ -42,5 +42,8 @@
     "posttest": "npm run postpublish",
     "svgo": "svgo --config=./.svgo.yml",
     "get-filename": "node scripts/get-filename.js"
+  },
+  "engine": {
+    "node": ">=>=0.12.18"
```
ericcornelissen, line 47: I think this should be:

```suggestion
    "node": ">=0.12.18"
```

ericcornelissen (changes requested):
This is okay with me, if it turns out to be incorrect I guess we will find out soon enough :upside_down_face: 

Nevertheless, here are two language features we use that might potentially be missing from 0.12.18, though i wasn't able to verify this:

- The use of getters in icon objects

  https://github.com/simple-icons/simple-icons/blob/9cee719fcb487d7a43c9e4def631c82f8271c8bd/scripts/templates/icon-object.js#L5-L7

- The use of `enumerable: false` in index.js:

  https://github.com/simple-icons/simple-icons/blob/9cee719fcb487d7a43c9e4def631c82f8271c8bd/scripts/templates/index.js#L4

If you can verify this @mondeja that would be great. Otherwise, you can merge this after fixing the requested change and we will find out later whether or not these are (or something else is) a problem.

package.json
mondeja: What a mistake :sweat_smile: Sorry, it's fixed now. I will investigate these features before merge.

mondeja: - Getters are supported in 0.12.18 (tested manually and defined in [node.green](https://node.green/#ES2017-annex-b-Object-prototype-getter-setter-methods-__lookupGetter__)).
- `enumerable` property was been added at the beggining of the [`Object.defineProperty` support](https://node.green/#ES2017-annex-b-Object-prototype-getter-setter-methods-__lookupGetter__). More info [here](https://www.ecma-international.org/ecma-262/6.0/#sec-property-attributes).

PR #4000, opened 2020-11-20
Updated Boeing

fbernhart:
boeing

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted (changes requested):
Icon looks perfect! 🎉
Any chance you can reduce it to 3 decimals?

fbernhart:
@adamrusted I gave it a few tries and even rebuilt from the source, but I couldn't decrese it to 3. Seems like we have to stick to a precision of 4 for this one.

adamrusted:
Looks like the Wikipedia logo used as source draws from [THIS PDF](http://www.boeing.com/resources/boeingdotcom/history/pdf/Boeing_Chronology.pdf), would it be worth updating source - or do you think we should keep it as Wikipedia?

fbernhart:
@adamrusted Quoting our [contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#source-guidelines): The `source` attribute is as well used as a "jumping-off point to find updates if the logo changes".

So probably if Boeing is ever going to change their logo, they probably won't updated the old, already existing PDF file. That's why I would keep Wikimedia as source. (It's based on the same PDF anyway).

I've only slightly changed the source now to: https://commons.wikimedia.org/wiki/File:Boeing_full_logo.svg, which gives a better overview of the logo.

adamrusted:
> It passes the linter. **BUT** it's the logo of Baidu and not Boeing. 😄 

That's what I get for working on multiple simultaneous icons! 😬😂

PR #4001, opened 2020-11-20
Updated Boost

fbernhart:
boost

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (changes requested):
Seems that the logo has been updated. Could you rebuilt using the new version?

image

fbernhart:
boost

@mondeja Done. As new color I've used `#F7901E`, which is as well used in the SVG file from the header of [their website](https://www.boostmobile.com/).

mondeja (approved):
Good catch @fbernhart :+1: 
Seems that has imperfections, but coming from the original :sweat_smile:

PR #4004, opened 2020-11-20
Updated Buffer

fbernhart:
buffer

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted (changes requested):
Could you look to rebuild from source? Looks like the logo may have changed.
I was able to get away with a decimal of 2 when testing this, due to the simplistic nature of the logo.

image

fbernhart:
I've now used the `buffer-logo-mark.svg` file from the brand assets at https://buffer.com/press to rebuild it.

I've as well changed the color to `#231F20`, which is mentioned in their brand guidelines as first color and is as well used as color for the SVG file.

In my opinion a precision of 3 should be the default. Only in very rare cases (like here) we can even reduce it further to 2 or below. Reducing this icon to a precision of 2 saves 32 bytes (which isn't much), but it might be that the icon will land on a "Icons to review checking their simplicity of precision" list like the one in #3169. So we would have to check it again in the future.

I don't think that 32 bytes is worth the extra work we might have in the future to double check that the precision is ok. So I'd be in favour of keeping it at 3.

What do you think @adamrusted?

adamrusted (approved):
LGTM 🎉 - I agree that 3 decimals is better for consistency, and we can always review to 2 decimals if we ***need*** to save bytes of space in the future.

PR #4009, opened 2020-11-20
Updated CodeChef

fbernhart:
codechef

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted:
Do you know if there's an SVG available at source?
I can only see the PNGs while browsing - and obviously we'd be better to compare to an SVG if available.

mondeja:
It seems that was created in #1921 vectorizing manually a PNG, so I would only compare against the previous vector if there is not a vector version now in the source, and I can't find it either.

fbernhart:
@adamrusted  How come you can't see the SVG file in the header of https://www.codechef.com? 🤔 

It's this one:
https://s3.amazonaws.com/codechef_shared/sites/all/themes/abessive/logo.svg


I'll rebuild the icon from it.

fbernhart:
codechef

adamrusted:
> @adamrusted How come you can't see the SVG file in the header of [codechef.com](https://www.codechef.com)? 🤔

My bad 😅 must have thought it was a PNG/JPG due to it's complexity.

> Note: I had to keep a precision of 4. I couldn't make the SVG pass the linter with a precision of 3.

I'll compare it against source now, but I imagine it's absolutely fine! 🎉

mondeja (approved):
> My bad 😅 must have thought it was a PNG/JPG due to it's complexity.

Wow, I must have thought the same :sweat_smile: 

It seems that these gaps are reasonable, so looks perfect for me :100: Thanks for this update @fbernhart!

fbernhart:
@adamrusted Yes, I've slightly cleaned up the bottom of the hat a bit. So that there were no very thin lines hanging around without any reason.

PR #4043, opened 2020-11-20
Updated Fandom

fbernhart:
fandom

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (changes requested):
The geometry is perfect :gem:

Could you double check the precision @fbernhart?

mondeja (approved):
Awesome :bulb:

fbernhart:
@mondeja I couldn't decrease the precision with the icon that was already part of Simple Icons. But rebuilding it from source, I was able to decrease it to 3. 💪🏼 

I've as well changed the `source` so that it points to the correct section in the branding guidelines. https://fandomdesignsystem.com/identity/assets

If you want to compare it again: I've used the embedded SVG from the section "Fandom heart" (the one with the heart alone).

PR #4051, opened 2020-11-20
Updated GIMP

fbernhart:
gimp

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja:
The icon it's outdated. You can update it in this pull or we open another one for this, as you prefer @fbernhart

gimp-outdated

fbernhart:
@mondeja I'll update it here. 👍🏼

fbernhart:
@mondeja That's a pretty tough logo 😅 And I'm not really sure how to best approach it to make it monochrome. Any opinions on which paths to fill? And which ones should use a stroke instead?

That's my first try. I've created gaps to separate the nose and the paint brush from the face.

image

Maybe the other @simple-icons/maintainers could help out as well and give some opinions on how to best handle this logo. 😊

ericcornelissen:
I'd say the separations like this are a good choose, nice work 👍 The only thing I would change is remove the "spikes" from the eye (iris) and instead somehow include the reflection (the solid white circle top-left of the iris) in there.

fbernhart:
I was thinking that it might look pretty weird having the reflection included. 🤔 Especially in the eyes. That's why I've removed it.

Opening GIMP, I can see that they are using a very similar monochrome version inside their programm.
image


I've looked through their repo and I guess it should be this one:
https://github.com/GNOME/gimp/blob/master/data/images/wilber-ui-shadow.svg

It's back from 2010 (and therefore without the paint brush), but in my opinion we could use it as inspiration.

So how about using the style of this 2010 version, but with the paint brush?

ericcornelissen:
> 
> 
> I was thinking that it might look pretty weird having the reflection included. 🤔 Especially in the eyes. That's why I've removed it.

I guess that depends on how you do it, if you just remove the reflected part from the iris I would say it still looks okay. But again it was just a suggestion

> Opening GIMP, I can see that they are using a very similar monochrome version inside their programm.
> image
> 
> I've looked through their repo and I guess it should be this one:
> https://github.com/GNOME/gimp/blob/master/data/images/wilber-ui-shadow.svg
> 
> It's back from 2010 (and therefore without the paint brush), but in my opinion we could use it as inspiration.
> 
> So how about using the style of this 2010 version, but with the paint brush?

Seems reasonable to me 👍

fbernhart:
image


@ericcornelissen How about this one now?

ericcornelissen:
Not a full review, but I think this style looks good :+1: Nice work @fbernhart

mondeja:
I'm looking it very good :+1:
Just a small suggestion/question: I note that the right part of the mouth looses a bit of expression because the gap cover the curvature. Have you tried how the icon is displayed in small sizes if the gap is less wide?

fbernhart:
gimp


@mondeja Thanks for pointing that out. I've now used smaller gaps and moved the right gap as well slightly - in order to keep as much details of the mouth as possible.

I've pushed the new version - feel free to have a look at it and merge it, if you think it's o.k. 🖌️

mondeja (approved):
Wow, very nice! :art:
I think that it's more more loyal to the original as the geometry of the commissure of the mouth remains. I'm seeing the icon good at small sizes also. Anyway, in the future could be improved if someone wants another version. Thanks @fbernhart :tada:

PR #4066, opened 2020-11-20
Updated Haskell

fbernhart:
haskell

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis (approved):
LGTM 💯 
Is the precision of 4 is necessary?
Since they use a SVG file in the header of their [official website](https://www.haskell.org/), should we change the source for it? What do you think?

fbernhart:
@service-paradis I've fixed the precision. 😊

I think we should keep the source as it is. What if they upload a new logo to their website `haskell-logo-new.svg` and we're still having `haskell-logo.svg` as source? 🤔 Maybe we won't even realise the logo change. What do you think?

service-paradis (approved):
Perfect, thank you 💯 
Though it might be better to use the official website as source instead of Wikimedia, but I'm ok with it 😄

PeterShaggyNoble:
@service-paradis, the SVG on Wikimedia is lifted directly from [this page](https://wiki.haskell.org/Thompson-Wheeler_logo) but there's also an SVG in the header of their [website homepage](https://www.haskell.org/).

PR #4113, opened 2020-11-22
Updated NuGet

fbernhart:
nuget

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (changes requested):
I'm seeing some differences between [this source](https://github.com/NuGet/Media/blob/ef5ae34c768af3f17ee91c0815e7c9a252041b73/Images/MainLogo/Vector/nuget.svg) (which I suppose that is the original, since is not clear) and your version (red below). Can you rebuilt from that source and change the URL in `/_data/simple-icons.json`, so if the NuGet media repository change, we can track the original icon?

nuget-diff

fbernhart:
@mondeja Done. Should be fine now

mondeja (approved):
Another one perfect :tada:

PR #4125, opened 2020-11-24
Skip release job in non-release pull request review

mondeja:
I've seen a lot of times that the `release` job is executed when a pull request is reviewed, even if is not a pull with `release` label. This spend minutes of Github Actions (now free) which can be saved, time reviewing pulls and can generate confusion.

In order to accomplish this, I'm skipping the job if the event triggered is `pull_request_review` but the pull request does not [contains](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#contains) the label `release`.

You can see this behaviour tested at https://github.com/mondeja/test-si-release/pulls

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    if: (github.event_name == 'pull_request_review' && contains(github.event.pull_request.labels.*.name, 'release')) || github.event_name == 'schedule'
     steps:
     - uses: simple-icons/release-action@master
       with:


ericcornelissen (commented):
This seems fine to me, can't believe I didn't think of this myself 🤦‍♂️ But perhaps an idea is to test whether the pull request targets `master` instead of checking for the `release` label. Not really sure... What do you think @mondeja? If you feel like testing for the label is fine you can merge this as is 👍

mondeja:
:thinking: In theory would be more secure check if is targeting master. I'm going to try a little with it...

mondeja:
> But perhaps an idea is to test whether the pull request targets master instead of checking for the release label. Not really sure... What do you think @mondeja?

If an user targets `master` and someone aproves the pull can be released a version by mistake, not? But an user without priviledges can't add the `release` label. I think that is more secure by this way.

ericcornelissen:
Fair point :+1: the action that does the merge has protection against this as well (it checks if the reviewer is of a certain type) but having some extra security in this case is not a bad idea.

PR #4147, opened 2020-11-28
Correct Logstash icon

service-paradis:
image

**Issue:** #3169
**Alexa rank:**
  

### Checklist

### Description
- I took the original logo from https://www.elastic.co/fr/brand since the current icon looks wrong.

**_Note: I think the color needs to be updated, but not sure which one to use..._**


fbernhart:
@service-paradis The icon looks great! 👍🏼

I would keep the color for now. It's consistent with all the other "elastic" icons of Simple Icons. I'd suggest that if we ever find a better color, we could update all of them at once.

PR #4167, opened 2020-11-29
Updated Pi-hole

fbernhart:
pi-hole

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted (changes requested):
Source updated to [here](https://github.com/pi-hole/pi-hole/blob/0d8ece1be249fdc5993f01a45f8d0464257df819/README.md), previous repo has been archived. Could you compare new source to current SVG, and try and reduce down to 3 decimals? 🙏🏻

fbernhart:
@adamrusted I've found the same link to the same SVG on their documentation website: https://docs.pi-hole.net
I'm not sure why, but I would prefer linking to a website rather than an embedded SVG in a readme file. Do you agree?

I've now rebuilt if from this new source (precision = 3). And changed the color as well, as the SVG from the new source uses `#FF0000`.

adamrusted (approved):
LGTM!

PR #4179, opened 2020-11-29
Updated Read the Docs

fbernhart:
readthedocs

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja (changes requested):
After play a bit with the source, I want to point that the treatment of current version in simple-icons is very poor. You can see in next comparison that the current version (black) has too much coordinates, which seems innecessary:

image

Also, it has some inconsistencies in the path. Check it:

image

So, I think that would be worth to rebuild from source. Also, the source must be changed to use https://github.com/readthedocs/readthedocs.org as the repository (currently pointing to old https://github.com/rtfd/readthedocs.org organization). What do you think @fbernhart?

fbernhart:
readthedocs

@mondeja Done. The style is a bit different now, as the logo isn't "filled" anymore. But I think it is closer to the original logo, as the outline of the logo can be seen.

Or would you prefer a filled version?

mondeja:
I'm afraid that this could break a lot of badges, since the ReadtheDocs icon is not well seen at small sizes using a non filled version... Could you try with a filled one?

### Before

image


### After

image

fbernhart:
@mondeja You're right! Thanks for pointing this out.

I've updated it.

mondeja (approved):
Ouh yeah, really amazing job! :trophy:
Thank you so much @fbernhart!

Issue #3270, opened 2020-06-29
Solve possible issues for (sub-)brands with the same icon


PeterShaggyNoble:
### Kind of issue
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### This issue concerns
  - [x] The NPM Package
  - [x] The Packagist Package
  - [x] The Website

### Description
Spinning this out of #2260 from an issue that arose in #3028 and, more pressingly, #3219 and that's what do do in cases where 2 brands differ in name and colour but share the same icon. We initially decided against adding them in #2260, at least until such a time as someone requested one or more of the subsidiaries which is what has now happened in #3219.

My initial suggestion was to use the new `slug` being added in #2260 to point the subsidiary to the parent's SVG file but quickly realised that this would create problems (including for our linting process) with the `<title>` of the SVG being incorrect for the subsidiary, something we _could_ change dynamically in a few places but not everywhere.

Another possibility would be somehow use the `<use>` tag in the subsidiary's SVG and reference it to the parent's. Issues with implementation aside, though, this would also create a problem in a scenario where a company may change the logo or icon for the parent brand but not for the subsidiaries.

A final possibility would be to continue to _not_ add subsidiaries that differ only in name and colour and leave it up to the user to implement themselves.

ericcornelissen:
Having thought about this a bit more, I actually believe we should just add the same SVG, with the `<title>` changed, if subbrands use the same logo but different color (maybe *even* if it's the same color) **and** are truly separate brands. So, for #3219, I think we should indeed add the Svelte SVG with the `<title>` changed to "Sapper icon".

This sidesteps all problems that come with trying to re-use the existing SVG and only adds a near identical SVG to the collection. For us, this is not really that much of a problem.

PeterShaggyNoble:
> For us, this is not really that much of a problem.

One small problem I foresee is that if, using Svelte as an example again, they update their logo, we'd have to remember to update Sapper, too. The more sub-brands or subsidiaries using the same logo (e.g., Japan Railways, if we eventually added them all), the more icons we'd need to remember to update.

ericcornelissen:
> One small problem I foresee is that if, using Svelte as an example again, they update their logo, we'd have to remember to update Sapper, too. The more sub-brands or subsidiaries using the same logo (e.g., Japan Railways, if we eventually added them all), the more icons we'd need to remember to update.

Though not ideal, I wouldn't call it a _problem_ per se.

[...]

PR #4184, opened 2020-11-29
Updated Runkeeper

fbernhart:
runkeeper

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted (changes requested):
Runkeeper have updated their icon, and have a proper [press kit](https://runkeeper.com/cms/press-kit/) section on their website. Could you rebuild this one, and point source to this page?

fbernhart:
@adamrusted Done 🏃🏼  For reference: I've used the `ASICS Runkeeper logo - Intl - blue and white.svg` file.

runkeeper

adamrusted:
Here's a fun one @simple-icons/maintainers - just having a quick glance at the [Brand Guidelines](https://runkeeper.com/cms/wp-content/uploads/sites/3/2020/06/Brand_Guidelines.pdf) for Runkeeper, and it's saying that ® should only be used in the USA, otherwise `TM` should be used. It's also unclear as to whether the icon can be used without the wordmark from the documentation there. I imagine given that an icon option isn't presented, we ***should*** use the full wordmark and icon?

fbernhart:
Hmm ... as Simple Icons is clearly an international project, we can use the international version of the Runkeeper logo (and as they don't requite the trademark symbol, remove the (TM) sign 😝).

Probably we should go for "logo + wordmark". Because actually the logo itself is the logo of ASICS, the sports company that bought Runkeeper some time ago. So that we differenciate between ASICS and Runkeeper.

fbernhart:
runkeeper

adamrusted (approved):
Perfect! 🏃🏻‍♂️

PeterShaggyNoble:
Generally, adding the wordmark simply to differentiate it from another brand, wouldn't be the correct call. If the brand colours are the same then adding the parent brand instead of the sub-brand would usually be the way to go. Although, there are, of course, plenty of exceptions which get addressed on a case-by-case basis. If the brand colours are different then we'd need to address #3270 first.

That said, though, given the following line from their brand guidelines, I think it's clear that we _had_ to include the wordmark in this case:

> The ASICS Runkeeper® master logo should be used at all times.

However, on the trademark symbol, the brand guidelines do state that it _must_ be included as well so I think we, unfortunately, will need to add it in:

> Don't remove the (R) or (TM) from the logo

PR #4185, opened 2020-11-29
Updated Salesforce

fbernhart:
salesforce

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis:
Thanks again @fbernhart 
Just want to make sure if precision of 4 is necessary. Also, we could probably update source to https://www.salesforce.com/styleguide/elements/logos/. What do you think?

fbernhart:
@service-paradis Thanks for pointing out. It passed now with a precision of 3.

I've updated the source as well. 😊

PR #4201, opened 2020-11-30
Updated Steem

fbernhart:
steem

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis:
According to https://steem.com/brand/, I think we can update the source and the color. I would probably choose `#171fc9`.
Logos can also be downloaded here. I'm asking if we should keep the circle.
Maybe, we can merge this one and open a new issue about the update? What do you think?

fbernhart:
@service-paradis I would keep the circle as well.

Color and source have indeed to be updated.

I prefer pushing the new icon to this PR. Is that ok?

service-paradis:
> I prefer pushing the new icon to this PR. Is that ok?

Sure 😄

fbernhart:
steem

service-paradis (approved):
Perfect, thank you! 💯

PR #4126, opened 2020-11-24
Switch from Travis CI to GitHub Actions

ericcornelissen:
This replaces the Travis CI configuration file by (hopefully) equivalent GitHub Actions Workflows. The result is two GitHub Actions Workflows, one that lints, tests, and builds on pushes and Pull Requests, and one that deploys to NPM and creates a release tag on pushes to `master`. ~~Note that I did not yet test the latter.~~

I now (1 day after opening) tested - and fixed - the publishing workflow as well. An example of a working run is [this one](https://github.com/ericcornelissen/simple-icons/actions/runs/382865128). You can see that the NPM deployment (should) work in [these logs](https://github.com/ericcornelissen/simple-icons/runs/1452632770#step:6:167) and that the GitHub release is created in [these logs](https://github.com/ericcornelissen/simple-icons/runs/1452632826#step:10:73) - and [here is the release on my fork](https://github.com/ericcornelissen/simple-icons/releases/tag/3.13.0).

The GitHub release title is `Release {{ VERSION }}` and the GitHub release text is taken directly from the commit on which the job is running (and our [release-action](https://github.com/simple-icons/release-action) ensures that this is the list of new/edited/removed icons). Note that this slightly changes the release title, currently it is formatted roughly as `Release {{ NEW_ICONS }} and {{ UPDATED_ICONS }} and {{ REMOVED_ICONS }}`. If we want that format back we can obtain it by looking at the commit title of the commit that triggers the job since our [release-action](https://github.com/simple-icons/release-action) ensures the commit title has that format.

~~Things that still need to be looked in to:~~

- [x] Git clone depth (similar to #3847)
- [x] Caching (similar to #3847)

Other things to do:

- [x] Add `NPM_TOKEN` secret to repository
- ~~Add `GITHUB_USERNAME` secret to repository~~ No longer required
- ~~Add `GITHUB_EMAIL` secret to repository~~ No longer required
- [x] Update [the build badge](https://github.com/simple-icons/simple-icons#status) in the README

* * *

Closes #4124. But, as per the discussion there, we might still decide to stick with Travis if we get the free plan for OSS projects.

PR #4213, opened 2020-12-02
Update setup Ruby step in build job of verification workflow

mondeja:
As you can see in [this run](https://github.com/simple-icons/simple-icons/pull/4210/checks?check_run_id=1484519338#step:5:6), the `bundle` command does not exists in the instance for ruby which executes the "Build Website" job. This is happening in all new pull requests because Github Actions now uses Ubuntu 20.04 specifying `ubuntu-latest`. I've been investigating this and seems that the action [actions/setup-ruby](https://github.com/actions/setup-ruby) is not maintained, with [bugs in Ubuntu 20.04](https://github.com/actions/setup-ruby/issues/70), and do not explicitly specify Bundler availability, so I've changed by [ruby/setup-ruby](https://github.com/ruby/setup-ruby) action that seems better maintained, supports a lot of versions and [explicitly declares Bundler availability](https://github.com/ruby/setup-ruby#bundler).

Also, I've updated the Ruby version to 2.7 because [Ruby v2.5 will reach their EOL](https://www.ruby-lang.org/en/downloads/branches/) at 2021-03-31.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -27,9 +27,9 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Use Ruby 2.5
-      uses: actions/setup-ruby@v1
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.5
+        ruby-version: 2.7
     - name: Setup cache for gems
       uses: actions/cache@v2
       with:
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Use Ruby 2.5
+    - name: Use Ruby 2.7
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: 2.7


ericcornelissen:
Oops... seems like [I had too much fate](https://github.com/simple-icons/simple-icons/pull/4126#issuecomment-735214521) in GitHub :sweat_smile: I'm very surprised [actions/setup-ruby](https://github.com/actions/setup-ruby) is already not being maintained any more, especially given that most of GitHub's infrastructure is build on Ruby!

Anyway, thanks for the investigation and quick fix @mondeja. I agree with your decision given the discussion in https://github.com/actions/setup-ruby/issues/70, I'm also completely okay with updating to Ruby 2.7 - with Ruby I'm not as worried since it only builds our website and doesn't affect users of the packages.

And my apologies to everyone who experienced any difficulty due to this problem :disappointed:

ericcornelissen (approved):

mondeja:
> And my apologies to everyone who experienced any difficulty due to this problem

I don't think it's your fault, I think that Github Actions is not well supported by the platform. Beyond this, I've seen this kind of lack of maintenance in many of the Github related repositories. Perhaps the most sensible thing to do is to stop trusting that "official" actions are the best.

ericcornelissen:
> I don't think it's your fault, I think that Github Actions is not well supported by the platform. Beyond this, I've seen this kind of lack of maintenance in many of the Github related repositories. Perhaps the most sensible thing to do is to stop trusting that "official" actions are the best.

I'm afraid so... :disappointed:

PR #4214, opened 2020-12-02
Update strongSwan

fbernhart:
strongswan

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


mondeja:
See #4213

adamrusted (changes requested):
Looks like there's a fair few differences between this and source, though there is no source SVG. Would you be able to take a look at this when you get a chance?

image

fbernhart:
I've now created a new vectorized version from scratch. And yeah, that logo is DIFFICULT to vectorize. 😅 

I can see a clear conflict here:
- Either go for accuracy (so the logo will fit as much the source as possible) --> this will mean a lot of points and a pretty big file size. I've chosen this option now for the commit that I've just pushed, which is a SVG with 80,1 kB.

image


- Or: Try to simplify the logo a bit and drastically reduce the file size. Maybe to around 5-8 kB?

image

Which way should we go?

I guess as file size is a pretty big deal and as SVGs are typically displayed at small sizes, we should probably go for the second option and simplify it as much as possible. But that's just my opinion.

fbernhart:
And should we include the wordmark "strong swan"? I would say yes, as it's pretty much part of the logo.

By the way: You can see the PNG source here: https://www.strongswan.org/images/strongswan_neu.png

fbernhart:
After a lot of manual work, I've now pushed a simplified version with a lot less points. It still looks pretty good in my opinion and now "only" is 9,7 kB big.

strongswan

adamrusted:
This looks great to me, well done @fbernhart 💯
Could another @simple-icons/maintainers take a look at this one before we merge it, just to make sure we're all happy regarding accuracy and file size?

image

mondeja (approved):
Wow, great work @fbernhart :100:

PR #4215, opened 2020-12-02
Update Stylus

fbernhart:
stylus

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


service-paradis:
Looks perfect 💯 , but was it necessary to use a precision of 4?

fbernhart:
@service-paradis Probably not.

When I started working on the list in #3169 I saw that about 1/3 of the icons will fail the linter with a precision of 3.

And as a precision of 3-5 is accepted by the linter, I started using 4 as default precision. Because with a higher precision it is less likely that the linter will fail with something like "size is 24.001".

But @mondeja already told me, that that was probably a wrong decision. And that a precision of 4 shouldn't be used as default, but only if a precision of 3 isn't working.

Sorry for that. If you find any other icon with a precision of 4, just let me know and I'll optimize it again with a precision of 3. 😊

service-paradis:
Ok, I understand, no problem at all!

mondeja:
I would like to point out the same, the linter must be used to detect errors, it must not condition the precision of the icons. Since this icon it was optimized to 3 decimal numbers, I would leave it with 3.

fbernhart:
Thanks for the feedback. I guess this can be merged now. 😊

mondeja (approved):
Looks perfect. Thank you for the fix @fbernhart! :tada:

PR #4220, opened 2020-12-02
Update Synology

fbernhart:
synology

**Issue:** #3169 
**Alexa rank:** -

### Checklist

### Description
I took the already used SVG.


adamrusted (commented):
Should this be updated to the full wordmark, as shown on their logo guide?
What do other @simple-icons/maintainers think?

mondeja:
> Should this be updated to the full wordmark, as shown on their logo guide?

I would say yes, quoting brand guidelines: 'The "Synology" logo consists of the word Synology as illustrated below.'

adamrusted (changes requested):
> > Should this be updated to the full wordmark, as shown on their logo guide?
> 
> I would say yes, quoting brand guidelines: 'The "Synology" logo consists of the word Synology as illustrated below.'

@fbernhart - as mentioned above, could you please regenerate this using the full wordmark for Synology? 🙏🏻

fbernhart:
@adamrusted Done. I've as well changed the color slightly. [Their brand guidelines](https://www.synology.com/en-global/company/branding) mention rgb(181,181,182) which should be `#B5B5B6`.

adamrusted:
> @adamrusted Done. I've as well changed the color slightly. [Their brand guidelines](https://www.synology.com/en-global/company/branding) mention rgb(181,181,182) which should be `#B5B5B6`.

Amazing - do we need to include the &reg; symbol in the logo also? I don't see any explicit mention on the logo guide page about the mark, but it *is* featured in every use of the logo on the page, even the incorrect ones!

fbernhart:
> Amazing - do we need to include the ® symbol in the logo also?

@adamrusted Quoting our contribution guidelines:

> If there is no explicit requirement that a symbol must be included, it must not be included.

That's why I've removed it.

ericcornelissen:
I'm not familiar with the brand, but from the brand guidelines this seems to me to fall into the same category as Intel :thinking: 

> If the symbol is incorporated into the design of the logo (e.g. Intel), it must be included.

fbernhart:
I wouldn't say that the (R) is really integrated into the design of the logo.

But I'm happy to add it, if this is the consensus. 😊

ericcornelissen:
> I wouldn't say that the (R) is really integrated into the design of the logo.

If it isn't can you show me a place where they use the logo without the (R)?

fbernhart:
You're right. Seems like they are using it everywhere.

Let me add it. 👍🏼

fbernhart:
Fixed

synology

adamrusted (approved):
Perfect! @fbernhart 💯

PR #4262, opened 2020-12-03
Add a new linter rule for ineffective commands in paths

### LINKED ISSUE #4257: Check ineffective path commands like 'm0 0'
I've found in #4024 that the icon `disroot.svg` has a command in their path `m0 0zm0 0zm0 0zm0 0zm0 0zm0 0zm0 0z`. This does not render nothing and could be hard to debug manually. Maybe, we can extend the linting process ~~(or the tests?)~~ to check that the command `m0 0z` does not appear in icons paths. Thoughs about this?

## List of ineffective paths

- `m0 0`
- `M{x} {y}` where `{x}` `{y}` is the last coordinate ~~(at the beggining `0` `0`)~~
- `l0 0`
- `h0 0`
- `v0 0`
- `L{x} {y}` where `{x]` `{y}` is the last coordinate
- `H{x} {y}` where `{x}` `{y}` is the last coordinate
- `V{x} {y}` where `{x}` `{y}` is the last coordinate

service-paradis:
Close #4257

I made some tests locally. I though I might as well create a PR as a starting point.

### Proposed changes

- Add a new linter rule for invalid command in path

#### List of ineffective paths

- [x] m0 0
- [x] l0 0
- [x] h0
- [x] v0
- ~~M0 0 at the beginning~~
- [x] M{x} {y} where {x} {y} is the last coordinate
- [x] L{x} {y} where {x] {y} is the last coordinate
- [x] {prevX} {prevY} H{x} where {x} is equal to {prevX}
- [x] {prevY} V{x} where {x} is equal to {prevY}
- [x] `C x1 y1, x2 y2, x y` where `x1 y1` is the last coordinate and where `x2 y2` is equal to `x y`. It covers the following cases
  - [x] `M 0 12 C 0 12, 0 12, 0 12`: unnecessary
  - [x] `M 0 12 C 0 12, 24 12, 24 12`: should be `L24 12` instead
- [x] `c 0 0, 0 0, x y`: should be `lx y` instead
- [x] `c 0 0, 0 0, 0 0`: unnecessary
- [x] `S x2 y2, x y` where `x2 y2` is the last coordinate and where `x2 y2` is equal to `x y`: unnecessary
- ~~`S x2 y2, x y` where `x2 y2` is equal to `x y`: should be `lx y` instead~~
- [x] `s 0 0, x y` : should be `lx y` instead
- [x] `s 0 0, 0 0` : unnecessary

_Since english is not my native language, I have some difficulties finding a sentence that would be clear for the end user._

_I also need to do some more tests to be sure there are no false positives. If not, there are ~30 icons with ineffective paths._


.svglintrc.js
```diff
@@ -163,6 +163,33 @@ module.exports = {
...
+                  return true;
+                }
+              }
+            };
+            if (segments.some(isSegmentInvalid)) {
+              reporter.error(`Unexpected segment in path.`);
+              if (updateIgnoreFile) {
+                ignoreIcon(reporter.name, iconPath, $);
+              }
+            }
```
mondeja, line 191: - We can improve the failure message reconstructing the command since all are made by one point.
- Also, we can show all failed commands (segments) for each failed path.
- Note that here I'm executing the `isSegmentInvalid` function only for segments that starts with possible invalid commands.
 - I've changed the "segment" word in the message replacing it with "command" because the SVG specification [describes them as commands](https://www.w3.org/TR/SVG/paths.html#Introduction).

```suggestion
            _invalidSegment = false;
            for (let s=0; s<segments.length; s++) {
              if (
                ['M', 'L', 'H', 'V', 'm', 'l', 'h', 'v'].includes(segments[s][0]) &&
                isSegmentInvalid(segments[s], s)
              ) {
                let readableSegment = `${segments[s][0]}${segments[s][1]} ${segments[s][2]}`;
                reporter.error(`Unexpected command ${readableSegment} in path.`);
                _invalidSegment = true;
              }
            }
            if (_invalidSegment && updateIgnoreFile) {
              ignoreIcon(reporter.name, iconPath, $);
            }
```

mondeja (changes requested):
Very good @service-paradis :clap:

However, it can be improved. What do you think about each one of this changes?

.svglintrc.js
service-paradis: Thanks for your review! I made some changes. Not exactly as you proposed, but inspired from. What do you think?

mondeja: Better than my proposal :+1:

.svglintrc.js
```diff
@@ -163,6 +163,44 @@ module.exports = {
...
+                    return true;
+                  }
+                }
+              }
+            };
+            const invalidSegments = segments.filter(getInvalidSegments);
+
+            if (invalidSegments.length) {
+              invalidSegments.forEach(([command, coord1, coord2]) => {
+                const readableSegment = `${command}${coord1} ${coord2}`;
```
mondeja, line 196: Something I haven't realized before, is that the H and V commands ony has a number, so something like this would be necessary:

```suggestion
                let readableSegment = `${command}${coord1}`;
                if (coord2) {
                   readableSegment += ` ${coord2}`;
                }
```

.svglintrc.js
```diff
@@ -163,6 +163,44 @@ module.exports = {
...
+            }
+
+            const { segments } = parsePath(iconPath);
+
+            const lowerCommand = ['m', 'l', 'h', 'v'];
+            const upperCommand = ['M', 'L', 'H', 'V'];
+            const commands = [...lowerCommand, ...upperCommand];
+            const getInvalidSegments = ([command, coord1, coord2, ...rest], index) => {
+              if (commands.includes(command)) {
+                if (lowerCommand.includes(command) && coord1 === 0 && coord2 === 0) {
```
service-paradis, line 181: > Something I haven't realized before, is that the H and V commands ony have a number, so something like this

@mondeja I will need to rethink this part of the code. There is more than that.

- For `h` and `v`, `coord2` will always be undefined so `coord2 === 0` is never true in this case.
- For `h` and `v`, when we compare with a previous segments, we need to compare `coord1` with `coord2`
- As you said, I also need to adapt the reported error according to this

.svglintrc.js
```diff
@@ -163,6 +163,55 @@ module.exports = {
...
+                }
+              }
+            };
+            const invalidSegments = segments.filter(getInvalidSegments);
+
+            if (invalidSegments.length) {
+              invalidSegments.forEach(([command, coord1, coord2]) => {
+                let readableSegment = `${command}${coord1}`;
+                if (coord2 !== undefined) {
+                  readableSegment = `${readableSegment} ${coord2}`
```
mondeja, line 206: This can be simplified:

```suggestion
                  readableSegment += ` ${coord2}`
```

mondeja (changes requested):
Good work :+1: Only one minor suggestion.

service-paradis:
Only one thing is missing is the Absolute command `H` or `V` comparison with the previous segment. Will try do something for it soon!

mondeja:
> I have the feeling that this kind of message might be hard to correct for new contributors since most of them are hard to see in tools (inkscape or illustrator) and are kept by optimization.

In an ideal world, this could be added to SVGO as a plugin but it seems that is not maintained. The only feasible solution I can think of is to deocuple `parsePath` function from `svgpath` in another package and include the index of the character that starts each segment into the parsing results, so we could improve the error message here and, at least, be able to debug this quickly.

.svglintrc.js
```diff
@@ -163,6 +163,65 @@ module.exports = {
...
+                  return true;
+                }
+                // Relative movement (m or l) having a distance of 0
+                if (lowerCommand.includes(command) && coord1 === 0 && coord2 === 0) {
+                  return true;
+                }
+                if (index > 0) {
+                  const [prevCoord2, prevCoord1, ...rest] = [...segments[index - 1]].reverse();
+                  // Absolute horizontal direction (H) having the same x coordinate as the previous segment
+                  if (upperHorDirectionCommand === command && coord1 === prevCoord1) {
```
service-paradis, line 196: This does not work as expected right now. For example, it gave the following error for Zendesk: `Unexpected segment H12.915 in path.` that looks like 
image

If I remove this segment from the path, the icon looks like
image

mondeja: Check the previous command, is `m` (relative): `m12.915 0H12.915`

service-paradis: Yes, not that easy to have a good inefficiency checker... I would need to keep the current computed absolute coordinate when looping over segments... I think it will become an overcomplicated rule...

`l10,0 l10,10 l10,0 H30` is inefficient. But to know it is, I would need to compute the current absolute coordinate after each segments:
1. [l, 10, 0]: (10, 0)
2. [l, 10, 10]: (20, 0)
3. [l, 10, 0]: (30, 0)
4. [H, 30]: invalid since the previous coordinate was (30, 0)

Can look easy while taking lineTo or moveTo, but computing the final coordinate when using arc or curve and when mixing absolute and relative command, it becomes complicated.
Maybe i'm also missing something 😛

mondeja: I think that you are well on your way. I ran into a similar problem developing `svg-path-bbox`. The solution would be to absolutify the points in another array to compare with it. You can use the `abs()` function that `svgpath` provides, something like:

```javascript
svgPath(d).abs().unarc().unshort().iterate((seg, dc, x, y) => { ... });
```

service-paradis: Thank you very much for your help on this!
`svgPath(d).abs().unarc().unshort()` is working well to get only absolute coordinate, but the number of segments is not always the same as the parsePath results. In this case, how can I compare the 2 arrays since item index wont necessarily be the same between those arrays?

mondeja: If I don't get lost, you "just" need to do for `l`, `L`, `m`, `M`, `v`, `V`, `h` and `H` commands right? For `Q` and `C` segments returned by `abs().unarc().unshort()`, the last coordinate will be those numbers in the indexes 5 and 6 for `C` and 3 and 4 for `Q`.

mondeja: :thinking: You have tried with `svgPath(d).abs().unshort().iterate((seg, dc, x, y) => { ... }` (without `unarc()`)?

service-paradis: Ha, I should have think of that!
Now, both returns the same number of segments.
Will try to see if I'm able to use this!

mondeja: By the way, keep in mind that whitout unarc() you must handle also `S` and `T` commands.

service-paradis: Looks better now! Thanks again!
There is another edge case 😅 when I compare with the previous path. If the previous path is an absolute direction command (H or V), I only have a part of the coordinate (X or Y). 

So if I have `['L', 0, 24]` and the previous command was a direction one, I'm not able to tell if it is inefficient. In those cases, I would need to check the previous of the previous until I find the missing coordinate.

ex.
- `['L', 12, 24]` (12, 24)
- `['H', 12]` (12, ?)
- `['V', 24]` (12, 24)

mondeja: Wow that's interesting! So you must iterate backwards more for `H` and `V` commands until `H` and `V` (both) have been found or another command, not?. Don't worry if the code is too long, we can always separate it into another package. I'm sure that this implementation could be very reusable.

service-paradis: Exact! It could be any command that contains the missing coordinate. So if the last one is `H24` (`x` = 24), I need to know what was the previous value for `y`. So I need to iterate back to find a `V` or any other command that contains a coordinate for `y`.
I have something in mind. Stopping for today (friday 5pm here, time for a beer 😛)

service-paradis: More tests are needed, but I think I have something working!

runxel:
First of all: Wow!
Thanks for the nice work you guys pour into that. I'm really amazed!

> In an ideal world, this could be added to SVGO as a plugin but it seems that is not maintained. 

I wonder if we could maintain a fork of SVGO. There are many great PRs lying around there.
How great would it be if could have Github Action that does all the stuff any contributor has to do by themself right now.

ericcornelissen:
> I wonder if we could maintain a fork of SVGO. There are many great PRs lying around there.

I was wondering this too, isn't it possible to load plugins manually? In any case, probably most contributors use SVGOMG rather than the SVGO CLI...

> How great would it be if could have Github Action that does all the stuff any contributor has to do by themself right now.

You mean [this one](https://github.com/ericcornelissen/svgo-action)? 😄

runxel:
> You mean [this one](https://github.com/ericcornelissen/svgo-action)? 😄

Big oof. Haha 😅
But wait, we don't use it at the moment, do we? I thought we only do linting at the moment...

ericcornelissen:
>But wait, we don't use it at the moment, do we? I thought we only do linting right now...

We don't currently. We looked into this in the context of #2380 but at the time it didn't seem possible to get this working with forks. We might try this now with [`pull_request_target`](https://github.com/simple-icons/simple-icons/blob/2d254acf73bf46a7c0def3228245667504076544/.github/workflows/labeler.yml#L3), though I'm still not sure that allows for pushes to a fork. Alternatively, I just realized, we could try to do this using a Personal Access Token of one of the maintainers.

But the main problem here is that of the precision. This will always be something that the contributor has to consider. We could make it configurable per PR but it would be hard to recover automatically after the precision was set too low. We could set the precision of [`.svgo.yml`](https://github.com/simple-icons/simple-icons/blob/develop/.svgo.yml) to a [the maximum precision we allow](https://github.com/simple-icons/simple-icons/blob/2d254acf73bf46a7c0def3228245667504076544/.svglintrc.js#L13) so that precision will never be reduced by the automation. Or someone with a better idea might come along and actually solve the problem 🙃 

In any case, I suggest we leave that discussion for #2380 and leave the thread here to the new linter implementation.

.svglintrc.js
mondeja: I will try to review this as soon as possible.

mondeja (changes requested):
Can you specify the `svgpath` dependency v2.3.0 in `package.json` @service-paradis?

.svglintrc.js
```diff
@@ -163,6 +164,87 @@ module.exports = {
               }
             }
           },
+          function(reporter, $, ast) {
+            reporter.name = "icon-path";
```
mondeja, line 168: I would change this name by... `ineffective-segments` maybe? Seems to me that is too general :thinking:

.svglint-ignored.json
```diff
@@ -122,10 +121,46 @@
     "M4.78 24c-.88 0-1.36-.2-1.62-.36-.36-.21-1.02-.75-1.62-2.33A27.06 27.06 0 0 1 0 12V12c.02-3.66.59-6.76 1.54-9.3C2.14 1.1 2.8.56 3.16.35 3.42.21 3.9 0 4.78 0c.33 0 .72.03 1.18.1a26.1 26.1 0 0 1 8.7 3.3h.01a26.4 26.4 0 0 1 7.16 6.01c1.06 1.32 1.19 2.17 1.19 2.59 0 .42-.13 1.27-1.19 2.59a26.4 26.4 0 0 1-7.16 6h-.01a26.03 26.03 0 0 1-8.7 3.3c-.46.08-.85.11-1.18.11z": "Zalando",
     "M7.014 4.984A7.02 7.02 0 0 0 0 12.002a7.017 7.017 0 0 0 13.271 3.174h2.915c.696 0 1.324-.044 1.962-.553.461-.365.749-.884.883-1.56v2.103h1.336v-2.473h3.153v-1.1h-3.16l.02-.445c.005-.724.226-1.162 1.277-1.162H24V8.873h-2.818c-1.517 0-2.141.85-2.141 2.18v.129c-.254-1.565-1.185-2.309-2.889-2.309h-2.855a7.018 7.018 0 0 0-6.283-3.889zM8.02 8.873h3.436c1.742 0 1.992 1.219 1.992 1.9 0 .725-.298 1.873-1.992 1.873h-.844c-1.056 0-1.281.38-1.281 1.104v.336h3.945v1.074H7.982v-1.558c0-1.335.625-2.123 2.137-2.123h.873c.691 0 1.1-.14 1.1-.725 0-.605-.409-.772-1.12-.772h-2.95V8.873zm6.63 1.113h1.472c1.157 0 1.574.496 1.574 2.041 0 1.541-.412 2.035-1.574 2.035H14.65V9.986z": "ZDF"
   },
+  "icon-path": {
```
mondeja, line 124: Here too

mondeja (changes requested):

service-paradis:
> Can you specify the `svgpath` dependency v2.3.0 in `package.json` @service-paradis?

It is already [v2.3.0](https://github.com/simple-icons/simple-icons/blob/develop/package.json#L28)
image

.svglintrc.js
service-paradis: Sure!

mondeja:
Oh, it's true :facepalm:

mondeja:
The next list tracks all icons failing that I've rewieved manually:

- [x] `academia.svg`
- [x] `apacheant.svg`
- [ ] `asciidoctor.svg`
- [ ] `codeclimate.svg`
- [ ] `coronaengine.svg`
- [ ] `filezilla.svg`
- [ ] `fiverr.svg`
- [ ] `fossa.svg`
- [ ] `githubsponsors.svg`
- [ ] `gnuicecat.svg`
- [ ] `googlenews.svg`
- [ ] `guangzhoumetro.svg`
- [ ] `gutenberg.svg`
- [ ] `helm.svg`
- [ ] `itch-dot-io.svg`
- [ ] `landrover.svg`
- [ ] `kyocera.svg`
- [ ] `lenovo.svg`
- [ ] `lyft.svg`
- [ ] `neovim.svg`
- [ ] `opel.svg`
- [ ] `rubyonrails.svg`
- [ ] `sat-dot-1.svg`
- [ ] `sketch.svg`
- [ ] `sparkfun.svg`
- [ ] `theregister.svg`
- [ ] `toyota.svg`
- [ ] `trustpilot.svg`
- [ ] `tunein.svg`
- [ ] `unilever.svg`
- [ ] `wish.svg`
- [ ] `xmpp.svg`
- [x] `zalando.svg`

mondeja (commented):
I've found another edge case in the `opel.svg` icon: `c0,0,0,0,0,0`. I have solved it successfully by removing a coordinate from the path.

So, two more cases could be added: if all coordinates of a `C` or `S` command are the same as the last coordinate. What do you think @service-paradis?

service-paradis:
> I've found another edge case in the `opel.svg` icon: `c0,0,0,0,0,0`. I have solved it successfully by removing a coordinate from the path.
> 
> So, two more cases could be added: if all coordinates of a `C`, `S`, `c` or `s` command are the same as the last coordinate, the linter must fail. What do you think @service-paradis?

Yes, I can take a look at it. I think I would need to check for:

- `C x1 y1, x2 y2, x y` where `x1 y1` is the last coordinate and where `x2 y2` is equal to `x y`. It covers the following cases
  - `M 0 12 C 0 12, 0 12, 0 12`: unnecessary
  - `M 0 12 C 0 12, 24 12, 24 12`: should be `L24 12` instead
- `c 0 0, 0 0, x y`: should be `lx y` instead
- `S x2 y2, x y` where `x1 y1` where `x2 y2` is equal to `x y`
- `s 0 0, x y`

What do you think @mondeja ? Am I missing something?

mondeja:
Much better! :100:

service-paradis:
> Much better! 💯

Good, will give this a try!

> By the way, can you comment each possible case in the code? So we don't go crazy in the future...

I already add some comments so far 😃

service-paradis:
> The next list tracks all icons failing that I've reviewed manually

Thanks a lot for your help and your review! I think I have manage the Curve cases mentioned above. I need to do some manual tests, but looks promising I think.

mondeja:
> Thanks a lot for your help and your review! I think I have manage the Curve cases mentioned above. I need to do some manual tests, but looks promising I think.

Mark the pull as ready for review when you want another review.

.svglintrc.js
```diff
@@ -163,6 +164,136 @@ module.exports = {
...
+                    readableSegment += ` (should be "L${x2Coord} ${y2Coord}" or removed)`;
+                  }
+                  if (command === lowerCurveCommand && (xCoord !== 0 || yCoord !== 0)) {
+                    readableSegment += ` (should be "l${xCoord} ${yCoord}" or removed)`;
+                  }
+                  if (command === upperCurveCommand) {
+                    readableSegment += ` (should be "L${xCoord} ${yCoord}" or removed)`;
+                  }
+                }
+                reporter.error(`Unexpected segment ${readableSegment} in path.`);
```
mondeja, line 290: What about improve the error message wrapping all failed segments between double quotes and showing the resolution tip at the end?

So the message `Unexpected segment c0 0, 0 0, 0 0.007 (should be "l0 0.007" or removed) in path.` would be converted to `Unexpected segment "c0 0, 0 0, 0 0.007" in path (should be "l0 0.007" or removed).`.

```suggestion
                let readableSegment = `${command}${x1Coord}`,
                    resolutionTip = '';
                if (y1Coord !== undefined) {
                  readableSegment += ` ${y1Coord}`;
                }
                if (curveCommands.includes(command)) {
                  const [x2Coord, y2Coord, xCoord, yCoord] = rest;
                  readableSegment += `, ${x2Coord} ${y2Coord}`;
                  if (yCoord !== undefined) {
                    readableSegment += `, ${xCoord} ${yCoord}`;
                  }
                  if (command === lowerShorthandCurveCommand && (x2Coord !== 0 || y2Coord !== 0)) {
                    resolutionTip = `should be "l${x2Coord} ${y2Coord}" or removed`;
                  }
                  if (command === upperShorthandCurveCommand) {
                    resolutionTip = `should be "L${x2Coord} ${y2Coord}" or removed`;
                  }
                  if (command === lowerCurveCommand && (xCoord !== 0 || yCoord !== 0)) {
                    resolutionTip = `should be "l${xCoord} ${yCoord}" or removed`;
                  }
                  if (command === upperCurveCommand) {
                    resolutionTip = `should be "L${xCoord} ${yCoord}" or removed`;
                  }
                }
                let errorMessage = `Unexpected segment "${readableSegment}" in path`;
                if (resolutionTip) {
                  errorMessage += ` (${resolutionTip})`
                }
                errorMessage += '.';
                reporter.error(errorMessage);
```

mondeja (changes requested):
What a great job, @service-paradis! :1st_place_medal:

Although I have not reviewed all the paths yet, I leave you a small suggestion to improve the error message.

.svglintrc.js
```diff
@@ -163,6 +164,136 @@ module.exports = {
...
+                    readableSegment += ` (should be "l${x2Coord} ${y2Coord}" or removed)`;
+                  }
+                  if (command === upperShorthandCurveCommand) {
+                    readableSegment += ` (should be "L${x2Coord} ${y2Coord}" or removed)`;
+                  }
+                  if (command === lowerCurveCommand && (xCoord !== 0 || yCoord !== 0)) {
+                    readableSegment += ` (should be "l${xCoord} ${yCoord}" or removed)`;
+                  }
+                  if (command === upperCurveCommand) {
+                    readableSegment += ` (should be "L${xCoord} ${yCoord}" or removed)`;
```
mondeja, line 287: Another improvement for resolution tips. What about replace `0.xx` coordinate numbers by their simplified form `.xx`? So, as we are telling the user how to improve the path, improve it by making it efficient.

I've created next function at the start of the file that does the job:

```javascript
function removeLeadingZeroes(number) {
  // convert 0.03 to .03
  return number.toString().replace(/^(-?)(0)(\.?.+)/, '$1$3');
}
```

```suggestion
                    readableSegment += ` (should be "l${removeLeadingZeroes(x2Coord)} ${removeLeadingZeroes(y2Coord)}" or removed)`;
                  }
                  if (command === upperShorthandCurveCommand) {
                    readableSegment += ` (should be "L${removeLeadingZeroes(x2Coord)} ${removeLeadingZeroes(y2Coord)}" or removed)`;
                  }
                  if (command === lowerCurveCommand && (xCoord !== 0 || yCoord !== 0)) {
                    readableSegment += ` (should be "l${removeLeadingZeroes(xCoord)} ${removeLeadingZeroes(yCoord)}" or removed)`;
                  }
                  if (command === upperCurveCommand) {
                    readableSegment += ` (should be "L${removeLeadingZeroes(xCoord)} ${removeLeadingZeroes(yCoord)}" or removed)`;
```

So the message `Unexpected segment "c0 0, 0 0, 0.007 -0.007" in path (should be "l0.007 -0.007" or removed).` (see previous suggestion) would be `Unexpected segment "c0 0, 0 0, 0.007 -0.007" in path (should be "l.007 -.007" or removed).`.

.svglintrc.js
mondeja, line 290: Another minor suggestion. "Unexpected" seems a bit ambiguous... If we use "Ineffective" to maintain the consistence?

```suggestion
                reporter.error(`Ineffective segment ${readableSegment} in path.`);
```

service-paradis: Much better, thanks for the suggestion!

service-paradis: Thanks again for the review. I applied your suggestions!

mondeja: Sorry for the mix of variable names of three reviews. I'm seeing now meesages like `Ineffective segment "c0 0, 0 0, 0.007 -0.007should be "l.007 -.007" or removed" in path (should be removed).`. Seems that you should use `resolutionTip` variable in the conditionals as it's shown in this review.

service-paradis: Oh, it is ok. I already fixed it, but did not pushed it 🤦‍♂️

mondeja (approved):
"fix brain" :rofl: 

Now, it's perfect to me! Great work @service-paradis :tada: :muscle:

There may be an edge case that we are ignoring, but after another review I can assure that there are no false positives. If we are ignoring any case, we will fix it in the future. Feel free to merge it as is.

service-paradis:
> "fix brain" 🤣
> 
> Now, it's perfect to me! Great work @service-paradis 🎉 💪
> 
> There may be an edge case that we are ignoring, but after another review I can assure that there are no false positives. If we are ignoring any case, we will fix it in the future. Feel free to merge it as is.

Yes, I think, there are probably some other improvements and edge cases to cover. As you said, we can keep those for the future for now. 

Thank you very much for your help!! Merging 😄

PR #4267, opened 2020-12-04
Add Jaguar

### LINKED ISSUE #3174: jaguar
**Name:roger**
**Website:support27**
  

https://www.jaguar.com/market-selector.html

fbernhart:
jaguar

Closes #3174
**Alexa rank:** [81,199](https://www.alexa.com/siteinfo/jaguar.com)

### Checklist

### Description
**Source:** I couldn't find any SVG version of the logo [on the website](https://www.jaguar.com/index.html). Only PNGs.
But they do have [a media center](https://media.jaguar.com/en/press-kit), where you can download the press kits of the new released cars. So I simply downloaded one of these ZIP files. Inside are some images and two PDF files, that contain a vectorized version of the logo. So I've extracted the logo from the PDF file with Inkscape.

 **Color:** The logo [on the website](https://www.jaguar.com/index.html) has a white background. But they also use the kind of 3D logo used [here](https://media.jaguar.com/en) sometimes, which has all kind of grey/silver in it.

For the sake of simplicity and because I don't know which one to choose out of the 50 shades of grey, I went for white `#FFFFFF`.


fbernhart:
The initial PR is without the wordmark "Jaguar". With the wordmark, it would look like this:
jaguar (2)

Any opinions on which one would be better?

I'd slightly prefer the one with the wordmark, as the logo is more recognisable that way.

adamrusted:
> I'd slightly prefer the one with the wordmark, as the logo is more recognisable that way.

I think I agree with you - while the jaguar is recognizable on its own in most settings, it may be difficult to distinguish as an icon. @simple-icons/maintainers - any other opinions?

ericcornelissen:
I'd usually be against including the wordmark, but somehow I feel like for Jaguar the wordmark should be included... That said, I'd be fine either way :+1:

fbernhart:
Done 🐅

adamrusted:
@fbernhart which PDF did you download? I've just tried `
PRESS KIT: NEW JAGUAR F-TYPE 21MY (PDF)` and can't find an SVG in there...

adamrusted (approved):
LGTM 🎉🐱‍👤

PR #4271, opened 2020-12-04
Update Dockerfile

ericcornelissen:
This is just a couple of small suggestions to slightly improve the Docker experience. I don't think using Docker for simple icons is super common, but I tried it once and noticed some things are not that nice if anyone tries. Hence, I figured I submit a PR with some quality of life changes.

Most notably, this changes the Docker container base from `alpine` to `node:12` which should make the NodeJS version for development a bit more consistent.


mondeja (approved):
Faster than current configuration. Looks good @ericcornelissen :clap:

PR #4278, opened 2020-12-06
Specify guidelines on (registered) trademarks

### LINKED ISSUE #2460: On (Registered) Trademark Symbols

ericcornelissen:
Closes #2460, this adds our guidelines on whether or not to include (registered) trademark symbols in logos to the Contributing Guidelines as per the discussion there.

This also slightly changes the document structure of the Contributing Guidelines with the goal of making it a little bit easier to find this new information.

Suggestion regarding the wording, location, and document structure are welcome 🙂

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,19 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Wikimedia (which should provide a source)
 1. GitHub repositories
 
+#### Icon Guidelines
+
 Working with an SVG version of the logo is best. In the absence of an SVG version, other vector filetypes may work as well (e.g. EPS, AI, PDF). In the absence of vector logos, a vector can be created from a high quality rasterized image, however this is much more labor intensive.
+
+If the icon includes a (registered) trademark icon we follow the guidelines below to decide whether to include the symbol or not:
+
+* If brand guidelines explicitly require including the symbol, it must be included.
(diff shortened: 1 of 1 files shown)


mondeja (approved):
For me it's ok :+1:

fbernhart:
LGTM! 👍🏼

service-paradis (approved):

PR #4309, opened 2020-12-10
Fix errors in website search and remove uneeded statements

mondeja:
- Remove a repeated assign: `{% assign hex = icon.hex %}`
- Cleaned double chained `toLowerCase` in website Javascript `normalizeSearchTerm` function.
- Solves errors with search functionality shown in GIF below:
  - When you searches for the first time (previous query is empty) and the **selected** sorter is "order-by-color" , loaded from local storage, icons are ordered by relevance but the "order-by-color" sorter is not **selected**.  (GIF)
  - When you searches for the first time (previous query is empty) and there is no **selected** sorter because your local storage does not contains simple-icons data, icons are ordered by relevance but the "order-by-color" sorter is not **selected**.
  - When you clears the input after a second search, icons are unsorted and no sorter is **selected**. (GIF)

## Before

simple-icons-web-search-error

## After

simple-icons-web-fixed


site_script.js
```diff
@@ -111,7 +111,7 @@
     $body.classList.toggle('search__active', matchedIcons.length < icons.length);
 
     if (query === '') {
-      if ($orderByRelevance.classList.contains('active')) {
+      if ($body.classList.contains('order-by-relevance')) {
```
ericcornelissen, line 114: Relatedly, [could you remove the `active` class here](https://github.com/simple-icons/simple-icons/blob/0b54463efd721c008b387f1b2bfabb8537280cd9/index.html#L217) and add `class="order-by-color"` to the body? This bug causes none of the ordering buttons to be highlighted on the first load.

ericcornelissen (approved):
Great, thanks! I have just one request for an additional change if you don't mind.

<sup>Some of those mistakes are quite embarrassing :sweat_smile:</sup>

site_script.js
ericcornelissen: @mondeja, it seems you didn't `class="order-by-color"` to the body?

mondeja: Right, it's done! :hammer_and_wrench:

PR #4318, opened 2020-12-11
Fix multiple braces in single lines in '_data/simple-icons.json'

mondeja:
Fix a minor formatting error introduced in #4310.

PR #4320, opened 2020-12-11
Add lint to check JSON data file prettification

mondeja:
To avoid #4318, I've been looking for something that can check JSON files format, but I haven't found it, so I've created a simple test to keep the data file prettification consistency. Also I've removed a redundant configuration in `.editorconfig`. If anyone finds another cleaner solution, I will be happy to use it.

Next failure would be raised for #4310: 

<details>
  <summary>Failed test message</summary>

```
 FAIL  tests/meta.test.js
  ● Data file must be consistent › Data file is prettified

    expect(received).toEqual(expected) // deep equality

    - Expected  - 4
    + Received  + 2

    @@ -1667,17 +1667,15 @@
              },
              {
                  "title": "DHL",
                  "hex": "FFCC00",
                  "source": "https://www.dpdhl-brands.com/dhl/en/guides/design-basics/logo-and-claim.html"
    -         },
    -         {
    +         }, {
                  "title": "diagrams.net",
                  "hex": "F08705",
                  "source": "https://github.com/jgraph/drawio/blob/4743eba8d5eaa497dc003df7bf7295b695c59bea/src/main/webapp/images/drawlogo.svg"
    -         },
    -         {
    +         }, {
                  "title": "Diaspora",
                  "hex": "000000",
                  "source": "https://wiki.diasporafoundation.org/Branding"
              },
              {

      15 |   test(`Data file is prettified`, () => {
      16 |     const prettyObject = `${JSON.stringify(jsonDataObject, null, '    ')}\n`;
    > 17 |     expect(jsonDataString).toEqual(prettyObject);
         |                            ^
      18 |   });
      19 | })
      20 | 

      at Object.<anonymous> (tests/meta.test.js:17:28)
```

</details>


ericcornelissen:
I find it a bit odd that this is implemented as a test rather than a linting step. I would suggest a check like this is instead added to the [script/lint.js](https://github.com/simple-icons/simple-icons/blob/25213c0cb09e960a848227c7c09848d3f305a605/scripts/lint.js) file so that it is run when you do `npm run our-lint`.

mondeja:
> I find it a bit odd that this is implemented as a test rather than a linting step. I would suggest a check like this is instead added to the script/lint.js file so that it is run when you do npm run our-lint.

:thinking: Let me check if I can implement it keeping an error message similar to Jest...

mondeja:
Fixed, I've used [jest-diff](https://github.com/facebook/jest/tree/master/packages/jest-diff) directly. Now the error message is even cleaner:

<details>
  <summary>Error message</summary>

```diff
Data file is not prettified:
- Expected
+ Received

@@ -1667,17 +1667,15 @@
          },
          {
              "title": "DHL",
              "hex": "FFCC00",
              "source": "https://www.dpdhl-brands.com/dhl/en/guides/design-basics/logo-and-claim.html"
-         },
-         {
+         }, {
              "title": "diagrams.net",
              "hex": "F08705",
              "source": "https://github.com/jgraph/drawio/blob/4743eba8d5eaa497dc003df7bf7295b695c59bea/src/main/webapp/images/drawlogo.svg"
-         },
-         {
+         }, {
              "title": "Diaspora",
              "hex": "000000",
              "source": "https://wiki.diasporafoundation.org/Branding"
          },
          {
```
</details>

mondeja:
I've inverted the diff so now shows how to solve the problem. What do you think?

<details>
  <summary>Error message</summary>

```diff
Data file is not prettified:

- Received
+ Expected

@@ -1667,15 +1667,17 @@
          },
          {
              "title": "DHL",
              "hex": "FFCC00",
              "source": "https://www.dpdhl-brands.com/dhl/en/guides/design-basics/logo-and-claim.html"
-         }, {
+         },
+         {
              "title": "diagrams.net",
              "hex": "F08705",
              "source": "https://github.com/jgraph/drawio/blob/4743eba8d5eaa497dc003df7bf7295b695c59bea/src/main/webapp/images/drawlogo.svg"
-         }, {
+         },
+         {
              "title": "Diaspora",
              "hex": "000000",
              "source": "https://wiki.diasporafoundation.org/Branding"
          },
          {
```
</details>

ericcornelissen:
I'm having a small issue running this on my Windows machine where it reports the complete file as being improperly formatted due to different line endings (which are automatically correct when committing). It doesn't seem jest-diff has a built-in solution for that, I wonder if we should manually normalize the newlines in the script... 🤔 

> I've inverted the diff so now shows how to solve the problem. What do you think?

That makes more sense 👍 


> Fixed, I've used jest-diff directly. Now the error message is even cleaner:

I think we should add `jest-diff` explicitly as a devDependency then. I tried the [diff](https://www.npmjs.com/package/diff) package but for some reason it is super slow so I'm fine with using the jest-variant. Also, given the [jest-diff options](https://www.npmjs.com/package/jest-diff#options) I'd say we can omit explicitly setting the colors (and potentially annotations).

mondeja:
> I'm having a small issue running this on my Windows machine where it reports the complete file as being improperly formatted due to different line endings (which are automatically correct when committing). It doesn't seem jest-diff has a built-in solution for that, I wonder if we should manually normalize the newlines in the script... 🤔 

Since we use `editorconfig-checker` as a linter, adding `end_of_line=lf` in global section of `.editorconfig` should solve the problem. Can you try that?

mondeja:
> I think we should add jest-diff explicitly as a devDependency then.

Right, this is done :+1:

> I'd say we can omit explicitly setting the colors (and potentially annotations).

I've removed colors and annotations in last commit. I think that now is much less understandable, but I don't know. What are your concerns regarding the exclusion of colors and annotations? Now it's shown as (all in red):

<details>
  <summary>Error message</summary>

```
Data file is not prettified:

@@ -1667,15 +1667,17 @@
          },
          {
              "title": "DHL",
              "hex": "FFCC00",
              "source": "https://www.dpdhl-brands.com/dhl/en/guides/design-basics/logo-and-claim.html"
-         }, {
+         },
+         {
              "title": "diagrams.net",
              "hex": "F08705",
              "source": "https://github.com/jgraph/drawio/blob/4743eba8d5eaa497dc003df7bf7295b695c59bea/src/main/webapp/images/drawlogo.svg"
-         }, {
+         },
+         {
              "title": "Diaspora",
              "hex": "000000",
              "source": "https://wiki.diasporafoundation.org/Branding"
          },
          {
```
</details>

ericcornelissen:
> > I'm having a small issue running this on my Windows machine where it reports the complete file as being improperly formatted due to different line endings (which are automatically correct when committing). It doesn't seem jest-diff has a built-in solution for that, I wonder if we should manually normalize the newlines in the script... 🤔
> 
> Since we use `editorconfig-checker` as a linter, adding `end_of_line=lf` in global section of `.editorconfig` should solve the problem. If you are using a text editor with `editorconfig` configured, could you try that?

I do but I have never touched the data file so my editor never had the change to convert them. Moreover, I share a filesystem between Ubuntu and Windows for programming projects, and even though I have edited the file on Ubuntu, Windows is still reporting them as CRLFs...

I also cloned a fresh copy to verify the problem and it is there:

```shell
git clone git@github.com:simple-icons/simple-icons.git
cd simple-icons
# checkout this PR
git merge develop
npm install
npm run our-lint
# diff of the complete file
```


> Also, consider to set LF line endings working with GIT in Windows with (I think):
> 
> ```
> git config core.eol lf
> git config core.autocrlf input
> ```

I tried this as well and I still have the same problem. Even after cloning freshly after updating the configuration.


> I've removed colors and annotations in last commit. I think that now is much less understandable, but I don't know. What are your concerns regarding the exclusion of colors and annotations? Now it's shown as (all in red):

Does it not work if you just omit the color fields entirely? The documentation says the default values are exactly as you provided them originally.

mondeja:
> Does it not work if you just omit the color fields entirely? The documentation says the default values are exactly as you provided them originally.

Not, if I omit the color fields is like if I pass the default values defined [here](https://www.npmjs.com/package/jest-diff#properties-of-options-object). Where says "the default values are exactly as you provided them originally"?

> I tried this as well and I still have the same problem. Even after cloning freshly after updating the configuration.

I've replaced `\r\n` with `\n` in the prettified string. Can you try with this change?

scripts/lint.js
```diff
@@ -22,11 +29,31 @@ const TESTS = {
...
+      const dataDiff = diffLinesUnified(simpleIconsDataString.split("\n"),
+                                        simpleIconsDataPretty.split("\n"),
+                                        {
+                                          expand: false,
+                                          aColor: (v) => v,
+                                          bColor: (v) => v,
+                                          commonColor: (v) => v,
+                                          patchColor: (v) => v,
+                                          omitAnnotationLines: true
+                                        });
```
ericcornelissen, line 54: > Not, if I omit the color fields is like if I pass the default values defined here. Where says "the default values are exactly as you provided them originally"? If I'm not lost, options object uses color settings by default.

Sorry for the confusion, I meant essentially the suggestion below. I tested this on Ubuntu and it seems to work as expected.

```suggestion
                                        {
                                          expand: false,
                                          omitAnnotationLines: true
                                        });
```

> I've replaced \r\n with \n in the prettified string. Can you try with this change?

Will try this later when I'm on my Windows boot again :upside_down_face:

ericcornelissen (approved):
> I've replaced \r\n with \n in the prettified string. Can you try with this change?

Can confirm it work on both Windows and Ubuntu now. Let's hope it works for everyone on Windows now 🤞

mondeja:
Great! Thank you for the review @ericcornelissen :tada:

PR #4333, opened 2020-12-12
Minify icon HEX colors in website

mondeja:
Only affects `background-color` style property, not to color HEX strings displayed under icon names.


ericcornelissen (approved):
Looks good to me and seems to work just fine :+1: Nice optimization @mondeja, thanks :tada:

PR #4366, opened 2020-12-15
Add workflow to remove outdated labels

### LINKED ISSUE #4332: Remove impossible labels automatically
After reviewing a pull request, it is very common that I forget to remove some tags from it, such as when the pull was marked as `changes requested`, it becomes accepted, the merge is done and ... have I removed the tag? :thinking:

So, I've created [a simple Github Action](https://github.com/mondeja/remove-labels-gh-action) that just removes tags from issues and pull requests. Combining it with github events and filtering by their states, we can forget the need of remove the impossible tags for each context, shown below. There is in their README a possible configuration for `simple-icons`.

> You can see running with [this action](https://github.com/mondeja/test-gh-actions-webhooks/runs/1542804434?check_suite_focus=true) for [this issue](https://github.com/mondeja/test-gh-actions-webhooks/issues/9) when I close it.

What do you think? I'm too lazy :sweat_smile: and this is something too intrusive, or could improve the review speed? Is there a label that I am forgetting for some context?

### Pull request closed

- `in discussion`

### Pull request merged

- `awaiting reply`
- `changes requested`
- `duplicate`
- `in discussion`
- `invalid`
- `out of scope`
- `pending`
- `won't add`

### Issue closed

- `in discussion`
- `pending`
- `tracking`

mondeja:
Closes #4332

This add a workflow to remove automatically impossible labels of pull requests and issues. I've renamed the `labeler.yml` file by `add-labels.yml` so both workflows are more clearly differentiated.

Because there is no way to fully test this on a branch, I've created [this repo](https://github.com/mondeja/test-gh-actions-webhooks) which has this same configuration included. Just let me know if you want admin permissions to play labeling, opening, closing/merging issues and pull requests.


fbernhart:
Great work! 💪🏼 Let me create some issues.

mondeja:
> Great work! 💪🏼 Let me create some issues.

Invited, keep in mind that only is executed closing pull requests and issues.

ericcornelissen:
One initial comment: Did you consider using the v1 tag for your action to keep the action up-to-date within the v1.x.x range automatically?

mondeja:
> Did you consider using the v1 tag for your action to keep the action up-to-date within the v1.x.x range automatically?

I'm releasing the action with `v1` tag also and I'm not against specify it like so here because it's my own action, but since renovates and dependabot also updates actions versions, I'm always using the full version as a good practice so as not to be surprised if someone releases a buggy version and to be more explicit for facilitate the investigation if something fails.

ericcornelissen:
That's fair I'm also not against leaving it like it is currently. My only argument against that would be that I don't think Renovate's PR would really allow us to spot any bugs...

mondeja:
I've used [`pull_request_target`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) event for pull requests because `pull_request` event is executed in the branch of the user that open the pull, and the `GITHUB_TOKEN` only have read permissions for the repository, the labels can't be deleted in these events. Using `pull_request` event, raises [this error](https://github.com/actions/labeler/issues/12).

ericcornelissen:
> I've used [`pull_request_target`](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) event for pull requests because `pull_request` event is executed in the branch of the user that has opened the pull, and the `GITHUB_TOKEN` only have read permissions for the repository, so the labels can't be deleted in these events, raising [this error](https://github.com/actions/labeler/issues/12).

Good catch, we do the same in [the existing labaler workflow](https://github.com/simple-icons/simple-icons/blob/096199cfea4dc94a4f4f67d2314cef11d6c29130/.github/workflows/labeler.yml#L3) :+1:

mondeja:
> I don't think the comment is necessary, but you don't have to remove it.

I didn't realize. If the other workflow has the same event, better remove it :+1:

.github/workflows/remove-labels.yml
```diff
@@ -0,0 +1,50 @@
...
+
+  remove-closed-pr-labels:
+    name: Remove closed pull request labels
+    if: github.event_name == 'pull_request_target' && (! github.event.pull_request.merged)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mondeja/remove-labels-gh-action@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          labels: in discussion
```
ericcornelissen, line 37: Thinking about it more, should we include `pending` here as well?

```suggestion
          labels: |
            in discussion
            pending
```

mondeja: Absolutly!

ericcornelissen (approved):
Looks good to me :+1:

> Great work! 💪🏼 Let me create some issues.

@fbernhart, did you find anything interesting that we should consider here before merging this in?

fbernhart:
@ericcornelissen  No. I think it's ready to be merged.

I haven't looked at the code myself, as I'm not familiar with GitHub actions. But I've tested it a bit here https://github.com/mondeja/test-gh-actions-webhooks and it seems to work as expected. 👍🏼

ericcornelissen:
In that case, let's merge this and see how it works in practice, we can always refine the configuration later :+1: Thanks once more for your work @mondeja :tada:

mondeja:
Thanks for the help testing this @fbernhart! :100:

mondeja:
Seems that [it's working](https://github.com/simple-icons/simple-icons/pull/4086#event-4130500273) for closed pull requests :tada:

fbernhart:
@mondeja I've been wondering why the "in discussion" label wasn't removed automatically from #3428 after closing it. Do you have any idea? 🤔

mondeja:
Because the workflow [has not been added to the branch](https://github.com/FeedsPub/simple-icons/tree/develop/.github/workflows). Only should work it the *develop* branch has been merged into the pull.

fbernhart:
Oh, ok. Thanks for the explanation! 👍🏼

Issue #4205, opened 2020-11-30
Google Icons Review (again)


service-paradis:
Google recently updated a lot of their logos. Following on from #1919, #3725 & #4134, we should do a full review of all our Google icons (again #2364).

For Google Brands, we should follow Visual Identity Guidelines found at https://partnermarketinghub.withgoogle.com/#/brands or their new site https://thepartnermarketinghub.withgoogle.com/brands/

For Google products, it might be a little bit harder to find good icons to based our work on since there is no specific guidelines pages. We should use the icons from the new redesign. This page was still outdated a few days ago, but looks to have been updated now: https://about.google/products/#all-products

Some of the item listed below might not need to be updated, but should be reviewed.

### Here is the list of Google Brands we currently have:

- [x] Android - #4399
- [x] Android Auto - #4400
- [x] Google Assistant -  #4401
- [x] Google Fit - #4506
- [x] Google Lens - #4402
- [x] Google News - #4474
- [x] Google Pay - #4403
- [x] Google Play - #4505
- [x] Google Stadia - #4503
- [x] Wear OS - doesn't need editing @adamrusted 
- [x] YouTube Music - #4404
- [x] YouTube TV - #4405

### Here is the list of Google Products we currently have:

- [x] Android Studio - #4469
- [x] Gmail - #4470
- [x] Google - #4508
- [x] Google Ads - #4471
- [x] Google AdSense - #4472
- [x] Google Analytics - #4504
- [x] Google Calendar - #4522
- [x] Google Cardboard - #4058
- [x] Chromecast (formerly Google Cast) - #6512
- [x] Google Chat - #4695
- [x] Google Chrome - #4525
- [x] Google Classroom - #6512 
- [x] Google Cloud - Seems to be the same @jorgeamadosoria 
- [x] Google Drive - #6596 
- [x] Google Earth - Seems to be the same @jorgeamadosoria 
- [x] Google Hangouts - Seems to be the same @jorgeamadosoria 
- [x] Google Keep - #6599 
- [x] Google Maps - Seems to be the same @jorgeamadosoria 
- [x] Google Marketing Platform - #6599 
- [x] Google Meet - #6599 
- [x] Google Messages - #6599 
- [x] Google My Business - Seems to be the same @jorgeamadosoria 
- [x] Google Nearby - #4059
- [x] Google Podcasts - Seems to be the same @jorgeamadosoria 
- [x] Google Scholar - #4686
[...]

PR #4403, opened 2020-12-18
Updated Google Pay

adamrusted:
googlepay

**Issue:** #4205
**Alexa rank:** N/A

### Checklist
  - [ ] I updated the JSON data in `_data/simple-icons.json` --> Not Necessary

### Description
Slight change to logo (Added 0.35px lines where colours change in the 'G').
Generated in Adobe Illustrator.
Ran through SVGO CLI & SVGlint.


mondeja (changes requested):
Good work with these gaps @adamrusted :+1:

I'm seeing some differences against the source in the letter "G". Can you check them?

image

mondeja:
Before continuing with the review, although Google doesn't prohibit it, I want to know what other @simple-icons/maintainers think about that gaps. We should allow them or not? Note that the previous version did not include them and they could be a subjective distortion of the symbol :thinking:

adamrusted:
> Before continuing with the review, although Google doesn't prohibit it, I want to know what other @simple-icons/maintainers think about that gaps. We should allow them or not? Note that the previous version did not include them and they could be a subjective distortion of the symbol 🤔

Totally agree, I added the gaps purely to distinguish where colours would change, as I've seen discussed on other icons. If you want to stick with the current icon, the path itself is absolutely fine.

ericcornelissen:
Our [current version of just the Google logo](https://github.com/simple-icons/simple-icons/blob/6379ae7022b7bf8bca9e46c74dc53874265f66d8/icons/google.svg) also doesn't have it. I'd say the logo is recognizable without the separations and in my opinion looks better. Since there are no explicit guidelines on how to do the logo monochrome I would say we leave out the gaps. Moreover, I think there are examples of Google using the "G" monochrome, e.g. on [the back of Pixel phones](https://images.app.goo.gl/7bLJaXU1kxiFmRwaA).


> If you want to stick with the current icon, the path itself is absolutely fine.

To clarify, you mean [the existing path](https://github.com/simple-icons/simple-icons/blob/6379ae7022b7bf8bca9e46c74dc53874265f66d8/icons/googlepay.svg) is fine and doesn't need to be updated?

adamrusted:
> To clarify, you mean [the existing path](https://github.com/simple-icons/simple-icons/blob/6379ae7022b7bf8bca9e46c74dc53874265f66d8/icons/googlepay.svg) is fine and doesn't need to be updated?

Yep. Only difference between the two is the addition of the gaps.

PR #4416, opened 2020-12-20
Formalize website Ruby dependencies and installation process

mondeja:
- The documentation is explaining how to build the website using a global-installed Jekyll version. An user could be using another Jekyll version and we are probably forcing to install another version for website dependencies.
- We are using the latest version of Jekyll, also in the CI. This should bring problems to us in the future, when next major versions of Jekyll be released.

So this pull request tries to solve these things by:

- Adding `Gemfile` and `Gemfile.lock`, so the user only must configure locally the location where Jekyll dependency will be installed, using that version to build with `bundle`. Renovate [supports them](https://docs.renovatebot.com/ruby/) so we let it do the work of updating versions.
- Updating CI because now the Gemfile it's included yet.
- Updating `.gitignore`, adding Ruby environment related directories. I've taken the opportunity to eliminate ignorances of SASS files, because now are not used in the project.
- Updating documentation explaining how to build locally properly.


CONTRIBUTING.md
```diff
@@ -244,8 +244,10 @@ If you have an affiliation to the brand you contributing that allows you to spea
 ## Building Website Locally
 
 * Make sure you have [Ruby](https://www.ruby-lang.org/en/downloads/) installed.
-* Make sure you have [Jekyll](https://jekyllrb.com/) installed (using `$ gem install jekyll bundler`). At least version 3.5 is required.
-* Build and run the website locally using `$ jekyll serve`.
+* Make sure you have [Bundler](https://bundler.io/) installed (using `$ gem install bundler`).
+* Set locally the gems installation location for simple-icons website using `$ bundle config set path 'vendor/bundle'`.
```
ericcornelissen, line 248: Currently the sentence is a bit awkward. Not sure if _"locally"_ is needed here, so I'd suggest removing it.

```suggestion
* Set the gems installation location for the simple-icons website using `$ bundle config set path 'vendor/bundle'`.
```

If you do want it I suggest

```suggestion
* Set your local gems installation location for the simple-icons website using `$ bundle config set path 'vendor/bundle'`.
```

ericcornelissen (changes requested):
Not sure how I feel about these changes, but I haven't use Jekyll in a while so it might be that all this is necessary, so I would be fine with merging this.

One separate suggestion, could you add the `Gemfile.lock` to the `.gitattributes` file before `package-lock.json` with the same `-diff` flag:

https://github.com/simple-icons/simple-icons/blob/e0753b06a116316f57f50fbb63267aa43bb8d8cd/.gitattributes#L6-L8

CONTRIBUTING.md
mondeja: Sure!

.gitattributes
```diff
@@ -5,6 +5,7 @@
 
 # Don't diff machine generated files
 package-lock.json  -diff
+Gemfile.lock  -diff
```
ericcornelissen, line 8: Minor nitpick, but given that everything is aligned here can you align this one as well? :sweat_smile: 

```suggestion
package-lock.json  -diff
Gemfile.lock       -diff
```

ericcornelissen (approved):
Looks good to me. The changes to the Contributing Guidelines might warrant a review from at least one more @simple-icons/maintainers, even if just to check for typos :slightly_smiling_face:

adamrusted:
>  even if just to check for typos 🙂

Looks good for typos. Got to start learning how to use Jekyll / Ruby myself so I actually know what half this stuff is doing! 😅

fbernhart:
As someone else that has no clue about Jekyll, I was wondering why the step "Install Jekyll" is taking so long in the "Build website" process (usually around 2,5 minutes) and if there would be ways to speed it up. 🤔

mondeja:
> I was wondering why the step "Install Jekyll" is taking so long in the "Build website" process (usually around 2,5 minutes) and if there would be ways to speed it up.

I'm asking this myself also. Thanks to your comment I have realized that, in theory, we can get the hash from the `Gemfile.lock` file that I'm adding in this pull. Perhaps the cache was not well configured, I've found [this issue](https://github.com/actions/cache/issues/348) opened in the cache action repository that could be related.

.github/workflows/verify.yml
```diff
@@ -34,12 +34,11 @@ jobs:
       uses: actions/cache@v2
       with:
         path: vendor/bundle
-        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        key: ${{ runner.os }}-gems-${{ hashFiles('Gemfile.lock') }}
```
mondeja, line 37: This should be more reliable.

mondeja:
I have ended up using the cache strategy included in `ruby/setup-ruby` action: https://github.com/ruby/setup-ruby/blob/master/README.md#caching-bundle-install-automatically
It's doing the same, but should be more reliable than configuring it manually. Seems that the Bundler example in `actions/cache` [is outdated](https://github.com/actions/cache/pull/396).

ericcornelissen:
> I have ended up using the cache strategy included in `ruby/setup-ruby` action: https://github.com/ruby/setup-ruby/blob/master/README.md#caching-bundle-install-automatically
> It's doing the same, but should be more reliable than configuring it manually. Seems that the Bundler example in `actions/cache` [is outdated](https://github.com/actions/cache/pull/396).

Though I'm not against the changes related to this comment, everything actually already started working better once you added `Gemfile.lock` in this PR - Just check the build logs for this PR. Personally I like the explicit "Setup cache" and "Install Jekyll" steps better as it is more clear about what is happening. It does not seem at all obvious that an option `bundler-cache: true` configures and restore a cache **and** installs Jekyll (and other gems).

Also: hindsight being 20-20, it actually makes some sense that the cache didn't work before as the key was configured based on a non-existent `Gemfile.lock`.

mondeja:
> Personally I like the explicit "Setup cache" and "Install Jekyll" steps better as it is more clear about what is happening. It does not seem at all obvious that an option bundler-cache: true configures and restore a cache and installs Jekyll (and other gems).

This is not hiding nothing @ericcornelissen, [you can still see all the commands](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:14).

Also, this configuration has improvements that we can avoid to dig into, for example [install gems without documentation](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:9), [more explicit configuration](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:14), configuration of [deployment](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:15). Although I personally like explicitement also, I think that the use in this case of this nice configuration could help more than hide.

ericcornelissen:
> This is not hiding nothing @ericcornelissen, [you can still see all the commands](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:14). Also, this configuration has improvements that we can avoid to dig into, for example [install gems without documentation](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:9), [more explicit configuration](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:14), configuration of [deployment](https://github.com/simple-icons/simple-icons/pull/4416/checks?check_run_id=1595094335#step:3:15). Although I personally like explicitement also, I think that the use in this case of this nice configuration could help more than hide.

Again, I'm fine with this PR as is, and indeed these additional improvements (which I hadn't noticed) are nice to have. Perhaps "hiding" was not the best word to use, I meant that it is hard(er) to tell from looking just at the Workflow file.

mondeja:
> I meant that it is hard(er) to tell from looking just at the Workflow file.

Maybe adding a comment pointing to the documentation of that configuration?

ericcornelissen:
> Maybe adding a comment pointing to the documentation of that configuration?

Could be helpful in this case yes :+1:

mondeja:
Done :writing_hand:

fbernhart:
Just wanted to give some quick feedback:

AMAZING how the step "Build website" went from around 3 - 3,5 minutes down to just about 10 seconds! 🤯 

🥇 🏆 🍾

PR #4422, opened 2020-12-21
Update Apache Cordova

mondeja:
apachecordova

**Issue:** #2887
**Alexa rank:** n/a

### Checklist

### Description
Updated logo using https://apache.org/logos/ as source.


_data/simple-icons.json
```diff
@@ -443,7 +443,7 @@
         {
             "title": "Apache Cordova",
             "hex": "E8E8E8",
-            "source": "https://cordova.apache.org/artwork/"
+            "source": "https://apache.org/logos/"
```
adamrusted, line 446: I'd leave the source as cordova.apache.com/artwork purely because there's a decent PSD of the monochrome version there.

mondeja: But the main purpose of #2887 is to unify all Apache icons under the same source.

adamrusted: As mentioned in #2887 though:
> As with the Google review, I think that if a project maintains its own design guidelines with vector artwork (e.g., [Flink](https://flink.apache.org/material.html)) then they should take precedence.

The PSD available [here](https://cordova.apache.org/artwork/) does contain proper paths for the monochrome version, whereas the Apache link doesn't.

mondeja: I see... So let's return to the old source!

mondeja:
:thinking: Unfortunately, I'm not able to work with `.psd` files, so I'm closing this. Maybe someone else can take care of it. It's a bit weird that Apache uses privative software to share their vectors, taking into account their licenses.

PR #4427, opened 2020-12-21
Unify 'svgpath' calls in linter

mondeja:
Minor optimization and code refactoring in linter. This avoid redundant calls to `svgpath` parser and replace calls to make use of their public API.


service-paradis (approved):
LGTM 💯

PR #4433, opened 2020-12-22
Suggest using "Treeless clone" to contributors

ericcornelissen:
I recently had the _pleasure_ of cloning this repository, but with close to 3.500 commits in the project that is getting pretty slow. So, following a recent [GitHub blog post](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/), this Pull Request adds a suggestion to the Contributing Guidelines to use "Treeless clone" when cloning this repository. Put simply, this creates a clone that knows everything about the latest commit, as well as the existence of al other commits. However, the clone doesn't know anything about the content of other commits. Since the commit history is generally not super important to contributors of this project (I think!?), this can significantly speed up cloning without any real downside

I fear that many contributors won't see this (or only see it after they clone) and so still have to sit through the long cloning process for this repo. I'm not sure if we can put it in a place where more people can see it... Any suggestions?

For completeness, I compared normal cloning and all the cloning options from the blog past using the [`time` command](https://man7.org/linux/man-pages/man1/time.1.html). Note that this doesn't tell you the whole picture because, as the blog post explains, each method comes with different down and upsides.

| Method | Name | `time` |
| --- | --- | --- |
| - | Normal clone | 48.649 total |
| `--depth=1` | Shallow clone | 5.491 total |
| `--filter=tree:0` | Treeless clone | 9.915 total |
| `--filter=blob:none` | Blobless clone | 12.111 total |

<sup>Apologies about the unrelated changes (removal of whitespace at the end of a line)</sup>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,6 +3,10 @@
 Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below. If you decide to contribute anything, please do the following:
 
 1. Fork this repository
+1. (Optional) Clone the fork
+   ```
+   git clone --filter=tree:0 git@github.com:simple-icons/simple-icons.git
+   ```
 1. Create a new branch from the latest `develop` (read more [here](https://guides.github.com/introduction/flow/))
 1. Start hacking on the new branch
 1. Commit and push to the new branch
@@ -31,11 +35,11 @@ We welcome icon requests. Before you submit a new issue please make sure the ico
        - Allowed: Space agencies
     - Symbols, including flags and banners
     - Sport clubs
-       - Allowed: Sport organizations 
+       - Allowed: Sport organizations
     - Yearly releases
     - Universities or other educational institutions
     - Any brands representing individuals rather than an organization, company or product. This includes musicians, bands, and social media personalities.
-    
+
 If you are in doubt, feel free to submit it and we'll have a look.
 
 When submitting a request for a new or updated icon include helpful information such as:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,9 @@ Simple Icons welcomes contributions and corrections. Before contributing, please
 
    # Using HTTPS
    git clone --filter=tree:0 https://github.com/simple-icons/simple-icons.git
+
+   # Using GitHub CLI
+   gh repo clone simple-icons/simple-icons -- --filter=tree:0
    ```
 1. Create a new branch from the latest `develop` (read more [here](https://guides.github.com/introduction/flow/))
 1. Start hacking on the new branch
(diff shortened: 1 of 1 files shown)


adamrusted:
Looks like it'll cut out a lot of the clone time for people who may only commit 1 or 2 changes, and there's always the option to clone using the default parameters should you *need* the full history. I think this is a good shout 👌

adamrusted (approved):
Looks good to me! 👌

CONTRIBUTING.md
```diff
@@ -3,6 +3,10 @@
 Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below. If you decide to contribute anything, please do the following:
 
 1. Fork this repository
+1. (Optional) Clone the fork
+   ```
+   git clone --filter=tree:0 git@github.com:simple-icons/simple-icons.git
```
mondeja, line 8: I think that the `git@` protocol could be troublesome for some users (like for example me :sweat_smile:). I'm not sure if something must be configured in GIT to make this works and if brings some benefit, but I always use HTTPS:

```suggestion
   git clone --filter=tree:0 https://github.com/simple-icons/simple-icons.git
```

ericcornelissen: Good catch, you're probably right. I'd suggest we provide both versions, that should cover 99.9% of contributors, agreed?

FYI. To use the `git@` protocol you need to configure a key (SSH or PGP) in your account (I'd recommend doing that, you can read more [in this article](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/adding-a-new-ssh-key-to-your-github-account))

mondeja: Yes, agreed with providing both versions :+1:

ericcornelissen: Alright updated it, added comments to be extra clear about which one is which. I also decided to add a GitHub CLI option for those using it, even if just to keep in line with the clone URLs the GitHub interface gives you :arrow_down: 

image

ericcornelissen: Also, I was very surprised by how fast HTTPS was for cloning compared to SSH :open_mouth: Though using `--filter=tree:0` still cut the time in half for HTTPS.

mondeja (approved):

CONTRIBUTING.md
```diff
@@ -3,6 +3,17 @@
 Simple Icons welcomes contributions and corrections. Before contributing, please make sure you have read the guidelines below. If you decide to contribute anything, please do the following:
 
 1. Fork this repository
+1. (Optional) Clone the fork
+   ```bash
```
mondeja, line 7: Just one minor suggestion more. Seems that if you add a newline here the render looks better, more separated from code block:

```suggestion

   ```bash
```



### Before this

image


### After this

image

ericcornelissen: Hmm, not sure if I agree as the large spacing between list elements looks worse to me. But if other @simple-icons/maintainers or contributors agree with you I'm happy to change it :slightly_smiling_face:

adamrusted: I personally prefer the larger spacing, but not a deal breaker either way for me.

ericcornelissen: Alright, I updated it in 74408f27d74b61fed82e64f5ee90dfb3b31eed21 :+1:

PR #4436, opened 2020-12-22
Update 'svg-path-bbox'

mondeja:
I've released `svg-path-bbox@0.2.0` and this version has slight differences in the computation of bounding boxes (differences lower than `0.0000001`), but this minor difference is causing that two bounds reach to `0.005` and the rounding to 3 decimals cause the linter to fail. So, before renovates could bring this update, breaking the lint, I've updated manually and added these two icons to `.svglint-ignored.json` file.


service-paradis (approved):
Thank you @mondeja!

Issue #3829, opened 2020-11-07
Netflix update


mondeja:
**Name:** Netflix
**Official resources for icon and color:** `#E50914`
[Symbol from brand guidelines](https://brand.netflix.com/en/assets/brand-symbol/)

Current Netflix icon in simple-icons is slightly different from symbol icon shared by Netflix brand guidelines.

fbernhart:
![netflix](https://user-images.githubusercontent.com/70264417/101504153-2e8dcf80-3973-11eb-9e09-9244b7593e93.png)

I gave it a first quick try. How about separating the color changes with gaps? Similar to #4281

mondeja:
Since Netflix does not prohibit it, that could be a good solution. At first glance it seems to me that the gaps are very wide, you've tried with less?

fbernhart:
![path1838](https://user-images.githubusercontent.com/70264417/101524226-f3e36180-3989-11eb-97bb-93e6a91ffc74.png)


@mondeja

mondeja:
Oh, very nice @fbernhart! I would opt by the third icon since seems that is the one that looks more like the original. Maybe the third with a bit less gap wide? :thinking:  I don't know. Feel free to select the one that suits you best!

fbernhart:
Really? I would have gone for the second one.

The gaps are as following: 0.65px, 0.5px, 0.35px, 0.2px

So I would choose 0.5px. I would be as well ok with 0.35px as gap (the third one). But I wouldn't choose smaller gaps.

Would it be possible to define a default value for all the new icons that we add with these kind of gaps? So that they kind of look similar? I would adopt #4281 accordingly then.

Maybe the other @simple-icons/maintainers could as well give their opinions, if we want to define some kind of "standard".

mondeja:
> Would it be possible to define a default value for all the new icons that we add with these kind of gaps?

I suppose that if this kind of standard could be defined, would be relative to surrounding lines.

ericcornelissen:
I would be inclined to go with a smaller gap as well (not 0.2px though), I agree that 0.5px seems like a good compromise :+1:

fbernhart:
Ok, let's go with 0,35px then! 👍🏼  I've even changed my mind: 0,35px now looks better to me than 0,5px.

I'll create a PR.

runxel:
> I suppose that if this kind of standard could be defined, would be relative to surrounding filled geometries.

[...]

Issue #1141, opened 2018-12-24
Sketch App icon


LucasZapico:
**Name:** __Sketch App__
**Website:** [Sketch App Website](https://www.sketchapp.com/)
**Official resources for icon and color:**
  
[Press kit](https://www.sketchapp.com/press/)

ericcornelissen:
We considered adding it in #788, but their logo guidelines explicitly state the following:

> We would appreciate it if you don’t modify or embellish Sketch brand assets in any way, such as by changing the design or color of our symbol, thanks.

Hence we decided not to add Sketch

davidklebanoff:
If anyone feels strongly about getting this icon added, I would suggest contacting Sketch and asking how the icon should be presented in a monochromatic format. Sketch is indeed a popular brand and we'd like to add it to Simple Icons, but we make a point not to add unofficial icons to the project. Currently they don't seem to offer a monochromatic (or monochromatic-able) version.

LucasZapico:
@ericcornelissen @davidklebanoff, 
I'll reach out to them and inquire in regards to a monochromatic version.

PeterShaggyNoble:
@LucasZapico, did you ever get a response to your enquiry?

PeterShaggyNoble:
I'm closing this out for now as it has been inactive for 10 months, with no response from Sketch. @LucasZapico, if you do get a response then please reopen this issue.

PeterShaggyNoble:
Looks like I may have too hasty in closing this - despite the lack of reply from them to @LucasZapico, Sketch have since added a monochrome version of their logo to their [press page](https://www.sketch.com/press/). However, there is no explicit mention on that page of using the monochrome symbol on its own, without the wordmark, other than:

> Use the mark only if the space is limited.

Reopening this for further discussion.

ericcornelissen:
Good catch @PeterShaggyNoble :ok_hand:

> However, there is no explicit mention on that page of using the monochrome symbol on its own, without the wordmark, other than:
>
> >    Use the mark only if the space is limited.

From that note I would say we are safe if we omit the wordmark from the SVG

PeterShaggyNoble:
Next thing we have to figure out then, is the colour. Options from their logo are:
- `#FDB300`
- `#EA6C00`
- `#FDAD00`
- `#FDD231`
- `#FEEEB7`

[...]

PR #4480, opened 2020-12-26
Remove Sketch

adamrusted:
sketch

**Alexa rank:** [5,670](https://www.alexa.com/siteinfo/sketch.com)

See https://github.com/simple-icons/simple-icons/pull/4480#issuecomment-752518958 for reason behind removal.


mondeja:
I like more this version than the previous one, but these gaps are not visible in small sizes. Have you considered using wider gaps?

adamrusted:
> I like more this version than the previous one, but these gaps are not visible in small sizes. Have you considered using wider gaps?

I can use wider gaps - set 0.35px following previous discussions I've seen. Will try with 0.5px when I'm in front of my PC next.

mondeja:
It's something that we should discuss to clarify these cases. See [this comment](https://github.com/simple-icons/simple-icons/issues/3829#issuecomment-749561677).

ericcornelissen:
Regarding this icons, the discussion in #1141 says that the current version was, at the time, found in the Sketch brand guidelines as is So in that regard I think we should stick with it. However, the new brand guidelines at https://www.sketch.com/about-us/#press seem to specify that a monochrome version like this is not allowed...

adamrusted:
> However, the new brand guidelines at [sketch.com/about-us/#press](https://www.sketch.com/about-us/#press) seem to specify that a monochrome version like this is not allowed...

Well spotted! 👀
Should I look at marking this for removal in v5.0.0 then?

PeterShaggyNoble:
They're getting as bad as Medium for constantly changing their branding!

PeterShaggyNoble:
Looks like we might actually be able to keep this one as they now include monochrome versions in [the logos & guidelines file](https://cdn.sketchapp.com/assets/pages/about-us/logos.zip).

adamrusted:
If other @simple-icons/maintainers are happy to keep it, maybe if we properly link to the guidelines, then this can be closed - as our version matches up perfectly to the logo they provide.

ericcornelissen:
The presence of a monochrome version in the logos & guidelines file is definitely enough for keeping a monochrome version here, I would say :+1: Good catch @PeterShaggyNoble, glad we can keep this one :smile: 

With three votes in favor of keeping the Sketch SVG I will close this for now - should be easy enough to reopen/replicate this PR later anyway.

PR #4484, opened 2020-12-26
Updated Known

adamrusted:
known

**Issue:** Contributes to #3169
**Alexa rank:** [26,299](https://www.alexa.com/siteinfo/withknown.com)

### Checklist

### Description
Changed source to GitHub repo where logos are stored.
Regenerated icon using `logo_k_192.png`.
Ran through SVGO CLI.


mondeja (changes requested):
Good work @adamrusted :+1:

I'm seeing some details that could be improved to more closely match the original. In next images, black is the PNG source, green is my version and red is yours:

1. This corner could be rounded:

image

2. These two corners could be aligned to make the central line of the "K" more consistent:

image

image

Take into account that in this coordinate there are two points overlapping:

image

3. This peak is too steep, it should be 90 degree or slightly rounded:

image

4. This corner is too much rounded:

image

and it seems to me that this corner in the same place could be rounded instead of using two coordinates:

image

5. Since this is a rect line, these two coordinates could be removed improving the match and optimizing the resulting path:

image

adamrusted:
So I'm just in the process from rebuilding the logo from scratch - Illustrator's Image Trace function doesn't want to work with me this evening. Here's a comparison of my path against the original PNG for reference.

image

I'll submit this path in my next PR - so feel free to go through it 🤞🏻

mondeja (changes requested):
Great work @adamrusted :muscle:

I'm seeing only two minor errors in the geometry:

- Here there are two overlapping coordinates:

image

- This part of the geometry is not properly aligned:

image

mondeja (approved):
Awesome :tada:

I'm sure it could be improved, but without better references any suggestion seems subjective to me. Thanks for the effort @adamrusted :pray:

PR #4499, opened 2020-12-27
Updated Delicious

adamrusted:
delicious (1)

**Issue:** Contributes to #3169
**Alexa rank:** [74,351](https://www.alexa.com/siteinfo/icio.us)

### Checklist

### Description
Changed source to official website.
Built from scratch, added border to define edges of shape.
Color updated to blue used on website.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1662,8 +1662,8 @@
         },
         {
             "title": "Delicious",
-            "hex": "3399FF",
-            "source": "https://en.wikipedia.org/wiki/Delicious_(website)"
+            "hex": "0000FF",
+            "source": "http://del.icio.us/"
         },
         {
             "title": "Deliveroo",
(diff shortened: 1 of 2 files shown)


fbernhart:
Just wondering, how about such an approach with a gap of 0.5px between color changes? Would this better represent [Delicous](http://del.icio.us/) or is this already Microsoft? 😂  @simple-icons/maintainers opinions?

delicious

PeterShaggyNoble:
Yeah, that's definitely too close to Microsoft to me, @fbernhart. I also don't think that outlining the lighter regions is the way to go either; our current treatment works better for me.

We should also probably flag this one for monitoring for potential removal if it doesn't come back online.

adamrusted:
> I also don't think that outlining the lighter regions is the way to go either; our current treatment works better for me.

Fine by me - I'll remove the edit to the SVG, and just update the source.

> We should also probably flag this one for monitoring for potential removal if it doesn't come back online.

Happy to hold fire on the merging on this until we get a better idea of what's going on. Given it's been down since mid-December, I'd say we give it until mid-February / early March for them to update in some way.

fbernhart:
I would still merge it now and just recheck before releasing v5.0 at the end of May.

If by then the website is still not up and the Alexa rank has dropped out of scope, we can still create a new PR to remove it.

adamrusted:
Should be good to go. Just removed the icon - left the website as source and updated colour though.

PeterShaggyNoble:
If, come v5, we _are_ able to keep this one then it should be renamed as "del.icio.us".

PR #4501, opened 2020-12-27
Updated Apache Airflow

adamrusted:
apacheairflow (1)

**Issue:** Contributes to #2887
**Alexa rank:** N/A

### Checklist

### Description
Updated source URL.
Rebuilt from source, with 0.35px lines.
Ran through SVGO CLI.


mondeja (changes requested):
Good work @adamrusted! :+1:
The geometry is almost perfect, I'm only seeing some minor imperfections that could be solved in the places marked with arrows:

image

It seems that the lines coming from the outside exceed the perpendicular, could you solve it?

image

mondeja (changes requested):
Perfect, just one minor suggestion more. I've found three overlapping coordinates here:

image

which are not necessary, could be removed

image

mondeja (approved):
And with this merge #2887 is completed! :tada:
Great work with this @adamrusted :muscle:

PR #4506, opened 2020-12-27
Updated Google Fit

adamrusted:
googlefit

**Issue:** Contributes to #4205
**Alexa rank:** N/A

### Checklist

### Description
Refined source link.
Rebuilt from source. 0.35px gaps between colours.
Ran through SVGO CLI.


fbernhart:
@adamrusted Overlaying your version with the source it seems like you've cut off quite a lot. 🤔 

image

How about just creating gaps in the following places? Instead of cutting off half of the red shape and half of the yellow shape, I'd just cut off all of the gap from the red shape (same between green/yellow and red/blue). Do you get what I mean? 😄 So we don't affect the width and height of the icon.

image

fbernhart:
Or maybe forget about that. Reading the [brand guidelines](https://partnermarketinghub.withgoogle.com/#/brands/1XjOvM2TCIH1P3V_vI-DPgr_m7dNa27GO/1RtQf7kzQQBNnmKIL2_au8FPBt_8Gq71oAqs_Zv1ffJk) which state "Do not break the mark apart.", maybe we should just go for the monochrome version without the gaps?

Google already provides a monochrome version. So probably this is their preferred way.

image

adamrusted:
@fbernhart the brand page wasn't letting me download any EPS files earlier, will keep trying and recommit once got the monochrome version working.

fbernhart:
@adamrusted Yeah, I can't download it either ... really annoying.

adamrusted:
@fbernhart I've just rebuilt using the Wikipedia SVG - as it matches up to the PNG (which still allows me to download it 🤔)

fbernhart:
@adamrusted 👍🏼  Great, thanks!

PR #4507, opened 2020-12-27
Updated Milestone in README.md

adamrusted:
Update README.md to include new milestone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 </a>
 <h3 align="center">Simple Icons</h3>
 <p align="center">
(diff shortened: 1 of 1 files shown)


fbernhart (approved):
Thanks! 👍🏼

PeterShaggyNoble:
I wonder if there's some way we could automate changes like this one and major version bumps in the README through GH Actions somehow?

mondeja:
Seems related to [simple-icons/release-action/issues/1](https://github.com/simple-icons/release-action/issues/1). I think that the best option is to use the `release-action` to do it.

PR #4509, opened 2020-12-27
Remove CEVO

adamrusted:
cevo

**Alexa rank:** [2,118,977](https://www.alexa.com/siteinfo/cevo.com)

Removing due to the severe dip in Alexa Rank.


mondeja:
The Alexa Rank has dropped a lot since #1244, so this should be removed not?

PeterShaggyNoble (approved):
Looks like this no longer exists so can definitely be removed; even if their website is only down temporarily, their Alexa rank has plummeted to 7.7m. Thanks, @adamrusted 👍🏻

Issue #2127, opened 2019-12-14
Cocos


runxel:
**Name:** Cocos (formerly Cocos2d)
**Website:** https://www.cocos.com/en/
**Official resources for icon and color:**  🤷‍♂ 
  
[Alexa rank](https://www.alexa.com/siteinfo/cocos.com): ~34k

adamrusted:
Following discussion with Cocos' marketing team (@Cocosluke), it would appear their icon is out of scope for Simple Icons due to the designer not wanting any changes made to their work (including removing / changing color). I'll be creating another PR to remove this in v5.0.0 - but have been given permission to export and share an SVG of their logo 'as is'. Below I've included the SVG code for an optimized SVG at full resolution:

```svg
[...]

PR #4528, opened 2020-12-29
Remove Cocos

adamrusted:
See #2127, #4340 
Remove Cocos logo following discussion with Marketing Director.
Designer does not want logo altered in any way, including removing / changing colors - therefore we're unable to use the monochrome version generated.


PeterShaggyNoble:
Just to clarify: from Luke's comments it seems to me that a monochrome treatment is allowed but _only_ in white - both our existing and new websites would violate that as both display it in black.

adamrusted:
@PeterShaggyNoble yep - and because they don't have documented brand guidelines online, we're unable to advise when #2846 is resolved.

PeterShaggyNoble:
If that's our only concern then we could, instead, link directly to Luke's comment regarding white being the only acceptable colour.

adamrusted:
> If that's our only concern then we could, instead, link directly to Luke's comment regarding white being the only acceptable colour.

Could do, though that'd require the 'brand guidelines' Issue / PR to be resolved first.

PeterShaggyNoble:
Let's ping, @CocosLuke for final say on this one before removing it. [Our new website](https://simple-icons.github.io/simple-icons-website/?q=cocos)'s dark theme would display your icon in white although it would be black in the light theme. But, if we were to link to [your comment](https://github.com/simple-icons/simple-icons/pull/4340#issuecomment-752365377) stating that the icon can only be used in white, would you be happy for us to keep it. Obviously, even with that link in place, it would be impossible for us to actually _enforce_ what colour our users display your icon in.

If either of those things are an issue for you then we will, of course, respect your wishes and this PR can be merged straight in once the conflicts are resolved.

PeterShaggyNoble:
Closing this out due to us now [having full approval](https://github.com/simple-icons/simple-icons/pull/4340#issuecomment-847505387).

PR #4564, opened 2021-01-01
Added FedEx

### LINKED ISSUE #3225: Request: Fedex Icon
Name: FedEx
Website: https://www.fedex.com
Official resources for icon and color:
As best I can tell, their brand site is https://brand.fedex.com/, but requires a login to get more detailed information.

adamrusted:
fedex

**Issue:** Closes #3225
**Alexa rank:** [229](https://www.alexa.com/siteinfo/fedex.com)

### Checklist

### Description
SVG pulled from recent PDF from the newsroom site, so set newsroom as source.
From a browse of their website, the only time the ® isn't included is when it's for internal use (uniforms etc), they even use it on their packaging and the front page of their website, so I felt it appropriate for us to include it too. Happy for another @simple-icons/maintainers to say otherwise.


ericcornelissen:
> From a browse of their website, the only time the ® isn't included is when it's for internal use (uniforms etc), they even use it on their packaging and the front page of their website, so I felt it appropriate for us to include it too.

I would be inclined to agree with you and would be in favor of including it. Especially given [the 3th bullet regarding trademark symbols in the Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#icon-guidelines):

_If there is ambiguity about the conditions under which the symbol is required, it must be included if it is a registered trademark symbol (®) but not if is a trademark symbol (™)._

fbernhart:
Unfortunately I can't find any PDF in the [newsroom](https://newsroom.fedex.com/) to compare against. Or am I just blind? 😅

fbernhart:
LGTM! 👍🏼 

In case we're in need of the exact source to compare against in the future, it was this one (page 2):

https://newsroom.fedex.com/wp-content/uploads/2020/10/FedEx-CorpOverview-2020.pdf

Issue #3255, opened 2020-06-24
Request: FreeNAS / TrueNAS


CoreyVidal:
**Name:** FreeNAS / TrueNAS
**Website:** freenas.org / ixsystems.com
**Alexa rank:** 81,462 / 32,613

**Official resources for icon and color:**
  https://www.ixsystems.com/blog/new-new-truenas-logo-unveiled/

CoreyVidal:
I originally came here only to request an icon for FreeNAS (which is a single-colour shark logo). But while searching for resources, I've discovered that it's about to be renamed TrueNAS, and they just recently released their new logos.

FreeNAS is free and open-source NAS (network-attached storage) operating system: https://en.wikipedia.org/wiki/FreeNAS

PeterShaggyNoble:
Thanks for the request, @CoreyVidal 

First of all, to summarise everything that was mentioned here and in the linked blog entry (it was a bit confusing on the first read-through!), as much for reference by anyone requesting any of the other logos in the future. The only point below that are really relevant to this issue are 2, 3 & 7 but please correct me if I'm wrong on any of it.

1. There's a company called iXsystems.
2. They have a product called FreeNAS which uses [this logo](https://www.freenas.org/wp-content/uploads/logo_flat_V2.png).
3. They are currently in the process of rebranding that product as "TrueNAS CORE", which will use the second logo in [this image](https://static.ixsystems.co/uploads/2020/06/Stacked_Primary_TrueNAS_Logos.png).
4. They have another product, currently called TrueNAS, which, as part of the rebranding, is being renamed as "TrueNAS Enterprise" and will use the third logo in [this image].(https://static.ixsystems.co/uploads/2020/06/Stacked_Primary_TrueNAS_Logos.png).
5. Both products are being bundled together as a suite to be called "TrueNAS Open Storage", which will use the first logo in [this image](https://static.ixsystems.co/uploads/2020/06/Stacked_Primary_TrueNAS_Logos.png).
6. There's an third product in development called "TrueNAS SCALE", which will use the last logo in that image.
7. This issue is a request for FreeNAS (soon to be TrueNAS Core) only.

Assuming point 7 to be correct then there are 2 possible approaches we can take, depending on the timescale of the rebranding.
1. If it's still a good few months out then we add the current shark logo as FreeNAS now and update to the new logo and name when the rebrand is complete.
[...]

PR #4574, opened 2021-01-01
Added TrueNAS

adamrusted:
truenas

**Issue:** Contributes to #3255
**Alexa rank:** [56,479](https://www.alexa.com/siteinfo/truenas.com)

### Checklist

### Description
SVG found in header of site. Used hex from text in SVG.


fbernhart:
The linter isn't complaining about collinear segments, but I think we can still remove these two nodes:

image


Do you agree @adamrusted ?

adamrusted:
@fbernhart should be sorted - thanks for spotting! 👌🏻

PR #4584, opened 2021-01-03
Updated Google Translate

adamrusted:
googletranslate

**Issue:** Contributes toward #4205
**Alexa rank:** N/A (Google subdomain)

### Checklist

### Description
Rebuilt SVG from source ~with gaps of 0.5px between different 'layers'.~


PeterShaggyNoble:
Note that [the monochrome treatment Google themselves use](https://material.io/resources/icons/?search=g_translate&icon=g_translate&style=baseline) for this one matches our current treatment.

adamrusted:
@PeterShaggyNoble just rebuilt to match existing styling, but with the 0.5px path around the right-hand section for consistency.

PeterShaggyNoble (approved):
Nice work, @adamrusted - looks good to me :+1: Note, though, that there is a slight gap in both your version and the original which I had "fixed" in my original contribution but I'm OK with or without it, if the other @simple-icons/maintainers are. The only change I'd suggest is using https://commons.wikimedia.org/wiki/File:Google_Translate_logo.svg for the source, in keeping with our current standards.

adamrusted:
> Note, though, that there is a slight gap in both your version and the original which I had "fixed" in my original contribution but I'm OK with or without it, if the other @simple-icons/maintainers are.

Have updated the source, and will fix the gap now, thanks for spotting @PeterShaggyNoble!

adamrusted:
@PeterShaggyNoble should be good to go, would you mind looking over it again before merging? 🙏🏻

PeterShaggyNoble (commented):
Looks perfect to me now, @adamrusted 👍🏻 Nice work.

PR #4585, opened 2021-01-03
Fix variable reference error in 'ineffective-segments' lint

mondeja:
Sorry for this error, my bad. This was introduced in #4427: `segments` and `absSegments` variables has the same reference and this is causing that reported errors are wrong.


adamrusted (approved):
LGTM! Will wait on another @simple-icons/maintainers before merging though.

service-paradis (approved):
Good catch! Missed it while reviewing the other PR, thanks!

PR #4604, opened 2021-01-04
Fix error in 'collinear-segments' lint

mondeja:
Sorry, my bad. The linter was reporting collinear segments at curves limits. There are 8 icons less to review.
Thanks for report it @adamrusted!


adamrusted (approved):
LGTM! Thanks @mondeja 🎉

PR #4616, opened 2021-01-05
Fixed naming of 'bigbasket'

adamrusted:
As mentioned by @fbernhart in #4572, I got the name styling wrong. To save this being a 'breaking change', created this PR to fix issue ahead of 4.6.0 release.


fbernhart:
@adamrusted Thanks! 👍🏼

ericcornelissen:
Has this been tested with the NPM package? That is, if I'm not mistaken, before this change `simpleIcons.get('Big Basket')` worked and after this change it won't. <sup>(The same goes for `simpleIcons['Big Basket']`, but as this is not in the API I don't think we consider this breaking)</sup>

ericcornelissen:
Never mind, didn't realize [the brand was added on the same day](https://github.com/simple-icons/simple-icons/pull/4572) 😅

PR #4623, opened 2021-01-05
Add Presto

### LINKED ISSUE #3569: Icon Request: Presto SQL
**Name:** Presto SQL
**Website:** https://prestodb.io/
**Alexa rank:** 125,234 (as of 2020-09-16)

PeterShaggyNoble:
Presto

**Issue:** Closes #3569
**Alexa rank:** [~148k](https://www.alexa.com/siteinfo/prestodb.io)

### Checklist

### Description
I'd prefer this without the box but they only time they seem to use that treatment is in conjunction with the wordmark so I think we'll have to keep it. I'm not entirely convinced the dark blue is the correct colour choice but it was the one I came across most often in their SVGs; the other option would be the lighter `#4967a9`.


_data/simple-icons.json
```diff
@@ -5370,6 +5370,11 @@
             "hex": "DF0067",
             "source": "https://www.prestashop.com/en/media-kit"
         },
+        {
+            "title": "Presto",
+            "hex": "0F2648",
```
adamrusted, line 5375: Looks good to me, however the 'most used' colour in the logo found [here](https://github.com/prestodb/presto/blob/414ab2a6bbdcca6479c2615b048920adac34dd20/presto-docs/src/main/resources/logo/web/main/black/FB_Presto_Logo_BlackBG-01.svg) is `#5890FF`, would you mind updating to that?

adamrusted (changes requested):

_data/simple-icons.json
PeterShaggyNoble: They seem to use loads of different shades of blue with no clear preference for one so I suppose that's as good a choice as any. Updated.

adamrusted (approved):
LGTM! Thanks @PeterShaggyNoble 💯

Issue #4590, opened 2021-01-03
The 2021 Icon Review


mondeja:
Starting with #4589, I open this to track those icons that should be reviewed this year. This issue should serve to review the icons that fail with the new linting rules added in 2021.

## Icons with collinear coordinates in straight lines (#4589)

- [x] Amazon Lumberyard (#4591)
- [x] Andela (#4592)
- [x] Apache JMeter (#7000)
- [x] Apache Maven (#4593)
- [x] Arch Linux (#5172)
- ~~Aseprite~~ (#4692)
- ~~Citroën~~ (#4604)
- [ ] Cocos (#6996)
- [x] Daimler (#4618)
- [x] Disroot (#5173)
- [x] DigitalOcean (#4648)
- [x] Elsevier (#5175)
- [x] Epic Games (#5178)
- [x] FedEx (#5180)
- ~~Fedora~~ (#5540)
- [x] Ferrari (#4778)
- [x] Firefox (#5203)
- [x] Fossa (#4473)
- [x] Furry Network (#6221)
- [x] Gatsby (#4620)
- [x] Giphy (#4621)
- [x] Google Ads (#4471)
- ~~Google Pay~~ (#4604)
- [x] Grubhub (#5168)
- [x] Harbor (#4598)
- [x] Jira Software (#4657)
- [x] Kotlin (#4485)
- [x] Lamborghini (#5169)
- [x] Lenovo (#4477)
- [x] Livestream (#4647)
- [x] ManageIQ (#5165)
- [x] Material Design Icons (#4642)
- [x] Matternet (#6222)
- [x] Microsoft (#4649)
- ~~Minds~~ (#4604)
- [x] Monzo (#5219)
- [x] Mozilla (#4656)
- ~~MusicBrainz~~ (#4604)
- [x] New Japan Pro-Wrestling (#4930)
- [x] ngrok (#5176)
- [x] Notepad++ (#5187)
- [x] OpenAPI Initiative (#6223)
- [x] Packagist (#5182)
- [x] PCGamingWiki (#6224)
- [x] Phabricator (#5143)
- [x] pr.co (#4655)
- [x] PrestaShop (#6225)
- [x] RabbitMQ (#5183)
- [x] Ruby (#6226)
- [x] Server Fault (#6227)
- [x] Showpad (#https://github.com/simple-icons/simple-icons/pull/6999)
- [x] SonicWall (#4654)
- [x] StackPath (#4652)
- [x] strongSwan (#6228)
[...]

PR #4122, opened 2020-11-23
Replace Postwoman by Hoppscotch

ericcornelissen:
### Checklist
  - [x] I updated the JSON data in `_data/simple-icons.json`
  - [x] I optimized the icon with SVGO or SVGOMG
  - [x] The SVG `viewbox` is `0 0 24 24`

### Description

Closes #3589, removes [Postwoman](https://simpleicons.org/?q=Postwoman) and replaces it by [Hoppscotch](https://hoppscotch.io/) in preparation for the upcoming [v4 release](https://github.com/simple-icons/simple-icons/milestone/3).

The new SVG is based on [the SVG version of the logo](https://github.com/hoppscotch/hoppscotch/blob/main/static/images/ufo_logo.svg) I found in the repository of the website of Hoppscotch. that version has tiny lines between the three components of the logo (UFO top part, UFO body, UFO "beam") that I removed - I'd be open to reintroducing them, but I feel like they don't work at small sizes... The color I took from the "alien" SVG in the sidebar of [the Webapp](https://hoppscotch.io/), given that the green in the logo SVG is a gradient. Suggestions for a better color source are welcome (@ChristianOellers any ideas?).

Issue #3219, opened 2020-06-14
Sapper icon


erianvc:
**Name:** Sapper
**Website:** https://sapper.svelte.dev/
**Official resources for icon and color:**
  
The same Svelte icon but with `#159794` brand color

NovaGL:
SVG is in the top of the main page. I suggest we just use the "S"
https://sapper.svelte.dev/sapper-logo-horizontal.svg

Alexa Rank  #46,637

PeterShaggyNoble:
Hmm ... we have in the past (most recently with the various subsidiaries of [JR group](https://github.com/simple-icons/simple-icons/issues/3028)) shied away from adding duplicate icons that differ only in colour, at most.

While I can definitely see the need for them, I think we'd need to come up with a different solution than having multiple, identical SVG files. One _possible_ way could be through the changes being proposed and discussed in #2260, which might allow us to have multiple JSON entries pointing to a single SVG file.

PeterShaggyNoble:
@erianvc, to assist is in dealing with and resolving this request, could you provide us with some detail on your use case for it? Do you need to be able to differentiate between Svelte & Sapper or just have the Sapper colour made available?

erianvc:
Yes, it was to have the brand color available.
I think should be able to point to the same SVG but different color, because there are brands that use the same icon but with different flavors...

PeterShaggyNoble:
Thanks for the feedback, @erianvc. I'm of the same opinion as you that we should have some means of including brands that use the same logo just with different colours, just without duplication of the SVG. I'm going to slap the "pending" label on this until #2260 is resolved (we're aiming to have it done for the v3 release, Sunday week) and we see if we can leverage what we're doing there to address this.

PeterShaggyNoble:
Discussion on a possible approach for this has been moved to #3270.

PeterShaggyNoble:
Thanks for your patience, @erianvc; current consensus from #3270 for this is that we're going to allow SVGs with duplicate paths so  would you like to have a read through [our contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and give this a try yourself?

Note that our existing Svelte icon will need to be updated first (see #3169) before it can be copied for Sapper.

PR #4647, opened 2021-01-06
Remove Livestream

fbernhart:
**Alexa rank:** [9,442](https://www.alexa.com/siteinfo/livestream.com)

### ~Checklist~
~- [x] I updated the JSON data in `_data/simple-icons.json`~
~- [x] I optimized the icon with SVGO or SVGOMG~
~- [x] The SVG `viewbox` is `0 0 24 24`~

### ~Description~
~- Seems like Livestream has been bought by vimeo back in 2017. See [their blog](https://livestream.com/blog/livestream-vimeo-acquisition) and [Wikipedia](https://en.wikipedia.org/wiki/Vimeo_Livestream#History)~
~- The new logo is a combination of the vimeo and livestream wordmark.~
~- SVG pulled from the header of https://livestream.com~
~- Changed source to https://livestream.com, as https://livestream.com/press doesn't exist anymore.~
~- Changed color to `#000000`. The SVG at source doesn't have a color, so black is probably the best choice.~

### Edit
This PR removes Livestream as the brand doesn't exist anymore. It's now called "Vimeo Livestream" and has a different logo. "Vimeo Livestream" is added in a different PR.


fbernhart:
@simple-icons/maintainers I've been wondering if we should as well rename this one (which would be a breaking change). I have the impression that `Vimeo Livestream` would be a better fit than just `Livestream`.

Any opinions?

ericcornelissen:
I agree @fbernhart, either

> Another option would be to actually just fix the collinear linter error of the old icon, creating a new icon called Vimeo Livestream and then removing Livestream with the next major release.

Or we update the icon under the current name and rename with the next release 👍

PeterShaggyNoble:
I don't think this is the right way to go on either front. In the absence of a colour in the source icon (or brand guidelines where available), we should go with our best guess from the website. In this case I'd say it should be either the purple `#0a0a20` that features throughout the site and in their favicon on the homepage.

However, and more pressingly, I don't think we need to use the full wordmark here as they themselves use the Vimeo "V" in a few places, including in the favicon and in the website's header when you scale down to smaller sizes. Given that and, as we've yet to fully resolve #3270 & #3219, I'd hold off on doing anything for now as the solution may end up being to remove this completely.

> Or we update the icon under the current name and rename with the next release

If it's decided that we're going to allow brands that share an icon with another one then this would also be my preferred approach, too, to avoid duplication.

ericcornelissen:
> If it's decided that we're going to allow brands that share an icon with another one then this would also be my preferred approach, too, to avoid duplication.

To me, the discussion here is a separate from #3270. There we consider icons that will be part of the collection "**permanently**" using the same icon. Here, we consider the option of providing the icon of a renamed brand in two ways in one version so as to make upgrading to the next major version potentially easier. That does not mean I think we should necessarily provide both icons, I just want to point out that these are two different scenarios.

Now, from a user's perspective, I would say it is ideal to have both "Livestream" and "Vimeo Livestream" available in v4, and remove "Livestream" when v5 comes around. This is because 1) it gives users of "Livestream" a larger window of time to update and 2) it allows users looking for "Vimeo Livestream" to find it already in v4. (This is arguably not as big a deal for "Vimeo Livestream" as it still contains the old name, but e.g. the rename from [Postwoman to Hopscotch](https://github.com/simple-icons/simple-icons/pull/4122) would be more relevant to this argument.)

---

<sup>I won't respond to the rest of the discussion, at least for now</sup>

PeterShaggyNoble:
To clarify: my reference to #3270 was in direct relation to my suggestion that their icon is now just the Vimeo "V".

adamrusted:
> I would say it is ideal to have both "Livestream" and "Vimeo Livestream" available in v4, and remove "Livestream" when v5 comes around. This is because 1) it gives users of "Livestream" a larger window of time to update and 2) it allows users looking for "Vimeo Livestream" to find it already in v4.

I agree with @ericcornelissen's comments above. I'd prefer to have ***both*** Livestream and Vimeo Livestream available in v4 *(even if it's a duplication of this newer icon)* to update the brand, and give any users of `livestream.svg` time to swap over to the newer icon, before the deletion of `livestream.svg` in v5.0.0.

fbernhart:
Ok 👍🏼  Let me make this one the "Remove Livestream" PR, that will remove `livestream.svg` with the next major release.

I'll as well create a new PR with "Add Vimeo Livestream".

PeterShaggyNoble (approved):
Looks good to me 👍🏻 Once the conflicts are resolved then this can be merged straight in.

PR #4649, opened 2021-01-06
Update Microsoft

fbernhart:
microsoft

**Alexa rank:** [21](https://www.alexa.com/siteinfo/microsoft.com)

### Checklist

### Description
- I've updated the source to an official site --> https://developer.microsoft.com
- I've rebuilt the icon with the SVG in the header of the source
- I've changed the color to `#5E5E5E`, which is used for the wordmark "Microsoft" in the header.

- The icon now passes the linter without collinear segments. 👍🏼


PeterShaggyNoble:
How have I never noticed our MS icon is grey before?! 😆 Am I the only one that would expect their brand colour to be one of the 4 from their logo?

fbernhart:
@PeterShaggyNoble As all of the four colours are used in the same amount, I can't spot a way to choose one. And as nobody has complained (or even noticed 😄) up to now, I would keep the gray color.

Unless someone can find a brand guideline that lists one of the four colors as primary. Unfortunately I couldn't.

adamrusted:
[Brand Guidelines for Microsoft](https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks/usage/logo.aspx)

ericcornelissen:
Regarding the color, this color was chosen back in 45da1abea8476cd38aa7980410d83597d005f081 without any explanation so I think we're fine changing it.

> 
> 
> [Brand Guidelines for Microsoft](https://www.microsoft.com/en-us/legal/intellectualproperty/trademarks/usage/logo.aspx) - states the wordlogo should be kept, but also:
> 
> > Don't make the logo a single color.

Interesting 🤔 On the one hand we should follow this, on the other hand they use the logo in a single color themselves, e.g. on the back of the surface laptops...

PeterShaggyNoble:
> On the one hand we should follow this

It is a pretty major icon for us to be removing and as it's used extensively by plenty of others, I think we should keep it until told otherwise.

> on the other hand they use the logo in a single color themselves

I think we've learned by now that a lot of brand guidelines come down to "do as I say, not as I do" - context is everything.

adamrusted:
> It is a pretty major icon for us to be removing and as it's used extensively by plenty of others, I think we should keep it until told otherwise.

Should we look to use the full wordmark in this case then?

Also, as far as colours go, Microsoft use shades of blue as their CTA colours on both their main and developer site. Might be worth picking one of those?

fbernhart:
With a quick Google search I've found a monochrome version of the logo used at https://code.visualstudio.com in the footer of the site. So there might be some very few cases, where they use a monochrome version themselves.

PeterShaggyNoble:
The wordmark gets a big fat "nope" from me, @adamrusted! 🙈 Separately, I've actually been giving consideration to suggesting we cull and start rejecting brands that explicitly require wordmarks be included, otherwise forbid their logo be simplified for the purposes of an icon or just don't have anything that can be used as an icon. As a monochrome icon library, we already reject brands that don't allow monochrome treatments of their logos and I think we should do the same for brands that don't allow icon treatments of their logo. But that's a broader discussion for another day! (I swear, that Elsevier "icon" annoys me every time I see it! 😆)

@fbernhart, we often come across companies that seem to violate their own publicly available brand guidelines and this is usually because they'll have different internal guidelines for situations where icons & logos are being used in a context where their brand is already well-established.

adamrusted:
@simple-icons/maintainers are we settled on the hex staying as is? If so, this can be merged once the conflict is resolved.

fbernhart:
In lack of a better alternative, I would keep the color as it is.

PR #4672, opened 2021-01-07
Ignore files that are not SVGs in the icons folder

service-paradis:
**Issue:** #4662
**Alexa rank:**
  

### Description

As discussed in #4662, I added a `.gitignore` rule to ignore non `.svg` files in the `icons` folder.


.gitignore
```diff
@@ -47,3 +47,8 @@ node_modules/
 # Files generated by build script
 icons/*.js
 /index.js
+
+# Ignore all files in the icons folder
+icons/*
+# Except SVG files
+!*.svg
```
ericcornelissen, line 54: Shouldn't it be ⬇️ ?

```suggestion
!icons/*.svg
```

We might, in the future, want to ignore SVGs elsewhere

service-paradis: Totally, I already fixed it but did not commit 🤦‍♂️ 
Coming back to work after the holiday is not so easy 😅 
Should be better now, thanks!

adamrusted (approved):
LGTM! Thanks @service-paradis 🎉

PR #4683, opened 2021-01-08
Update Google Street View

adamrusted:
googlestreetview

**Issue:** Contributes to #4205
**Alexa rank:** N/A (Google Product)

### Checklist

### Description
Rebuilt from [Wikipedia SVG](https://upload.wikimedia.org/wikipedia/commons/e/e0/Google_Street_View_icon.svg) as matched PNG. Updated gaps to 0.5px, and included curved line at bottom left, to complete 'icon'.


PeterShaggyNoble:
That rounded corner in the bottom left doesn't look right to me; I'd expect a gap there based on the original. What do the other @simple-icons/maintainers think?

adamrusted:
> That rounded corner in the bottom left doesn't look right to me;

Yeah - I wasn't sure on it, but the sharp edges of the cutout also didn't feel right in the bottom left. Tried to maintain the 'badge' shape as much as possible underneath, with the 'person' covering the top right, but happy to remove the curve if it's preferred.

fbernhart:
- Like @PeterShaggyNoble already mentioned, I would as well put a gap at the bottom left.
- The head and the body seem to be pretty close in your version @adamrusted. 🤔 Did you cut off half of each color?
- I would as well create a gap where the yellow color changes to orange.

As a quick draft, I would prefer a version like this (blue strokes represent gaps):

image

adamrusted:
googlestreetview

@fbernhart @PeterShaggyNoble - should be updated with better gaps now. I've always seen the orange triangle just as a shading thing, which when flattening should be removed. Happy to add it if we'd prefer it there though.

PeterShaggyNoble:
Sorry, only getting back to this now. As the version on Wikimedia is an original work, I think we should stick with the version found in [this PDF](https://developers.google.com/streetview/ready/pdf/svr-badge.pdf) which is available at [the original source](https://developers.google.com/streetview/ready/branding) as there are quite a few differences between it and your version (in red below).

adamrusted:
Give that a go @PeterShaggyNoble - I'm having some difficulty comparing, as the PDF appears to be password protected when trying to take into Illustrator...

PeterShaggyNoble:
That's odd 🤔 Resaved it as an SVG for you if it helps: [googlestreetview.zip](https://github.com/simple-icons/simple-icons/files/5952672/googlestreetview.zip)

adamrusted:
Amazing, thanks @PeterShaggyNoble!
Have re-generated using the SVG you provided.

PeterShaggyNoble:
Looks perfect to me, @adamrusted - nice work 👍🏻

PR #4695, opened 2021-01-08
Update Google Chat

PeterShaggyNoble:
Google Chat

**Issue:** #4205
**Alexa rank:** n/a

### Checklist

### Description
Icon from SVG in source header; I went with a 0.5dp cut-out. Without a GSuite account, I'm not sure on the colour. From what little I _can_ see, our existing colour is still in use but maybe the green from the bottom half of the front speech bubble might be a better choice.


adamrusted:
@PeterShaggyNoble looks good to me - should we be using the green in the icon (`#00AC47`) as opposed to the CTA colour of `#00897B`?

PeterShaggyNoble:
As I can't access the site, @adamrusted, I'll take guidance from someone who can. The source could probably also do with being updated to the homepage, if someone can confirm that the the SVG is used there.

adamrusted:
@PeterShaggyNoble Sure - I'm unable to access the 'homepage', but am able to get onto the [landing page](https://chat.google.com/) with embedded SVG.
The three hex codes used in the SVG there are `#00AC47`, `#00832D` in the centre, and `#6ABE82`.
Would it be worth adding gaps to show the 'change in gradient' like below? Happy to provide the path I used if that's preferred.

image#

PeterShaggyNoble:
Colour & source updated.

I don't think that's the correct treatment here, @adamrusted; the change in colour in that middle rectangle is to make it look like the front speech bubble is semi-transparent and overlaps that back bubble.

adamrusted (approved):
LGTM! Thanks @PeterShaggyNoble 🎉

adamrusted:
> the change in colour in that middle rectangle is to make it look like the front speech bubble is semi-transparent and overlaps that back bubble.

Yep, fine by me. Icon matches otherwise, so I'll merge this now.

PR #4703, opened 2021-01-10
Add Vimeo Livestream

fbernhart:
vimeolivestream

**Issue:** N/A
**Alexa rank:** [9,441](https://www.alexa.com/siteinfo/livestream.com)
 
 ### Checklist

### Description
- Relies on the discussion in #4647.
- SVG pulled from the header of https://livestream.com
- Color `#0A0A20`, which is used on the website and was mentioned in https://github.com/simple-icons/simple-icons/pull/4647#issuecomment-756094307


adamrusted (approved):
LGTM! Thanks @fbernhart 🎉

PeterShaggyNoble:
I still think just the Vimeo `v` is the way to go here, as they do in their favicon and in their website's header on smaller devices.

fbernhart:
@PeterShaggyNoble We already have the "Vimeo v" as the icon of Vimeo itself. https://simpleicons.org/?q=vimeo

So this would mean we're having the same icon twice.

PeterShaggyNoble:
Yes, but first we'd need to resolve #3270

For me, adding a wordmark simply to differentiate from an existing brand isn't ideal; we should always strive to add the icon only whenever possible.

PR #4707, opened 2021-01-10
Escape 'Get commit message' step output in 'Publish' workflow

mondeja:
If a brand name has the character `'` the release will fail because is not properly escaped in Publish workflow, as [has been happended in latest release](https://github.com/simple-icons/simple-icons/runs/1676956577?check_suite_focus=true#step:8:463) publication due to "De'Longhi" brand name.

- This is solved by using `printf` with `"` wrapped string to prevent string termination. This works for brand names with `'` character, but not if brand names have characters `"` (I've opened [an issue in the kceb/git-message-action](https://github.com/kceb/git-message-action/issues/7), but other suggestions are welcome). Anyways, there is no brands with `"` character in their name, but this does not mean that they cannot appear in the future.
- Additionally, the `wc` + expression computation of release body has changed by `tail` command, easier and not redundant.

You can check the reproducible error with a test of these new commands [here](https://github.com/mondeja/test-escape-step-output/blob/main/.github/workflows/test.yml), executed [here](https://github.com/mondeja/test-escape-step-output/runs/1677376113?check_suite_focus=true).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,10 +47,10 @@ jobs:
     - name: Get release title and body
       id: release
       run: |
-        COMMIT_MSG="$(printf "${{ steps.commit.outputs.git-message }}")"
-        RELEASE_TITLE="$(printf "$COMMIT_MSG" | head -n 1)"
+        COMMIT_MSG="$(echo "${{ steps.commit.outputs.git-message }}")"
+        RELEASE_TITLE="$(echo "$COMMIT_MSG" | head -n 1)"
         echo "::set-output name=title::$RELEASE_TITLE"
-        RELEASE_BODY="$(printf "$COMMIT_MSG" | tail -n +3)"
+        RELEASE_BODY="$(echo "$COMMIT_MSG" | tail -n +3)"
         echo "::set-output name=body::$RELEASE_BODY"
     - name: Get release version
       id: get-version
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -47,9 +47,10 @@ jobs:
     - name: Get release title and body
       id: release
       run: |
-        RELEASE_TITLE=$(echo '${{ steps.commit.outputs.git-message }}' | head -n 1)
+        COMMIT_MSG="$(printf "${{ steps.commit.outputs.git-message }}")"
+        RELEASE_TITLE="$(printf "$COMMIT_MSG" | head -n 1)"
         echo "::set-output name=title::$RELEASE_TITLE"
-        RELEASE_BODY=$(echo '${{ steps.commit.outputs.git-message }}' | tail -n $(expr $(echo '${{ steps.commit.outputs.git-message }}' | wc -l) - 1))
+        RELEASE_BODY="$(printf "$COMMIT_MSG" | tail -n +3)"
         echo "::set-output name=body::$RELEASE_BODY"
     - name: Get release version
       id: get-version


ericcornelissen:
> If a brand name has the character `'` the release will fail because is not properly escaped in Publish workflow, as [has been happended in latest release](https://github.com/simple-icons/simple-icons/runs/1676956577?check_suite_focus=true#step:8:463) publication due to "De'Longhi" brand name.

Awesome, thanks for investigating 🕵️ and fixing 🔧 
 
> * This is solved by using `printf` with `"` wrapped string to prevent string termination. This works for brand names with `'` character, but not if brand names have characters `"` (I've opened [an issue in the kceb/git-message-action](https://github.com/kceb/git-message-action/issues/7), but other suggestions are welcome). Anyways, there is no brands with `"` character in their name, but this does not mean that they cannot appear in the future.

Using `printf` seems like a good solution. Currently you use it three times shouldn't once (when assigning to `COMMIT_MSG`) be enough?


> * Additionally, the `wc` + expression computation of release body has changed by `tail` command, easier and not redundant.

Great improvement 👍

mondeja:
> Using printf seems like a good solution. Currently you use it three times shouldn't once (when assigning to COMMIT_MSG) be enough?

The variable `COMMIT_MSG` should be printed to pass it to `tail` and `head`.

ericcornelissen:
> The variable `COMMIT_MSG` should be printed to pass it to `tail` and `head`.

Again, I'm not very good with shell scripting, but isn't it enough to use `$COMMIT_MSG`, as in `echo $COMMIT_MSG`. So, something like?:

```
COMMIT_MSG=$(printf "${{ steps.commit.outputs.git-message }}")
RELEASE_TITLE=$($COMMIT_MSG | head -n 1)
echo "::set-output name=title::$RELEASE_TITLE"
RELEASE_BODY=$(echo '${{ steps.commit.outputs.git-message }}' | tail -n $(expr $(echo '${{ steps.commit.outputs.git-message }}' | wc -l) - 1))
RELEASE_BODY=$($COMMIT_MSG | tail -n +3)
```

ericcornelissen:
Ah I see now, apologies for my confusion. In that case, what is the advantage of using `printf "..."` over `echo "..."`? I.e. is there a reason for using `printf` instead of `echo` in addition to switching from `'` to `"`?

mondeja:
I always use `printf` because of [this](https://askubuntu.com/questions/467747/which-is-better-printf-or-echo), but if you want consistency with others `echo` commands, I can change it.

ericcornelissen:
> I always use `printf` because of [this](https://askubuntu.com/questions/467747/which-is-better-printf-or-echo), but if you want consistency with others `echo` commands, I can change it.

Didn't know, thanks for the link :slightly_smiling_face: In that case, can you update all uses of `echo` in this Workflow file? (And feel free to _fix_ this in other Workflow files in this and other simple-icons repos, separately from this PR.)

mondeja:
I've tested `echo` commands for `::set-output` in Github workflows but [does not work](https://github.com/mondeja/test-escape-step-output/runs/1680832813?check_suite_focus=true), although I don't know why, maybe they have patched `echo` to read their customized commands in their virtual machines. I think that we could here rely in `printf` to safely read the commit message and use `echo` for Github Actions commands.

ericcornelissen:
> I've tested `echo` commands for `::set-output` in Github workflows but [does not work](https://github.com/mondeja/test-escape-step-output/runs/1680832813?check_suite_focus=true), although I don't know why, maybe they have patched `echo` to read their customized commands in their virtual machines. I think that we could here rely in `printf` to safely read the commit message and use `echo` for Github Actions commands.

Looking at [the `printf` run](https://github.com/mondeja/test-escape-step-output/runs/1680832813?check_suite_focus=true#step:2:1) versus [the `echo` run](https://github.com/mondeja/test-escape-step-output/runs/1677261674?check_suite_focus=true#step:2:1), it seems that the missing part in the former is actually printed to the console whereas it is not in the latter - seems like a "problem" with `printf` to me :thinking: (I don't think GitHub created a custom `echo` binary just for Actions...)

I'd like to keep it consistent and use only `echo` or only `printf`. Since `echo` works for everything (right?) and is always run on Ubuntu I think we should be mostly safe using `echo`. But if you feel using `printf` where possible is the better approach, I trust you, I feel like you know more about this stuff than I do :smile:

ericcornelissen (approved):

ericcornelissen:
Hmm, seems something went wrong with the release message of [v4.7.0](https://github.com/simple-icons/simple-icons/releases/tag/4.7.0)... Any idea why @mondeja ?

mondeja:
It's hard to debug, could you consider the usage of `printf` in the commands that I've introduced in this pull?

ericcornelissen:
Worth a try 👍 Feel free to commit that change directly to `develop` and we see if it works with the next release . I'll manually set the release message for [v4.7.0](https://github.com/simple-icons/simple-icons/releases/tag/4.7.0) now.

mondeja:
Thank you @ericcornelissen, done in https://github.com/simple-icons/simple-icons/commit/70ead50f27b1e9f14547e6739de4f830a0600304.

mondeja:
It seems that doesn't work using `printf` :thinking: I'm going to revert this to the previous state, maintaining the `tail -n +3` command.

mondeja:
It seems that doesn't work using `printf` :thinking: I'm going to revert this to the previous state, maintaining the `tail -n +3` command.

PR #4708, opened 2021-01-10
Minor improvements

ericcornelissen:
Low priority PR with some improvements to the repository:

- Make some improvements to consistency in indentation
- Update `.editorconfig` (in line with the previous point)
- Update `.gitattributes`
- Remove unused variables in JavaScript
- Add semicolons in JavaScript where they were missing
- Prevent error in `npm run clean`
- Update indentation in YAML


mondeja (commented):
I've some suggestions:

- Could be worth add Prettier here? Since code contributors of this project are not often newcomers, I think it wouldn't hurt too much.
- Even if Prettier is not added, could you run it in YAML files to make sure that the correct indent is used?
- Even if Prettier is not added, we are using double `"` and single `'` quote characters in the code (apart from backticks), should we unify both?

ericcornelissen:
> I've some suggestions:
> 
>     * Could be worth add Prettier here? Since code contributors of this project are not often newcomers, I think it wouldn't hurt too much.
> 
>     * Even if Prettier is not added, could you run it in YAML files to make sure that the correct indent is used?
> 
>     * Even if Prettier is not added, we are using double `"` and single `'` quote characters in the code (apart from backticks), should we unify both?

I'm not against using Prettier in this repository if there is a consensus about this amongst the @simple-icons/maintainers. Up front, I think we should limit Prettier to the JavaScript and YAML of this project, as the HTML and CSS will likely be removed due to #980. Additionally, we could consider it for JSON as well, but should probably exclude `_data/simple-icons.json` because it still uses 4-spaces indentation (which I think we should not change unless there are no open Pull Requests from outside contributors).

I will update the indentation of the YAML files in this PR, but I won't fix the quotes here unless this is not merged before we have a decision on Prettier.

ericcornelissen:
@simple-icons/maintainers, how about we leave applying Prettier for another issue/PR and merge this as is?

adamrusted (approved):
Looks good to me. As @ericcornelissen says, we can add Prettier at a later date. Can another @simple-icons/maintainers review this too before we merge it?

mondeja (approved):

PR #4710, opened 2021-01-11
Remove Sprint

PeterShaggyNoble:
From [this article](https://en.wikipedia.org/wiki/Merger_of_Sprint_Corporation_and_T-Mobile_US), following their merger with T-Mobile:

> The Sprint brand was discontinued on August 2, 2020.

Their [website](https://www.sprint.com/) now also redirects to T-Mobile's so I think we can safely remove this one in v5.


fbernhart:
@PeterShaggyNoble Have you seen the discussion in #4196?

We've decided there to keep it for now.

PeterShaggyNoble:
Yes, @fbernhart; that's what prompted me to double check. It looks to me that the decision at that time to keep it was based solely on the fact that their website was still active, which is no no longer the case.

fbernhart:
If I remember correctly, the website still looks the same as it did 2 weeks ago. 🤔

I can for example still access https://www.sprint.com/en/login.html.


But having in mind that the icon won't be removed before end of May 2021 (release date of v5.0), I think we can remove it with this PR.

adamrusted (approved):
Returning to this one again, and while the Sprint login page is still available, visiting their homepage redirects you to T-Mobile. Given they're probably having to keep their auth system in place for some legacy users / systems, I'm going to say we're safe to remove it at this point. I'm going to merge this one, and we can always revert the PR if anyone feels super strongly that we should keep it.

PR #4716, opened 2021-01-11
Add Packagist README badge icon and place badges before 'Usage'

mondeja:
It doesn't look quite right due to the number of lines the icon has, but IMHO seems to be enough, and since simple-icons is the icon provider, I think it is useful to add it. You can see the rendered result [here](https://github.com/mondeja/simple-icons/tree/packagist-icon-readme-badge#status).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@
 Over 1500 Free SVG icons for popular brands. See them all on one page at <a href="https://simpleicons.org">SimpleIcons.org</a>. Contributions, corrections & requests can be made on GitHub. Started by <a href="https://twitter.com/bathtype">Dan Leech</a>.</p>
 </p>
 
+<p align="center">
+<a href="https://github.com/simple-icons/simple-icons/actions?query=workflow%3AVerify+branch%3Adevelop"><img src="https://img.shields.io/github/workflow/status/simple-icons/simple-icons/Verify/develop?logo=github" alt="Build status" /></a>
+<a href="https://www.npmjs.com/package/simple-icons"><img src="https://img.shields.io/npm/v/simple-icons.svg?logo=npm" alt="NPM version" /></a>
+<a href="https://packagist.org/packages/simple-icons/simple-icons"><img src="https://img.shields.io/packagist/v/simple-icons/simple-icons?logo=packagist&logoColor=white" alt="Build status" /></a>
+</p>
+
 ## Usage
 
 ### General Usage
@@ -138,9 +144,3 @@ Icons are also available as a [Vue package](https://github.com/mainvest/vue-simp
 ### WordPress
 
 Icons are also available as a [WordPress plugin](https://wordpress.org/plugins/simple-icons/) created by  [@tjtaylo](https://github.com/tjtaylo).
-
-## Status
-
-[![Build status](https://img.shields.io/github/workflow/status/simple-icons/simple-icons/Verify/develop?logo=github)](https://github.com/simple-icons/simple-icons/actions?query=workflow%3AVerify+branch%3Adevelop)
-[![npm version](https://img.shields.io/npm/v/simple-icons.svg?logo=npm)](https://www.npmjs.com/package/simple-icons)
-[![Packagist version](https://img.shields.io/packagist/v/simple-icons/simple-icons?logo=packagist&logoColor=white)](https://packagist.org/packages/simple-icons/simple-icons)
--- a/README.md
+++ b/README.md
@@ -143,4 +143,4 @@ Icons are also available as a [WordPress plugin](https://wordpress.org/plugins/s
 
 [![Build status](https://img.shields.io/github/workflow/status/simple-icons/simple-icons/Verify/develop?logo=github)](https://github.com/simple-icons/simple-icons/actions?query=workflow%3AVerify+branch%3Adevelop)
 [![npm version](https://img.shields.io/npm/v/simple-icons.svg?logo=npm)](https://www.npmjs.com/package/simple-icons)
-[![Packagist version](https://img.shields.io/packagist/v/simple-icons/simple-icons)](https://packagist.org/packages/simple-icons/simple-icons)
+[![Packagist version](https://img.shields.io/packagist/v/simple-icons/simple-icons?logo=packagist&logoColor=white)](https://packagist.org/packages/simple-icons/simple-icons)


fbernhart:
👍🏼 

- How about moving the badges to the top? Somewhere before "Usage". So people could see immediately which is the current version of Simple Icons.

- Strange that the `npm` logo is just the "n", when using it in a shields.io badge. 🤔 Because the [version we're having](https://simpleicons.org/?q=npm) is the full "npm" wordmark. I guess shields.io is preferring their own version over ours. 😄 See: https://shields.io
     > Insert one of the named logos from (bitcoin, dependabot, discord, gitlab, npm, paypal, serverfault, stackexchange, superuser, telegram, travis) or simple-icons.

mondeja:
> How about moving the badges to the top? Somewhere before "Usage". So people could see immediately which is the current version of Simple Icons.

I'm not against of that.

> Strange that the npm logo is just the "n", when using it in a shields.io badge. thinking Because the version we're having is the full "npm" wordmark.

It seems that they uses [these icons](https://github.com/badges/shields/tree/master/logo) as colored replacements for some brands.

ericcornelissen:
Cool, I was thinking this exact thing yesterday but forgot about it before I could open a PR so thanks! :smile: 

Wondering if we should make the NPM logo white as well, since both GitHub and Packagist are white now... :thinking: 


> How about moving the badges to the top? Somewhere before "Usage". So people could see immediately which is the current version of Simple Icons.

Seems good to me yes. Either before usage but after the centered part or before the logo even. Not sure what looks best... Is there anyone that is not too lazy to do some mockups of these two and possibly other options and share them here as screenshots :slightly_smiling_face: (or on a separate thread and merge this as is for now)

fbernhart:
> Wondering if we should make the NPM logo white as well, since both GitHub and Packagist are white now...

I was thinking the same and gave it a try. But it doesn't work ...

As Shields.io is preferring [these icons](https://github.com/badges/shields/tree/master/logo) over ours and their icon is colored and not monochrome. I guess the only way would be to open an issue with Shields.io and convince them to use our version as default. 😅

mondeja:
Here are some previews:

- [After "Usage"](https://github.com/mondeja/simple-icons/blob/test-badges-position-after-usage/README.md)
- [Before "Usage"](https://github.com/mondeja/simple-icons/blob/test-badges-position-before-usage/README.md)
- [Before "Usage" centered](https://github.com/mondeja/simple-icons/blob/test-badges-position-before-usage-centered/README.md)

adamrusted:
> * [Before "Usage" centered](https://github.com/mondeja/simple-icons/blob/test-badges-position-before-usage-centered/README.md?rgh-link-date=2021-01-12T12%3A21%3A38Z)

This gets my vote!

service-paradis:
> > Wondering if we should make the NPM logo white as well, since both GitHub and Packagist are white now...
> 
> I was thinking the same and gave it a try. But it doesn't work ...
> 
> As Shields.io is preferring [these icons](https://github.com/badges/shields/tree/master/logo) over ours and their icon is colored and not monochrome. I guess the only way would be to open an issue with Shields.io and convince them to use our version as default. 😅

The only way to use ours would be to convert our SVG to base64 and use the hardcoded version. Something like:
```
[![npm version](https://img.shields.io/npm/v/simple-icons.svg?logo=data:image/svg+xml;base64,PHN2ZyByb2xlPSJpbWciIHZpZXdCb3g9IjAgMCAyNCAyNCIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48dGl0bGU+TlBNIGljb248L3RpdGxlPjxwYXRoIGQ9Ik0wIDcuMzM0djhoNi42NjZ2MS4zMzJIMTJ2LTEuMzMyaDEydi04SDB6bTYuNjY2IDYuNjY0SDUuMzM0di00SDMuOTk5djRIMS4zMzVWOC42NjdoNS4zMzF2NS4zMzF6bTQgMHYxLjMzNkg4LjAwMVY4LjY2N2g1LjMzNHY1LjMzMmgtMi42Njl2LS4wMDF6bTEyLjAwMSAwaC0xLjMzdi00aC0xLjMzNnY0aC0xLjMzNXYtNGgtMS4zM3Y0aC0yLjY3MVY4LjY2N2g4LjAwMnY1LjMzMXpNMTAuNjY1IDEwSDEydjIuNjY3aC0xLjMzNVYxMHoiLz48L3N2Zz4=)](https://www.npmjs.com/package/simple-icons)
```
[npm version](https://www.npmjs.com/package/simple-icons)

> Before "Usage" centered

Also gets my vote 👍

ericcornelissen:
> Before "Usage" centered

Also gets my vote 😃

mondeja:
Since there is consensus, I take advantage of this pull to move the badges using the [Before "Usage" centered](https://github.com/mondeja/simple-icons/blob/test-badges-position-before-usage-centered/README.md) layout.

adamrusted (commented):
LGTM!

mondeja:
> Are we going to try using the NPM textmark, or just leaving it as is?

NPM textmark in the icon implies base64 data hardcoding using `fill="white"`, which means:

- Updates on the icon must be manually updated in the badge (release-action could be edited to accomplish that).
- The README file size increases.

If you still want that version, I'll be happy to do the demo I included above with that icon.

adamrusted:
> If you still want that version, I'll be happy to do the demo I included above with that icon.

I'd be happy to keep things as they are, and hope the lovely people at shields.io choose to use our icon instead - though I also see the argument for showing off our NPM icon within the shield.

mondeja:
> though I also see the argument for showing off our NPM icon within the shield.

I've created [another demo with with NPM wordmark](https://github.com/mondeja/simple-icons/blob/test-badges-position-before-usage-centered-npm-wordmark/README.md). Keep in mind that the version of NPM package should be [hardcoded in the `message`](https://github.com/mondeja/simple-icons/commit/13835ffc28ea244d5fb6b8620d138202c6811f46#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R12) parameter of the image URL, so the release-action, using this layout, should update it for every release of simple-icons.

adamrusted:
I'm noticing a weird black line when using the base64, and spotted it when trialing something the other day. Maybe we'd be best to stick to the `N` they use, and drop them a line to try the wordmark in the next version?

image

mondeja:
I'm not sure of what weird black line you are reffering to, the only thing that could be improved I see in the simple-icons NPM logo is a overlapping coordinate. Feel free to open an issue in [badges/shields](https://github.com/badges/shields), but if I were a maintainer, i would prefer the colored logo to this monochromated one that doesn't look so good at small sizes.

adamrusted:
Nevermind, must be an issue with my setup. I've gone back on a different device and the line is gone.
As mentioned above, I'd be happy to keep the existing NPM icon as it is, just in the new location.

Any other @simple-icons/maintainers have thoughts on this, or should we look to merge just the change in location and addition of the Packagist icon?

PeterShaggyNoble:
My view on it is we should be promoting our own icons in our badges so I think we should go with that.

ericcornelissen:
Given that it requires manually updating the badge every update, I'd be against it. At least the release action should be able to automate it before we change it.

So I think this PR can be merged as is. Agreed?

PeterShaggyNoble:
> Given that it requires manually updating the badge every update, I'd be against it

Ah, in that case then, yeah, we should probably go with Shields' one to avoid that.

adamrusted (approved):
Approving as we seem to have reached a consensus regarding the NPM icon.
Although it's a small change, can another @simple-icons/maintainers approve this before merging though?

fbernhart (approved):
LGTM 👍🏼

PR #4723, opened 2021-01-11
Add hotjar icon

service-paradis:
image

**Issue:** NA
**Alexa rank:** [~1200](https://www.alexa.com/siteinfo/hotjar.com)
  

### Checklist

### Description

- I took the SVG found in [website](https://www.hotjar.com/) footer
- I used Inkscape to scale and adjust
- I took the redish color from the flame found in the website header


PeterShaggyNoble:
I think we should be OK using the flame icon here, @service-paradis. They use it themselves in their favicon and in a few other places such as the header of [this page](https://www.hotjar.com/guides/) and use a few variants of it throughout their site.

service-paradis:
> I think we should be OK using the flame icon here, @service-paradis. They use it themselves in their favicon and in a few other places such as the header of [this page](https://www.hotjar.com/guides/) and use a few variants of it throughout their site.

They mostly use the full wordmark (website header, website footer, social medias), but I agree that the flame icon would be more appropriate for Simple Icons. It is surely recognizable enough.

I just pushed a new version that looks like this:
image

adamrusted (approved):
LGTM! Thanks @service-paradis 🔥🏺

PR #4744, opened 2021-01-12
Rename Bing as Microsoft Bing

PeterShaggyNoble:
"Bing" was renamed and rebranded as "Microsoft Bing" late last year.


adamrusted:
So, the homepage of Bing uses the `Microsoft Bing` logo, but the `<title>` of the page, as well as other mentions of Bing omit `Microsoft`. What do other @simple-icons/maintainers think?

PeterShaggyNoble:
Both [Wikipedia](https://en.wikipedia.org/wiki/Microsoft_Bing) and [Logopedia](https://logos.fandom.com/wiki/Microsoft_Bing) list it as "Microsoft Bing", noting that "Bing" is the outdated name. I'm guessing that, within the context of the site itself, they allow themselves to still refer to it as "Bing" in places as the full name is well established through the logo and in a few other places.

Perhaps more problematic (or just confusing!) for us is that Logopedia states that the Fluent design icon was never officially adopted, only ever trialled and that the actual logo is just the MS logo. This despite the fact that Bing still use the Fluent design icon as [their favicon](https://www.bing.com/sa/simg/favicon-2x.ico).

adamrusted:
I think the use as favicon warrants us using it, unless they have explicit guidelines stating the plain Microsoft logo should be used. I'm happy for this one to be merged in based on https://github.com/simple-icons/simple-icons/pull/4744#issuecomment-846931529 - but will wait for another maintainer to contribute first.

ericcornelissen:
I'd say that the `"title"` should be the _official_ name of the brand, which definitely seems to be "Microsoft Bing" in this case. "Bing" would be a good candidate as an alias for this one, I would say.

As for the SVG, I would agree that we can use the fluent design version of the logo but am not sure. In any case, we can merge this as is to include the rename in the v5 release and update the SVG later if needed.

adamrusted (approved):
Cool, taking on board feedback let's get this merged as is. As mentioned by @ericcornelissen we can look to set up aliases or updated the SVG at a later date, but the update to the `title` is what this PR is focussed on.

PR #4792, opened 2021-01-18
Add Chemex

PeterShaggyNoble:
Chemex

**Issue:** Templarian/MaterialDesign#5088
**Alexa rank:** [~229k](https://www.alexa.com/siteinfo/chemexcoffeemaker.com)

### Checklist

### Description
Icon & colour from [this PDF](https://www.chemexcoffeemaker.com/static/pdf/QuickStart_Otto_final_1.pdf) linked from the end of the source page. I went with just the "X", as seen in their favicon. They do have a [press kit](https://www.chemexcoffeemaker.com/presskit) but, even with an account, it's inaccessible and there may well be something in there that requires us to use the full wordmark or the registered trademark symbol.


ericcornelissen:
> [...] that requires us to use the full wordmark or the registered trademark symbol.

I'd say we can get away with just using the "X", but I'm pretty sure we should include the registered trademark symbol - especially given the following line from our [Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/b4fd6d4d7271492277e5158a7fdeef5ef1735c9d/CONTRIBUTING.md):

> If there is ambiguity about the conditions under which the symbol is required, it must be included if it is a registered trademark symbol (®) but not if is a trademark symbol (™).


---


_(I did not review the SVG or JSON data)_

PeterShaggyNoble:
I'd agree with that, @ericcornelissen as they even using it when referencing the name in text. However, looking back over things myself again, I do think we should go with the full wordmark as the _only_ place I see them using the "X" by itself is in their favicon.

ericcornelissen:
> However, looking back over things myself again, I do think we should go with the full wordmark as the only place I see them using the "X" by itself is in their favicon.

I don't have any strong opinions in that regard. I think just the "X" would look better in our format but indeed it seems they only use that as the favicon.

adamrusted:
Returning to this, it looks like they use just the `X` on top of their boxes (see [here](https://www.amazon.co.uk/Chemex-FS-100-Prefolded-Square-Filters/dp/B000N4W2SG) for an example). It would need the registered trademark icon next to it though.

jorgeamadosoria:
chemex

Added the registered trademark to the icon as pointed out by @adamrusted. 
took the icon from here: https://vtlogo.com/chemex-coffeemaker-vector-logo-svg/ 
not an official source, but the icon is identical to the official and is already an SVG, so I don't think it matters all that much.

IMO, this is ready for merging, but of course it needs a review.

_data/simple-icons.json
```diff
@@ -1506,6 +1506,11 @@
             "hex": "F09820",
             "source": "https://www.chef.io/"
         },
+        {
+            "title": "Chemex",
+            "hex": "4D2B1A",
+            "source": "https://vtlogo.com/chemex-coffeemaker-vector-logo-svg/"
```
sachinraja, line 1512: I don't like the idea of having an unofficial source here...

sachinraja (approved):

sachinraja:
Great job finishing this up @jorgeamadosoria!

PR #4799, opened 2021-01-18
Add Air China

PeterShaggyNoble:
Air China

**Issue:** #2587
**Alexa rank:** [~46.6k](https://www.alexa.com/siteinfo/airchina.com.cn)

### Checklist

### Description
There are a number of imperfections in the source SVG which, initially, seemed to be as a result of the vectorisation process but they're actually all present in [this raster file](https://www.airchina.co.uk/CAPortal/ca-ui_CAPortal_9.11.1.46_141220//portal/skin/img/CA-LOGO-chinese.png) too (among others) so I'm guessing they're actually intentional.


runxel:
Oh wow, that svg is really shoddy!

Tbh I would reject it as it is right now... 🙈
I can't think that this is really they intended their logo to be.

However in the absence of any better source I would slip this through. One demand tho: There is a place which really needs some fixes: (Zoom in very closely, there is a small bridge of points leading into nothing)

grafik

PeterShaggyNoble:
Did some more digging and found [a _much_ better source](http://www.airchina.com.cn/en/investor_relations/images/financial_info_and_roadshow/2020/09/22/B1084DA9D8C8450268D4404BCDE841E5.pdf), @runxel.

adamrusted (approved):
Spot on! Thanks @PeterShaggyNoble ✈

Issue #3443, opened 2020-08-13
Burger King Icon


bramleydev:
**Name: Burger King**
**Website: https://www.bk.com/**
**Alexa rank: 26,260**
  

[...]

PR #4804, opened 2021-01-18
Add Hungry Jack's

PeterShaggyNoble:
Hungry Jack's

**Issue:** Follow-up to #3443 & #4801
**Alexa rank:** [~120k](https://www.alexa.com/siteinfo/hungryjacks.com.au)

### Checklist

### Description
Icon from [SVG](https://www.hungryjacks.com.au/App_Themes/HJ/assets/images/HJLogo.svg) in website header - it looks like they use the boxed version consistently across their website & packaging. I went with the more vibrant red from their website stylesheet for the colour rather than the one used in the source SVG.


adamrusted:
Do we need to keep the ® being as it's in both shown versions of the logo?

NovaGL:
I wouldn't remove the registered sign I think that's a requirement for logos with registered trademarks.

PeterShaggyNoble:
Our stance on symbols is:
1. If there is a explicit requirement to include it then we include it, regardless of the symbol.
2. If there is a requirement to include except under specific circumstances then we base our decision on whether our usage falls within those circumstances.
3. If there is a requirement to include except under specific circumstances but it's unclear or ambiguous whether our usage falls within those circumstances then we include it if it's a registered trademark symbol but omit it otherwise.
4. If there is nothing stated anywhere regarding its inclusion then we omit it, regardless of the symbol.

So, unless we can find some guidelines covering its inclusion, I think we're safe to omit it here under point 4 above.

adamrusted:
As @PeterShaggyNoble mentioned above, given there isn't an official trademark document readily available for Hungry Jack's - I'm _happier_ to look at merging this one without the ® for now - though I would note that even the tiny icons on their [voucher books](https://www.hungryjacks.com.au/Upload/HJ/Media/HJ00428_National_Voucher_December_WEB.pdf), they include the ® symbol. If the other @simple-icons/maintainers could weigh in on this, that'd be great.

PeterShaggyNoble:
Just realised: We did decide a while back that it the symbol is integrated into the logo/icon, as it was in the old Intel icon and in the Chupa Chups icon, then we should _include_ it if we're not explicitly allowed to omit it. The fact that they themselves include it when displaying the logo a small sizes would also lead me to believe we need to keep it. Will get this PR updated later today.

PeterShaggyNoble:
Updated.

adamrusted (approved):
Looks good to me! Thanks @PeterShaggyNoble 🍔

Issue #1368, opened 2019-04-21
Create custom Open Graph image for SimpleIcons


ericcornelissen:
As per [this blogpost](https://github.blog/2019-04-17-custom-open-graph-images-for-repositories/) by GitHub it is now possible specify a custom [Open Graph](http://ogp.me/) image for the repository. It currently looks like ⬇️ 

![open-graph repository preview](https://user-images.githubusercontent.com/3742559/56468908-4bdaf680-643b-11e9-8c00-024454962695.png)

Similarly, we might think about this for the website, [simpleicons.org](https://simpleicons.org/), as well (though I don't know what that currently looks like).

phatbhoy67:
> Similarly, we might think about this for the website, [simpleicons.org](https://simpleicons.org/), as well (though I don't know what that currently looks like).

The image for the website is (https://simpleicons.org/images/og.png)
`<meta property="og:image" content="https://simpleicons.org/images/og.png">`

![og](https://user-images.githubusercontent.com/18670428/67239281-cd412f00-f446-11e9-9261-887e30858764.png)

@ericcornelissen are you still looking to get this updated/added?

ericcornelissen:
Oh 😅 I had never realized that...

Given [the template GitHub gives](https://github.com/simple-icons/simple-icons/settings/og-template) we may consider changing it, I would personally be a fan of an image with some more detail/larger icons as well. 🤔

ericcornelissen:
> looks like a good place to start: icon | name | tag line | blurb?

I doesn't have to be to fancy I think, continuing on [the current OG image](https://simpleicons.org/images/og.png) is already fine in my opinion. Below is a quick preview of what I was describing in my previous comment.

That said, if you want to design something @phatbhoy67, go right ahead 😄 

| New OG Image | New OG Image w/ template overlayed |
|---|---|
| ![og-1](https://user-images.githubusercontent.com/3742559/67316402-08a83000-f511-11e9-8056-2cf6a15ea151.jpg) | ![og](https://user-images.githubusercontent.com/3742559/67316404-09d95d00-f511-11e9-9658-25ad9d36a195.jpg) |

PeterShaggyNoble:
Something I was playing around with this morning:

![](https://user-images.githubusercontent.com/15157491/76525578-0022a880-6464-11ea-92af-2e1b5be55194.png)

PeterShaggyNoble:
Take 2:

[...]

PR #4816, opened 2021-01-19
Add some icons in Readme

service-paradis:
**Issue:** N/A
**Alexa rank:** N/A
  

### Checklist
  - [ ] ~~I updated the JSON data in `_data/simple-icons.json`~~

### Description

I think SimpleIcons `README` could be a good place to show some icons we have. 
The results can be seen [here](https://github.com/service-paradis/simple-icons/blob/readme-icons-proposal/README.md).


adamrusted:
I'm having the below issue due to GitHub dark mode - is there any way we could add `@media (prefers-color-scheme:dark)` to the readme, and have it take effect to recolour the SVGs?

image

service-paradis:
> I'm having the below issue due to GitHub dark mode, where the icons are appearing black (when really, a white would be better) - is there any way we could add `@media (prefers-color-scheme:dark)` to the readme, and have it take effect to recolour the SVGs?

I have the same here with dark mode, and I agree that it is not ideal. Unfortunately (as discussed a little bit [here](https://github.com/simple-icons/simple-icons/issues/1368#issuecomment-743002189)), there are no easy alternatives to detect scheme in a Readme.

mondeja:
I would really like this improvement, but as @adamrusted has pointed, the dark mode doesn't helps. We first must solve that issue for the simple-icon icon at the top. I've done some tests ([see here](https://github.com/simple-icons/simple-icons/issues/1368#issuecomment-743429184)), also using `@media (prefers-color-scheme:dark)` with examples that I've found in the web, without success. Maybe, Github should provide a clear solution for this inconsistencies of the READMEs in dark mode before we can continue with this :thinking:

service-paradis:
> I would really like this improvement, but as @adamrusted has pointed, the dark mode doesn't helps. We first must solve that issue for the simple-icon icon at the top. I've done some tests ([see here](https://github.com/simple-icons/simple-icons/issues/1368#issuecomment-743429184)), also using `@media (prefers-color-scheme:dark)` with examples that I've found in the web, without success (lack of browsers support). Maybe, Github should provide a clear solution for this inconsistencies of the READMEs in dark mode before we can continue with this 🤔

You are right! Now that Github supports dark mode, they should give people the ability to adapt their README file!
There could be a possibility by using shields.io. Something like `https://img.shields.io/badge/%20-%20-white?logo=typescript` that gives typescript

adamrusted:
@mondeja there's a gist [here](https://github.community/t/dark-mode-detection-in-profile-readme/149403) if that's any help?

mondeja:
> @mondeja there's a gist here if that's any help?

Nope, as far as I know, we can't add Javascript to the README, I wish it could.

service-paradis:
There is an interesting discussion [here](https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/37) on this subject.

fbernhart:
@service-paradis I just found the same link. 👍🏼 Seems like A LOT of people are requesting this and seems like GitHub is really working on this. So probably we'll get an official solution in a few weeks.

fbernhart:
For now I would merge it with the SVGs in black. And we can create another PR once GitHub fully supports dark mode.

@service-paradis How about adding the `npm` or `Node.js` icon to the "Node Usage" section? And maybe the `JSDelivr` icon to the "CDN Usage" section? Maybe we could need the `Unpkg` icon as well. 🤔

PeterShaggyNoble:
We could put white circles behind them, and our own logo at the top of the page, in the meantime so that they can be seen when using the dark theme.

ericcornelissen:
Cool, nice work @service-paradis. I'm personally not such a big fan of the icons in-text (especially not if it is inconsistent because `Unpkg` does not yet have an icons...), but in front of headers is definitely nice!

Also, I wonder if the margin between the icon and titles can be reduced?
 
> We could put white circles behind them, and our own logo at the top of the page, in the meantime so that they can be seen when using the dark theme.

I would agree that this is a sufficient solution for now 👍 Relatedly, I would be in favor of a PR implementing @mondeja's [suggestion for the simple-icons logo at the top](https://github.com/simple-icons/simple-icons/issues/1368#issuecomment-743429184).

ericcornelissen (changes requested):
Just noticed this now but could you change the URLs to be relative in the repository, rather than absolute to the website? I.e. replace `https://simpleicons.org/icons` by `./icons`, e.g.:

```diff
-  https://simpleicons.org/icons/npm.svg
+  ./icons/npm.svg

**Comment by service-paradis** [comment]
> Cool, nice work @service-paradis. I'm personally not such a big fan of the icons in-text (especially not if it is inconsistent because `Unpkg` does not yet have an icons...), but in front of headers is definitely nice!

Yes, I agree.

> Also, I wonder if the margin between the icon and titles can be reduced?

Unfortunately, It is part of Github Stylesheet. I do not have any way to change this.
```css
.markdown-body img[align=left] {
    padding-right: 20px;
}
```

> > We could put white circles behind them, and our own logo at the top of the page, in the meantime so that they can be seen when using the dark theme.
> 
> I would agree that this is a sufficient solution for now 👍 Relatedly, I would be in favor of a PR implementing @mondeja's [suggestion for the simple-icons logo at the top](https://github.com/simple-icons/simple-icons/issues/1368#issuecomment-743429184).

We would need to create images especially for the README file though instead of using the existing SVG files directly.

> Just noticed this now but could you change the URLs to be relative in the repository, rather than absolute to the website? I.e. replace https://simpleicons.org/icons by ./icons, e.g.:

Sure!

adamrusted:
> We would need to create images especially for the README file though instead of using the existing SVG files directly.

Not 100% up to speed on Markdown/README files, but would there be a way for us to wrap the existing icons you've added in a div/span that has a white background? Other option would be some sort of call to an API like Shields.io that could combine the current version of the icon with a circular path - but that's a hell of a lot of work for an aesthetic issue that may be fixed soon by GitHub.

service-paradis:
> > We would need to create images especially for the README file though instead of using the existing SVG files directly.
> 
> Not 100% up to speed on Markdown/README files, but would there be a way for us to wrap the existing icons you've added in a div/span that has a white background? Other option would be some sort of call to an API like Shields.io that could combine the current version of the icon with a circular path - but that's a hell of a lot of work for an aesthetic issue that may be fixed soon by GitHub.

Unfortunately, we dont have much flexibility to style component in MArkdown. I made an example [here](https://github.com/simple-icons/simple-icons/pull/4816#issuecomment-763091412) using a simple shields.io config.

As you said, we can probably wait a little bit to see Github solution for this.

adamrusted:
Looks like there's not yet been an official implementation from GitHub regarding Light/Dark mode images. Only thing I can see is by using a `<picture>` tag, with a `<source>` specifically targeting dark mode users - though that's unlikely to translate nicely to MD.

service-paradis:
> Looks like there's not yet been an official implementation from GitHub regarding Light/Dark mode images. Only thing I can see is by using a `<picture>` tag, with a `<source>` specifically targeting dark mode users - though that's unlikely to translate nicely to MD.

Will mark this issue as `pending` for now. No official workaround and no easy fix.

adamrusted:
@service-paradis - just a thought for when GitHub get round to adding adaptive images, might it help tidy up the `third party extensions` to have the logo of the brand, a link to the extension, and the creators in a table, as opposed to various headings?

service-paradis:
> @service-paradis - just a thought for when GitHub get round to adding adaptive images, might it help tidy up the `third party extensions` to have the logo of the brand, a link to the extension, and the creators in a table, as opposed to various headings?

I like the idea! What do you think of [this](https://github.com/service-paradis/simple-icons/blob/readme-icons-proposal/README.md)?

adamrusted:
> I like the idea! What do you think of [this](https://github.com/service-paradis/simple-icons/blob/readme-icons-proposal/README.md?rgh-link-date=2021-02-27T19%3A59%3A03Z)?

Looks great - though obviously subject to us being able to re-colour the SVGs...

adamrusted:
@service-paradis I've just seen the folks over at SVGO solve our issue with the SVGs on Dark Mode (see https://github.com/svg/svgo/blob/new-logo/logo/logo-web.svg?short_path=a7df137)

It may require some duplication of icons into a `Readme Icons` folder, but with a bit of CSS built into those SVGs you're able to change the colour scheme. Might I suggest we set them to brand colour, unless Dark Mode is specified?

service-paradis:
> @service-paradis I've just seen the folks over at SVGO solve our issue with the SVGs on Dark Mode (see https://github.com/svg/svgo/blob/new-logo/logo/logo-web.svg?short_path=a7df137) using a media query embedded in the SVG file itself.
> 
> It may require some duplication of icons into a `Readme Icons` folder, but with a bit of CSS built into those SVGs you're able to change the colour scheme. Might I suggest we set them to brand colour, unless Dark Mode is specified?

Unfortunately, it wont solve every cases. `prefers-color-scheme` try to detect the System color scheme preference. We can also manually toggle Github Scheme. I can have the system configured as light and use Github dark theme (or dark system with light Github). It could solve the issue for most of the users though, but until Github comes with a solution, it wont fix Readme for everyone.

Examples:

##### SVGO Readme on a dark System Scheme with Github Dark theme (looking good)
image

##### SVGO Readme on a dark System Scheme with Github Light theme (looking ??)
image

adamrusted:
@service-paradis, given the recent boom in third-party extensions, would it be worth breaking the 'table' into it's own PR to allow us to tidy the entries up ahead of adding the icons to the README?

service-paradis:
> @service-paradis, given the recent boom in third-party extensions, would it be worth breaking the 'table' into it's own PR to allow us to tidy the entries up ahead of adding the icons to the README?

Done #5563 😃

jorgeamadosoria:
Would it make sense just having a special set of icons just for this purpose? The Readme looks positively fantastic with these icons added. Adding a few icons that are not exported and only exists for this is maybe not that big of a deal?

jorgeamadosoria:
image

Is this good enough?

jorgeamadosoria:
> > @service-paradis, given the recent boom in third-party extensions, would it be worth breaking the 'table' into it's own PR to allow us to tidy the entries up ahead of adding the icons to the README?
> 
> Done #5563 😃

another possibility:

image

image

mondeja (commented):
Really stange, the images are not loading, nor in Github UI nor in local with two different Markdown viewers:

### Brave browser (Chromium based)

image


### Firefox

image

I'm seeing the next message trying to access [the URLs of the images](https://raw.githubusercontent.com/service-paradis/simple-icons/readme-icons-proposal/readme-icons/simpleicons.svg) directly:

image

Can you load them @jorgeamadosoria? If so, what viewers and/or browsers are you using?

service-paradis:
If we go with this solution, we'll also need to think of a way to simplify it for [npm](https://www.npmjs.com/package/simple-icons). I'm not sure they support this feature.

mondeja:
> If we go with this solution, we'll also need to think of a way to simplify it for [npm](https://www.npmjs.com/package/simple-icons). I'm not sure they support this feature.

Yes, I think the better solution would be to replace the links before the release, just [after this line](https://github.com/simple-icons/simple-icons/blob/cfb81045db29ace972eea43142172460a7d223d8/.github/workflows/publish.yml#L52). Maybe using an action like https://github.com/marketplace/actions/replace-action

@jorgeamadosoria could you use absolute links and we'll change them the branch to which they would point just before doing the merge?

jorgeamadosoria:
> If we go with this solution, we'll also need to think of a way to simplify it for [npm](https://www.npmjs.com/package/simple-icons). I'm not sure they support this feature.

we discussed that with @mondeja , he has a solution.

jorgeamadosoria:
Fixed the issue. The problem was not absolute paths, but rather that the SVGs were corrupted on download, so they were never going to show.
I substitute them and I can see the readme just fine now

jorgeamadosoria:
I don't know how to work with Github Actions.
How can that be tested?

mondeja (dismissed):
Great!

Drupal and Home Assistant white icons are missing in the [readme-icons](https://github.com/service-paradis/simple-icons/tree/ad783322d698df6e7af4856340697a0eb233c9e0/readme-icons) directory.

Also, could you add [`diagrams.net` icon](https://simpleicons.org/?q=diagrams.net) for the Drawio library and [`hexo` icon](https://simpleicons.org/?q=hexo) for the Hexo plugin in the extensions table?

To test that part of the workflow, the better option would be to create another temporal repository with the README file and a simple workflow with the affected steps. I can create it if you don't know how to, but first let's prepare the readme as it will be 👍🏼

sachinraja:
Can we maybe move `readme-icons` to `assets/readme` to avoid cluttering the root directory with any unnecessary files later? `assets` is more generic so other files can be added there if necessary later on.

jorgeamadosoria:
I've added the icons for Drawio, Hexo and Jetpack Compose.
I also moved them to assets/readme.

Only thing missing now would be the readme handling for npm.

mondeja (commented):
Great 👍🏼 

> Only thing missing now would be the readme handling for npm.

I've been thinking about this two weeks. It seems to me that the best for us and for the community overall would be to create a Github Action that would strip `#gh-light-mode-only` and `#gh-dark-mode-only` parts of links and dark or light links (with a `style` parameter to keep in files one of the alternatives choosing between `dark` or `light`).

This separate action would be more benefical than adding a custom script here and should be more maintainable at the long term. I think I'll work on this soon, I'll let you know here when I have it @jorgeamadosoria

jorgeamadosoria:
> Great 👍🏼
> 
> > Only thing missing now would be the readme handling for npm.
> 
> I've been thinking about this two weeks. It seems to me that the best for us and for the community overall would be to create a Github Action that would strip `#gh-light-mode-only` and `#gh-dark-mode-only` parts of links and dark or light links (with a `style` parameter to keep in files one of the alternatives choosing between `dark` or `light`).
> 
> This separate action would be more benefical than adding a custom script here and should be more maintainable at the long term. I think I'll work on this soon, I'll let you know here when I have it @jorgeamadosoria

whatever you think is easier to make / maintain, @mondeja. I have zero experience with Github actions so I can't offer an opinion on it.
Thank you for helping out on this.

mondeja:
Created an action for this at [mondeja/strip-gh-theme-links](https://github.com/mondeja/strip-gh-theme-links
). I'm going to continue with this PR making a stable release of the action testing these replacements in a separate repo. Thank you for the great work @jorgeamadosoria!

mondeja:
This is almost ready, it needs just two changes more:

- [ ] Add Jetpack Compose icon to the library. Adding it in #6983
- [ ] Create a test to maintain white icons synced with correspondent library icons (check that they have an additional attribute `fill="white"` as the first of `svg` tags, compared with the standard version in the library)

jorgeamadosoria:
> 

I've removed the icons from assets/readme. Sourcing the icons from the icons folder and adding a style of fill:white on the readme removes the need for any testing (there is no need to keep the icons synchronized nor apart from the library, so there is no need for additional testing).
Only thing left is to merge jet compose and we should be good togo.

mondeja:
> I've removed the icons from assets/readme. Sourcing the icons from the icons folder and adding a style of fill:white on the readme removes the need for any testing (there is no need to keep the icons synchronized nor apart from the library, so there is no need for additional testing).
Only thing left is to merge jet compose and we should be good to go.

Tested in https://github.com/simple-icons/simple-icons/pull/4816/commits/1f3f848466141ce258176a05a9602eebcebfbfcb and  https://github.com/simple-icons/simple-icons/pull/4816/commits/e3b1a8373917fb3128e7755b88e52582f72bfbee and it seems that is not true. Are you sure @jorgeamadosoria?

jorgeamadosoria:
> > I've removed the icons from assets/readme. Sourcing the icons from the icons folder and adding a style of fill:white on the readme removes the need for any testing (there is no need to keep the icons synchronized nor apart from the library, so there is no need for additional testing).
> > Only thing left is to merge jet compose and we should be good to go.
> 
> Tested in [1f3f848](https://github.com/simple-icons/simple-icons/commit/1f3f848466141ce258176a05a9602eebcebfbfcb) and [e3b1a83](https://github.com/simple-icons/simple-icons/commit/e3b1a8373917fb3128e7755b88e52582f72bfbee) and it seems that is not true. Are you sure @jorgeamadosoria?

Yes, I'm sure.
I just missed the ones outside the 3rd party table.
But the folder is removed, and all the icons are now pulling from the icons folder, the only difference is that the white ones have a filter:invert(1) style applied.

mondeja:
It doesn't work @jorgeamadosoria:

image

Is working for you?

jorgeamadosoria:
Locally, yes.

mondeja:
It doesn't works in Github. Reverted the changes.

jorgeamadosoria:
Style doesn't work in Github in any way except into a foreignObject in the SVGs, it turns out. And that's overly complicated.
I've added the test for the light icons vs dark icons, and readded the icons in assets/icons.

assets/readme/vuedotjs.svg
```diff
@@ -0,0 +1,4 @@
+<svg fill="white" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
```
mondeja, line 1: Please minimize the icons. Also, why do you've changed the names from `<icon>-white` to `<icon>`? It seems to me more understandable with `-white` because are the white versions.

README.md
```diff
@@ -105,16 +106,16 @@ This is useful if you want to do a computation on every icon:
...
 }
 ```
 
-#### TypeScript Usage
+#### TypeScript Usage <img src="./icons/typescript.svg#gh-light-mode-only" alt="Typescript" align=left width=24 height=24><img src="./assets/readme/typescript.svg#gh-dark-mode-only" alt="Typescript" align=left width=24 height=24>
 
 Type definitions are bundled with the package.
 
-### PHP Usage
+### PHP Usage <img src="./icons/php.svg#gh-light-mode-only" alt="Php" align=left width=24 height=24><img src="./assets/readme/php.svg#gh-dark-mode-only" alt="Php" align=left width=24 height=24>
```
mondeja, line 118: Paths must be absolute or will not work in NPM not Packagist.

tests/readme-icons.test.js
```diff
@@ -0,0 +1,34 @@
...
+import * as assert from 'uvu/assert';
+import path from 'path';
+import * as fs from 'fs';
+import { fileURLToPath } from 'url';
+
+(async () => {
+  test(`Check readme light and dark icons`, () => {
+    const __dirname = path.dirname(fileURLToPath(import.meta.url));
+    const lightIconsPath = path.resolve(__dirname, '../assets/readme');
+    const darkIconsPath = path.resolve(__dirname, '../assets/readme');
```
mondeja, line 11: These path-related variables can be placed outside or they will be executed multiple times.

tests/readme-icons.test.js
```diff
@@ -0,0 +1,34 @@
...
+    fs.readdirSync(lightIconsPath).forEach((file) => {
```
mondeja, line 12: We are in async context, please use the async version of the function.

Also, `.forEach` adds more uneeded complexity here, perhaps a `for of` loop is enough.

mondeja (changes requested):

tests/readme-icons.test.js
```diff
@@ -0,0 +1,34 @@
+import { test, exec } from 'uvu';
+import * as assert from 'uvu/assert';
+import path from 'path';
+import * as fs from 'fs';
```
mondeja, line 4: Use the `node:` protocol:

```suggestion
import path from 'node:path';
import { promises as fs } from 'node:fs';
```

tests/readme-icons.test.js
```diff
@@ -0,0 +1,34 @@
...
+        'utf8',
+      );
+
+      //compare light svgs to dark ones. The only difference should be a fill="white" attribute
+      assert.ok(
+        lightContent
+          .replace(/ /g, '')
+          .replace('fill="white"', '')
+          .localeCompare(darkContent.replace(/ /g, '')),
+      );
```
mondeja, line 28: This is not testing nothing, they are the same file.

jorgeamadosoria: changed darkIconsPath for lightIconsPath in line 14.

tests/readme-icons.test.js
jorgeamadosoria: fixed the path.
The PR is salvageable as is, no need to revert.

tests/readme-icons.test.js
jorgeamadosoria: Sync is fine for what is needed, no need for an async + callback.

assets/readme/vuedotjs.svg
jorgeamadosoria: Icons are taken from the icons folder as is. If the icons are not minimized here the only thing I can think of is that they are not minimized there.
There is no need for white since these are the only versions in assets/readme, and it makes comparison much easier in test.

README.md
jorgeamadosoria: Why absolute paths?

jorgeamadosoria: The icons should not work in NPM or Packagist anyways. Weren't they supposed to be stripped before publishing to anything other than Github??
Since I don't have I restored them to absolute paths since you preferred it like that. I don't have strong opinions on the urls.

tests/readme-icons.test.js
jorgeamadosoria: Don't know what difference does this makes, but sure.

assets/readme/vuedotjs.svg
mondeja: > There is no need for white since these are the only versions in assets/readme, and it makes comparison much easier in test.

Sorry, I don't agree. How makes more easy for tests the addition of `-white` to the files? File names must express their meaning.

README.md
mondeja: The action strips the dark URLs, not absolutify also the white ones.

tests/readme-icons.test.js
mondeja: > Since this is test code, there is no need for the performance advantage of using promises.
> A for..of is probably good enough, yes, but this code has no need for optimization.

So why we've changed the test framework from jest to esbuild in #6898? Sorry, optimizations matters a lot since our tests grows as every icon added to the library. Also, there is no need for callback, use a promise resolving with `await`. I'm going to change it.

mondeja (approved):

mondeja (approved):
Now looks good to me

jorgeamadosoria (approved):

PR #4817, opened 2021-01-19
Draft: Detect potential duplicated issues

### LINKED ISSUE #4128: Detecting potential duplicated issues
@simple-icons/maintainers 

Following an internal discussion about detecting potential duplicated issues. There is an existing action in the marketplace for this kind of thing.:
https://github.com/marketplace/actions/potential-duplicates

This could be evaluated/discussed.

service-paradis:
Close #4128
**Alexa rank:** N/A
  

### Checklist
  - [ ] ~~I updated the JSON data in `_data/simple-icons.json`~~

### Description

Following on the https://github.com/service-paradis/service-paradis/issues discussion and the changes made in https://github.com/bubkoo/potential-duplicates, I made some tests.

You can see the results of my test [here](https://github.com/service-paradis/service-paradis/issues?q=is%3Aissue+is%3Aclosed). In this case, I used a threshold of 0.6 which was a little too low for my expectation. According to the results, a threshold of, at least, 0.7 seems better.

### DIFF
--- a/.github/workflows/potential-duplicates.yml
+++ b/.github/workflows/potential-duplicates.yml
@@ -1,0 +1,56 @@
+name: Potential Duplicates
+on:
+  issues:
+    types: [opened, edited]
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: bubkoo/potential-duplicates@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Label to set, when potential duplicates are detected.
+          label: potential duplicate
+          # Exclude keyworlds in title before detecting.
+          exclude: |
+            icon
+            Icon
+            icons
+            Icons
+            logo
+            Logo
+            logos
+            Logos
+            request
+            request:
+            Request
+            Request:
+            add
+            Add
+            adding
+            Adding
+            update
+            Update
+            updating
+            Updating
+            rename
+            Rename
+            renaming
+            Renaming
+            remove
+            Remove
+            removing
+            Removing
+          # Get issues with state to compare. Supported state: 'all', 'closed', 'open'.
+          state: all
+          # If similarity is higher than this threshold([0,1]), issue will be marked as duplicate.
+          threshold: 0.6
+          # Reactions to be add to comment when potential duplicates are detected.
+          # Available reactions: "-1", "+1", "confused", "laugh", "heart", "hooray", "rocket", "eyes"
+          reactions: 'eyes, confused'
+          # Comment to post when potential duplicates are detected.
+          comment: >
+            Potential duplicates:
+            {{#issues}}
+              - [#{{ number }}] {{ title }} ({{ accuracy }}%)
+            {{/issues}}
--- a/.github/workflows/potential-duplicates.yml
+++ b/.github/workflows/potential-duplicates.yml
@@ -44,10 +44,10 @@ jobs:
           # Get issues with state to compare. Supported state: 'all', 'closed', 'open'.
           state: all
           # If similarity is higher than this threshold([0,1]), issue will be marked as duplicate.
-          threshold: 0.6
+          threshold: 0.7
           # Reactions to be add to comment when potential duplicates are detected.
           # Available reactions: "-1", "+1", "confused", "laugh", "heart", "hooray", "rocket", "eyes"
-          reactions: 'eyes, confused'
+          reactions: eyes
           # Comment to post when potential duplicates are detected.
           comment: >
             Potential duplicates:


service-paradis:
@ericcornelissen 
As discussed [here](https://github.com/bubkoo/potential-duplicates/issues/14), he added the case insensitivity to the excluded words.
I updated my PR accordingly.

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: bubkoo/potential-duplicates@v1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Label to set, when potential duplicates are detected.
+          label: potential duplicate
+          # Exclude keyworlds in title before detecting.
```
ericcornelissen, line 14:
```suggestion
          # Exclude keywords in title before detecting.
```

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+          exclude: |
+            icon
+            icons
+            logo
+            logos
+            request
+            request:
```
ericcornelissen, line 21: Are these needed separately as well? (quite possible if the exclusion is "word" based)

If so, ideally the action would be updated to ignore punctuation before and after a word - or at least have an option for this. What do you think @service-paradis

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+            add
+            adding
+          # Get issues with state to compare. Supported state: 'all', 'closed', 'open'.
+          state: all
+          # If similarity is higher than this threshold([0,1]), issue will be marked as duplicate.
+          threshold: 0.7
+          # Reactions to be add to comment when potential duplicates are detected.
+          # Available reactions: "-1", "+1", "confused", "laugh", "heart", "hooray", "rocket", "eyes"
+          reactions: eyes
```
ericcornelissen, line 30: Personally not so much of a fan of this, what does it add that the label doesn't? :thinking:

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+          exclude: |
+            icon
+            icons
+            logo
+            logos
+            request
+            request:
+            add
+            adding
+          # Get issues with state to compare. Supported state: 'all', 'closed', 'open'.
```
ericcornelissen, line 24: I don't think we should keep the inline "documentation" when this gets merged.

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+          threshold: 0.7
+          # Reactions to be add to comment when potential duplicates are detected.
+          # Available reactions: "-1", "+1", "confused", "laugh", "heart", "hooray", "rocket", "eyes"
+          reactions: eyes
+          # Comment to post when potential duplicates are detected.
+          comment: >
+            Potential duplicates:
+            {{#issues}}
+              - [#{{ number }}] {{ title }} ({{ accuracy }}%)
+            {{/issues}}
```
ericcornelissen, line 36: I would prefer something closer to a complete English sentence here, what about?:

```suggestion
            This issue is potentially a duplicate of one of the following issues:
            {{#issues}}
              - [#{{ number }}] {{ title }} ({{ accuracy }}%)
            {{/issues}}
```

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
+name: Potential Duplicates
+on:
+  issues:
+    types: [opened, edited]
+jobs:
```
ericcornelissen, line 5: Minor, but for consistency with the other Workflow files could you insert a newline here :smile: 

```suggestion
    types: [opened, edited]

jobs:
```

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,36 @@
...
+            adding
+          # Get issues with state to compare. Supported state: 'all', 'closed', 'open'.
+          state: all
+          # If similarity is higher than this threshold([0,1]), issue will be marked as duplicate.
+          threshold: 0.7
+          # Reactions to be add to comment when potential duplicates are detected.
+          # Available reactions: "-1", "+1", "confused", "laugh", "heart", "hooray", "rocket", "eyes"
+          reactions: eyes
+          # Comment to post when potential duplicates are detected.
+          comment: >
```
ericcornelissen, line 32: If I'm not mistaken, the `>` here tells YAML to remove the newlines in the comment template below, but I would expect it to keep them in which case I think this should be `|`.

(see https://yaml-multiline.info/)

ericcornelissen (changes requested):
I know this is still a draft but here's a review anyway - mainly because I think I found another potential improvement to the Action you're using - if we could get this to work that would be amazing :confetti_ball:

.github/workflows/potential-duplicates.yml
service-paradis: With the current action, it would be needed to have them separated. I can remove `request:`. `request` will still be removed from the comparison string, but it will keep `:`. It could lead to have a little bit more false positive than expected.

I can ask the action author if it would be possible to ignore punctuations.

.github/workflows/potential-duplicates.yml
service-paradis: Totally agree, I kept it there to test it out, but it does not add any real value. Will remove it!

.github/workflows/potential-duplicates.yml
service-paradis: Totally agree! Kept them here during my tests. Will remove it now :)

.github/workflows/potential-duplicates.yml
service-paradis: Totally agree!

.github/workflows/potential-duplicates.yml
service-paradis: Hum... Good question!
Newlines are kept right now using the current configuration as seen [here](https://github.com/service-paradis/service-paradis/issues/10)
image

> Lines with extra indentation are also not folded.

I think this is the reason why newlines are kept right now. I can try it out with literal style (`|`) to see if it works the same way!

service-paradis: I can confirm that using the pipe gives the same expected results. Will use it!
image

.github/workflows/potential-duplicates.yml
ericcornelissen: Ah okay, thanks for the explanation - it's a slightly different problem than I expected. I would indeed expect the action to ignore punctuation (or, again, at least have an option to do so).

If the Action's maintainer(s) agree I think we should remove it already now, otherwise I would suggest we ignore most punctuation manualy - am I correct that a including just `:` in the list would remove all colons?

Relatedly, if this works the way you say than I actually think the configuration is currently invalid. The action would first remove the string "request" from the issue text, and then try to remove "request:". But "request:" won't be in the issue text because it just removed "request".

service-paradis: > Relatedly, if this works the way you say than I actually think the configuration is currently invalid. The action would first remove the string "request" from the issue text, and then try to remove "request:". But "request:" won't be in the issue text because it just removed "request".

You're right, my bad 🤦‍♂️ 

> am I correct that a including just : in the list would remove all colons?

You're correct. We might be ok if we list potential punctuation marks (other than colon, not much used in issue title) as exclusion.

ericcornelissen: Should we still not exclude "all" punctuation marks. Beyond the template, in an extreme case, if one issue contains punctuation and an "identical" issue is opened that is just missing the punctuation, I think the issues should be considered identical (of course, it is unlikely that punctuation by itself is going to bring the similarity value below the threshold of 0.7, but still...)

service-paradis: I just checked again the source of the action. It does, in fact, remove a lot of punctuation and common words (https://github.com/bubkoo/potential-duplicates/blob/master/src/dic.ts).
Colo (`:`) is missing though 😅

ericcornelissen: Can you open an issue or submit a Pull Request to expand that list?

service-paradis: Sure, that is on my list :)

service-paradis: https://github.com/bubkoo/potential-duplicates/pull/18 is now merged!

ericcornelissen: Sound good, @service-paradis :+1: 

Wondering if this should be updated given the updates to the issue templates from #6303

ericcornelissen: In that case, I think we can give this a try. Perhaps you can ask for a review from newer members to see (1) make them aware and (2) see if they have any objections?

service-paradis:
With the latest dependency updates and the changes made, I think this PR would be ready to merge if we want to test this out.

I would like to have opinion of other @simple-icons/maintainers about this and if there are any objections.

sachinraja (approved):

.github/workflows/potential-duplicates.yml
```diff
@@ -0,0 +1,29 @@
...
+          label: potential duplicate
+          exclude: |
+            icons
+            icon
+            logos
+            logo
+            adding
+            add
+            requesting
+            request
```
mondeja, line 22: I think that we can add some more words here:

- update
- updating
- need
- for
- remove
- create
- please
- and? (not sure about it)

service-paradis: Good idea for a lot of them, thank you!

I voluntarily omit update and delete variations since I dont want those to be flagged with the original request.

Some common words are also already ignored (ex. `and`):
https://github.com/wow-actions/potential-duplicates/blob/master/src/dic.ts

I will add the other mention from your list

runxel (approved):
Looks good! 💪🏼

mondeja (approved):
Great! 👍🏼

adamrusted:
Merging based on previous approvals.

PR #4821, opened 2021-01-20
Add ANA

PeterShaggyNoble:
ANA

**Issue:** #2587
**Alexa rank:** [~2.4k](https://www.alexa.com/siteinfo/ana.co.jp)

### Checklist

### Description
Icon from [SVG](https://www.ana.co.jp/common-layout2/images/ana-logo-stacked.svg) in source header, colour from [website](https://www.ana.co.jp/en/eur/) stylesheet. I went with "ANA" over "All Nippon Airways" as it looks like that's how they refer to themselves almost exclusively.

A couple of possible alternative treatments, one without the cut-out and one without the shape, which is what the use on they tails of their planes and on social media:


adamrusted:
Only other treatment I can think of is similar to the treatment of `Google Translate`, whereby the lighter colour has an inner stroke. Preview below, I think I prefer yours to this treatment however. Happy to go with this PR if no other @simple-icons/maintainers have any objections?

image

PeterShaggyNoble:
That one doesn't work as well for me as it does for Google Translate, which has the "fold" effect, and because one side is so much thinner than the other here

adamrusted (commented):
LGTM! Thanks @PeterShaggyNoble ✈

PR #4833, opened 2021-01-21
Add more characters to normalisation scripts

PeterShaggyNoble:
**NOTE:** I haven't had a chance to test this myself as yet.

This PR adds a few more accented characters to our normalisation scripts, to pre-empt future issues with brand names such as we had with ŠKODA. If anyone thinks of any others, feel free to push a commit directly to this branch.

@ericcornelissen, if you let me know which files on the new site need to be updated to match this change, I'll put a PR together for that tomorrow.


mondeja:
Have you considered in something like [this implementation](https://stackoverflow.com/questions/990904/remove-accents-diacritics-in-a-string-in-javascript)?

PeterShaggyNoble:
Good point, @mondeja :+1: I'm guessing it was originally done the way it is because support for `normalize` wasn't great at the time. It looks to be pretty widely supported now, though, and, with the new site coming soon, I don't think we need worry too much about compatibility issues in minor browsers.

Doing this would also allow us to remove _all_ "special" characters after the normalisation with

```
.replace(/[^a-z0-9\-]/g, "')
```

`index.html` will still need to be updated manually for each new character replacement or removal we require.

ericcornelissen:
> Good point, @mondeja 👍 I'm guessing it was originally done the way it is because support for `normalize` wasn't great at the time. It looks to be pretty widely supported now, though, and, with the new site coming soon, I don't think we need worry too much about compatibility issues in minor browsers.

The browser support of the new website should be roughly the same as the current website, potentially even better due to the use of automated backwards compatibility tools.


> Doing this would also allow us to remove _all_ "special" characters after the normalisation with
> 
> ```
> .replace(/[^a-z0-9\-]/g, "')
> ```

You mean **before** the normalization with that regular expression? Shouldn't we first normalize accented characters and then do this replace?




> `index.html` will still need to be updated manually for each new character replacement or removal we require.

For now yes, but with the redesign that is no longer necessary 👍

PeterShaggyNoble:
> Shouldn't we first normalize accented characters and then do this replace?

Yep, that's what I'm suggesting. So to take care of all accented and special characters we'd now have:

```
.normalize("NFD")
.replace(/[\u0300-\u036f]/g,"")
.replace(/[^a-z0-9\-]/g,"")
```

We may even be able to skip the second line completely.

I'll also add an additional replacement to convert `ß` to `ss`.

ericcornelissen:
> Yep, that's what I'm suggesting. So to take care of all accented and special characters we'd now have:

Ok, then we're on the same page 👍 


> I'll also add an additional replacement to convert ß to ss.

Good call. Moreover, [the implementation linked by @mondeja](https://stackoverflow.com/questions/990904/remove-accents-diacritics-in-a-string-in-javascript/37511463#37511463) does not seem to cover everything in your script of 68664d1ab52d84c7b64c7360999f6adb2f37d801! Given the description in the StackOverflow post I immediately tested `ŧ` and it doesn't get replace by `t` for me (Ubuntu 19.04, Firefox 84.0.2), and there may be others.

PeterShaggyNoble:
Yeah, as you can see [here](https://tio.run/##TcnLSgIBAIXhd5mVQo4DQU8QLXuBbDHYJMY0I9oFWnW/ECISIiJiZmKT2WBmlk22OOc810S0afXD92/Ze3YpW8wXdlKev@HEcdb3Sr7rmK6fSxi4RQd36OKeRzzhGfq84BWvWWaFVTwgwCMGvGGNdTbYZIttdthlj308YYhnhAw44JAhR5xwynfOGHGuQx1jpFOd6xIvGOMVE7yprIqqqqmuhppqqa2Ouuqpjw/M8IlIgQYaKtRIY03whW9NNVOkuWF6fnHbdvMHTsJYXVk2kmbRKbh21kmk1zK71qJlpX6ztLmezi0Y/3dm/0@ScfwD), these characters will still need to be replaced manually: `đħıĸŀłŧ`. I also need to triple check that `toLowerCase` can handle the uppercase variants of all the characters in the list, as well as `ẞ`.

PeterShaggyNoble:
[Ran a test](https://tio.run/##ZY/LSgIBAEX3fYarGaj5A1fRDwStIpjBBlPMGUZ7LFr0fhAiEiIiYmZiZiZmZtlki3vvd01Cm6DlhXPg3LSz6@QSQcrPL2W9TTeKEl4252VcK@MljcxeEI/hFk3coYV7HvGEZ@jwgle8ZoFFlvCALh7R4w3LrLDKGutssMkW2@zgCX08Y8Aue@xzwCHHnPCdU4ac6VDHGOpU57rEC0Z4xRhvKqioksqqqKqa6mqoqZba6uADU3wiVFc99TXQUCON8YVvTTRVqFnMXPjbv@MH8fkHK@@t@b4bLDs51zD/Ida2k09sGatucmXfN9Yty5o7G1baS2XtA3vRTtqm@btsM4p@AA) using `toUpperCase` instead ('cause I'm a few cans deep, on my phone and too lazy to manually generate the uppercase string to test with `toLowerCase`) and all characters a

ericcornelissen:
> Yeah, as you can see here, these characters will still need to be replaced manually: đħıĸŀłŧ. I also need to triple check that toLowerCase can handle the uppercase variants of all the characters in the list, as well as ẞ.
> 
> Ran a test using toUpperCase instead ('cause I'm a few cans deep, on my phone and too lazy to manually generate the uppercase string to test with toLowerCase) and all characters are supported. Ignore the lone ĸ in the results of the match; Kra doesn't have an uppercase variant.

Nice work @PeterShaggyNoble, even if it is incomplete that should be good enough for now (and a while to come) 👍 

> What's interesting is that ß gets converted to SS but I do think we should explicitly replace it anyway, just in case.

If I'm not mistaken, that's how it is supposed to work in German (I'm sure one of the other maintainers can confirm this).

fbernhart:
As a German native speaker I can confirm that `ß` and `ẞ` (yes, that's the uppercase version. Which was added to the German language just in 2017 😄) are normally replaced by `ss` and `SS` for all the "computer stuff".

PeterShaggyNoble:
OK, so, while `toUpperCase` converts `ß` to `SS`, `toLowerCase` converts `ẞ` to ... `ß`! Yay, JavaScript! 🙄

So, to be safe, we'll need to replace it explicitly. I also decided to keep the `replace`ments of the diacritics separate from the other special characters as we only use the latter in `utils.js` and it makes things easier to change if we ever decide we want to handle accented characters or special characters differently.

Just pushed an update to the PR but I don't have the necessary setup to test locally, particularly `utils.js`. Everything _should be_ fine but can someone double check before I take this out of draft.

@ericcornelissen, would I be correct that [this](https://github.com/simple-icons/simple-icons/blob/260080370a8c9cb67dacf062a9d42160c3927bf2/scripts/utils.js#L15-L25) is the only code block that needs to be updated on the new site?

-----

> Which was added to the German language just in 2017

Thanks for that, @fbernhart. Thought I was going mad when the uppercase version appeared on my phone's keyboard a while back; I _knew_ I never remembered there being an uppercase `ß`! It didn't help that my phone's font rendered them both identically. After doing some research, what's interesting is that it's actually existed since 1969 and was added to Unicode in 2008 but wasn't added to the German language until, as you said, 2017.

scripts/utils.js
```diff
@@ -12,17 +12,17 @@ module.exports = {
...
+            .replace(/ħ/g, "h")
+            .replace(/ı/g, "i")
+            .replace(/ĸ/g, "k")
+            .replace(/ŀ/g, "l")
+            .replace(/ł/g, "l")
+            .replace(/ß/g, "ss")
+            .replace(/ŧ/g, "t")
+            .normalize("NFD")
+            .replace(/[\u0300-\u036f]/g, "")
+            .replace(/[^\w\-]/g, "")
```
ericcornelissen, line 25: I was a bit surprised that `_` is included in `\w` so I would personally prefer we use the RegExp below to be more explicit (note that the hyphen does not need to be escaped):

```suggestion
            .replace(/[^a-zA-Z0-9_-]/g, "")
```

But if there's consensus on using `\w` I'm happy with that as well, its behaviour is well documented after all.

PeterShaggyNoble: Yeah `[^a-z0-9_-]` (`A-Z` isn't needed) is probably better for clarity and easier updating. Updated. I always feels it's safer to escape the `-` here and makes it a bit clearer to people who may not be as familiar with JS.

While I'm at it, do you want me to dupe `normalizeSearchTerm` into `utils.js` and update `titleToFilename` accordingly so you can import `normalizeSearchTerm` from this repo for use on the site?

ericcornelissen: > Yeah `[^a-z0-9_-]` (`A-Z` isn't needed) is probably better for clarity and easier updating. Updated. I always feels it's safer to escape the `-` here and makes it a bit clearer to people who may not be as familiar with JS.

Fair enough, I also only recently realized it's optional :upside_down_face: 


> While I'm at it, do you want me to dupe `normalizeSearchTerm` into `utils.js` and update `titleToFilename` accordingly so you can import `normalizeSearchTerm` from this repo for use on the new site?

I'm not sure if I follow. [`utils.js`](https://github.com/simple-icons/simple-icons/blob/develop/scripts/utils.js) is not exposed through the NPM package of simple-icons so that wouldn't help. Moreover, shouldn't `titleToFilename` and `normalizeSearchTerm` be the same function (perhaps one being the alias of the other for clarity)?

I think we should use your suggestion @PeterShaggyNoble and put the titleToFilename logic in a separate package so that it can easily be shared between repos. We just need to keep `normalizeSearchTerm` in this repo in sync for a bit until the new site is in use (*).

---

(*): because here we can't easily import JS modules on the website - without resorting to the latest browser features - while with the new repo that is easy.

PeterShaggyNoble: > Fair enough, I also only recently realized it's optional

In that case, this should be ready to go now, pending someone testing it.

> I'm not sure if I follow. utils.js is not exposed through the NPM package of simple-icons so that wouldn't help.

Ah, OK, I thought we might be able to pull it in for the new website with some dependency "magic"!

> I think we should use your suggestion @PeterShaggyNoble and put the titleToFilename logic in a separate package so that it can easily be shared between repos

If that what it takes 🙂 I'll leave that to someone with more GH savvy than I to do, separately to this PR.

> Ow and by the way, yes

Once this PR gets tested & merged, I'll submit one to the website's repo to keep the 2 scripts in sync for now.

ericcornelissen (approved):
I did some manual testing and this seems to work as expected, not breaking anything w.r.t. [v4.8.0](https://github.com/simple-icons/simple-icons/releases/tag/4.8.0).


> > I'm not sure if I follow. utils.js is not exposed through the NPM package of simple-icons so that wouldn't help.
>
> Ah, OK, I thought we might be able to pull it in for the new website with some dependency "magic"!
>
> > I think we should use your suggestion @PeterShaggyNoble and put the titleToFilename logic in a separate package so that it can easily be shared between repos
>
> If that what it takes :slightly_smiling_face: I'll leave that to someone with more GH savvy than I to do, separately to this PR.

I'll look into this after this PR has been merged :+1:

ericcornelissen:
Linking https://github.com/simple-icons/simple-icons/issues/1531 here as it is relevant

ericcornelissen:
Merging this with the knowledge that [this line of the build script](https://github.com/simple-icons/simple-icons/blob/de8bff418b63b989597fddc8bdb0ad1572cdc32e/scripts/build-package.js#L51) would fail if an automatic conversion doesn't match a manually selected `icons/*.svg` file name. Of course we should always keep this in mind when reviewing PRs and not blindly follow what the script outputs and instead follow a name that makes sense (and can be automated).

If the website turns out to break from this, that is easy to recover, but hopefully we can switch to the redesign soon enough :smile:  As for sharing this logic between this repository and [simple-icons/simple-icons-website](https://github.com/simple-icons/simple-icons-website), I think we can either **a**) keep the duplicate (we have survived that way for this long) or **b**) extract the logic of `normalizeSearchTerm` into a third repo with a `package.json` in its root and and add it as a dependency both here and in [simple-icons/simple-icons-website](https://github.com/simple-icons/simple-icons-website) (similar to the way [simple-icons/simple-icon-pdf](https://github.com/simple-icons/simple-icons-pdf) is [a dependency](https://github.com/simple-icons/simple-icons-website/blob/d82268f064382c8e9e078b190a939953b8676a38/package.json#L30) of [simple-icons/simple-icons-website](https://github.com/simple-icons/simple-icons-website)). I'm honestly fine either way so would like some feedback on what the preferences of others are.

<sup>For the record, I don't think we should somehow expose the `titleToFilename` function in the NPM package just to make it available to the website, unless it is something that could also be useful to other users of the package - even ignoring package size and the option of making `titleToFilename` tree-shakable.</sup>

PeterShaggyNoble:
> Merging this with the knowledge that this line of the build script would fail if an automatic conversion doesn't match a manually selected icons/*.svg file name.

Doesn't it already fail if `titleToFilename` doesn't match an SVG's file name, regardless of character conversion?

> For the record, I don't think we should somehow expose the titleToFilename function in the NPM package just to make it available to the website, unless it is something that could also be useful to other users of the package

One _possible_ use case I can think of is someone using `simple-icons.json` in their project; having it available in the package would make it easier to convert the `title` entry to the appropriate file name. Or maybe that's too niche a use case? I don't know that having a completely separate repo just for one function is worth it so if we're not including it in the package; I'd be OK with keeping the duplication - it's rare enough that we update it and it will be easier to remember just the 2 locations changes need to be made when it does need to be updated.

ericcornelissen:
> Doesn't it already fail if titleToFilename doesn't match an SVG's file name, regardless of character conversion?

Not sure if I follow but I think we're trying to say the same thing. Basically, changes to `titleToFilename` cannot (should not be able to?) introduce breaking changes as that would result in the build script failing 

<sup>(ignoring edge cases where someone changes `titleToFilename` to output a specific brand slug)</sup>


> One possible use case I can think of is someone using simple-icons.json in their project; having it available in the package would make it easier to convert the title entry to the appropriate file name. Or maybe that's too niche a use case? 

That would be a valid use case I think, but even though we expose the raw simple-icons.json file in the NPM package, if you instead `require('simple-icons');` you get the data of the data file + more data, including the `slug` (which is `titleToFileName(title)`) so I'm not convinced we should expose the function as well.


> I don't know that having a completely separate repo just for one function is worth it so if we're not including it in the package; I'd be OK with keeping the duplication - it's rare enough that we update it and it will be easier to remember just the 2 locations changes need to be made when it does need to be updated.

Would be inclined to agree, we can always change our minds later (e.g. when we update it in one place and forget the other :upside_down_face:) or reconsider if we have more functionality that we want to share.

If anyone is strongly in favor or avoiding the duplication, feel free to let us know here (or in a new issue) and if you have an alternative solution (beyond _GitHub-repo-as-dependency_ and _include-in-existing-npm-package_) please include it in your comment :slightly_smiling_face:

PR #4834, opened 2021-01-21
Add Logopedia

PeterShaggyNoble:
Logopedia

**Issue:** n/a
**Alexa rank:** n/a (see below)

### Checklist

### Description
No Alexa rank (or any other metric) to go on here so instead I'd like to propose the addition of a new metric for Fandom & Gamepedia wikis: Fandom's [Wiki Activity Monitor](https://community.fandom.com/wiki/WAM), where Logopedia currently ranks 31st out of 385k+ wikis with a score of 98.9 & rising. It's also previously been ranked 1st overall and is the top wiki in Fandom's "Lifestyle" category. Even before deciding on a cut-off point, I think we can say that makes it popular enough for inclusion.

On the cut-off, we'll need to decide whether to go by current rank or current score. I vote score as I think it gives a more accurate reflection of the wiki's current popularity and also allows us to see whether it's increasing or decreasing so we can monitor additions that might be hovering around our cut-off point. I'd be in favour of setting that cut-off fairly high so we can avoid the possibility, however remote, of being swamped with requests/contributions of wiki logos so my initial suggestion would be a score of 90.


fbernhart:
Just to provide some more information:

There are currently 343 wikis with a score higher than 90.

Would that be manageable? Or are these already too many?

As the rank seems to be correlated to the score, we could as well pick the rank as metric. And just allow the top 300, 250, 200 (or whatever) wikis.

PeterShaggyNoble:
> Would that be manageable? Or are these already too many?

I was originally toying with suggesting 95 but that felt like too harsh a cut-off. ~340 possibilities wouldn't be _all_ that many when you consider:
1. Our (theoretical) limit for icons on Alexa rank alone is 500,000.
2. Many wikis will use the logo of their subject matter.
3. Many other either use a graphic for a logo or don't have one at all.ck the rank as metric. And just allow the top 300, 250, 200 (or whatever) wikis.

PeterShaggyNoble:
Feck! Looks like that activity monitor is now gone 🤔

ericcornelissen:
Any further input on this one? Should we close it out for now given the lack of a popularity metric we can use here + given the fact that there's no demand for the SVG? (Or is there demand, and I'm not aware of it?)

ericcornelissen:
As part of trying to keep the list of open Pull Requests manageable I will close this Pull Request. This does not necessarily mean this won't be added. If you're interested in seeing this icon being added, please leave a comment here 🙂

PeterShaggyNoble:
Sounds reasonable to me, @ericcornelissen - this one was more of a "nice to have" one as I rely on them a lot when researching and updating icons, thought it'd be nice to give them the recognition. We can revisit it should anyone _actually_ request it.

PR #4852, opened 2021-01-24
Add Okta

### LINKED ISSUE #4802: Request Okta (identity management)
**Brand Name:** okta
**Website:** https://www.okta.com/
**Alexa rank:** [127](https://www.alexa.com/siteinfo/okta.com)

**Official resources for icon and color:** https://en.wikipedia.org/wiki/Okta_(identity_management)#/media/File:Okta_logo.svg

adamalston:
okta

Closes #4802 

### Checklist

### Description
Color taken from the source of the Okta website

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5192,8 +5192,8 @@
         },
         {
             "title": "Okta",
-            "hex": "338CCC",
-            "source": "https://www.okta.com/press-room/media-assets/auras/"
+            "hex": "007DC1",
+            "source": "https://www.okta.com/press-room/media-assets/"
         },
         {
             "title": "OnePlus",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5191,6 +5191,11 @@
             "source": "https://okcupid.com/press"
         },
         {
+            "title": "Okta",
+            "hex": "338CCC",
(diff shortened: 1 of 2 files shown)


_data/simple-icons.json
```diff
@@ -5190,6 +5190,11 @@
             "hex": "0500BE",
             "source": "https://okcupid.com/press"
         },
+        {
+            "title": "Okta",
+            "hex": "338CCC",
+            "source": "https://www.okta.com/press-room/media-assets/auras/"
```
adamrusted, line 5196: Given that the SVG used is on the homepage, I think we'd be best to link there as the source. While the 'auras' show the treatment we give it, that page in particular doesn't include the SVG used. The hex is also `#007DC1` when pulled straight from the SVG, so maybe we should use that as our colour?

adamrusted (changes requested):
Thanks for your work @adamalston 🎉
Couple of tiny changes below though if possible?

_data/simple-icons.json
PeterShaggyNoble: I think https://www.okta.com/press-room/media-assets/ would suffice for the source as it includes the SVG logo in the header, confirms our treatment of it and may be the best source for future contributors to find changes to the treatment of the colour.

adamalston: Updated

PR #4418, opened 2020-12-20
Auto Assign PRs to priorisation project

service-paradis:
As discussed internally before, we use a priorisation project. This workflow would help us auto assign PRs to this project.

- New PRs to Unprioritised column
- PRs labeled as `icon outdated` to Priority 1

### Missing

I'm not able to generate an Organization token to add as Secrets:
https://github.com/srggrs/assign-one-project-github-action#organisation-or-user-project

PR #4863, opened 2021-01-25
Auto assign to project workflow

service-paradis:
### Take 2 (see https://github.com/simple-icons/simple-icons/pull/4418 and https://github.com/simple-icons/simple-icons/commit/bfe305f6c43f1bec882b2dfa6f58d45bd26c974f)

Add new PR to the Organisation Priorisation Project:

- New PRs to Unprioritised column
- PRs labeled as icon outdated to Priority 1

Now using a PAT with necessary permissions. 🤞

### DIFF
--- a/.github/workflows/auto-assign-to-project.yml
+++ b/.github/workflows/auto-assign-to-project.yml
@@ -17,7 +17,6 @@ jobs:
       with:
         project: https://github.com/orgs/simple-icons/projects/2
         column_name: Unprioritised
-
     - name: Assign `icon outdated` pull requests to "Priority 1"
       uses: srggrs/assign-one-project-github-action@1.2.0
       if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
--- a/.github/workflows/auto-assign-to-project.yml
+++ b/.github/workflows/auto-assign-to-project.yml
@@ -11,15 +11,15 @@ jobs:
     runs-on: ubuntu-latest
     name: Assign to Project
     steps:
-    - name: Assign pull requests to "Unprioritised"
-      uses: srggrs/assign-one-project-github-action@1.2.0
-      if: github.event.action == 'opened'
-      with:
-        project: https://github.com/orgs/simple-icons/projects/2
-        column_name: Unprioritised
-    - name: Assign `icon outdated` pull requests to "Priority 1"
-      uses: srggrs/assign-one-project-github-action@1.2.0
-      if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
-      with:
-        project: https://github.com/orgs/simple-icons/projects/2
-        column_name: Priority 1
+      - name: Assign pull requests to "Unprioritised"
+        uses: srggrs/assign-one-project-github-action@1.2.0
+        if: github.event.action == 'opened'
+        with:
+          project: https://github.com/orgs/simple-icons/projects/2
+          column_name: Unprioritised
+      - name: Assign `icon outdated` pull requests to "Priority 1"
+        uses: srggrs/assign-one-project-github-action@1.2.0
+        if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
+        with:
+          project: https://github.com/orgs/simple-icons/projects/2
+          column_name: Priority 1
--- a/.github/workflows/auto-assign-to-project.yml
+++ b/.github/workflows/auto-assign-to-project.yml
@@ -1,0 +1,26 @@
+name: Auto Assign to Project(s)
+
+on:
+  pull_request_target:
+    types: [opened, labeled]
+env:
+  MY_GITHUB_TOKEN: ${{ secrets.AUTO_ASSIGN_WORKFLOW_TOKEN }}
+
+jobs:
+  assign-to-project:
+    runs-on: ubuntu-latest
+    name: Assign to Project
+    steps:
+    - name: Assign pull requests to "Unprioritised"
+      uses: srggrs/assign-one-project-github-action@1.2.0
+      if: github.event.action == 'opened'
+      with:
+        project: https://github.com/orgs/simple-icons/projects/2
+        column_name: Unprioritised
+
+    - name: Assign `icon outdated` pull requests to "Priority 1"
+      uses: srggrs/assign-one-project-github-action@1.2.0
+      if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
+      with:
+        project: https://github.com/orgs/simple-icons/projects/2
+        column_name: Priority 1


.github/workflows/auto-assign-to-project.yml
```diff
@@ -0,0 +1,26 @@
...
+    name: Assign to Project
+    steps:
+    - name: Assign pull requests to "Unprioritised"
+      uses: srggrs/assign-one-project-github-action@1.2.0
+      if: github.event.action == 'opened'
+      with:
+        project: https://github.com/orgs/simple-icons/projects/2
+        column_name: Unprioritised
+
+    - name: Assign `icon outdated` pull requests to "Priority 1"
```
ericcornelissen, line 21: Minor nitpick, but I suggest we remove this extra newline to keep the formatting consistent with [existing workflow files](https://github.com/simple-icons/simple-icons/tree/880c706a50e4d3101dcb869d04d88ea2fcb0a77c/.github/workflows)

```suggestion
        column_name: Unprioritised
    - name: Assign `icon outdated` pull requests to "Priority 1"
```

.github/workflows/auto-assign-to-project.yml
```diff
@@ -0,0 +1,25 @@
...
+      if: github.event.action == 'opened'
+      with:
+        project: https://github.com/orgs/simple-icons/projects/2
+        column_name: Unprioritised
+    - name: Assign `icon outdated` pull requests to "Priority 1"
+      uses: srggrs/assign-one-project-github-action@1.2.0
+      if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
+      with:
+        project: https://github.com/orgs/simple-icons/projects/2
+        column_name: Priority 1
```
ericcornelissen, line 25: Another minor nitpick, but these should all be indented one more level to keep the formatting consistent with [existing workflow files](https://github.com/simple-icons/simple-icons/tree/880c706a50e4d3101dcb869d04d88ea2fcb0a77c/.github/workflows)

```suggestion
      - name: Assign pull requests to "Unprioritised"
        uses: srggrs/assign-one-project-github-action@1.2.0
        if: github.event.action == 'opened'
        with:
          project: https://github.com/orgs/simple-icons/projects/2
          column_name: Unprioritised
      - name: Assign `icon outdated` pull requests to "Priority 1"
        uses: srggrs/assign-one-project-github-action@1.2.0
        if: contains(github.event.pull_request.labels.*.name, 'icon outdated')
        with:
          project: https://github.com/orgs/simple-icons/projects/2
          column_name: Priority 1
```

ericcornelissen (approved):
:+1: Let's hope it works this time

ericcornelissen:
It appears not, I tested the Worfklow on https://github.com/simple-icons/simple-icons/pull/4843 (apologies for the mess :sweat_smile:) with both `target: pull_request` and `target: pull_request_target` and here are my findings:

- With `target: pull_request` we get the error `MY_GITHUB_TOKEN not defined`, see https://github.com/simple-icons/simple-icons/pull/4843/checks?check_run_id=1769489133 (this is likely for security reasons, generally secrets are not exposed in the context of Pull Requests from forks, can't find the docs for this right now though)
- With `target: pull_request_target` we get the error ``, see https://github.com/simple-icons/simple-icons/pull/4843/checks?check_run_id=1769489179 (seems to be because it is not supported by [the action we're using](https://github.com/srggrs/assign-one-project-github-action))

I reverted the changes of merging this PR in https://github.com/simple-icons/simple-icons/commit/7181ea0b31adce9364fc6e131471be8e17b0781d

service-paradis:
Well, I do not understand what is going on or what is missing exactly!
This actions is working fine in a private project I'm working on! Sorry for the trouble 😅

ericcornelissen:
> Well, I do not understand what is going on or what is missing exactly!
> This actions is working fine in a private project I'm working on!

Are you using forks there as well?


> Sorry for the trouble

No problem

service-paradis:
> Are you using forks there as well?

You're right, no I'm not! The problem is probably there!

ericcornelissen:
> You're right, no I'm not! The problem is probably there!

Thanks for the confirmation, then that is almost 100% the culprit (it is for many Actions, as it is very annoying to work around). I'll try to get this fixed up-stream :+1:

mondeja:
This should be fixed in the action. The "ugly token name" problem is tracked in https://github.com/srggrs/assign-one-project-github-action/issues/61 and the `pull_request_target` event, as @ericcornelissen has pointed, is not handled by the action, must be added to [here](https://github.com/srggrs/assign-one-project-github-action/blob/980675fe5ab544e620c79a4d480305058376adc1/entrypoint.sh#L100) and [here](https://github.com/srggrs/assign-one-project-github-action/blob/980675fe5ab544e620c79a4d480305058376adc1/entrypoint.sh#L125). When the action would support the `pull_request_target` event we could add this workflow.

mondeja:
Adding `pull_request_target` support for the action in https://github.com/srggrs/assign-one-project-github-action/pull/62

ericcornelissen:
> This should be fixed in the action. The "ugly token name" problem is tracked in https://github.com/srggrs/assign-one-project-github-action/issues/61

Yes and no, it does solve the problem if you can use `secrets.GITHUB_TOKEN`. But I think we will still require a PAT, which cannot be loaded automatically by the action, so we still need to configure it.

A more important issue than it being ugly is that it is global. Which is not a problem for us currently as all the steps in this workflow need the secret, but if this Workflow has steps that don't need the secret, we should consider configuring it as [a step `env` variable](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsenv).

On a related note, the next time we try to implement this workflow, I wonder if we can define the project URL once and reuse it as an environment variable :thinking: 


> the `pull_request_target` event, as @ericcornelissen has pointed, is not handled by the action, must be added [here](https://github.com/srggrs/assign-one-project-github-action/blob/980675fe5ab544e620c79a4d480305058376adc1/entrypoint.sh#L100) and [here](https://github.com/srggrs/assign-one-project-github-action/blob/980675fe5ab544e620c79a4d480305058376adc1/entrypoint.sh#L125). When the action would support the `pull_request_target` event we could add this workflow.
>
> Adding `pull_request_target` support for the action in [srggrs/assign-one-project-github-action#62](https://github.com/srggrs/assign-one-project-github-action/pull/62)

You beat me to it, thanks @mondeja :smiley:

mondeja:
Given that https://github.com/srggrs/assign-one-project-github-action/pull/62 has been merged, could you reopen this pull using `pull_request_target` as the event @service-paradis?

service-paradis:
> Given that [srggrs/assign-one-project-github-action#62](https://github.com/srggrs/assign-one-project-github-action/pull/62) has been merged, could you reopen this pull using `pull_request_target` as the event @service-paradis?

@mondeja The current version already use `pull_request_target` 😃

mondeja:
I'm not sure if I follow you, the file was removed in https://github.com/simple-icons/simple-icons/commit/7181ea0b31adce9364fc6e131471be8e17b0781d

service-paradis:
> I'm not sure if I follow you, the file was removed in [7181ea0](https://github.com/simple-icons/simple-icons/commit/7181ea0b31adce9364fc6e131471be8e17b0781d), would you mind to reopen this as is?

@mondeja My bad, sorry! Will reopen another one! 🤦‍♂️

PR #4880, opened 2021-01-27
Update Anchor

PeterShaggyNoble:
Anchor

**Issue:** n/a
**Alexa rank:** [~2.7k](https://www.alexa.com/siteinfo/anchor.fm)
### Checklist

### Description
Icon from SVG in [website](https://anchor.fm/) header, colour from stylesheet. See discussion below on the colour choice.

Incidentally, we had the name of this one wrong; before this rebranding it was called "Anchor FM".


adamrusted:
They use a more vibrant purple (`#5000B9`) on their [Facebook](https://www.facebook.com/anchor.fm/), [Twitter](https://twitter.com/anchor), and [Instagram](https://www.instagram.com/anchor.fm/) profiles. Another variation I can find on their homepage is `#4A04AA`. I think given the consistency of use across all 3 platforms, we should perhaps go for `#5000B9` as our icon colour?

PeterShaggyNoble:
Interesting; I hadn't noticed the colour change as you scroll down the page. The last shade is actually also `#5000b9` so, given the lighter purples used in the locations you mentioned and in their favicon, let's go with that.

adamrusted (approved):
With the hex updated, this icon is spot on. Thanks @PeterShaggyNoble 🎉

PR #4904, opened 2021-02-01
Update openSUSE

PeterShaggyNoble:
openSUSE

**Issue:** n/a
**Alexa rank:** [~60.7k](https://www.alexa.com/siteinfo/opensuse.org)

### Checklist

### Description
They [allow the use](https://en.opensuse.org/openSUSE:Artwork_brand#Buttons) of the Gecko's head in circle, which would suit us better. They have a few different options so I went with the one they use themselves in the header of their wiki.


adamrusted (approved):
LGTM, like the addition of the guidelines link too. Thanks @PeterShaggyNoble 👌🏻

Issue #4735, opened 2021-01-12
Define different threshold for removing a brand


fbernhart:
As suggested in #3465, it would be helpful in my opinion to define a different threshold for removing a brand than for adding it.

### Problem
As we've probably all seen in the past, the Alexa rank isn't a very "stable" measurement of popularity sometimes. Once in a while it happens that we're adding a brand that is within scope, just to see it's Alexa rank dropping in the next few months or even weeks to `out of scope`.

Example: #3465
- Alexa rank when the PR was created and merged in August 2020: 219,404
- Alexa rank now (January 2021):  [683,075](https://www.alexa.com/siteinfo/hackclub.com)

So according to our contribution guidelines right now, we would already have to remove it - throwing all the work that has been done 5 months ago away.

And probably as soon as we've removed it, the Alexa rank will climb up again to 450k and someone will request the same icon again.

### Suggestion
I'd suggest defining a different threshold for removing a brand. Something like:

- Alexa rank of 500k and below in order to be added to Simple Icons
- If the icon is already part of Simple Icons and if it ever drops beyond an Alexa rank of 1 million, we'll remove it.
This would give us a bit of "free space".

1 million is just a number here. I'm open to discuss different thresholds for removing a brand. 750k? 1.5 million?

@simple-icons/maintainers any opinions or other ideas how to solve this problem?

adamrusted:
I think it may be worth mentioning #4725 here - as the icon was proposed, and initial work put into it, before the Alexa rank plummeted. Should this fall into the same rank as 'icons already added', as similar amounts of work have gone into constructing the icon for the collection?

PeterShaggyNoble:
Here were my thoughts on it over the weekend, all assuming that the Alexa rank is the only thing we have to go on.

1. A brand dipping below 500k should not be grounds for _immediate_ consideration for removal.
2. Only when it reaches a threshold of 600k* should it be considered for removal and an issue opened for it.
3. If, after 30* days, there is no sign of recovery (rank is stable or still declining) then a PR should be opened for its removal in the next major release.
4. If, after those initial 30* days, its rank is instead increasing but not yet at 500k then we give it another 30* days before reviewing it on both points 3 and 4.
5. On the week of the major release, the rank should be reviewed with points 3 and 4 in mind prior to merging the PR.

[...]

PR #4910, opened 2021-02-02
Remove PJSIP

PeterShaggyNoble:
**Issue:** #2018
**Alexa rank:** [~829k](https://www.alexa.com/siteinfo/pjsip.org)

### Checklist

### Description
PJSIP's Alexa ranking has taken a beating since it was added and, with only 742 stars on [GitHub](https://github.com/pjsip/pjproject), is therefore a candidate for removal from our library in `v5`.

PJSIP


PeterShaggyNoble:
Alexa rank has recovered significantly but is still below 500k.

PeterShaggyNoble:
OK, so, with the decision in #4735 to set a secondary cut-off of 750k for brands that have already been added we can now keep this based on it's current rank of 635k. As it has only recovered in the past few days I won't close it yet so that we can continue monitoring it between now and Sunday. If it is still within scope come Sunday then I would suggest that this PR be moved to the v6 milestone, again because the recovery is only recent.

ericcornelissen:
The Alexa rank has currently at ~1.5M - while it was at ~400k 90 days ago, the fact that it was already due for possible removal in [v5](https://github.com/simple-icons/simple-icons/milestone/4) I think we can somewhat safely say that it's unlikely going to meet the popularity threshold within the planned v6 lifetime.

Unless a @simple-icons/maintainers disagrees, we can merge this in somewhere before November 27, 2021.

sachinraja (approved):

mondeja (approved):

Issue #4797, opened 2021-01-18
Adding Aliases


PeterShaggyNoble:
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### Description
In somewhat of a follow-up of #3695, although stopping short of adding alternative spellings & keywords, and in light of #3573, I'd like to propose that we somehow add aliases to icons for (sub-)brands that share the same icon _**and**_ colour (see #3270 for brands that share an icon but _not_ a colour). Only **_official_** names should be considered covering the following points:

1. A brand that shares the same icon and colour as an existing icon (e.g., React & React Native)
2. Alternative, localised brand names. Some companies in Ireland have both an English and an Irish name, for example. (Note: this would _not_ include translating non-English names into English or vice-versa _unless_ it's something the company themselves do)
3. Expanding acronyms & abbreviations (e.g., National Basketball Association for NBA) as it's not inconceivable that someone might search for a brand using its full name
4. Adding acronyms or abbreviations but _only_ in cases where the company abbreviates the brand name itself and, somewhat more subjectively, is in common usage (e.g., NJPW for New Japan Pro-Wrestling)

I would not suggest that this data be added to `simple-icons.json` as I only envisage it being useful for the search feature of the website so it would only lead to unnecessarily bloated files in the package and possible confusion regarding the ability to use icons by calling them by their aliases - to do so would require all aliases, as well as titles, to be unique. I'm open to suggestion on exact implementation, though.

Thoughts from @simple-icons/maintainers and community members on whether this would be a useful addition to the project? Any suggestions for additions to the list above?

adamrusted:
I think the icons should always be found first by the primary version used by the brand themselves, regardless of whether that's the abbreviation or the full title. Having an array of "alisases" as a second level of searching could work quite well, if proper guidelines are applied, as @PeterShaggyNoble lays out above.

fbernhart:
I think it's a great idea! 👍🏼 

I would be in favour of adding the aliases to the `simple-icons.json`, if that's technically possible and doesn't have any major disadvantages. Would something like this work?

```
        {
            "title": ["NBA", "National Basketball Association", "some other name"],
            "hex": "253B73",
            "source": "https://nba.com/"
        },
```

The benefit of putting everything into the `simple-icons.json` file is in my opinion, that everything is "at one place". Which will as well help keeping it easy for outsiders to contribute to Simple Icons.

[...]

Issue #43, opened 2013-01-25
Spelling: codeacademy


michaelrbock:
It's actually spelled "codecademy" (http://www.codecademy.com/)

Thanks!

PR #4920, opened 2021-02-03
Remove get-by-title from NPM package API

ericcornelissen:
**Issue:** Relates to #2260 and #4797
**Alexa rank:** n/a

### Checklist

### Description

Based on the discussion in #2260, with the intent to allow for brands with the same name (read `"title"`) to be included in the project, this removes the ability to `simpleIcons.get()` an icon from the NPM package based on the title.

This does beg the question whether we should convert back to the old API:

```js

console.log(simpleIcons['simpleicons']); // Use the filename of the SVG

/*
{
    title: 'Simple Icons',
    slug: 'simpleicons',
    hex: '111111',
    source: 'https://simpleicons.org/',
    svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>',
    path: 'M12 12v-1.5c-2.484 ...'
}
*/
```

I would be okay with that, but I would also like to explore the possibility of making the `simpleIcons` object both [enumerable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) and [iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) and am unsure if these three things can exist at the same time.


PeterShaggyNoble:
Thought: to minimise the number of times a user may potentially need to check [the look-up table](https://github.com/simple-icons/simple-icons/pull/5002), I wonder if this could be the excuse we're looking for to expose `titleToSlug` in the package and, thus, to the website. It would allow users to continue using the `title` as much as they like. For example:

```
let slug=titleToSlug(title);
simpleIcons.get(slug);
```

ericcornelissen:
> Thought: to minimise the number of times a user may potentially need to check [the look-up table](https://github.com/simple-icons/simple-icons/pull/5002), I wonder if this could be the excuse we're looking for to expose `titleToSlug` in the package and, thus, to the website. It would allow users to continue using the `title` as much as they like. For example:
> 
> ```
> let slug=titleToSlug(title);
> simpleIcons.get(slug);
> ```

I'd personally not be a fan of that as it doesn't work 100% of the time, what does it even return for brands that have a custom slug and how does it know that?

I think the best we can do is improve the experience of the look-up table, e.g. with a website or a CLI script made available through NPM. But those seem more like above-and-beyond features to me.

PeterShaggyNoble:
> I'd personally not be a fan of that as it doesn't work 100% of the time

Yeah, it was just a though, trying to find _some_ excuse to add `titleToSlug` to the package so we don't have to keep it manually synced on the website! 😆

> what does it even return for brands that have a custom slug and how does it know that?

It would return the one without the custom slug which, of course, could lead to confusion if people don't get the icon they're expecting. At that stage, though, they could go to the lookup table. The thought was to try to reduce the amount of times people have to look up the slug if they're using a lot of icons but that's probably the lesser of 2 evils compared to them not getting the icon they're expecting.

> I think the best we can do is improve the experience of the look-up table, e.g. with a website

Displaying it somewhere on the new website would be useful, perhaps as part of [#43](https://github.com/simple-icons/simple-icons-website/issues/43) if it gets implemented.

ericcornelissen:
> > I'd personally not be a fan of that as it doesn't work 100% of the time
> 
> Yeah, it was just a though, trying to find _some_ excuse to add `titleToSlug` to the package so we don't have to keep it manually synced on the website! laughing

[The `titleToSlug` function](https://github.com/simple-icons/simple-icons/blob/4887643364e4ddda1638460bb1297df403addb05/scripts/utils.js#L17-L37) and  [the `normalizeSearchTerm` function](https://github.com/simple-icons/simple-icons-website/blob/45309e58a3a2075dc869262d1983482d976104d6/public/scripts/utils.js#L33-L46) are different anyway, so unless we split the overlapping parts into a third function here, exposing `titleToSlug` doesn't help the website I'm afraid.


> > what does it even return for brands that have a custom slug and how does it know that?
> 
> It would return the one without the custom slug which, of course, could lead to confusion if people don't get the icon they're expecting. At that stage, though, they could go to the lookup table. The thought was to try to reduce the amount of times people have to look up the slug if they're using a lot of icons but that's probably the lesser of 2 evils compared to them not getting the icon they're expecting.

Fair enough. I would definitely go crazy if that function gives me the wrong name and hence icon though, hence my confusion as to what should be returned. Your suggestion actually makes a lot of sense if you think about it, but I don't think it does as an end-user. That said, if others agree with you I'm happy to expose `titleToSlug` for this purpose :slightly_smiling_face: 


> > I think the best we can do is improve the experience of the look-up table, e.g. with a website
> 
> Displaying it somewhere on the new website would be useful, perhaps as part of [#43](https://github.com/simple-icons/simple-icons-website/issues/43) if it gets implemented.

Definitely :+1:

PeterShaggyNoble:
Couple of conflicts to be resolved in this one, @ericcornelissen.

scripts/build/package.js
```diff
@@ -34,11 +34,7 @@ function escape(value) {
   return value.replace(/(?<!\\)'/g, "\\'");
 }
 function iconToKeyValue(icon) {
-  let iconName = escape(icon.title);
-  if (icon.slug !== titleToSlug(icon.title)) {
-    iconName = icon.slug;
-  }
-
+  const iconName = icon.slug;
   return `'${iconName}':${iconToObject(icon)}`;
```
mondeja, line 38: Minor suggestion:
```suggestion
  return `'${icon.slug}':${iconToObject(icon)}`;
```

mondeja (approved):
The next release will be faster, LGTM!
I've a minor suggestion but not important at all, good work :+1:

PeterShaggyNoble:
Merging this in based on @mondeja's approval above. Nice work, @ericcornelissen 👍🏻

PR #4921, opened 2021-02-03
Disallow underscores in slugs and filenames(*)

ericcornelissen:
**Issue:** Relates to #2260 and #4918
**Alexa rank:** n/a
  

### Checklist

### Description

In simple terms, this prevents underscores (`_`) from showing up in slugs (that are automatically generated, see #4918). And renames any existing SVGs with an underscore accordingly, this currently includes:

- Pop!_OS (from `pop_os.svg` to `popos.svg`)

In technical terms, this updates the `titleToSlug` function to strip out any underscores. That way underscores are "reserved" for custom slugs, again see #4918.


ericcornelissen:
I made this PR non-breaking by using the change of #4918 to preserve the original slug of "Pop_OS!". I would still be in favor of changing this slug to `popos` (I think) but if we do that in a separate PR we can merge this change w.r.t. generated slugs now.

PeterShaggyNoble (approved):
This looks good to me, Eric 👍🏻

And I agree that we should change the slug for Pop_OS! in v5.

PR #4929, opened 2021-02-04
added protonvpn icon

xtlmnz:
protonvpn

**Issue:**
**Alexa rank:**
  

### Checklist


adamrusted (changes requested):
Thanks for your work on this @xtlmnz!
Looks like there's a few issues with the SVG itself, which is why it's not passing the linter. Firstly, could you look to remove the blank background from the SVG, and filling the frame as best as you can with the icon (whilst keeping the aspect ratio intact)? I've recreated the icon in Illustrator (green) and put it against your icon (black) and there's plenty more space that can be filled!
image

It also looks like you're missing the `role` and `title` attributes, that are mentioned in the [contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md), could you have another look over those, and make sure they're included in the SVG?

adamrusted:
Marked as abandoned, as the account associated with this PR has been deleted, so I'm unable to update the path. Will submit a new PR shortly.

Issue #1167, opened 2019-01-13
Request: (GNU) Savannah


MrWils:
**Name:** (GNU) Savannah
**Website:** https://savannah.gnu.org/
**Official resources for icon and color:**
- https://savannah.gnu.org/images/Savannah.theme/floating.png
- https://en.wikipedia.org/wiki/GNU_Savannah#/media/File:Gnu_meditate_levitate.png

**Note:**
I couldn't find anything in an another format so both images are PNG.
I am not sure what was meant with color but I doubt that I could even link it since I couldn't find anything else as a PNG.

ericcornelissen:
That brings up an interesting point because we already [provide GNU logos (namely _GNU_ and _GNU social_)](http://simpleicons.org/?q=gnu) which I imagine also use the GPL license - although I could not quickly confirm this....

MrWils:
It is not a problem since:

- The GNU head is CC BY-SA 2.0 licensed. 
source: https://en.wikipedia.org/wiki/GNU#/media/File:Heckert_GNU_white.svg

- The GNU Social logo is CC0 licensed.
source: https://en.wikipedia.org/wiki/GNU_social#/media/File:GNU-social-logo.svg


The GNU Savannah would be a problem. 
It is GPLv3 licensed.

ericcornelissen:
Thanks for investigating this further @MrWils, am I correct that in the case of a CC BY-SA 2.0 license of the GNU head we should "give appropriate credit" as per https://creativecommons.org/licenses/by-sa/2.0/ ?

MrWils:
Yes, I think so.   
It also says that it should be under the same license.  
I am not sure if it has to be since a lot of things seem to use the GNU head.  

I found this [page](https://www.gnu.org/graphics/heckert_gnu.html) on the GNU site after a bit of searching.

> If you want to use the GNU head to link to a website run by the Free Software Foundation or the GNU project, feel free, or if you're using it in contexts talking about GNU in a supportive and accurate way, you can also do this without permission. For any other requests, please ask licensing@fsf.org for permission first.

I am pretty sure that they would be ok with this but I honestly would send them a mail to be 100% sure.

ericcornelissen:
> ShareAlike — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original. 

The above comes from the [CC-2.0](https://creativecommons.org/licenses/by-sa/2.0/) page, I would argue we don't "remix, transform, or build upon the material" 🤔

[...]

PR #4945, opened 2021-02-06
Add icon license to JSON linting (and change JSON linting dependency)

### LINKED ISSUE #1168: Icon licenses
Recent activity (in particular #1154 and #1167) has triggered a more in-depth discussion about licenses on icons and in general. In this issue I want to focus on icon licenses.

I think we should be more active in trying to find out under what license an icon is available when issues or pull requests are opened. This would probably be the easiest by updating the templates and contribution guidelines.

In general I think most licenses require at most "giving appropriate credit" (e.g. [CC BY-SA 2.0](https://creativecommons.org/licenses/by-sa/2.0/)). This would mean that we need to find an appropriate way to include credits in the SVG source material.

However, for example [GNU GPLv3](https://choosealicense.com/licenses/gpl-3.0/) restricts making the icon available without the license. Theoretically we could specify and provide the license text on [the website](http://simpleicons.org/), but I don't think this is feasible for [the NPM package](https://www.npmjs.com/package/simple-icons).

ericcornelissen:
**Issue:** closes #1168, relates to #1167 and #4797
**Alexa rank:** n/a
  

### Checklist

### TODO

- ~~Forgot to update the Contributing Guidelines, will do that later~~

### Description

This allows for a `"license"` field in the data file as per the discussion in #1168 and adds the license of two icons (based on  #1167). But it does a few more things :sweat_smile: If all the extra stuff turns out to be to complicated I can open a separate PR to add the `"license"` field so we can start using it sooner...

One thing about the addition of the `"license"` field, I explicitly added an enum list of [SPDX license ids](https://spdx.org/licenses/) that are allowed because I didn't manage to import a list remotely.  The downside is that the list may be(come) incomplete and requires manual updating - I don't think this is a big problem as that should be fairly uncommon and can easily be changed. The upside is that I was able to reasonable restrict the list to exclude e.g. company-specific licenses or licenses that just don't seem to be applicable to logos.

A list of the other stuff in this Pull Request:

- I updated the descriptions of each field in `.jsonlintschema` as they were poorly written before.
- I started using `"additionalProperties": false` so that icon/license properties that are not specified in the schema are reported as error. So for example, the following icon entry would be fine before but will now throw some errors (the comment text is not how the error is actually reported):
  ```js
  [
    ...
    {
      "tile": "Google",  // ERROR: "tile" is not allowed (as it should be "title")
      "hex": "4285F4",
      "source": "https://partnermarketinghub.withgoogle.com/",
      "contributor": "Eric Cornelissen" // ERROR: "contributor" is not a field we want to have
    }
    ...
  ]
  ```
- I want to reduce the duplication of what we consider valid URLs, but unfortunately our existing JSON linter ([jsonlint2](https://www.npmjs.com/package/jsonlint2) - or [jsonlint](https://www.npmjs.com/package/jsonlint) for that matter) doesn't seem to support that. Hence, I decided to switch to [jsonschema](https://www.npmjs.com/package/jsonschema) instead (*). JSON schema is used under the hood by jsonlint and [`.jsonlintschema`](https://github.com/simple-icons/simple-icons/blob/develop/.jsonlintschema) followed that format already). There is an existing [CLI package for jsonschema](https://www.npmjs.com/package/jsonschema-cli), but it's old and limited - crucially, it doesn't exit with a non-zero exit code if there is an error in the data file - so I created a custom script for our purposes.
- To support the above further, the `.jsonlintschema` file has been renamed to `.jsonschema.json` and internally restructured, but it still does the exact same thing (excluding the changes I explicitly mentioned before).

---

(*): this has the added benefit of being actively maintained.


.jsonschema.json
```diff
@@ -0,0 +1,156 @@
...
+              "additionalProperties": false
+            },
+            {
+              "type": "object",
+              "required": ["type"],
+              "properties": {
+                "type": {
+                  "description": "An SPDX License Identifier",
+                  "type": "string",
+                  "enum": [
```
mondeja, line 52: Since this implementation already uses a script to run the linting process, why not just download [this JSON file](https://github.com/spdx/license-list-data/blob/master/json/licenses.json) (extracted from [SPDX web repository](https://github.com/spdx/license-list-data)) and build the enum in-place?

ericcornelissen: I'm not totally against it but in my opinion this is still not a good solution. First, the fact that we have a script now does not mean that we should stick with it, and I'm not sure if being able to pull this list dynamically is a good reason. Secondly, at least the repository you linked to, does not interface well with NPM. Sure, we _could_ pull the git repo manually or even the file directly, but we have no guarantees on the structure of the repository, hence this (the JSON linter!) might break overnight just because they changed something (*).

What we could do instead is have some automation that pulls the list on a schedule and updates the list for is. Still not ideal, but at least the `npm run jsonlint` script never breaks - if the scheduled workflow breaks that doesn't affect anyone directly.


---


(*): I'm not so worried that the structure of the JSON file itself changes, but this is a concern as well and ideally we have a source that provides a JSON schema for the JSON data they're providing us. Speaking of which, we may consider providing our JSON schema alongside `_data/simple-icons.json` in the NPM package 🤔

PeterShaggyNoble:
Are we forcing, or do we want to force, the inclusion of the `url` entry if the `type` is "custom"?

ericcornelissen:
> Are we forcing, or do we want to force, the inclusion of the `url` entry if the `type` is "custom"?


With the current schema we are forcing that. My reasoning being that the value "custom" is meaningless on its own and one needs the URL to find out what the license actually is. Note that I would expect we just update the enum of licenses if we encounter a well-known license not currently in the list, hence we should never have to use "custom" for those cases.

PeterShaggyNoble:
> My reasoning being that the value "custom" is meaningless on its own and one needs the URL to find out what the license actually is.

Agreed. Just wanted to make sure.

In that case, this all looks good to me. But I'll let one of the other @simple-icons/maintainers who's more knowledgeable of the linting process and dependencies throw an eye over it before merging.

CONTRIBUTING.md
```diff
@@ -189,16 +189,23 @@ Icon metadata should be added to the `_data/simple-icons.json` file. Each icon i
   * A `hex` color value that matches the brand's primary color. All uppercase and without the `#` pound symbol.)
   * The `source` URL of the logo being used. There are [more details below](#source-guidelines).
 
-There is also an optional field that may be used to specify the brand guidelines/presskit/etc. This is useful if the SVG file was sourced from a different place.
+Additionally, there are also optional fields that may provided for an icon:
```
PeterShaggyNoble, line 192: I wonder should we create a new section for adding icon data an link to it from here. The new section would cover all the optional entries such as guidelines, license, the upcoming aliases plus any others we add in the future.

My reason for suggesting it is twofold:
1. We avoid overwhelming new contributors as we add more entries to the JSON - yes, when adding icons, the more data that is included the better but as long as the bare minimum of `title`, `hex`, `source` and, where applicable, `slug` are included then we're off to a good start, and,
2. The addition of all this new data for existing icons is going to be separate to the addition of new icons and I'm hoping that with a dedicated section in the guidelines it might encourage community members to start helping us out in adding all the data we need.

ericcornelissen: Good idea, I'll draft something of that nature :+1:

PeterShaggyNoble: Yeah, it can wait; it's only really going to become an issue when we add the aliases.

mondeja (approved):
All looks good to me :sparkles:

CONTRIBUTING.md
ericcornelissen: See https://github.com/simple-icons/simple-icons/pull/5068 :+1:

.jsonschema.json
ericcornelissen: Can't believe I didn't think of this earlier, but we should be pretty much fine with an approach like this if we use [this NPM package](https://www.npmjs.com/package/spdx-license-ids).

> First, the fact that we have a script now does not mean that we should stick with it, and I'm not sure if being able to pull this list dynamically is a good enough reason to stick with it.

Coming back to this, I guess I changed my mind and if it benefits us to have a custom CLI interface, than using a script for that is worthwhile, I think... Though it would be even more ideal if a potential CLI for [jsonschema](https://www.npmjs.com/package/jsonschema) supported `.js` config files so that the configuration can be constructed dynamically.

In that regard, do we want to change [`.jsonschema.json`](https://github.com/simple-icons/simple-icons/blob/4c9cb1102f36c7f0af544e6eb19686d680d5f39c/.jsonschema.json) to `.jsonschema.js` and handle the logic there with [the NPM package](https://www.npmjs.com/package/spdx-license-ids) there, or inject it's data in [the `jsonlint.js` script](https://github.com/simple-icons/simple-icons/blob/4c9cb1102f36c7f0af544e6eb19686d680d5f39c/scripts/jsonlint.js)?

ericcornelissen: @PeterShaggyNoble @mondeja

Issue #4885, opened 2021-01-28
Icon request: Processing


Galarzaa90:
**Brand Name:** Processing
**Website:** https://processing.org/
**Alexa rank:** 47,511
  

**Official resources for icon and color:** This is a bit complicated because they don't have a dedicated site for these, and they have one icon for their website and one icon for their software.

Website: https://processing.org/img/processing2-logo-256x256.jpg
Software: https://processing.org/img/processing3-logo.png

adamrusted:
Searching through their [GitHub Organisation]() brought these SVGs to light. I'd go with the icon variant, providing we can match it to the company, and not just their community day.


`https://github.com/processing/processing-community-day-website/blob/a59dab73f5ad401e33d8995f936e2d28e183e4d6/img/logo.svg`
![](https://raw.githubusercontent.com/processing/processing-community-day-website/a59dab73f5ad401e33d8995f936e2d28e183e4d6/img/logo.svg)

`https://github.com/processing/processing-pi-website/blob/8cdce6763250d529a73a75abd570872dd5883f2b/themes/processing/static/img/processing-logo.svg`
![](https://raw.githubusercontent.com/processing/processing-pi-website/8cdce6763250d529a73a75abd570872dd5883f2b/themes/processing/static/img/processing-logo.svg)

service-paradis:
I'm a little bit confused with this request. Should we close it according to #4963?

PR #4963, opened 2021-02-10
Add Processing

codyduong:
processing)

**Issue: #4885**
**Alexa rank: 45,186**
  

### Checklist

### Description
Used the approximate HEX value from this [image's](https://processingfoundation.org/content/1-home/processing-social.jpg) background, sourced from the Processing Foundation

It does seem that there are many varying logos from Processing, from their [favicon](https://user-images.githubusercontent.com/24193268/107449750-bf1a9480-6b09-11eb-8d9d-507bdaa6952f.png) on their website, versus their logo specifically for Processing language, which looks like [this](https://user-images.githubusercontent.com/24193268/107449318-f9cffd00-6b08-11eb-9a5b-252af2277e02.png) (At least while it is still on Processing 3). As a result of this ambiguity, I don't know whether to call this icon Processing or Processing Foundation.

Either way, I stuck to the original issue request and added that logo, and called it Processing.


PeterShaggyNoble:
Welcome to Simple Icons, @codyduong and thanks for taking this one on 👍🏻

Their branding, especially in relation to their icons does indeed seem to be bit all over the place. From what I can see, they have following, each with different branding:
- Processing Foundation
- Processing Software
- Different versions of that software
- Language specific implementations
- A community day

I think it's the second one we're looking for here so I agree with both the name chosen and the fact that it should be square.
- For [the foundation](https://processingfoundation.org/), I'd expect a circular version
- Icons for individual versions have already been highlighted here and in the original issue
- Language specific implementations aren't really relevant here and we can address them as and when they're ever requested

Which leaves the community day and, given the repo the source SVG was found in, it looks to me like it's the icon for that. That branding also seems to be outdated if you check https://day.processing.org/ before it redirects back to the Foundation's site.

So, the question is: which icon should we be using here to represent Processing itself? Personally, with just their favicon to go on, I'd expect a square with rounded corners and the "P" knocked out of it. But, given the inconsistent branding and my unfamiliarity with Processing, let's hear what others think before you change it.

With regard to the colour, I'd go with the `#1d517e` from the website's menu.

PeterShaggyNoble:
To add to the confusion [Wikipedia](https://en.wikipedia.org/wiki/Processing_(programming_language)) uses the icon for Processing 3 so perhaps _that_ is what we should be using, too?

And, on top of all that, there's a completely different & newer set of icons available [here](https://github.com/processing/processing-website/tree/master/src/images) which makes me wonder if they're not in the middle of a rebranding. Although that shouldn't stop us from adding the current icon; if they do change it then we'll update when the time comes.

adamrusted:
Hmm.. Which of the various squiggly lines do we think is the best icon 😂
I guess we either go for `logo-processing.svg` for the 'overall product' or `logo-p5js.svg` for the product 'P5'.

PeterShaggyNoble:
You mean [this P5](https://github.com/simple-icons/simple-icons/pull/3507), @adamrusted?!

adamrusted:
Oh yeah, that's a thing isn't it?
Given there's a completely different logo for P5.js within that folder then, we may need to update that alongside this one.

PeterShaggyNoble:
@codyduong or @Galarzaa90, if either of you are involved in the Processing community, would you be able to reach out to get their input as to which icon we should be using here?

PeterShaggyNoble:
Taking a fresh look at this and I'm still no clearer on what we should be using! I would still lean towards the square with the rounded corners but not for any conclusive enough reasons.

Let's ping @processing directly to see if anyone there can steer us in the right direction.

adamrusted:
Returning to this one, given lack of response from Processing themselves. There is now in fact an SVG logo on their homepage that we can use, which is the `P` in a circle, in a lighter shade of blue. Given that seems to be their new colour (see CTA buttons on homepage) I'd suggest we go with that variation. @codyduong, would you mind looking at updating your icon to suit?

codyduong:
Alright, I went ahead and made the changes:
image

PeterShaggyNoble (approved):
Thanks, @codyduong, everything looks good to me - nice work 👍🏻

I wonder, though, should we now go with "Processing Foundation" for the name to avoid any confusion or ambiguity? 🤔

PeterShaggyNoble:
Oop, turns out the GitHub bot is better at reviewing than me! 😆 That error is simply due to the missing `<title>` tag in the SVG.

codyduong:
@PeterShaggyNoble Got it.

adamrusted (commented):
This looks spot on to me! Thanks for your work on this @codyduong 🚀
Before we merge though, what are your thoughts on changing the title to 'Processing Foundation' as mentioned by @PeterShaggyNoble? Will need updating in the JSON and the filename of the SVG will need changing also.

codyduong:
@adamrusted Originally, I was of the opinion that it should maintain the Processing name, since I could see that the Processing Foundation might be used less. But on further thought, Processing Foundation could do the same thing, and also be technically correct. At a later point we can add the actual Processing icon (once we get through the ambiguity). Until then, people can use the Processing Foundation as the icon for a placeholder of Processing and Processing.py. I went ahead and pushed those changes.

PeterShaggyNoble:
OK, seeing as we're all in agreement, let's go ahead with this one as "Processing Foundation" and, when we figure out the icons for the rest, we can add them separately, using the original issue for tracking.

Thanks again for your work on this one, @codyduong 👍🏻

Issue #4578, opened 2021-01-02
Stellantis Review


adamrusted:
Following the merge of FCA and PSA to form Stellantis - we should try and review all relevant brands when an official media source becomes available. Brands are listed below:

- [x] ~Abarth~ - Not Popular Enough
- [x] Alfa Romeo - #4990
- [x] Chrysler - #4991
- [x] Citroen - #5005
- [x] Dodge - #5003
- [x] DS Automobiles - #5006
- [x] Fiat - #5008
- [x] Jeep - #5011
- [x] ~Lancia~ - Not Popular Enough
- [x] Maserati - #5009
- [x] Opel - #5012
- [x] Peugeot - #5013, #5119
- [x] Ram Trucks - #5014
- [x] Vauxhall - #5015

PeterShaggyNoble:
With the recent merger of [PSA](https://www.groupe-psa.com/) & [FCA](https://www.fcagroup.com/) to form [Stellantis](https://www.stellantis.com/), we should keep an eye out for a more up-to-date source for these.

adamrusted:
@PeterShaggyNoble, looks like Stellantis have made this CRAY easy for us, SVG files are included on the homepage, in the mega-menu. Will go through our existing entries, and see if they need updating.

PeterShaggyNoble:
Wow! Why can't everything be that easy?! 🙂

And there's the elusive Lancia logo we could never find - Sweet!

adamrusted:
I know right? There's an arguably better (larger) SVG on the per-brand page, so I'm linking to those in the ones I do.

gizmecano:
I think Peugeot have to be unchecked (cf. https://github.com/simple-icons/simple-icons/pull/5013#issuecomment-786513237)

PeterShaggyNoble:
Closing this out as we now have PRs for all these updates.

PR #4990, opened 2021-02-12
Updated Alfa Romeo

adamrusted:
**Alexa rank:** [181,640](https://www.alexa.com/siteinfo/alfaromeo.com)

### Description
Updated HEX to match Pantone found in Style Guide. Removed Stellantis version, as existing icon in package matches perfectly to EPS available at source.

~Updated source to Stellantis website, as they use monochrome SVGs on the brand pages. Updated colour to a hex found on the Alfa Romeo homepage.~


PeterShaggyNoble:
Looks like you forgot to push the changes to `simple-icons.json` plus there's a change to `.svgo.yml` that needs to be reverted.

adamrusted:
Must've reverted the wrong file. Bear with.

PeterShaggyNoble:
SVG looks spot-on to me, @adamrusted 👍🏻 Although, there is an imperfection at the bottom of the first circle in the snake's crown that's present in the original if you want to try cleaning that up before merging.

On the colour, although my original choice came from [official brand guidelines](http://www.fcaci.com/v15-images/guidelines/Alfa-Romeo-Guidelines-Rev201510.PDF), they are slightly dated now and, as they use your choice throughout their site, I think we're OK to change it here.

adamrusted:
@PeterShaggyNoble copied the middle circle to replace the first on the crown. Otherwise, SVG should be exactly the same 🤞🏻

adamrusted:
@PeterShaggyNoble - just re-doing this one from the FCACI portal, and only thing that's different is the Pantone reference for the hex. Have amended this PR accordingly.

PeterShaggyNoble:
If we're going with the colours from [the PDF](http://www.fcaci.com/v15-images/guidelines/Alfa-Romeo-Guidelines-Rev201510.PDF) rather than from their website then we should probably use the hex reference provided on page 28, which is actually the original colour we had.

adamrusted:
> If we're going with the colours from [the PDF](http://www.fcaci.com/v15-images/guidelines/Alfa-Romeo-Guidelines-Rev201510.PDF) rather than from their website then we should probably use the hex reference provided on page 28, which is actually the original colour we had.

Missed that one - always confuses me when people define a Pantone, and have a hex that is different to the one calculated [by Pantone](https://www.pantone.com/uk/en/color-finder/201-C)! Will revert now.

fbernhart:
I'm a bit confused.

So if the hex is still the same and the SVG is still the same. What is this PR still about? 😅 

Seems like we can just close it. Or am I missing something?

adamrusted:
> So if the hex is still the same and the SVG is still the same. What is this PR still about?

The SVG is different to the one already in Simple Icons - the comment above was talking about differences between my PR versions 🤣

fbernhart:
@adamrusted How is it different? 😄

Have you maybe pushed the wrong file?

adamrusted:
> @adamrusted How is it different? 😄
> 
> Have you maybe pushed the wrong file?

Never mind, I'm obviously confusing this and other PRs. Let's close this one for now, shall we? 🤣

PR #5002, opened 2021-02-13
Document slugs of brands

### LINKED ISSUE #4946: Document what the slug of a brand is
### Kind of issue 
  - [ ] Improvement
  - [ ] Bug
  - [x] Other, namely: Documentation

### This issue concerns 
  - [x] The NPM Package
  - [x] The Packagist Package
  - [x] The CDN usage

### Description

Currently it is pretty hard to find out what the slug (i.e. the SVG filename among other things) of a brand is given the title - and issues such as #2260, #4918, #4921, and #4936 won't make it any easier - so I think we should somehow document the slug for each brand.

Two things that come to my mind immediately:

1) Show the slug on [the website](https://simpleicons.org/). The main downside here is that website users are the only group that don't need to know the slug...
2) Generate a Markdown table with the mapping with every release. This takes some work to set up, but once it is automated it is done and we don't need to think about it any more. Also, users can use <kbd>Ctrl</kbd>+<kbd>f</kbd> to search for the brand they want.

ericcornelissen:
**Issue:** Closes #4946, relates to #4918
**Alexa rank:** n/a
  

### Checklist

### Description

This adds a script to generate a MarkDown table of brand names and their respective slugs, and references this in the README. 

~~The only thing left to do is figure out how to automatically update it. I think the best approach is it automatically update it with every release. Thoughts @simple-icons/maintainers ?~~

I updated [the `create-release.yml` workflow](https://github.com/simple-icons/simple-icons/blob/2a550a7dd7f30473132772b6fa841193cdd4a318/.github/workflows/create-release.yml) to update the slugs table automatically with every release. This required some changes w.r.t. to [the release action](https://github.com/simple-icons/release-action):

- **BEFORE**: the release action would create a version bump commit updating the version number of the package in `package.json`
- **NOW**: when this is merged, https://github.com/simple-icons/release-action/pull/35 **must** be merged and then then release action will ONLY create the release PR and [the `create-release.yml` workflow](https://github.com/simple-icons/simple-icons/blob/2a550a7dd7f30473132772b6fa841193cdd4a318/.github/workflows/create-release.yml) will create a commit with both a version bump and updates to the slugs table.


scripts/build-slugs-table.js
```diff
@@ -0,0 +1,33 @@
...
+# Simple Icons slugs
+
+| Brand name | Brand slug |
+| :--- | :--- |
+`;
+
+data.icons.forEach(icon => {
+  const brandName = icon.title;
+  const brandSlug = titleToSlug(icon.title);
+  content += `| \`${brandName}\` | \`${brandSlug}\` |\n`
```
service-paradis, line 30: Should we also display the icon there?
```
content += `| <img src="./icons/${brandSlug}.svg" alt="${brandName}" width=24 height=24> | \`${brandName}\` | \`${brandSlug}\` |\n`
```

service-paradis:
LGTM 💯 
It would be great to call this script when creating a new release! Would it be possible in the current creation workflow?

scripts/build-slugs-table.js
ericcornelissen: Interesting idea! Not sure if it is useful but if more people are in favor of this I don't mind adding it :slightly_smiling_face:

That is, provided that it doesn't make the table (significantly) bigger or more ugly.

ericcornelissen:
> LGTM 100
> It would be great to call this script when creating a new release! Would it be possible in the current creation workflow?

That would be ideal yes, it might not be too straightforward with our current set up but I think we can work something out. E.g., we might move the creation of [the version bump](https://github.com/simple-icons/simple-icons/pull/5020/commits/d0c949855d4f896dcdcc650c518b207300ec2453) commit from [the release action](https://github.com/simple-icons/release-action) to a local workflow that runs when a Pull Request targeting master is created.

scripts/build-slugs-table.js
PeterShaggyNoble: Probably one to come back to after GitHub come up with a solution for displaying different images for different themes.

service-paradis:
I'm not familiar enough yet with Github actions, but could we add a step in `create-release.yml` ?
Something like the following maybe:
```yml
jobs:
  slugs:
    name: Prepare Slugs documentation
    runs-on: ubuntu-latest
    steps:
      run: node ./scripts/build-slugs-table
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: simple-icons/release-action@master
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

ericcornelissen:
First, I prefer to keep the changes we are going to commit in one place so it is clear where the changes come from. Second, that doesn't work because the release-action doesn't work based on a cloned version of the repo (in fact, we don't clone the repo at all in that workflow so a step for that would need to be added.

What I would propose instead is something like:

```yml
jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - name: Create release PR
        uses: simple-icons/release-action@master # no longer creates a version bump commit
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
      - name: Checkout
        uses: actions/checkout@v2
      - name: Version bump
        run: node ./scripts/version-bump.js  # needs to be added
      - name: Update slugs table
        run: node ./scripts/build-slugs-table.js
      - name: Commit
        run: ??? # Probably use a third-party action
```

ericcornelissen:
Alright, I updated the release-action workflow for creating the release PR as per [my previous comment](https://github.com/simple-icons/simple-icons/pull/5002#issuecomment-779891088) and I added the output it expected in https://github.com/simple-icons/release-action/commit/dc0cc8e3f7e8fff78d6dc62d0054ad5d4a35d18c. If we decide to use this approach and merge this PR, we should also merge https://github.com/simple-icons/release-action/pull/35 to prevent the release action from creating a version bump commit.

ericcornelissen:
I updated the release-action workflow for creating the release PR a little more to make some further improvements - especially with regards to the edge case where no release PR is created because [there are no notable changes](https://github.com/simple-icons/release-action/blob/dad562c29d43d3e08f226d69cd83579f27d0e692/src/create.js#L612). Hence, I added another output value that tells you whether or not a release PR has been created, see https://github.com/simple-icons/release-action/commit/dad562c29d43d3e08f226d69cd83579f27d0e692, which I'm using here to check if a second job should be run. If it should, the version bump commit is prepared (by updating the version in the project manifest and updating the list of slugs) and created.

You can see [a test run of when no release PR is created](https://github.com/ericcornelissen/simple-icons/actions/runs/582306471) and [a test run of when a release PR was created](https://github.com/ericcornelissen/simple-icons/actions/runs/582359998)

ericcornelissen:
Alright, after some further tweaking I was able to get this working - I think. You can see a preview of this [in this run of the create-release.yml workflow](https://github.com/ericcornelissen/simple-icons/actions/runs/582495475) which resulted in [this Pull Request being created](https://github.com/ericcornelissen/simple-icons/pull/57) with [this version bump commit](https://github.com/ericcornelissen/simple-icons/pull/57/commits/8d7b98c9f083876bed4d43006ad8041a7493b376) :tada:

service-paradis (approved):
LGTM 💯 🎉

PeterShaggyNoble:
Should we also add something to `CONTRIBUTING.md` here about when and how to use slugs?

As part of that, I would suggest moving the "SVG Filename Convention" up above the "Update the JSON Data for SimpleIcons.org" section as:
1. It makes more sense within the workflow of creating an icon to tell people what to name their file after they've annotated & checked it, and,
2. They'll need to know what the filename is for the (potential) `slug` before adding the data to the JSON.

ericcornelissen:
> Should we also add something to `CONTRIBUTING.md` here about when and how to use slugs?
> 
> As part of that, I would suggest moving the "SVG Filename Convention" up above the "Update the JSON Data for SimpleIcons.org" section as:
> 
>     1. It makes more sense within the workflow of creating an icon to tell people what to name their file after they've annotated & checked it, and,
> 
>     2. They'll need to know what the filename is for the (potential) `slug` before adding the data to the JSON.

While I'm not against switching the order, it seems beyond the scope of this Pull Request - it currently doesn't even touch the Contributing Guidelines.

Now, when you say "how to use slugs" do you mean as a user as the package? If so, it shouldn't be in the Contributing Guidelines. If you mean how to use the `slug` property in the JSON data file, then it _could_ be in the contributing guidelines but I thought we decided against it so as to avoid newcomers using it while they shouldn't, and instead only use it when upon request from the maintainer.

PeterShaggyNoble:
> Now, when you say "how to use slugs" do you mean as a user as the package?

No, I meant telling contributors when a `slug` should be added and how to construct it.

> I thought we decided against it so as to avoid newcomers using it while they shouldn't

Was that not just in the context of making exceptions for filenames (which we're not 100% decide on yet)? I agree in that case we shouldn't include it in the contributing guidelines but I do think we need to at least cover how to resolve conflicts when 2 brands share the same name through the use of the `[brand]_[modifier]` slug.

ericcornelissen:
> Was that not just in the context of making exceptions for filenames (which we're not 100% decide on yet)? I agree in that case we shouldn't include it in the contributing guidelines but I do think we need to at least cover how to resolve conflicts when 2 brands share the same name through the use of the `[brand]_[modifier]` slug.

That sounds reasonable and I would support adding it. Though again, this seems out of scope for this PR to me. I can add it when I find time but it will either hold this PR back as it is, or that improvement to the contributing guidelines if this PR needs more work...

PeterShaggyNoble:
> this seems out of scope for this PR to me

Fair enough 👍🏻 I'll to get a PR together over the weekend if I've time and you don't beat me to it.

README.md
```diff
@@ -166,3 +169,5 @@ Icons are also available as a [Vue package](https://github.com/mainvest/vue-simp
 ### WordPress
 
 Icons are also available as a [WordPress plugin](https://wordpress.org/plugins/simple-icons/) created by  [@tjtaylo](https://github.com/tjtaylo).
+
+[icon slug]: ./slugs.md
```
mondeja, line 173: This link will be broken in the NPM page, should be absolute, right?

```suggestion
[icon slug]: https://github.com/simple-icons/simple-icons/blob/develop/slugs.md
```

ericcornelissen: From my experience NPM converts these kinds of URLs correctly. I propose we try it like this (for the reasons below) and change it in the next release if indeed it doesn't work on NPM (or Packagist) - in which case I think it should point to `master`, not `develop`.

For the record, I prefer the local URL as 1) it doesn't need to be changed if the base URL or remote changes and 2) it works locally.

mondeja: Oh, I didn't know this. It seems that [Packagist also converts relative links to absolute ones](https://github.com/composer/packagist/blob/main/src/Package/Updater.php#L691), so I'm fine with this :+1:

mondeja (approved):

README.md
```diff
@@ -70,11 +70,11 @@ console.log(icon);
 */
 ```
 
-Alternatively you can import the needed icons individually.
+Alternatively you can import the needed icons individually, where `[ICON NAME]` is replaced by the [icon slug].
```
PeterShaggyNoble, line 73: Given that we have now moved away from passing the icon's title to `simpleIcons.get()` in favour of the slug, should we not update the example code to use `[ICON SLUG]` instead of `[ICON NAME]` allowing us to simplify these lines and avoid any potential confusion?

ericcornelissen: I'd be in favour of that, though I still think we need some text to link to the table :+1:

PeterShaggyNoble: Oh, definitely 👍🏻

README.md
```diff
@@ -24,11 +24,11 @@ Icons can be downloaded as SVGs directly from [our website](https://simpleicons.
...
 
 ```html
-<img height="32" width="32" src="https://cdn.jsdelivr.net/npm/simple-icons@v4/icons/[ICON NAME].svg" />
-<img height="32" width="32" src="https://unpkg.com/simple-icons@v4/icons/[ICON NAME].svg" />
+<img height="32" width="32" src="https://cdn.jsdelivr.net/npm/simple-icons@v4/icons/[ICON SLUG].svg" />
+<img height="32" width="32" src="https://unpkg.com/simple-icons@v4/icons/[ICON SLUG].svg" />
 ```
 
-Where `[ICON NAME]` is replaced by the [icon slug], for example:
+Where `[ICON SLUG]` is replaced by the [icon slug] of the icon you want to use, for example:
```
PeterShaggyNoble, line 31: I think, in all cases, "[icon slug]" can now be replaced by simply "[slug]":
```suggestion
Where `[ICON SLUG]` is replaced by the [slug] of the icon you want to use, for example:
```

ericcornelissen: Good one, updated :+1:

PeterShaggyNoble:
This looks good to me now, thanks, Eric 👍🏻

PR #5003, opened 2021-02-13
Added Dodge

adamrusted:
dodge

**Alexa rank:** [26,400](https://www.alexa.com/siteinfo/dodge.com)

### Checklist

### Description
Use of the wordmark on it's own is permitted by Dodge (see [this PDF](http://www.fcaci.com/x/v10_images/guidelines/Dodge_Brand_Mark_Guidelines.pdf)), so figured it's inclusion made sense until we can look at incorporating the full duotone icon (hopefully).

SVG used found on Stellantis website. Think we can get away with just the stripes, as it's used in the `Dodge` favicon. Hex pulled from Dodge stylesheet.

#### Previous Treatment
dodge


PeterShaggyNoble:
From page 3 of [this PDF](http://www.fcaci.com/x/v10_images/guidelines/Dodge_Brand_Mark_Guidelines.pdf) it looks to me like we need to include the wordmark. However, doing so would necessitate cutting the letters out of the lines which would also seem not to be permitted by those guidelines and I think the fact that this is the _only_ icon on the Stellantis website to be duotone probably backs that up.

adamrusted:
> From page 3 of [this PDF](http://www.fcaci.com/x/v10_images/guidelines/Dodge_Brand_Mark_Guidelines.pdf) it looks to me like we need to include the wordmark. However, doing so would necessitate cutting the letters out of the lines which would also seem not to be permitted by those guidelines and I think the fact that this is the _only_ icon on the Stellantis website to be duotone probably backs that up.

Yeah I took my lead from the favicon, which is just the stripes. Happy to skip this one for now if it requires a duotone treatment.

adamrusted:
@PeterShaggyNoble - updated to use the FCACI source.

PeterShaggyNoble:
Looking over their guidelines again, I don't think we're going to be able to add this one without violating them. The fact that their parent company can't even make an exception for their own brand is pretty damning too.

PR #5008, opened 2021-02-13
Updated Fiat

adamrusted:
fiat

**Alexa rank:** [78,446](https://www.alexa.com/siteinfo/fiat.com)

### Checklist

### Description
SVG pulled from http://www.fcaci.com/x/v10_images/fiat/FIAT_500_LOCKUP_VERT_BW_POS.zip.
Hex updated to one mentioned on pg14 of the brand guidelines, available at source.

~Stellantis and Fiat both use the car badge, with Stellantis providing a monochrome treatment, which is a *mess* of points. Tried cleaning up as best I could, but needed a precision of 4 to get it close.~

#### Previous Treatment
fiat


PeterShaggyNoble:
Could you take another look at this one, @adamrusted? It looks like the clean-up introduced a number of differences when compared with [this file](https://www.stellantis.com/content/dam/stellantis-corporate/assets/brand-logo-negative/brand-logos-same-width-white/brand-fiat-SW-white-07.svg). Some examples below, with your version in red. Alternatively, given our icon size, we may be better off going with the monochrome treatment found [here](http://www.fcaci.com/x/FIATv15).

adamrusted:
Yeah, figured the simplification would've done that. I'll look at the simpler one on the site you linked, and see how well that translates.

adamrusted:
@PeterShaggyNoble - updated to the FCACI version of the brand guidelines.

fbernhart:
@adamrusted The SVG looks spot on now! 👍🏼

But concerning the color: I think it's better to use the colors on [page 15](http://www.fcaci.com/v15-images/guidelines/FIAT-Brand-Mark-Guidelines-Rev051815.pdf) instead, which are already HEX values. Instead of converting Pantone/RGB/CMYK (page 14) to HEX. 

So I'd suggest updating the color to `#941711`.

adamrusted:
@fbernhart should be good to go 🤞

PR #5011, opened 2021-02-13
Updated Jeep

adamrusted:
jeep

**Alexa rank:** [9,857](https://www.alexa.com/siteinfo/jeep.com)

### Checklist

### Description
Updated source to FCACI now it's been updated with Stellantis branding.
Used SVG embedded in pg11 of the PDF, ® icon seems to be pretty clear on that one.

Given Jeep use the ® symbol _EVERYWHERE_, I figured we should probably look to include it ourselves. Happy to maintain existing icon and just update source if others prefer that.

jeep


PeterShaggyNoble:
With regard to the "®", you are, unfortunately, right about its inclusion; I'd missed the following paragraph on page 3 of [this PDF](http://www.fcaci.com/v15-images/jeep/Jeep-Brand-Mark-Guidelines-Rev10.15.pdf) when I originally added it:
> The Jeep brand mark consists of the Jeep logotype and the registration mark ® as one unit. The ® is an integral part of the Jeep brand mark design and may not be altered or deleted. The trademark registration symbol must appear with the Jeep brand mark in every application.

Looks like there are a few excess points that can be cleaned up here:
- Bottom of the inner curve of the "J"
- Top left of the "J"
- Top of the 2 "e"s
- In the "®"

And, as a follow-up to that last one, I wonder if we shouldn't create and supply our own versions of "®" & "™" for contributors to use in their SVGs as it's surprising the number of source files we come across where they are riddled with imperfections and/or excess points?

adamrusted:
> Looks like there are a few excess points that can be cleaned up here:

Will pick this up tomorrow - I bashed these out pretty quickly yesterday - so figured there'd be a fair bit to tidy up.

> I wonder if we shouldn't create and supply our own versions of "®" & "™" for contributors to use

I would've imagined there _are_ default versions provided by the relevant trademark authorities, though of course there's likely to be optimizations we can make on our end. May be worth finding the 'best one' from our existing icons, and look at applying them across the board?

adamrusted:
@PeterShaggyNoble give that one a go - rebuilt from FCACI using a decent SVG found in the PDF.

fbernhart:
LGTM 👍🏼  Let me merge it.

PR #5014, opened 2021-02-13
Added Ram Trucks

adamrusted:
ram

**Alexa rank:** [19,995](https://www.alexa.com/siteinfo/ramtrucks.com)

### Checklist

### Description
Added as part of the Stellantis review, given the Alexa rank is high enough for inclusion.
~Alternative name is just `RAM`, though Stellantis and Ram trucks seem to go by the full name in most places.~

#### Previous Treatment
ramtrucks


PeterShaggyNoble:
Looks like we may need to include the wordmark here, too: http://www.fcaci.com/x/v10_images/ram/Ram_Brand_Mark_Guidelines.pdf

For the name, I think we can just go with "Ram" as they only use the full name in a few places on their site and usually as a noun rather than part of their name. [Wikipedia](https://en.wikipedia.org/wiki/Ram_Trucks) claims it should be stylised as "RAM" but I'm not seeing that treatment anywhere on Ram's own site, nor Stellantis'.

adamrusted:
@PeterShaggyNoble - should be updated.

PeterShaggyNoble:
SVG looks spot on to me when compared to [this EPS](http://www.fcaci.com/x/v10_images/ram/RAM_Blk_Vert.zip). However, could you see if you could clean up any of the excess & duplicated points before we merge it?

adamrusted:
@PeterShaggyNoble - that should be a bit better. Reduced down as many points as possible without ruining the quality of the image.

PeterShaggyNoble:
🤔 Sorry, @adamrusted, I think we may need to bring back some of the points in the shield as removing them has introduced some minor discrepancies in [the bottom](https://user-images.githubusercontent.com/15157491/109041321-cbd0e800-76c6-11eb-9e86-d0cdb1d6c478.png) (yours in red) and on the inside of the vertical lines of the outer shape.

adamrusted:
@PeterShaggyNoble that should be a bit better. Not been _quite_ as ruthless with the excess points this time round 😅

PeterShaggyNoble (approved):
Looks perfect to me now, @adamrusted; nice work 👍🏻

PR #5039, opened 2021-02-16
Add Simkl

### LINKED ISSUE #5034: Request: Simkl icon
**Brand Name:** Simkl
**Website:** https://simkl.com
**Alexa rank:** [#63903](https://www.alexa.com/siteinfo/simkl.com)

https://us.simkl.in/img_favicon/v2/safari-pinned-tab.svg

alagos:
simkl

**Issue:** Closes #5034
**Alexa rank:** [#61675](https://www.alexa.com/siteinfo/simkl.com)

### Checklist

### Description
Just took the base svg file from the source code in https://simkl.com, specifically [this file](https://us.simkl.in/img_favicon/v2/safari-pinned-tab.svg), then I modified with inkscape removing all the unrequired nodes.


service-paradis:
Thank you @alagos for your contribution!
They have an official media kit linked in the footer of their website:
https://drive.google.com/drive/folders/0B-54ieAFggpUMmY5a195alFxSVE

We can found SVG version of their logo. I think, we should keep the square behind the `S`. It look to be part of the logo.

alagos:
Ohh, didn't notice that @service-paradis , good catch.
About the SVGs in the media kit, there's a squared version, rounded square and with no square at all (that would be pretty much my version), then in the PNGs I don't see any of the small logos using the squared version, only the rounded one.
I don't have problems to change it, but just want a confirmation, as to me looks that the rounded square version is more used.

PeterShaggyNoble:
As they themselves provide an SVG with no background, I think we're OK to go without it too but, then, I am always be biased towards the most simplistic interpretation of an icon. However, if it were to be included, I'd vote for the one with the rounded corners as that looks to be the most consistent usage; they have a directory within their media kit consisting solely of PNGs with rounded corners.

alagos:
Ok, second chance and indeed this version seems to look better

Let me know guys your thoughts on this

PeterShaggyNoble:
Thanks for the update, @alagos. However, I am seeing [some differences](https://user-images.githubusercontent.com/15157491/109034107-367e2580-76bf-11eb-9612-863ff8bbe8c9.png) when comparing your version (in red) to the `logo_square_rounded_S_white_on_black_bg.svg` file found [here](https://drive.google.com/drive/folders/1DIfMlC0CHBSI2p6bPwnbOFRzZ_mDapEZ). Did you base your icon on the same file or a different one?

I'd also like to get some feedback from the other @simple-icons/maintainers to break the deadlock on the choice of icon before we proceed with further updates or reviews.

adamrusted:
I'd be happy with either version - though agree that if the majority of icons presented in the Media Kit are in the square with rounded corners, we should use that as the most consistent representation of the brand.

Assuming other @simple-icons/maintainers are of a similar opinion, @alagos would you mind looking at this again over the next few days? Feel free to ping me here if you're unsure of anything.

fbernhart:
> though agree that if the majority of icons presented in the Media Kit are in the square with rounded corners, we should use that as the most consistent representation of the brand.

Agreeing

PeterShaggyNoble:
OK, looks like we have a consensus 🙂

@alagos, did you base your version on the same source I compared with? If you didn't, could you let us know which file you did use so we can check again and if you did then could you look into fixing the differences seen in the comparison?

alagos:
@PeterShaggyNoble hey sorry, I've been busy these days.
I based on [this file](https://us.simkl.in/img_favicon/v2/safari-pinned-tab.svg) instead of the files from the media kit.
Why? because the svg in the media kit are two objects, one is the black rounded square and on top of that is the white `S`, whereas the other file are both the square and the `S` a single object, so it was easier for me to edit it as I had no idea how to merge the two objects and make the `S` transparent without displaying the black square background.
I'll make the changes, but I had no experience with Inkspace up to this moment I started to tweak the icon, so bear with me as I'll probably have to check one or two Inkspace tutorials.

PeterShaggyNoble:
Thanks for the clarification, @alagos 👍🏻 We will generally always prefer vector files provided by a company explicitly for third party use over any they may use themselves as, for many different reasons, there can often be differences between the two and internal brand guidelines can sometimes differ from those made available to the public. So, for that reason, we should go with the ones provided in Google Drive.

With regard to cutting the "S" out of the square: you can do this in Inkscape by selecting them both and using the "Difference" tool ("Path -> Difference" in the menu or `Crtl + -`).

PeterShaggyNoble:
Are you still working on this one, @alagos?

alagos:
@PeterShaggyNoble sorry for the long delay. I followed your "difference" advise with the other svg and this is the result.
simkl
Let me know if it's something else to do, btw how do you do that image with the diffs in red?

PeterShaggyNoble:
Thanks, @alagos, the path looks perfect to me now 👍🏻

Just need to remove the `defs` element & its children and the `fill` attribute on the `path` element in order to pass the linter and this should be ready to merge.

alagos:
@PeterShaggyNoble oops, didn't notice the gradient came in the original svg. 
All linters are passing this time and last revision looks the same as the previous one.
simkl

kweis52 (approved):

PeterShaggyNoble (approved):
Sorry for the delay, @alagos. This looks perfect to me now, nice work 👍🏻

Issue #4936, opened 2021-02-04
On Removing Hyphens & Symbols from File Names


PeterShaggyNoble:
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### This issue concerns 
  - [x] The NPM Package
  - [x] The Packagist Package

### Description
This issue is a follow-up to @ericcornelissen's [comment](https://github.com/simple-icons/simple-icons/issues/2260#issuecomment-772689951) below:
> which, actually, raises the question of why hyphens are currently also kept when converting from [`titleToFilename`](https://github.com/simple-icons/simple-icons/blob/ac633dc120e4aa28d181f6d02bda2e417110b0be/scripts/utils.js#L6-L26)

With the addition of `slug`s in #4918 we could now, conceivably, remove `-`s from file names, freeing up their use for another purpose in the future*. Up until now they have served a purpose in, potentially, helping avoid potential name clashes - consider, for example, the completely hypothetical brands "S&Y" and "Sandy - but, with the `slug`s, that's now redundant. So _should_ we remove them? It wouldn't be a _huge_  job; by my count there are 23 icons in the current release with a `-` in their file name so I say we may as well.

If we do decide to do it, it could be rolled into #4921

(* One possible use case I'm thinking of is if we decide to generate SVGs with localised `title`s for some purpose, we can use the `-` to append the locale to the filename, e.g. `brandname-en-ie.svg`.)

ericcornelissen:
Thanks for converting this into an issue @PeterShaggyNoble :+1: 

Just to clarify the issue a bit. Do we want to remove all  `-`s from all automatically created slugs, including [those like `-dot-` and `-and-`](https://github.com/simple-icons/simple-icons/blob/e8ba23c0e3a888878cce72741caf817ef279ba55/scripts/utils.js#L9-L14), or just those that appear in the brand name? Here's my take:

Removing hyphens entirely is the best & simplest approach, firstly because automatically created slugs just don't have hyphens which is simpler to reason about. However, we do lose the "elegance" of the automatically inserted `-dot-` and `-and-`. Which is a shame, but is easily counteracted by the introduction of custom slug in #4918 (admittedly the PR needs some changes to support this, but I would be in favor of this), which would not only allow us to keep this format for icons that currently use it but also allows us to omit them in new additions if need-be - the only downside being that reviewers need to keep in mind to check whether or not it makes sense to keep a symbol in the slug or not.


[...]

PR #5047, opened 2021-02-17
Rename Delicious as del.icio.us

### LINKED ISSUE #4791: Rename or Remove Delicious
Following on from the discussion in #4499, [the del.icio.us website](http://del.icio.us/) was actually shut down in 2019 but, as of December, there is now a holding page up stating that it should be back online this year without providing a firm date. Therefore, we should check whether it's back online yet prior to the `v5` release and consider it for removal if not.

If we are able to keep it then we should instead rename it as "del.icio.us", consistent with how it's stylised pretty much everywhere.

PeterShaggyNoble:
**Issue:** Fixes #4791
**Alexa rank:** [~78.5k](https://www.alexa.com/siteinfo/icio.us) (I'm assuming `del` is the only, or at least primary, subdomain)

### Checklist

### Description
Updates our existing "Delicious" icon to use the stylised "del.icio.us" name. As the new site is in an archived state, we should check the Alexa rank again before merging this.


ericcornelissen:
I guess this goes against what I said in #4936, but I think that when #4918 is merged the (custom) slug should be set to "delicious" (as you don't, presumably, call it "del-dot-icio-dot-us").

<sup>or we do go with removing `-` and `_` entirely and instead use custom slugs for e.g. ".Net"</sup>

PeterShaggyNoble:
After reviewing a few other slugs for #5611, I am now of the opinion that would should, in rare cases, make allowances when the `.`s in a brand's name are not pronounced so I've updated this PR accordingly. This is probably no longer a breaking change now as it's just the `title`s in the JSON & SVG that are being changed, with the addition of a `slug` to allow the filename to remain unchanged.

ericcornelissen (approved):
Can confirm this would not be breaking :+1: 

I would be in favour of this approach. As it concerns a custom slug, I will leave it for at least one more @simple-icons/maintainers to approve this.

adamrusted (approved):
LGTM!

adamrusted:
> Separately, we may want to once again consider removing this as their website has been down for a few days as their SSL cert expired on the 4th.

If you're willing to visit without the SSL, the site's still there - so I think we're okay to keep it for now. Good to keep an eye on though. Were we waiting for v5 on this one, or is it ready for merge now?

ericcornelissen:
> Added "Delicious" as an alias for people searching for the un-stylised version. I added it as an aka rather than an old as they use it themselves in a few places.

Good call :+1: 


> > Separately, we may want to once again consider removing this as their website has been down for a few days as their SSL cert 
> > expired on the 4th.
> 
> If you're willing to visit without the SSL, the site's still there - so I think we're okay to keep it for now. Good to keep an eye on though. Were we waiting for v5 on this one, or is it ready for merge now?

I agree that an (recently) expired SSL certificate is not a reason for us to remove the icon.

This is good to merge, but I'll leave some time for @PeterShaggyNoble to respond to this discussion on removing it or not.

PeterShaggyNoble:
Yup, agreed. If it's still down come the week of v5, we can discuss it then.

Merging this in, so.

PR #5048, opened 2021-02-17
Update InVision

PeterShaggyNoble:
InVision

**Issue:** n/a
**Alexa rank:** [~4.2k](https://www.alexa.com/siteinfo/invisionapp.com)

### Checklist

### Description
- Their icon now has square corners not rounded.
- SVG from `\InVision\icon\svg\InVision-icon-pink.svg` in [this zip](https://marketing.invisionapp-cdn.com/www-assets.invisionapp.com/InVision-Media-Kit.zip) which can be found by right clicking on any of the logos or icons on their site.
- I left the colour as it is rather than using the `#ff0066` in the source icon as it's still used throughout the site, or through the source URL.
- Updated the source to their news page and added the media kit that links to as `guidelines`.


service-paradis (commented):
LGTM when compared to the icon found in their zip file 💯 

Looks like they dont use this logo on their website though, which is a little bit weird. When compared to the SVG found in their website footer, it gives this:
image

PeterShaggyNoble:
@service-paradis, as with #5045, we should go with the version provided for 3rd party use im their media kit rather than any version they might use themselves.

service-paradis:
> @service-paradis, as with #5045, we should go with the version provided for 3rd party use im their media kit rather than any version they might use themselves.

Sure, just wanted to note the difference :)
I will merge this one!

PR #5054, opened 2021-02-18
Add Passport (PassportJS)

aadelgrossi:
passport preview

**Issue:** n/a
**Github stars:** 18.5k [passport](https://github.com/jaredhanson/passport)

### Checklist

### Description
Hex color is one of the primary colors in the logo. I have joined the colored sections into one creating a single path, since the icon must be monochromatic.


PeterShaggyNoble:
Welcome to Simple Icons, @aadelgrossi and thanks for the contribution.

Often, with icons of this nature, we'll add cut-outs between the different coloured sections (see below) to maintain some sort of differentiation between them but I think, in this case, your approach also works just as well. For the colour, there doesn't look to be a clear cut choice as they use all 3 on their site but I would lean towards the green simply because that's the one used on their homepage.

Going to ping @jaredhanson for feedback here to see if they have an existing or preferred monochrome treatment of the icon, or even allow a monochrome version at all, and also to provide input on the colour choice.

**Alexa rank:** [~72.7k](https://www.alexa.com/siteinfo/passportjs.org)

aadelgrossi:
> Welcome to Simple Icons, @aadelgrossi and thanks for the contribution.
> 
> Often, with icons of this nature, we'll add cut-outs between the different coloured sections (see below) to maintain some sort of differentiation between them but I think, in this case, your approach also works just as well. For the colour, there doesn't look to be a clear cut choice as they use all 3 on their site but I would lean towards the green simply because that's the one used on their homepage.
> 
> Going to ping @jaredhanson for feedback here to see if they have an existing or preferred monochrome treatment of the icon, or even allow a monochrome version at all, and also to provide input on the colour choice.
> 
> **Alexa rank:** [~72.7k](https://www.alexa.com/siteinfo/passportjs.org)
> 
> 

You're right, the green may suit the icon better since it ends up more pervasive on other sections of the home page, the remaining colors are sprinkled throughout additional pages but the green definitely stands out.

As for the cut between the colors, that was my initial approach in an attempt to differentiate colored sections, but when I reduced it to the 24 24 viewbox the cut becomes so minimal it's barely visible. Yours is a little more evident since the gap is a bit wider than what I had done, but it's still quite tiny when applied on the smallest size. 

Thanks for the feedback and suggestions :clap:

PeterShaggyNoble:
We try to stick to a `0.5dp` cut-out if we do need one as anything larger than that and we begin to lose too much of the icon and anything smaller is barely visible at our native size - my version above was just a very quick mock-up.

adamrusted:
@PeterShaggyNoble I've applied your hex and styling to the icon. Could you or @fbernhart review this one so we can get it merged?

PeterShaggyNoble:
Thanks, @adamrusted 🙂 Looks like we're having the same problem here as we had with the Namebase icon, though (e.g.) but they might be fixable just by dragging the points across to meet the curve of the original. There are also a couple of duplicate points at the top of the "L" shape and very top of the icon.

No mention of the logo or artwork in their `LICENSE.md` so I think, to be safe, we should omit it here.

adamrusted:
Weird - no clue what's going on with those paths again. Should be fixed now though.

jaredhanson:
Just chiming in here to say the icon looks great, and thanks for including it!

PeterShaggyNoble:
Thanks for the seal of approval, @jaredhanson 🙂

Given that, I'm happy to merge this in as-is, if you are @adamrusted. Although, I'm still seeing some minor differences, most notably down the left edge of the inner part of the icon and in the vertical line at the bottom. Gave it a try myself and came up with this (unoptimised) path if you want to run with it instead:

```
M 11.875,0 C 6.6714098,0.06703894 2.4670256,4.2709337 2.4,9.4746094 H 7.2 C 7.2667969,6.9266223 9.3273115,4.866872 11.875,4.8 Z M 12.125,0 V 4.8 C 14.672987,4.8667969 16.733067,6.9269829 16.8,9.4746094 H 21.6 C 21.532961,4.2710192 17.328676,0.06702552 12.125,0 Z M 2.4003906,9.7246094 V 24 H 12 V 19.199219 H 7.1992188 V 9.7246094 Z M 12,19.199219 C 17.261827,19.199219 21.532433,14.970328 21.6,9.7246094 H 16.8 C 16.732089,12.315131 14.606809,14.400391 12,14.400391Z
```

adamrusted:
Thanks for the updated path @PeterShaggyNoble - I've optimized it down to 3 decimals and pushed to this PR. Could another @simple-icons/maintainers take a look over this before we merge it?

service-paradis (approved):
LGTM 💯 
Thanks everyone!

PR #5061, opened 2021-02-18
Add proper error messages to linter; fix bug in 'collinear-segments' rule

mondeja:
## Summary

This pull request solves two problems:

- Fixes a bug in `collinear-segments` lint rule. My bad, in the current implementation, a lot of hidden positives are being ignored because the function `unshort` of svgpath changes the number of segments. This adds ~~23~~ 22 more icons for the 2021 review.
- Adds proper error messages for `icon-precision`, `ineffective-segments` and `collinear-segments` rules of the linter.

## Detailed explanation

In this pull request, I'm replacing the SVG parsing code of the linter, currently handled by [svgpath](https://github.com/fontello/svgpath/blob/master/lib/path_parse.js), with [svg-path-segments](https://github.com/mondeja/svg-path-segments), a library that I've created to deal with the lack that svgpath has: failed segments can't be recreated to show the user where the problem is. Note that the maintainer of svgpath [does not plan to include this improvement](https://github.com/fontello/svgpath/issues/49). So, after this, all failed segments are properly shown in error messages, as well as its absolute indexes inside the file. This will help a lot debugging some strange linter errors.

Taking advantage of this new implementation, parsing performance has been improved about ~30% because svg-path-segments doesn't make some checks that svgpath parser does, but these are not needed in the context of simple-icons as, if there is an error on the drawing, this should be noted in the review. Anyway, the result of both implementations parsing the icons of simple-icons is the same, as is tested by [this file (part of svg-path-segments test suite)](https://github.com/mondeja/svg-path-segments/blob/master/tests/si-integration.test.js). There are some more differences between svgpath and svg-path-segments parsers, if you want to check which are the exact differences [check this table](https://github.com/mondeja/svg-path-segments#comparison-against-other-implementations).

~~Regarding to the implementation itself, the most weird thing is a function named `segmentsSVGPath` that wraps the SVG parsing logic to use `svg-path-segments` building the `segments` array of the `svgpath` instance.~~ The rest of the code is almost self explanatory, but surely you may find [`svg-path-segments` reference](https://github.com/mondeja/svg-path-segments#reference) useful reviewing it.

---

### TODO after merge

- [ ] Update #4590 list.
- [ ] Update "Solving common linter problems" guide in discussions.


.svglintrc.js
```diff
@@ -33,6 +34,19 @@ function sortObjectByValue(obj) {
     .reduce((r, k) => Object.assign(r, { [k]: obj[k] }), {});
 }
 
+/**
+ * Parses a SVG path using svg-path-segments library and building a svgpath
+ *   instance with the result, returning a convenient object with two properties:
+ *   - `segments`: array of path segments as are returned by svg-path-segments.
+ *   - `svgPath`: instance of svgpath object that can be manipulated easily.
+ **/
```
ericcornelissen, line 42: I'm unfamiliar with this style of this style of documentation. Throughout the simple-icons project we use [JSDoc](https://jsdoc.app/)-style documentation, so I suggest we do the same here.

And one general guideline for writing this kind of documentation that I always try to go by is to describe only what the function does, not how it does that.

```suggestion
/**
 * Parses an SVG path into segments and builds an svgpath instance.
 *
 * @param iconPath The path value of the SVG.
 * @returns An object with two properties
 *     - segments Array of path segments as are returned by svg-path-segments.
  *     - svgPath Instance of svgpath object that can be manipulated easily.
 */
```

.svglintrc.js
```diff
@@ -46,6 +60,27 @@ function collinear(x1, y1, x2, y2, x3, y3) {
     return (x1 * (y2 - y3) + x2 * (y3 - y1) +  x3 * (y1 - y2)) === 0;
 }
 
+/**
+ * Returns the number of decimals that have a number.
+ **/
```
ericcornelissen, line 65:
```suggestion
/**
 * Returns the number of digits after the decimal point.
 *
 * @param num The number of interest.
 */
```

.svglintrc.js
```diff
@@ -46,6 +60,27 @@ function collinear(x1, y1, x2, y2, x3, y3) {
...
+    let elen = parseInt(trail, 10);
+    let idx = base.indexOf('.');
+    return idx == -1 ? elen : base.length - idx - 1 + (op === '+' ? -elen : elen);
+  }
+  return 0;
+};
+
+/**
+ * Returns the index in which a SVG path starts given the content of its file.
+ **/
```
ericcornelissen, line 78:
```suggestion
/**
 * Get the index at which the first path value of an SVG starts.
 *
 * @param svgFileContent The raw SVG as text.
 */
```

ericcornelissen (commented):
If it works, I'm fine with merging this. This should definitely be synced up with `develop` before we merge it though, and ideally I would like to see one of the new SVGs with an error fixed to get some more certainty that the errors are indeed valid.

The only review I can give are the inline comments regarding the documentation and that what is happening in [the `segmentsSVGPath` function](https://github.com/mondeja/simple-icons/blob/a1ed7f2f14c6f4ddeedc3723fe90d74fb4443906/.svglintrc.js#L43-L48) seems incredibly fragile and easy to break.

That said, and not talking about this PR specifically or to you individually @mondeja, I feel like it is increasingly difficult to read this file and after this is merged we should definitely look into refactoring it for clarity (and consistency with other scripts - this script might be what convinces me to start using Prettier in this repo as well :upside_down_face:). Some of these custom linters are incredibly lengthy, seemingly contain a lot of duplication. The amount of unrelated changes in this PR is an extra indication that this file is not in good health...

mondeja:
Rebased, `segmentsSVGPath` function removed and fixed collinear segment error in `icons/automatic.svg` icon, let me know if you want more icons to be fixed.

ericcornelissen (approved):

PR #5066, opened 2021-02-19
Simplify how scripts get icon slugs

ericcornelissen:
**Issue:** n/a
**Alexa rank:** n/a
  

### Checklist

### Description

~~This changes and renames the `titleToSlug` (formerly `titleToFilename`) function once more to avoid having to do `icon.slug || titleToSlug(icon.title)` whenever you're calling the function.~~

~~There are two exceptions to this, one in the `get-filename.js` script where all we have is the brand name; and one in the `build-package.js` script which will be removed in #4920. For this reason we might choose to go with a slightly different approach to refactoring this, namely:~~

- ~~Keep `titleToSlug` in the utilities file.~~
- ~~Add `getIconSlug` and define it as `getIconSlug = icon => icon.slug || titleToSlug(icon.title)`.~~

~~Let me know if that is preferred, I'm happy either way.~~

This adds a utility function called `getIconSlug` that is a basically a shorthand for doing `icon.slug || titleToSlug(icon.title)`.

The changes in `scripts/get-filename.js` are purely for consistency.


scripts/utils.js
```diff
@@ -5,11 +5,11 @@
 
 module.exports = {
   /**
-   * Converts a brand title into a slug/filename.
-   * @param {String} title The title to convert
+   * Get the slug/filename for an icon.
+   * @param {String} title The icon data as it appears in _data/simple-icons.json
```
mondeja, line 9:
```suggestion
   * @param {Object} icon The icon data as it appears in _data/simple-icons.json
```

mondeja (changes requested):
> - Add `getIconSlug` and define it as `getIconSlug = icon => icon.slug || titleToSlug(icon.title)`.

For consistency, I think that this would be the right choice.

ericcornelissen:
> For consistency, I think that this would be the right choice.

Fair enough, updated the Pull Request :+1:

mondeja (approved):
Perfect, lets go ahead with the merge :+1:

PR #5068, opened 2021-02-19
Update text on optional JSON data in Contributing Guidelines

ericcornelissen:
**Issue:** #4945
**Alexa rank:** n/a
  

### Checklist

### Description

This updates the section in the Contributing Guidelines regarding optional data fields in `_data/simple-icons.org` following [this comment](https://github.com/simple-icons/simple-icons/pull/4945#discussion_r576921333). Also made some minor changes to the text of both optional and required values.

Suggestions are welcome :slightly_smiling_face:


CONTRIBUTING.md
```diff
@@ -185,16 +185,30 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
+```json
+{
+    "title": "A Fictional Brand",
+    "hex": "123456",
+    "source": "https://www.a-fictional-brand.org/logo"
+}
+```
 
-Here is the object of a fictional brand as an example:
+Make sure the icon is added in alphabetical order. If you're in doubt, you can always run `npm run our-lint` - this will tell you if any of the JSON data is in the wrong order.
```
mondeja, line 202: Perhaps adding here a link pointing to [Testing Package Locally](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#testing-package-locally) section?

 ```suggestion
Make sure the icon is added in alphabetical order. If you're in doubt, you can always run `npm run our-lint` - this will tell you if any of the JSON data is in the wrong order (refers to [Testing Package Locally](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#testing-package-locally) if you don't know how to setup the environment locally).
```

**Review by mondeja — COMMENTED** [review]
I'm seeing all good. Just a minor suggestion that could be useful for newcomers...

File: CONTRIBUTING.md (hunk 1 above)
Comment (ericcornelissen) [reply CONTRIBUTING.md:202]: I'm not sure if we should link there every time we mention NPM scripts - if we decide to do that we should also add it throughout [section 3](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#3-optimize-the-icon) and link to the docker section every time docker is mentioned. I think this relates a bit to https://github.com/simple-icons/simple-icons/issues/5018 and to what extend we should go to help beginners. That said, I think it's still a good suggestion! But I think the best we can reasonable do in this regard is link to the "Testing Package Locally" section in the [Table of Contents](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#table-of-contents). (And perhaps improving the name of the section.) But I think we should leave that for a separate PR.

Comment (mondeja) [reply CONTRIBUTING.md:202]: Okay, lets leave this in a separate PR

**Review by mondeja — APPROVED** [review]
For me is good, I'll wait for @PeterShaggyNoble's review before the merge.

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -185,16 +185,30 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
 Icon metadata should be added to the `_data/simple-icons.json` file. Each icon in the array has three required values:
 
-  * The `title` of the new SVG.
-  * A `hex` color value that matches the brand's primary color. All uppercase and without the `#` pound symbol.)
-  * The `source` URL of the logo being used. There are [more details below](#source-guidelines).
+* The `title` of the new SVG.
+* A `hex` color value that matches the brand's primary color. All uppercase and without the `#` symbol.
+* The `source` URL of the logo being used. There are [more details below](#source-guidelines).
 
-Additionally, there are also optional fields that may provided for an icon:
```
PeterShaggyNoble, line 192: I'd be in favour of leaving this line in and linking it to the relevant section so that those contributing an icon don't miss it.

CONTRIBUTING.md
```diff
@@ -185,16 +185,30 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
+    "title": "A Fictional Brand",
+    "hex": "123456",
+    "source": "https://www.a-fictional-brand.org/logo"
+}
+```
 
-Here is the object of a fictional brand as an example:
+Make sure the icon is added in alphabetical order. If you're in doubt, you can always run `npm run our-lint` - this will tell you if any of the JSON data is in the wrong order.
+
+#### Optional Data
```
PeterShaggyNoble, line 204: My original intention was for this to be it's own section entirely, at the same level as "Requesting an Icon".

ericcornelissen: Hmm, not sure how well that works with the current structural approach we have :thinking: This information seems to me to be at the same level as e.g. the [SVG Filename Convention](https://github.com/simple-icons/simple-icons/blob/355454cb6caa02aba70638631c557d4e06205710/CONTRIBUTING.md#svg-filename-convention) and [Source Guidelines](https://github.com/simple-icons/simple-icons/blob/355454cb6caa02aba70638631c557d4e06205710/CONTRIBUTING.md#source-guidelines) below. And I would definitely find it strange if this is a higher-level section than [6. Update the JSON Data for SimpleIcons.org](https://github.com/simple-icons/simple-icons/blob/355454cb6caa02aba70638631c557d4e06205710/CONTRIBUTING.md#6-update-the-json-data-for-simpleiconsorg).

If your concern is that it is easy to miss, as per [another comment of mine in this thread](https://github.com/simple-icons/simple-icons/pull/5068#discussion_r580233976), we could update [the Table of Contents](https://github.com/simple-icons/simple-icons/blob/355454cb6caa02aba70638631c557d4e06205710/CONTRIBUTING.md#table-of-contents) and include these lower-level subsections.

PeterShaggyNoble: My reasoning for suggesting we have a separate section for it, at least for now, is that we don't just need contributors adding the additional entries for new icons, we also need help adding it to all our existing icons and a separate section would give it more prominence and not just confine it to the workflow of adding a new icon.

Including the lower level sections in the ToC may be a good compromise, though.

ericcornelissen: > My reasoning for suggesting we have a separate section for it, at least for now, is that we don't just need contributors adding the additional entries for new icons, we also need help adding it to all our existing icons and a separate section would give it more prominence and not just confine it to the workflow of adding a new icon.

Ah, I see now, makes sense. Still not sure if a dedicated level-2 header makes sense though mainly as it would result in this section being somewhere completely different from where it _should_ be (unless, perhaps, we restructure the entire document).

One other thing we can consider that is at least a little closer to your original intention is to somehow include a reference to this section between [Adding or Updating an Icon](https://github.com/simple-icons/simple-icons/blob/86c8dd25a7f7cd151c585fb640ce98493bb65cc1/CONTRIBUTING.md#adding-or-updating-an-icon) and [1. Identify Official Logos and Colors](https://github.com/simple-icons/simple-icons/blob/86c8dd25a7f7cd151c585fb640ce98493bb65cc1/CONTRIBUTING.md#1-identify-official-logos-and-colors).


> Including the lower level sections in the ToC may be a good compromise, though.

I recommend we leave that to a separate Pull Request though.

CONTRIBUTING.md
```diff
@@ -185,32 +185,46 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
+}
+```
+
+Make sure the icon is added in alphabetical order. If you're in doubt, you can always run `npm run our-lint` - this will tell you if any of the JSON data is in the wrong order.
+
+#### Optional Data
+
+Additionally, each icon in the `_data/simple-icons.json` file may be given any of the following optional values:
+
+* The `guidelines` may be used to specify the URL of the brand's guidelines/press kit/etc. This is useful if the SVG file was sourced from a different place.
```
PeterShaggyNoble, line 210: I think we should add a sentence here clarifying that if the `source` is the guidelines then `guidelines` should still be added with the URL duplicated.

PeterShaggyNoble:
> Ah, I see now, makes sense. Still not sure if a dedicated level-2 header makes sense

I think I've come up with a possible solution/compromise which was born out of a bigger concern of when we get around to adding the aliases; that's going to require a considerable write up which I worry might end up dominating the section on adding data to the JSON: When we _do_ add the aliases, we add a new top level section called something like "Adding Additional Icon Data & Aliases" which would have a brief introduction mentioning adding `guideline`s & `license`s for existing icons which would link back to the section being added in this PR before going into detail on the aliases. The also makes sense from the point of view that aliases are more likely to be added after the fact rather than when adding an new icon.

So, given that, I'm OK to run with this PR as-is.

> unless, perhaps, we restructure the entire document

Been thinking about that as well this week as we seem to be getting more people who don't understand or are confused by our contributing guidelines so I think it's definitely a conversation we need to have, separately to this PR.

PeterShaggyNoble (approved):
Looks good to me now, @ericcornelissen 👍🏻

PR #5069, opened 2021-02-19
Ensure that icon paths start with moveto command

mondeja:
If you add a character like `<SPACE>` as the first in an icon path, linting and testing processes pass. Although that addition does not affects drawn icon geometry, that character would be unnecessary. After this pull request, all icons *must* start with `m` or `M` or the tests will fail. If you want this check to be moved to linting or to other new test just let me know.

> Note that all icons currently available in simple-icons start with `m` or `M`.


ericcornelissen (commented):
Assuming all paths must start with `m` or `M` this looks good enough to me. A source for that would be nice, otherwise, ideally we change the regular expression to allow the path the start with, e.g., any letter. Alternatively, we can update this when necessary but that is not ideal from the perspective of contributors.

I do have some suggestions:

1. For completeness we should add `$` at the end of the regular expression. 
2. Potentially we should move this check to the SVG linter instead/as wel.

mondeja:
> For completeness we should add $ at the end of the regular expression.
> Potentially we should move this check to the SVG linter instead/as well.

I agree with both suggestions. Any suggestion for the name of the new rule?

> Assuming all paths must start with m or M this looks good enough to me. A source for that would be nice, otherwise, ideally we change the regular expression to allow the path the start with, e.g., any letter. 

Sorry, I forgot to include it, the [SVG specification says](https://www.w3.org/TR/SVG/paths.html#PathDataMovetoCommands) "A path data segment (if there is one) must begin with a "moveto" command". However, it is not very explicit since it can lead to interpretations of whether it refers to "a path data segment" as all the commands in a path (I think so) or to each command (which anyways doesn't has sense at all). Note that drawing implementations could assume implicitly that, if a path doesn't start with `M` or `m`, `M0 0` would be the first command, but as this is not defined in the specification, IMHO we should force to declare it explicitly.

ericcornelissen:
> I agree with both suggestions. Any suggestion for the name of the new rule?

How about "path-format"?

mondeja:
> How about "path-format"?

I like it! :100:

mondeja:
The error messages are:

- `Invalid path format: should start with "moveto" command ("M" or "m"), but starts with "X"`
- `Invalid path format: unexpected characters found ("X" at index 549, "Y" at index 558)`

.svglintrc.js
```diff
@@ -473,6 +473,43 @@ module.exports = {
...
               }
             }
+          },
+          function(reporter, $, ast) {
+            reporter.name = "path-format";
+
+            const iconPath = $.find("path").attr("d");
+            if (!updateIgnoreFile && isIgnored(reporter.name, iconPath)) {
+              return;
+            }
```
ericcornelissen, line 483: Do we want to keep this around for every reporter? Given that there are currently no problems I don't think we ever want this rule to be ignored :thinking:

.svglintrc.js
```diff
@@ -473,6 +473,43 @@ module.exports = {
...
+
+            const validPathFormatRegex = /^[Mm][MmZzLlHhVvCcSsQqTtAaEe0-9-,.\s]+$/;
+            if (iconPath.match(validPathFormatRegex) === null) {
```
ericcornelissen, line 486: I would suggest using this instead

```suggestion
            if (validPathFormatRegex.test(iconPath)) {
```

To avoid an error if `iconPath` is somehow not a string.

.svglintrc.js
```diff
@@ -473,6 +473,43 @@ module.exports = {
...
+              let errorMsg = "Invalid path format: "
+
+              if ("Mm".indexOf(iconPath[0]) === -1) {
```
ericcornelissen, line 489: How about using a regular expression here as well?

```suggestion
              if (/^[Mm]/.test(iconPath) === -1) {
```

.svglintrc.js
```diff
@@ -473,6 +473,43 @@ module.exports = {
...
+                // doesn't start with moveto
+                errorMsg += `should start with \"moveto\" command (\"M\" or \"m\"), but starts with \"${iconPath[0]}\"`
+              } else {
```
ericcornelissen, line 492: I don't think these two issues are exclusive, shoulnd't this be

```suggestion
              }
              if (/^[MmZzLlHhVvCcSsQqTtAaEe0-9-,.\s]+$/.test(iconPath)) {
```

At which point, the top-level if-block becomes redundant, I think :thinking:

.svglintrc.js
mondeja: I'm not against it, but will be computationally more expensive, and is something that in the long term worries me

.svglintrc.js
mondeja: Note that in the `for` loop, I'm ignoring the first character, so it's exclusive.

.svglintrc.js
```diff
@@ -473,6 +473,40 @@ module.exports = {
...
+              const validPathCharacters = "MmZzLlHhVvCcSsQqTtAaEe0123456789-,. ",
+                    invalidCharactersMsgs = [],
+                    pathDStart = '<path d="',
+                    pathDIndex = $.html().indexOf(pathDStart) + pathDStart.length;
+
+              for (let i=1; i<iconPath.length; i++) {
+                if (validPathCharacters.indexOf(iconPath[i]) === -1) {
+                  invalidCharactersMsgs.push(`"${iconPath[i]}" at index ${pathDIndex + i}`);
+                }
+              }
```
ericcornelissen, line 501: This seems more complicated that it needs to be, can't we just loop over the characters in `iconPath`with a `for-in`/`for-of` loop? Or define the `validPathCharacters` as a "negative" regular expression (i.e. `/[^...]/`) and use [`RegExp.exec`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) instead?

The latter may be too complicated for this script, so I would be fine with either.

.svglintrc.js
```diff
@@ -473,6 +473,40 @@ module.exports = {
...
+
+              // contains invalid characters
+              if (invalidCharactersMsgs.length) {
```
ericcornelissen, line 504: I personally prefer to be explicit, but it's not important

```suggestion
              if (invalidCharactersMsgs.length > 0) {
```

ericcornelissen (commented):
The linter looks good to me, I have 2 suggestions to hopefully make the code a bit more intuitive but other than that this is good to merge (even as is, if other maintainers are fine with it)

.svglintrc.js
mondeja: Something like [this](https://github.com/simple-icons/simple-icons/pull/5069/commits/a69cedef897452ab1ac5f17d10edcf837ca5175b#diff-6a7a61d95c2af287ca7aae4861a6720e553f5aaebcfab1e90b1189bd6639f14cR497-R499)?

ericcornelissen (approved):

ericcornelissen:
Looking good to me, thanks @mondeja :tada:

PR #5074, opened 2021-02-19
Added Namebase

### LINKED ISSUE #3794: Icon request: Namebase / Handshake protocol
**Name: * Namebase / Handshake 
**Website:** https://www.namebase.io/
**Alexa rank:**  301,618 
  

Namebase Company : https://avatars1.githubusercontent.com/u/41350437?s=200&v=4
  Handshake Protocol : https://avatars0.githubusercontent.com/u/29678722?s=200&v=4

These go together as, handshake is the tld domain registry protocol, while namebase is a company that is building on-top of the protocol.

More details here: https://hsd-dev.org/

adamrusted:
namebase

**Issue:** Closes #3794
**Alexa rank:** [46,919](https://www.alexa.com/siteinfo/namebase.io)

### Checklist

### Description
Blue picked as primary hex, as it's used throughout rest of the site.
~Added the 0.5pt line between the sections of the `n` to differentiate between colours. Happy to remove if preferred.~


adamrusted:
@PeterShaggyNoble give that a go. I've upped the precision to 4 decimals, as the unoptimized path wouldn't easily scale.

PeterShaggyNoble:
I'm still seeing that same difference, @adamrusted. Gave it a try myself and the same thing happened, before any optimisation, after adding the cut-outs. So, given that this would appear to be a software issue, we need to decide whether we want to clean up the difference as best we can and run with the cut-out version or go with a version without the cut-out to try and stay as faithful to the source as possible.

In this case, as the colour isn't being used to differentiate parts of the icon, I'd be OK with just using a solid "n".

adamrusted:
@PeterShaggyNoble give that a go. Got rid of the gap, so should in theory work now 🤞

PeterShaggyNoble (approved):
Spot on, @adamrusted 👍🏻 That works for me.

If anyone feels we _do_ need the cut-out and can achieve it without the distortion Adam & I got when attempting it then we can address that in a separate PR.

PR #5106, opened 2021-02-23
Add license data to NPM package and documenation

### LINKED ISSUE #5067: Expose license data in NPM package
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### This issue concerns 
  - [x] The NPM Package
  - [ ] The Packagist Package

### Description

Following #4900 and with #4945 merged, this issue is to discuss if and how [the `"license"` field](https://github.com/simple-icons/simple-icons/blob/39269e7d3947c1280a96d4e6b996a746f3ff49f4/.jsonschema.json#L32-L143) should be exposed by the NPM package (beyond it being present in the `_data/simple-icons.json` file)

ericcornelissen:
**Issue:** Closes #5067
**Alexa rank:** n/a
  

### Checklist

### Description

This adds [the license data](https://github.com/simple-icons/simple-icons/blob/4c9cb1102f36c7f0af544e6eb19686d680d5f39c/.jsonschema.json#L51-L140) for every icon that has it to the NPM package.

Currently, the value of icons that don't have a license is explicitly set to `undefined` (that is, it appears as `license: undefined` in the `.js` source). We may want to avoid this, but this would require larger changes to the way we build the package that I don't want to include in this Pull Request.

I'm not sure how to add this to the documentation in the README. So far all fields are required, and this is the first optional value... I don't think we should provide separate examples of when the license is there or not there, as it would make the documentation unnecessarily long. But I'm not sure if what I choose here is understandable. I would be open to adding more information on what to expect in the license data, e.g. that type is a SPDX license ID or the value "custom".

### ToDo after merging

- [ ] Submit Pull Request to include the optional `license` field in [the simple-icons TypeScript type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/simple-icons).


PeterShaggyNoble:
I, too, would prefer no entry rather than `undefined` but, if that needs to be addressed in a separate PR, so be it.

As for the docs, I think including all entries in the sample JSON is probably the best option with a note underneath as to which are the optional ones (`guidelines` & `license` so far) and, until we change how the package is built, that they will be `undefined` if we have no data for them.

Is it possible for us to retrieve the full license name and include it in the entry?

Side note: we should probably add a license entry for our own icon.

ericcornelissen:
> I, too, would prefer no entry rather than `undefined` but, if that needs to be addressed in a separate PR, so be it.

I can prepare a Pull Request for that and leave this pending until that is resolved.

> As for the docs, I think including all entries in the sample JSON is probably the best option with a note underneath as to which are the optional ones (`guidelines` & `license` so far) and, until we change how the package is built, that they will be `undefined` if we have no data for them.

:+1:

> Is it possible for us to retrieve the full license name and include it in the entry?

I guess that's possible if there's an NPM package or similar for that. Not sure why we would want to add it though 🤔
 
> Side note: we should probably add a license entry for our own icon.

Good point, guess I could've added that in the PR adding the license field.

PeterShaggyNoble:
> I can prepare a Pull Request for that and leave this pending until that is resolved.

I'm OK with proceeding with this PR given that it deals with some legalities, which, to me, would be the more pressing issue.

> Not sure why we would want to add it though

My use case there was for users (including possibly us on the new website) who need/want to display license info on all the icons they use, e.g. "Company X's icon is released under [license y](link)"

ericcornelissen:
> My use case there was for users (including possibly us on the new website) who need/want to display license info on all the icons they use (e.g., `Company X's icon is released under [license y](link)`) but if we need to source a package for it I don't think we should let that block this PR; we can come back and add the name later.

I think the "type" is sufficient for that purpose, even more useful than `name` arguably. E.g., [CC-BY-1.0](https://github.com/simple-icons/simple-icons/blob/0684ba48bbabba359f000be1682998cd1a86410e/.jsonschema.json#L81) is nicer, in my opinion, than _Creative Commons Attribution 1.0 Generic_.

Though I'm not totally against including the full name as well at some point, my argument would be that we should our package as minimal as possible. If some of our users want to have the full name, they should be able to do the mapping themselves.

Also, we are already displaying license information on the website just from the JSON data, see for example [_GNU_ & _GNU Social_](https://simple-icons.github.io/simple-icons-website/?q=gnu). (That is also what I based the license URL creation in this PR on).

PeterShaggyNoble:
> Though I'm not totally against including the full name as well at some point, my argument would be that we should our package as minimal as possible

Agreed on keeping things as small as possible.

> If some of our users want to have the full name, they should be able to do the mapping themselves

I was trying to avoid that extra step as it might discourage people from using the licenses at all. But, given your example usage, let's go without it for now and, should we find we _do_ have people asking how to retrieve the full name, we can revisit adding it to the package then.

So, the only change I'd suggest is as above, just showing the following in the example JSON with a note underneath that in cases where we don't (yet) have a license for an icon that entry will be `undefined`.
```
license: {
    type: '...',
    url: 'https://example.com/'
}
```

ericcornelissen:
> > Though I'm not totally against including the full name as well at some point, my argument would be that we should our package as minimal as possible
> 
> Agreed on keeping things as small as possible.
> 
> > If some of our users want to have the full name, they should be able to do the mapping themselves
> 
> I was trying to avoid that extra step as it might discourage people from using the licenses at all. But, given your example usage, let's go without it for now and, should we find we _do_ have people asking how to retrieve the full name, we can revisit adding it to the package then.

I understood that, but why would the `type` value not suffice for people to use the license data?

In any case, we can indeed add the full name at a later point if there is indeed demand for that.


> So, the only change I'd suggest is as above, just showing the following in the example JSON with a note underneath that in cases where we don't (yet) have a license for an icon that entry will be `undefined`.
> 
> ```
> license: {
>     type: '...',
>     url: 'https://example.com/'
> }
> ```

Done :+1:

README.md
```diff
@@ -66,8 +66,13 @@ console.log(icon);
...
     svg: '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">...</svg>',
     path: 'M12 12v-1.5c-2.484 ...',
-    license: undefined || { type: '...', url: 'https://example.com/' }
+    license: {
+        type: '...',
+        url: 'https://example.com/'
+    }
 }
+
+NOTE: the license may be `undefined` if there is no license data for the icon.
```
PeterShaggyNoble, line 75: Just one small linguistic change that mainly avoids the implication that the absence of a license means there is no license:
```suggestion
NOTE: the `license` entry will be `undefined` if we do not yet have license data for the icon.
```

PeterShaggyNoble (approved):
Looks good to me now, @ericcornelissen, thanks 👍🏻

PR #5115, opened 2021-02-25
Require Sorting by Slug

### LINKED ISSUE #5098: Require brands of the same name to be sorted by slug
Currently [Handshake (Protocol) is listed before Handshake](https://github.com/simple-icons/simple-icons/blob/f464deebc2e6ecfe8da62fac3399a25141cb56f9/_data/simple-icons.json#L3432-L3442) but [Hive (Blockchain) is listed after Hive](https://github.com/simple-icons/simple-icons/blob/f464deebc2e6ecfe8da62fac3399a25141cb56f9/_data/simple-icons.json#L3558-L3568) in `simple-icons.json`.

For consistency, I suggest that we add a step to our linting process to ensure that entries with the same `title` are sorted by their `slug` with the one that doesn't have a `slug` coming first and the remainder sorted lexicographically.

PeterShaggyNoble:
**Issue:** Fixes #5098
**Alexa rank:** n/a

### Checklist

### Description
This PR adds a step to our linting that requires icons in our JSON with the same `title` to be sorted by their `slug` entry with the one without a `slug` coming first and the rest sorted alphabetically. If I've done this right, then the initial commit should fail on the 2 "Handshake" icons but, as I don't have a local setup to properly test with, I could really do with someone else throwing an eye over it.

Also, do we think it's enough to just log the `title` when icons aren't sorted by `slug` or should we somehow display the `slug` as well? One way we could do it would be to replace the whole `alphabetical` function with the following:
```
  alphabetical: function() {
    const collector = (invalidEntries, icon, index, array) => {
      if (index > 0) {
        const prev = array[index - 1];
        const compare = icon.title.localeCompare(prev.title);
        if (compare < 0) {
          invalidEntries.push(icon.title);
        }else if(compare === 0){
          if(prev.slug){
            if(!icon.slug){
              invalidEntries.push(icon.title);
            }else if(icon.slug.localeCompare(prev.slug) < 0){
              invalidEntries.push(icon.title+` (${icon.slug})`);
            }
          }
        }
      }
      return invalidEntries;
    };

    const invalids = data.icons.reduce(collector, []);
    if (invalids.length) {
      return `Some icons aren't in alphabetical order:
        ${invalids.join(", ")}`;
    }
  }
```

### DIFF
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -28,11 +28,9 @@ const TESTS = {
         const compare = icon.title.localeCompare(prev.title);
         if (compare < 0) {
           invalidEntries.push(icon);
-        }else if(compare === 0){
-          if(prev.slug){
-            if(!icon.slug){
-              invalidEntries.push(icon);
-            }else if(icon.slug.localeCompare(prev.slug) < 0){
+        } else if (compare === 0) {
+          if (prev.slug) {
+            if (!icon.slug || icon.slug.localeCompare(prev.slug) < 0) {
               invalidEntries.push(icon);
             }
           }
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -39,10 +39,10 @@ const TESTS = {
       return invalidEntries;
     };
     const format = icon => {
-            if (icon.slug) {
-                return `${icon.title} (${icon.slug})`;
-            }
-            return icon.title;
+      if (icon.slug) {
+        return `${icon.title} (${icon.slug})`;
+      }
+      return icon.title;
     };
 
     const invalids = data.icons.reduce(collector, []);
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -25,8 +25,17 @@ const TESTS = {
     const collector = (invalidEntries, icon, index, array) => {
       if (index > 0) {
         const prev = array[index - 1];
-        if (icon.title.localeCompare(prev.title) < 0) {
+        const compare = icon.title.localeCompare(prev.title);
+        if (compare < 0) {
           invalidEntries.push(icon);
+        }else if(compare === 0){
+          if(prev.slug){
+            if(!icon.slug){
+              invalidEntries.push(icon);
+            }else if(icon.slug.localeCompare(prev.slug) < 0){
+              invalidEntries.push(icon);
+            }
+          }
         }
       }
       return invalidEntries;
--- a/scripts/lint.js
+++ b/scripts/lint.js
@@ -38,11 +38,17 @@ const TESTS = {
       }
       return invalidEntries;
     };
+    const format = icon => {
+            if (icon.slug) {
+                return `${icon.title} (${icon.slug})`;
+            }
+            return icon.title;
+    };
 
     const invalids = data.icons.reduce(collector, []);
     if (invalids.length) {
       return `Some icons aren't in alphabetical order:
-        ${invalids.map(icon => icon.title).join(", ")}`;
+        ${invalids.map(icon => format(icon)).join(", ")}`;
     }
   },
 
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -3466,14 +3466,14 @@
         },
         {
             "title": "Handshake",
-            "slug": "handshake_protocol",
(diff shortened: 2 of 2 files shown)


scripts/lint.js
```diff
@@ -25,8 +25,17 @@ const TESTS = {
...
+        if (compare < 0) {
           invalidEntries.push(icon);
+        }else if(compare === 0){
+          if(prev.slug){
+            if(!icon.slug){
+              invalidEntries.push(icon);
+            }else if(icon.slug.localeCompare(prev.slug) < 0){
+              invalidEntries.push(icon);
+            }
+          }
```
ericcornelissen, line 38:
```suggestion
        } else if (compare === 0) {
          if (prev.slug) {
            if (!icon.slug || icon.slug.localeCompare(prev.slug) < 0) {
              invalidEntries.push(icon);
            }
          }
```

ericcornelissen (changes requested):
Apart from the formatting this looks good to me :+1: 

And as for:

> Also, do we think it's enough to just log the `title` when icons aren't sorted by `slug` or should we somehow display the `slug` as well? One way we could do it would be to replace the whole `alphabetical` function with the following:
> 
> ```
>   alphabetical: function() {
>     const collector = (invalidEntries, icon, index, array) => {
>       if (index > 0) {
>         const prev = array[index - 1];
>         const compare = icon.title.localeCompare(prev.title);
>         if (compare < 0) {
>           invalidEntries.push(icon.title);
>         }else if(compare === 0){
>           if(prev.slug){
>             if(!icon.slug){
>               invalidEntries.push(icon.title);
>             }else if(icon.slug.localeCompare(prev.slug) < 0){
>               invalidEntries.push(icon.title+` (${icon.slug})`);
>             }
>           }
>         }
>       }
>       return invalidEntries;
>     };
> 
>     const invalids = data.icons.reduce(collector, []);
>     if (invalids.length) {
>       return `Some icons aren't in alphabetical order:
>         ${invalids.join(", ")}`;
>     }
>   }
> ```

I think that's a thing we should look into. However, I would rather change [this line](https://github.com/simple-icons/simple-icons/blob/1dcec241332b77b7c8a4dd3e18835c6eafb878ac/scripts/lint.js#L38) (e.g. with a helper) function for formatting the string than doing that in the `collector` function. E.g. something along the lines of

```js
const collector = (invalidEntries, icon, index, array) => { ... };
const format = (icon) => {
  if (icon.slug) {
    return `${icon.title} (${icon.slug})`;
  }

  return icon.title
};

const invalids = data.icons.reduce(collector, []);
if (invalids.length) {
  return `Some icons aren't in alphabetical order:
    ${invalids.map(icon => format(icon)).join(", ")}`;
}
```

PeterShaggyNoble:
Thanks, @ericcornelissen, formatting updated. And I agree with your suggested change for including the slug in the error; will get that taken care of later on today.

PeterShaggyNoble:
Pushed the update to include the slug in the error log.

scripts/lint.js
```diff
@@ -25,17 +25,30 @@ const TESTS = {
...
         }
       }
       return invalidEntries;
     };
+    const format = icon => {
+            if (icon.slug) {
+                return `${icon.title} (${icon.slug})`;
+            }
+            return icon.title;
+    };
```
ericcornelissen, line 46:
```suggestion
    const format = icon => {
      if (icon.slug) {
        return `${icon.title} (${icon.slug})`;
      }
      return icon.title;
    };
```

ericcornelissen (changes requested):
Again, part from the formatting (:sweat_smile:) this looks good to me :+1:

scripts/lint.js
PeterShaggyNoble: Won't be the last time I do that! 😆 I use tabs for indentation almost exclusively; easy to make mistakes when using spaces instead.

ericcornelissen (approved):

ericcornelissen:
Apologies, messed up a bit but this is merged with 6c435fe7411182ea8e09c42d77ba827e37bb79f0 :sweat_smile:

PR #5119, opened 2021-02-26
Update Peugeot

adamrusted:
peugeot

**Issue:** #4578
**Alexa rank:** [46,068](https://www.alexa.com/siteinfo/peugeot.com)

### Checklist

### Description
As pointed out by @gizmecano in https://github.com/simple-icons/simple-icons/pull/5013#issuecomment-786513237, Peugeot are in the process of updating their branding. Thankfully, they've included SVGs on both the Stellantis and Peugeot websites. I've gone with the Stellantis source, as it seems to be a bit higher-fidelity than the header of Peugeot's own website. As pointed out by other @simple-icons/maintainers though, some of the Stellantis SVGs haven't been optimized effectively. Happy to switch source back to Peugeot.com if we think that's a better icon.


PeterShaggyNoble:
Oof! Both versions are a little rough, each with their own many and differing issues. If we have to choose between the 2, though, I'd opt for the one from Peugeot simply because the lettering is cleaner.

adamrusted:
> If we have to choose between the 2, though, I'd opt for the one from Peugeot simply because the lettering is cleaner.

Thought as much 😅
Give me a few minutes and I'll have it updated.

adamrusted:
@PeterShaggyNoble give that a go. I've removed as many of the excess points as possible from the Peugeot.co.uk website version.

fbernhart:
LGTM 👍🏼

PR #5121, opened 2021-02-26
Disallow Hyphens in File Names

PeterShaggyNoble:
**Issue:** #4936
**Alexa rank:** n/a

### Checklist

### Description
This is the first step in resolving #4936 by preventing any new icons with `-`s in their filenames from being added while using the `slug` entry to avoid this being a breaking change for existing icons. This will help minimise the number of files that will need to be renamed in the next step which _will_ be a breaking change and therefore need to be done in a separate PR as part of the v5 milestone.

That PR will need to do the following:
1. Rename all files with `-`s in their name to remove them.
2. Remove the corresponding `slug` entry added by this PR from the JSON.
3. Handle any exceptions we want to allow for, if any (see ongoing discussion in #4936) or revert [the change](https://github.com/simple-icons/simple-icons/pull/5121/commits/0c81292dfb34b3d127adc442ea45cab642cf7f22) this PR makes to `.jsonschema.json` if not.
4. Remove the hyphens from [the RegEx for the `slug` entry in `.jsonschema.json`](https://github.com/simple-icons/simple-icons/blob/86c8dd25a7f7cd151c585fb640ce98493bb65cc1/.jsonschema.json#L16)

After merging this PR, any open PRs that add or rename icons with `-`s in their filename will need to be merged with the upstream `develop` branch and the files renamed to remove the `-`s.

Hopefully I've accounted for everything in this PR and done it all correctly; if I have then the initial commit should fail on ".NET" and "Z-Wave". Again, as I don't have a local build that allows me to test these changes properly, I'll need someone else to triple check everything.

In addition to the changes related to the hyphens, I've also removed [the line](https://github.com/simple-icons/simple-icons/blob/86c8dd25a7f7cd151c585fb640ce98493bb65cc1/scripts/utils.js#L35) that replaces the diatrics in `titleToSlug` as, with the following line removing all non-alphanumeric characters, it was redundant. Happy to add it back in, though, if others feel it affords better future-proofing to explicitly remove them.


ericcornelissen:
The changes look fine. However, I'm afraid we're still disagreeing on one aspect: do we remove hyphens from brand names as well? If hyphens are not used for anything else than removing them is not necessary, and honestly many of these brand names look a bit weird without the hyphen... So I would prefer we keep them in the slug. On the other hand, I guess that's a bit inconsistent with other symbols, especially `_`... What do you think @PeterShaggyNoble @simple-icons/maintainers ?

PeterShaggyNoble:
> If hyphens are not used for anything else than removing them is not necessary

Did we not decide that we _were_ removing them so that they _would be_ available for us to use for other purposes with the suggested use being for adding modifiers if we generate SVG files based on aliases (e.g., `title-localename.svg`, `title-duplicatetitle.svg`)?

ericcornelissen:
> Did we not decide that we were removing them so that they would be available for us to use for other purposes with the suggested use being for adding modifiers if we generate SVG files based on aliases (e.g., title-localename.svg, title-duplicatetitle.svg)?

To me that discussion wasn't quite finished, we don't _have_ to use hyphens for that. For example, for locales I think using directories is a nicer solution (e.g. `/fr-CA/bureauengros.svg`) and a similar thing could be said for the other alias categories. But I guess that is slightly off-topic for this Pull Request. I would prefer to continue that discussion in #4797 or a separate issue so it's easier to find in the future - even though it is "blocking" this PR.

Then again, as I said, removing hyphens is more consistent and "the names look a bit off" is not a very good argument and the same can be said for other symbols too.

PeterShaggyNoble:
> To me that discussion wasn't quite finished, we don't have to use hyphens for that.

You're right, we hadn't decided on that, it was just the example usage we were working off as something we _could_ use them for. Even if we haven't come up with a confirmed use case for them yet, I think it's better to remove them completely at this stage so that it doesn't require another breaking change when we _do_ come up with a use for them.

> Then again, as I said, removing hyphens is more consistent and "the names look a bit off" is not a very good argument and the same can be said for other symbols too.

I guess the names looking a bit off could be chalked up to the likes of you & me looking at them with the hyphens in them for so long! 😆

ericcornelissen (approved):
Alright, this looks good to me :+1: Will let at least one other @simple-icons/maintainers give their OK as this is quite a significant change. Before we merge this we should be very careful that this branch is up-to-date with `develop` and all custom slugs are added!

> I guess the names looking a bit off could be chalked up to the likes of you & me looking at them with the hyphens in them for so long! :laughing:

That is probably true :sweat_smile:

mondeja:
> To me that discussion wasn't quite finished, we don't have to use hyphens for that.

I've a related question about this edge case: if a brand contains a `-` character? For example, if we have a brand named `-13` and other named `13`, these will collide between them, right? The slug would solve this, but as the names `+13` and `13` would not collide, we shouldn't replace `-` by `minus` as we are doing for `+` by `plus`?

ericcornelissen:
I'd be against replacing `-` by a string as it is not clear whether it should be called "minus", "dash", or "hyphen". Moreover, if it appears in a brand's name you (almost) never pronounce it, as opposed to `+` (`c++`), `&`, (`AT&T`), or `.` (`.NET`). There is indeed a risk of collisions - though this risk is also there when we replace it by a string. The only way to avoid this is by not replacing it, which to me is still on the table.

mondeja:
> The only way to avoid this is by not replacing it, which to me is still on the table.

Yyou mean not replacing any character?

ericcornelissen:
> You mean not replacing any character?

I mean specifically not replacing `-` by anything and just keeping it in the slug. We could do this for some other characters as well, but since slugs are used in file names and URIs the character set available is somewhat limited.


> In my language, at least, are commonly pronounced if they refer to "minus" (negative).

That is already two terms "minus" and "negative" :laughing: Additionally, I don't know if you can think of any brands in your language that has a `-` in it's name, but I would bet you don't think of it as a minus (at least not in all cases).

mondeja:
> Additionally, I don't know if you can think of any brands in your language that has a - in it's name, but I would bet you don't think of it as a minus (at least not in all cases).

That's right, could refer to another term, but that statement also applies to `+`.

mondeja (approved):
Apart from the previous inconsistency, these changes looks really better that the current approach, so feel free to merge this :+1:

PeterShaggyNoble:
I think if we ever did come across a brand that uses a `-` as a "minus" then we should spell it out, as we do with "plus", and use a custom slug to get around it.

ericcornelissen:
Seems a majority is in favor of merging this in, so @PeterShaggyNoble if you can merge `develop` in one last time to make sure everything is okay you can merge it yourself after that :+1:

PeterShaggyNoble:
Done and merged. Thanks everyone 👍🏻

@simple-icons/maintainers, don't forget about this one when reviewing any already open PRs.

PR #5124, opened 2021-02-26
Add CodeMirror

PeterShaggyNoble:
CodeMirror

**Issue:** n/a
**Alexa rank:** [~88.9k](https://www.alexa.com/siteinfo/codemirror.net)

### Checklist

### Description
Open to suggestion for alternative treatments of the mirror as well as for the colour, which isn't clear at all. They've a lot of different reds on [their website](https://codemirror.net/) so I went with the one their name is displayed in in the sidebar but the grey from the logo is probably an equally viable option.

The first in a (short) series of icons I could have sworn I added here ages ago but turns out I only ever added to my personal library!


adamrusted (changes requested):
I love what you've done with this @PeterShaggyNoble!

Comparing your PR (yellow) against the source SVG though does show some small differences (see the black bleeding through the sides [here](https://user-images.githubusercontent.com/1830867/109395266-63cb0d80-7923-11eb-8d98-0563b6af0686.png). Any chance you could take another look?

PeterShaggyNoble:
Well, spotted, @adamrusted. Looks like Inkscape is butchering the union of some paths 🙁 (e.g., [before](https://user-images.githubusercontent.com/15157491/109634468-2ba11600-7b41-11eb-8b46-8cc730943714.png) & [after](https://user-images.githubusercontent.com/15157491/109634529-3f4c7c80-7b41-11eb-91cf-c36ef8eed69d.png)) This is going to take some figuring out and a bit of trial and error 🤔

PeterShaggyNoble:
That _should_ be it now, @adamrusted. I think! Had to come at it from a completely different angle than I usually do.

adamrusted:
> That _should_ be it now, @adamrusted. I think! Had to come at it from a completely different angle than I usually do.

Just for my reference, what did you have to do? I've come across a fair few icons that just won't work with Illustrator - so having to learn Inkscape on the side 😅

PeterShaggyNoble:
> Just for my reference, what did you have to do?

Instead of just doing a straight up Union of all the paths, I:
1. Drew a square, slightly bigger than the canvas, on the bottom layer.
2. Deleted all the paths from the logo I didn't want to include.
3. Working bottom to top, selected each path individually along with the square and got the Difference of them.
4. Realised that while this was producing _better_ results, there were still imperfections so undid step 3!
5. Repeated step 3 but working through the paths from top to bottom instead.
6. Broke Apart the final path.
7. Deleted what was left of the outer square.
8. Selected everything that remained and Combined it back together again.
9. Cleaned up a couple of excess points and minor imperfections.

Having done that, I now wonder if I might not have achieved the same results by using Union on the paths one at a time rather than all together, either top to bottom or bottom to top!

This has also made me question _every_ icon I've _ever_ reviewed here - have these imperfect Unions happened to any other icons I've tried to recreate for the purposes of review?! 🙈 At least I've learned my lesson going forward: If I do need to recreate the icon, compare my version to the original _before_ comparing the contributor's version to mine.

adamrusted (commented):
There's a _tiny_ difference between your PR and the source 
([see here](https://user-images.githubusercontent.com/1830867/109991808-9188cb80-7d02-11eb-9e96-69cf6229a897.png)), though I'd be happy to let something that small slide - providing another @simple-icons/maintainers agrees on it too.

PeterShaggyNoble:
Thanks, Adam; I'll see if I can fix that up tomorrow.

PeterShaggyNoble:
That should be fixed now, @adamrusted.

mondeja (approved):
I'm seeing the geometry spot on. Nice work @PeterShaggyNoble :100:
I wonder if we should add MIT in `license` entry, as [is defined in their repository](https://github.com/codemirror/CodeMirror/blob/710949d765c8fb06ed6403fc98e7c7e457cbcf1b/LICENSE) :thinking:

PeterShaggyNoble:
Tricky one that, @mondeja, as the license doesn't make any specific mention of the logo, just the software. The logo _is_ included in the zip available at https://codemirror.net/ but does that necessarily mean it's considered a part of the software? I'm not sure and, in cases where it's not clear, I think we should opt to omit the license rather than risk attributing the wrong license to a logo or icon.

mondeja:
The license says textually:

> to any person obtaining a copy of this software and associated documentation files (the "Software")

Is the logo part of the documentation as it is in the `doc/` folder? I'm not sure, but definitely, I think I will have to check if the license I use in one of my projects includes an explicit mention of the logo, although I don't think there are then many open source licenses that I can use without modifying them :sweat_smile: It seems to me that, in a court, if there is no explicit license for the logo nowhere, the license in the repository would apply, but of course I'm not sure, I'm not a lawyer.

PeterShaggyNoble:
@mondeja, perhaps we should open a new issue to track all the icons where we're in doubt on the license so we can investigate further and, if necessary, reach out for clarification?

mondeja:
> @mondeja, perhaps we should open a new issue to track all the icons where we're in doubt on the license so we can investigate further and, if necessary, reach out for clarification?

I agree. Additionally, I'm going to open a question in StackExchange about the comprehensiveness of most common open source software licenses in relation with logos.

PR #5128, opened 2021-02-27
Update number of optimization ways in CONTRIBUTE.md

bbenno:


mondeja:
Good catch! Thanks for sending the pull request @bbenno :pray:

PR #5143, opened 2021-03-01
Updated Phabricator

adamrusted:
phabricator

**Alexa rank:** [347,956](https://www.alexa.com/siteinfo/phacility.com)

### Checklist

### Description
Took the liberty while reviewing #4590 to reduce the points down on the Phabricator icon. There were _A LOT_ of extraneous points, so while it might not _match_ the source PNG perfectly, it is a LOT cleaner for smaller screen sizes.


PeterShaggyNoble:
Nice one, @adamrusted; that's a huge improvement. However, it looks like there are still a few points you might be able to remove on all four sides of the eye ([example from top right](https://user-images.githubusercontent.com/15157491/109650412-17ffaa80-7b55-11eb-98d9-bb01f03e316a.png)).

Also, did you mean to update the `source`?

adamrusted:
@PeterShaggyNoble re: the source - I figured a link to the folder where I sourced both the full icon PNG and the gear SVG (from the various folders within that directory) would be ok as a source, while keeping the trademarks page as a `guidelines` entry - just for reference that it is trademarked.

adamrusted:
@PeterShaggyNoble give that a go - should've tidied up those points for you.

PeterShaggyNoble:
Looks good to me, @adamrusted :+1: I would have pointed out the slight difference in [the top & bottom of the inner circle](https://user-images.githubusercontent.com/15157491/109960445-ae140c00-7ce0-11eb-98c4-510c4f9142dd.png) but, as you used an SVG for that part, I don't think we need worry about it. As this is (partly) a manual vectorisation, we will need one of the other @simple-icons/maintainers to sign off on it before merging.

On the source, I think, to avoid confusion, we should just link to a PNG, either by using the original source or the commit URL of the PNG in their repo.

adamrusted:
@PeterShaggyNoble have updated the source, but kept the trademarks page as `guidelines`.

mondeja (commented):
It looks pretty good and I would happy to go with this as it is, but some improvements could be done, only if you want @adamrusted: giving that the source is a raster image I would not be so strict reviewing this. Anyways, I leave you here some imperfections that I have found in the gear.

There is an almost collinear coordinate here that could be removed:

image

There are some rounded corners and others that are not rounded, so could be unified based on a choice between the two styles:

image

Some protrusions (is the correct word?) are made up by 2, others by 3 and others by 4 coordinates. It seems to me that could be redrawn using 2, regardless the choice of corners style.

image

adamrusted:
@mondeja @PeterShaggyNoble I went with sharp corners on it, partly for the ease of editing, and partly to keep the icon as optimized as possible.

mondeja (approved):
Looks perfect now! I agree that  sharp corners is the best choide, mostly because the gear could slip if they were rounded :laughing: Nice work @adamrusted :trophy:

I'll wait for a final aprove from @PeterShaggyNoble before go ahead with the merge.

PeterShaggyNoble:
Let's run with that. Thanks, @adamrusted 👍🏻

PR #5146, opened 2021-03-01
Remove unneeded ignored entries

service-paradis:
**Issue:** #4590
**Alexa rank:** N/A
  

### Checklist

### Description

I was looking at the ignore file. It looks like there is 2 entries that do not need to be there anymore.

- Kahoot! was updated in #4731
- SparkFun size was fixed in #4194


PeterShaggyNoble:
Thanks, @service-paradis 👍🏻

PR #5149, opened 2021-03-02
Add Funding Option

adamrusted:
Awaiting feedback from Open Collective, I've set this PR up to just add the Funding option once we're good to go.


.github/FUNDING.yml
```diff
@@ -0,0 +1,3 @@
+# These are supported funding model platforms
```
ericcornelissen, line 1: This seems like a line from a `FUNDING.yml` template :thinking: I think we can either remove it, or if we want this file to make sense outside of GitHub (which I think is unnecessary as it is in a folder called `.github`) make it a bit more descriptive.

adamrusted: It is indeed from a template, happy to remove it 👌

PeterShaggyNoble:
Missed this coming out of draft, sorry. Merging it now.

PR #5150, opened 2021-03-02
Add Slugs to Contributing Guidelines

PeterShaggyNoble:
**Issue:** Follow-up to #4946, #4918 & #5002
**Alexa rank:** n/a

### Checklist

### Description
This PR adds a line to the optional JSON data section about how to add a custom slug to an icon's entry and a new bullet point to the filename convention section on how to resolve naming conflicts.

I also moved the section on file names up above the section on the JSON in an effort to improve the workflow. As it was, we would have been explaining how to add a slug before explaining what a slug is or how to construct it. This way the contributor will know what a slug is and what their specific icon's slug is before they get to the explanation on how to add it to the JSON.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -204,7 +204,7 @@ The filename of the SVG should correspond to the `<title>` used in the markup fi
 1. Replace the following symbols with their alias depending on their position in the title:
 
     | Symbol | Start | Middle | End  |
-    | :----: | ----- | ------ | ---- |
+    | :----: | :---- | :----: | ---: |
     |   +    | plus  | plus   | plus |
     |   .    | dot-  | -dot-  | -dot |
     |   &    | and-  | -and-  | -and |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -249,10 +249,9 @@ Make sure the icon is added in alphabetical order. If you're in doubt, you can a
 
 #### Optional Data
 
-If you needed to add a modifier to the icon's file name above to resolve a clash with the name of an existing icon then you will also need to add a `slug` entry to the icon's JSON and set the value to your chosen file name.
-
 Additionally, each icon in the `_data/simple-icons.json` file may be given any of the following optional values:
 
+* The `slug` must be used to specify the icon's file name in cases where a modifier has been added to it in order to resolve a clash with an exitsing icon's name.
 * The `guidelines` may be used to specify the URL of the brand's guidelines/press kit/etc. This is useful if the SVG file was sourced from a different place, still if the SVG file was sourced from the guidelines, the URL should be duplicated here.
 * The `license` may be used to specify the license under which the icon is available. This is an object with a `type` and `url`. The `type` should be an [SPDX License ID](https://spdx.org/licenses/) or `"custom"`, the `url` is optional unless the `type` is `"custom"`.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -207,6 +249,8 @@ Make sure the icon is added in alphabetical order. If you're in doubt, you can a
 
 #### Optional Data
 
+If you needed to add a modifier to the icon's file name above to resolve a clash with the name of an existing icon then you will also need to add a `slug` entry to the icon's JSON and set the value to your chosen file name.
```
ericcornelissen, line 252: Because you use "also" in this sentence, I feel like it should come after the list of optional fields. Also I think the sentence can be improved a bit, feel free to improve upon this

```suggestion
If you used a modifier for the icon's file name to resolve a clash with the name of an existing icon then you also need to set the value of `slug` to your chosen file name to the icon's data in `_data/simple-icons.json`.
```

I'm also fine with leaving it at the beginning of this subsection, in which case "also" should be removed. However, given that the custom slug should be very uncommon I think we should put it after the data that is more common.

PeterShaggyNoble: My thinking on putting it first was that, uncommon as it may be, in cases when it is needed it is arguably more important than `guidelines` or `license` as the exclusion of `slug` will cause the linting to fail while the exclusion of either or both of the others won't. That an it also comes before the other 2 in the JSON! 🙂

Changed the wording a bit, making it a bit more succinct and added it to the other 2 bullet points.

ericcornelissen: This seems like a good compromise :+1:

ericcornelissen (approved):

CONTRIBUTING.md
```diff
@@ -183,7 +183,49 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
 ```
 
-### 6. Update the JSON Data for SimpleIcons.org
+### 6. Name the Icon
+
+The filename of the SVG should correspond to the `<title>` used in the markup file mentioned above, and it should follow the rules below. If you're in doubt, you can always run `npm run get-filename -- Brand name` to get the correct filename.
```
mondeja, line 188: When I run `npm run get-filename -- Brand name` the executed command, as is displayed by NPM is `node scripts/get-filename.js "Brand" "name"`. I think that it's a bit weird and a bad practice, why not define the example using quotes?

```suggestion
The filename of the SVG should correspond to the `<title>` used in the markup file mentioned above, and it should follow the rules below. If you're in doubt, you can always run `npm run get-filename -- "Brand name"` to get the correct filename.
```

CONTRIBUTING.md
```diff
@@ -183,7 +183,49 @@ Here is the svg for the Adobe Photoshop icon as an example:
...
+
+    ```yml
+    title: Citroën
+    filename: citroen.svg
+    ```
+
+1. Replace the following symbols with their alias depending on their position in the title:
+
+    | Symbol | Start | Middle | End  |
+    | :----: | ----- | ------ | ---- |
```
mondeja, line 207: We can use the alignment for the table accordingly, what do you think?

```suggestion
    | :----: | :---- | :----: | ---: |
```

PeterShaggyNoble: Sure, why not? 🙂 Although this will be rendered moot by #5121

mondeja (approved):

PR #5175, opened 2021-03-04
Update Elsevier; fix edge case error in 'collinear segments' lint rule

mondeja:
image

**Issue:** #4590
**Alexa rank:** [663](https://www.alexa.com/siteinfo/elsevier.com)

### Checklist

### Description
- ~~Manually fixed all collinear segments.~~ Rebuilt from scratch using new version.
- Found in the process an error in `collinear-segments` rule that only affects Elsevier icon and fixed it. The problem was that, when an `m` or `M` is found in the middle of a path, the `startPoint` coordinate of the algorithm must be reinitialized as if a `Z` command would be found, so `m` can't be treated like `l`.


adamrusted:
Just looked on [their website](https://www.elsevier.com/en-gb), the footer SVG seems to be completely different to what we have for this brand. Thankfully looks a lot simpler too! Did you want to update it @mondeja?

mondeja:
I'm against that version, because:

- It is much simpler than the header icon, in big sizes looks pretty bad.
- The complexity of Elsevier icon has been very useful in some contexts of simple-icons development as a reference. Apart of this edge case in the linter only discovered thanks to this version, I remeber that when I was testing the simple-icons-font build system, I took as a reference the complexity of this icon to adjust the SVG font parameters well.

That said, if other @simple-icons/maintainers agree to use the footer version, I'll be happy to change it.

PeterShaggyNoble:
Yes! Finally they simplified that monstrosity somewhat! 😆

@mondeja, I'm seeing the same version in the header of their site (as a base-64 encoded PNG) so I think we can go ahead and update it.

mondeja:
It's true! In that case, let go with this simplified version :+1:

adamrusted (approved):
Icon looks spot on, and still plenty of points for us to use it as reference! Not 100% on the linter code, so will await another @simple-icons/maintainers looking at that before we merge.

PeterShaggyNoble:
Merging this based on Adam's & Eric's approvals. Nice work, @mondeja 👍🏻 And fair play to you even touching this monstrosity! 😆

PR #5177, opened 2021-03-04
Update Clubhouse icon

service-paradis:
image

**Issue:** #4590
**Alexa rank:** N/A
  

### Checklist

### Description

- I started from scratch using the graphite logo found at https://brand.clubhouse.io/
- I used Inkscape to scale
- I optimize it


mondeja (changes requested):
As our contributing guidelines says:

> The `guidelines` may be used to specify the URL of the brand's guidelines/press kit/etc. This is useful if the SVG file was sourced from a different place, still if the SVG file was sourced from the guidelines, the URL should be duplicated here.

So it seems to me that the source should be duplicated in this case.

Other than that, the geometry is perfect! 👍🏻

service-paradis:
Thanks for the review! I tend to forgot this new `guidelines` field 😅 
Should be good now :)

mondeja (approved):
Perfect, thank you @service-paradis!

PR #5180, opened 2021-03-05
Update Fedex

adamrusted:
fedex

**Alexa rank:** [198](https://www.alexa.com/siteinfo/fedex.com)

### Checklist
  - [ ] I updated the JSON data in `_data/simple-icons.json` --> Not Necessary

### Description
Looking at the `FedEx` icon for #4590 - I found our current version is _very_ angular. By reducing the paths down, it doesn't match the original perfectly, but does look a lot _better_ at a smaller file size. Happy to revert this PR to just fix collinear segments if the change isn't wanted.


PeterShaggyNoble (approved):
I'm OK with that clean-up as it's obviously closer to how it's _supposed_ to look.

As we are deviating from the source, though, we'll need one of the other @simple-icons/maintainers to sign off on it too.

mondeja (approved):
Much better this way. I'm seeing it perfect, thank you @adamrusted
Conflict fixed and merged :wrench:

PR #5206, opened 2021-03-08
Fix broken npm link

sampl:
Hiya :)

Clicking on the "homepage" field [on npm](https://www.npmjs.com/package/simple-icons) returns a privacy error for https://www.simpleicons.org/

Removing the www subdomain from that link sends you to https://simpleicons.org/ which works just fine.

You may also want to look into why the www cert doesn't work, here's a quick fix though.

Screen Shot 2021-03-08 at 8 44 27 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "simple-icons",
   "version": "4.14.0",
   "description": "SVG icons for popular brands https://simpleicons.org",
-  "homepage": "https://www.simpleicons.org",
+  "homepage": "https://simpleicons.org",
   "keywords": [
     "svg",
     "icons"


mondeja:
I think that this is a problem with `www` subdomain in our domain configuration. I'll take a look first.

mondeja:
@ericcornelissen It is worth to configure the DNS records to force `www` as landing page as explained in [Github docs](https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site)? Note that we are currently using an [apex domain](https://docs.github.com/en/github/working-with-github-pages/managing-a-custom-domain-for-your-github-pages-site#configuring-an-apex-domain) (the CNAME is included inside the repository). If so, should we wait for the migration to the new website before making the change, and merge this pull as is in the meanwhile?

ericcornelissen:
I think we can do it here now and "again" when we switch the website to https://github.com/simple-icons/simple-icons-website.

I'd definitely be in favor of "adding" support for the `www` subdomain. But I would actually prefer we switch to omitting the subdomain as it adds no values and only makes the URL longer - hence I would like to merge this PR - but that's not a strong preference and I'm fine either way.

I'm not sure if I follow your comment, @mondeja, are you advocating we should prefer `www.simpleicons.org` over `simpleicons.org` (e.g. redirecting `simpleicons.org` to the `www` subdomain), or just that we follow the docs to support `www.simpleicons.org`. Again, I'm fine either way.

And if fixing this takes too long I think we should just merge this in for the a few releases so at least the Homepage URL on NPM works.

mondeja:
> I'd definitely be in favor of "adding" support for the www subdomain. But I would actually prefer we switch to omitting the subdomain as it adds no values and only makes the URL longer - hence I would like to merge this PR - but that's not a strong preference and I'm fine either way.

Sure. I'm going to merge this pull as is and, after the migration to the new website, I'll revisit this problem again.

> I'm not sure if I follow your comment, @mondeja, are you advocating we should prefer www.simpleicons.org over simpleicons.org (e.g. redirecting simpleicons.org to the www subdomain)...

Yes, I would be in favour of redirect `simpleicons.org` to `www.simpleicons.org`, but I'm also fine either way.

mondeja:
Again, thank you for this fix @sampl. Really appreciated :pray:

sampl:
No problemo :)

PR #5217, opened 2021-03-10
Update Stack Overflow

PeterShaggyNoble:
Stack Overflow

**Issue:** n/a
**Alexa rank:** [51](https://www.alexa.com/siteinfo/stackoverflow.com)

### Checklist

### Description
Icon from [this SVG](https://stackoverflow.design/assets/img/logos/so/logo-stackoverflow.svg), which show some differences to our current version. Our current version also has some unnecessary points. A possible alternative source could be one of the smaller, icon sized `LogoGlyph` files available [here](https://stackoverflow.design/product/resources/icons/), which are considerably different to the logo I used. **Edited to add:** If we do choose to use one of those icons instead then our Stack Exchange icon will need to be updated to match as well.

Colour is from the same source SVG but a possible alternative could be the insignificantly different `#f48024`, which is the `--theme-primary-color` variable found in their [website](https://stackoverflow.com/)'s stylesheet.


fbernhart:
LGTM 👍🏼  Thanks @PeterShaggyNoble

PR #5235, opened 2021-03-13
Add PlatformIO

andre-fajar-n:
simpleicons

**Issue:**
**Alexa rank: [16.800](https://www.alexa.com/siteinfo/platformio.org)**
  

### Checklist

### Description

I get the file from [this link](https://iconify.design/icon-sets/file-icons/platformio.html)


PeterShaggyNoble:
Thanks for the contribution, @andre-fajar-n :+1: Did you mean to close this PR or are you still working on it?

andre-fajar-n:
i want to fix this, but I don't know how to fix it yet. Do you have any suggestions? Thanks

PeterShaggyNoble:
The checks are failing because the icon is slightly too large - the largest of the 2 dimensions (the height in this case) should be `24`, while it's currently `24.32` - and is also slightly off the centre of the canvas.

andre-fajar-n:
the dimensions must be 24 or can be less than 24? And the center also must be 12?

adamrusted:
Just having a quick look at this, there's a specific line [here](https://piolabs.com/legal/trademarks.html#use-of-logos) which makes me think we cannot include this as part of our project.

> Do not create imitations, facsimiles or reproductions of these design elements on your own. The design elements may only appear with the words, logotype, graphic elements and spacing which PlatformIO has previously approved for use and may not be modified, recombined or displayed in modified scenes, images or other artwork.

Unless @andre-fajar-n is able to get hold of an official monochrome icon we can include from a contact at PlatformIO - I don't think we can include this. Would you mind asking the question?

PeterShaggyNoble:
Based on that, @adamrusted, I would agree, unfortunately.

They do also state [further down](https://piolabs.com/legal/trademarks.html#websites-and-domain-name-uses), though:
> Proper use of our Marks on websites to name or accurately describe PlatformIO’s products, services or technology is permitted.

But I don't know if that's enough to cover our usage of it. The word "Proper" also suggests to me that it must be in adherence with the paragraph you quoted above.

Based on all that, I think we need to reject this one for now. Sorry, @andre-fajar-n - thanks for your contribution & work nonetheless 🙂 If you (or anyone else) can get permission from PlatformIO directly for us to include it then we can revisit this PR in the future.

ivankravets:
Hi guys,
PlatformIO's founder here. We would like to have a PlatformIO icon added to the "simple-icons" project. A similar icon was merged to https://github.com/jesseweed/seti-ui/blob/master/icons/platformio.svg

How could we help here?

ivankravets:
@andre-fajar-n, thanks for your contribution and this PR!

1. Could we ask you to update the company color to `#f5822a`?
2. How about using the SVG icon from https://github.com/jesseweed/seti-ui/blob/master/icons/platformio.svg ? (without circle around "ant")

andre-fajar-n:
@ivankravets ok, I will update it later

andre-fajar-n:
@ivankravets ok, I will update it later

ivankravets:
@PeterShaggyNoble is it possible to reopen this PR or @andre-fajar-n should create a new one?

adamrusted:
@ivankravets - it looks like @andre-fajar-n has deleted the branch / repo that this PR came from, so they'll have to open a new one. Thanks for confirming you're happy for it to be included though!

PR #5243, opened 2021-03-14
Updated AdGuard

adamrusted:
adguard

**Alexa rank:** [16,968](https://www.alexa.com/siteinfo/adguard.com)

### Checklist

### Description
Outlined version no longer available in footer of website, and AdGuard seem to be using the 'filled' version of the icon on all their socials. Updated the hex to the base colour available in the header SVG.


PeterShaggyNoble:
There's a media kit page [here](https://adguard.com/en/media-materials.html) (linked from the sitemap in the footer) which includes [this zip](https://cdn.adguard.com/public/Adguard/Mediakits/en/AdGuard_logo.zip) of logos & icons in which all the icons are square so we should probably do the same.

adamrusted:
Good spot @PeterShaggyNoble - I've updated the icon and source to match the link you found.

PeterShaggyNoble:
Looks perfect to me now, @adamrusted :+1: Although there's a duplicate point in the top right of the tick you might be able to remove.

adamrusted:
Well spotted! 👀
Point removed, should be good to go.

PeterShaggyNoble (approved):
Nice one, @adamrusted :+1:

PR #5248, opened 2021-03-14
Update Azure Pipelines

adamrusted:
azurepipelines

Contributes to #5251.

Source referred to a PR which has now been incorporated into the base of the git repo.
Have updated source to file in project base.

**Do we take this opportunity to update the icon, including some cut-outs?**


PeterShaggyNoble:
Not _entirely_ sure cut-outs are needed around the wings given the similarity of the colours used. But adding one around the fin may help it stand out a bit better.

adamrusted:
Try that @PeterShaggyNoble.

azurepipelines

PeterShaggyNoble:
That treatment looks good to me 🙂 However, there's [a minor difference](https://user-images.githubusercontent.com/15157491/111330155-9c7b1e80-8667-11eb-9381-fd7ecd01c880.png) running down both sides of the rocket, @adamrusted, more pronounced on the right than the left. Could you have a look at that?

I also gave it a go with the cut-outs in the wings and it works better than I thought it would. What do you think?

adamrusted:
Weird difference there @PeterShaggyNoble - I prefer the example given in your image above - so will try and rebuild with that gap included.

adamrusted:
Give that a shot @PeterShaggyNoble - had to go with 4 decimals, as the icon didn't want to work with 3.

PeterShaggyNoble:
Weird 🤔 I'm still seeing that same difference running down the sides of the rocket. I double checked against my version and the same thing is happening there.

fbernhart:
I can hardly see the difference, if I zoom in by a factor of 250. Given the _very, very_ small difference and the fact that both of you already gave it a try and couldn't remove the tiny difference, I'll merge it as it is. 😊

PR #5249, opened 2021-03-14
Removed Aventrix

adamrusted:
Contributes to #5251.

Aventrix's website is down, and they haven't used their [Instagram](https://www.instagram.com/aventrix/) or [Facebook](https://www.facebook.com/aventrix/) profile since December 2018.
Submitting this PR to keep an eye on [the site](https://www.aventrix.com/) (currently showing a CloudFlare DNS error) ahead of v5.0.0


fbernhart:
I think it's safe to be removed. Even if the website would be up again, it's absolutely out of scope.

The popularity is to bad that the Alexa rank doesn't even exist.
https://www.alexa.com/siteinfo/aventrix.com

PeterShaggyNoble (approved):
No change with this one since so I think it's safe to remove. Thanks, @adamrusted.

PR #5253, opened 2021-03-14
Updated Zerply Hex

adamrusted:
Hex code didn't match up to one shown at source. SVG matches up perfectly though.


fbernhart:
👍🏼

Issue #5251, opened 2021-03-14
Full Source Review


adamrusted:
I've started going through our collection a letter at a time using [_data/simple-icons.json](https://raw.githubusercontent.com/simple-icons/simple-icons/develop/_data/simple-icons.json) file to check our sources still match up. I'd appreciate any help in going through the list, to try and get through as many as possible ready for v5.0.0!

- [x] 0-9 (7 icons) @adamrusted
- [x] A (177 icons) @adamrusted
- [x] B (78 icons) @adamrusted 
- [x] C (130 icons) @adamrusted
- [x] D (77 icons) @adamrusted 
- [x] E (56 icons) @adamrusted 
- [x] F (73 icons) @adamrusted 
- [x] G (102 icons) @service-paradis 
- [x] H (65 icons) @service-paradis 
- [x] I (48 icons) @service-paradis 
- [x] J (38 icons) @service-paradis 
- [x] K (41 icons) @service-paradis 
- [x] L (59 icons) @service-paradis 
- [x] M (111 icons) @service-paradis 
- [x] N (57 icons) @service-paradis
- [x] O (50 icons) @service-paradis 
- [x] P (125 icons) @service-paradis 
- [x] Q (24 icons) @service-paradis 
- [x] R (72 icons) @service-paradis 
- [x] S (190 icons) @service-paradis
- [x] T (99 icons) @jorgeamadosoria 
- [x] U (27 icons) @adamrusted 
- [x] V (45 icons) @adamrusted
- [x] W (60 icons) @adamrusted
- [x] X (17 icons) @adamrusted
- [x] Y (16 icons) @adamrusted
- [x] Z (22 icons) @adamrusted

jorgeamadosoria:
Finished Letter T review. 19 icons are outdated in several ways:

https://github.com/simple-icons/simple-icons/issues/6523
https://github.com/simple-icons/simple-icons/issues/6524
https://github.com/simple-icons/simple-icons/issues/6525
https://github.com/simple-icons/simple-icons/issues/6526
https://github.com/simple-icons/simple-icons/issues/6527
https://github.com/simple-icons/simple-icons/issues/6528
https://github.com/simple-icons/simple-icons/issues/6529
https://github.com/simple-icons/simple-icons/issues/6530
https://github.com/simple-icons/simple-icons/issues/6531
https://github.com/simple-icons/simple-icons/issues/6532
https://github.com/simple-icons/simple-icons/issues/6533
[...]

PR #5254, opened 2021-03-14
Update Xcode

adamrusted:
xcode

**Alexa rank:** N/A (Apple Subdomain)

### Checklist

### Description
New monochrome icon found at source.


PeterShaggyNoble:
Thanks for this one, @adamrusted; SVG looks perfect to me 👍🏻

Before merging though, do we also want to take the opportunity to review the colour? Not sure where we sourced it originally. The blue they use on [their site](https://developer.apple.com/xcode/) is `#0066cc` but it doesn't look to be the primary colour to me. The blue from their full colour icon is probably what we should go for but that may be tricky to get right as it's a gradient and all versions seem to be raster files. Apple's black could be another option but that doesn't really feel right to me.

service-paradis:
Good job @adamrusted! Glad to see an official monochrome treatment for this one!
For the color, I dont find any official resources. There is color that comes often when searching for "ios palette" or "xcode palette": `#147EFB`.

adamrusted:
Good to go @service-paradis

service-paradis:
Thank you @adamrusted, 
since I dont find any official source supporting this color (or this palette), I will wait to see what @PeterShaggyNoble thinks about it.

PeterShaggyNoble:
In the absence of anything official stating otherwise, I think that's as good a choice as any so happy to merge this as-is. Nice work, @adamrusted 👍🏻

PR #5258, opened 2021-03-14
Add Vue.js License

adamrusted:
Added License Info to Vue.js icon entry - as per their README.md at source.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -8807,7 +8807,8 @@
             "hex": "4FC08D",
             "source": "https://github.com/vuejs/art",
             "license": {
-                "type": "CC-BY-NC-SA-4.0"
+                "type": "CC-BY-NC-SA-4.0",
+                "url": "https://github.com/vuejs/art/blob/master/README.md"
             }
         },
         {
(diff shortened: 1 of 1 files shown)


fbernhart:
I think it would be better to use the URL of the specific commit here, Could you update that?

And wouldn't https://github.com/vuejs/art/blob/a1c78b74569b70a25300925b4eacfefcc143b8f6/README.md as well fit for the `guidelines` attribute?

adamrusted:
> I think it would be better to use the URL of the specific commit here, Could you update that?
> 
> And wouldn't https://github.com/vuejs/art/blob/a1c78b74569b70a25300925b4eacfefcc143b8f6/README.md as well fit for the `guidelines` attribute?

Sorted, thanks @fbernhart!

PeterShaggyNoble:
As discussed elsewhere, I think we should use the default branch rather than the commit URL so that we're always linking to the most up-to-date version and, if the file is ever removed, we're not linking to outdated information. Would you mind submitting a new PR to update the license URL to https://github.com/vuejs/art/blob/master/README.md again? Also, should the `type` be "custom"? 🤔

fbernhart:
@PeterShaggyNoble I haven't thought about that. Thanks, I'm agreeing 👍🏼

PR #5259, opened 2021-03-14
Remove Vine

adamrusted:
Given Vine was shut down in 2019, I think we're safe to look at the removal of it's icon in v5.0.0.


PeterShaggyNoble:
Been toying with this one for a good while myself, too, but each time by hand was stayed by the fact that it still somehow maintains  [a healthy Alexa rank](https://www.alexa.com/siteinfo/vine.co) which makes me wonder if people aren't actually still using it in its archived state.

adamrusted:
> Been toying with this one for a good while myself, too, but each time by hand was stayed by the fact that it still somehow maintains [a healthy Alexa rank](https://www.alexa.com/siteinfo/vine.co) which makes me wonder if people aren't actually still using it in its archived state.

Videos previously posted on Vine are still accessible through links directly to the content - though the 'platform' itself has been inactive since the 2017 shutdown. Given no 'new' Vine content is being generated, I can't foresee a situation whereby a user of the package will need the icon in future.

PeterShaggyNoble:
👍🏻 Easy to add it back in if people do start clamouring for it.

ericcornelissen:
> Easy to add it back in if people do start clamouring for it

I agree with this point and agree that we can remove it.

However, this potentially brings up a difficult question. In the context of #4735, if we get a request to re-add it, should we use the threshold for a new icon or for removing an icon? I would expect the latter...

adamrusted:
Yeah, I guess for 're-additions' we could look at removals from major releases, and treat them differently to an entirely new icon.

PeterShaggyNoble (approved):

PR #5260, opened 2021-03-14
Update Viadeo

adamrusted:
viadeo

**Alexa rank:** N/A (Subdomain)

### Checklist

### Description
SVG at new source includes `JDN`, which I'm assuming is a company that bought out Viadeo. May be worth us updating the filename in v5.0.0.


PeterShaggyNoble:
I think that given the absence of any brand guidelines now and the fact that they still use the "o" for [their favicon](https://astatic.ccmbg.com/viadeo.journaldunet.com/faviconViadeo.ico) that we should still be OK to do the same. What do you think? We would need to update, though, as the favicon differs from our current version. It also differs from the "o" in the full logo, just to make things more interesting!

adamrusted:
> I think that given the absence of any brand guidelines now and the fact that they still use the "o" for [their favicon](https://astatic.ccmbg.com/viadeo.journaldunet.com/faviconViadeo.ico) that we should still be OK to do the same. What do you think?

Sounds good - any idea if there's an SVG version about? I don't quite fancy trying to vectorize that myself!

fbernhart:
@adamrusted Can't you just take your current SVG and remove all the letters apart from the "o"? 🤔😄

PeterShaggyNoble:
Not that I've been able to find yet and that favicon is going to be far too fiddly to recreate.

But, to muddy the waters further, both the French and English Wikipedias still use the old logo, despite them joining Journal du Net back in June. _And_  there's a completely different logo on [their Zendesk site](https://viadeo.zendesk.com/)! So now I'm totally confused!

adamrusted:
> Not that I've been able to find yet and that favicon is going to be far too fiddly to recreate.
> 
> But, to muddy the waters further, both the French and English Wikipedias still use the old logo, despite them joining Journal du Net back in June. _And_ there's a completely different logo on [their Zendesk site](https://viadeo.zendesk.com/)! So now I'm totally confused!

Shall we create an issue, and just track this icon for the time being? There's clearly a lack of clarity from the brand as to what we should be using.

PeterShaggyNoble:
Let's label this as pending for now. From the French Wikipedia article, it looks like the redesign may have been a lot more recent than June so it may still be in progress.

adamrusted:
Just had another look at this, there doesn't appear to be a decent source for the SVG logo yet. I'm happy to continue with our current icon if that works for other @simple-icons/maintainers given it's close enough to the favicon.

PeterShaggyNoble:
I'd agree with that, @adamrusted 👍🏻 But I do think we should update it to match the "o" and leaf in the [new logo](https://astatic.ccmbg.com/viadeo.journaldunet.com/img/viadeo/logo-viadeo.svg) (see comparison below, with our current version in red) and we'll need to update the colour and `source` too.

PeterShaggyNoble (approved):
Looks spot on to me, @adamrusted, and I'm happy to run with the difference created by fixing the imperfection in the source. Nice work 👍🏻

PR #5261, opened 2021-03-14
Updated VSCO Source

adamrusted:
Our current source URL leads to a 404 page. Updated to their new (?) press portal.


fbernhart:
How about using https://vscopress.co/media-kit instead? Their logo can as well be downloaded from there.

adamrusted:
> How about using https://vscopress.co/media-kit instead? Their logo can as well be downloaded from there.

Updated, thanks @fbernhart!

PR #5264, opened 2021-03-15
Remove Bancontact

adamrusted:
Removes Bancontact from the repository. Their guidelines specifically state that you cannot change the colours, and use a monochrome treatment as an example.


PeterShaggyNoble (approved):
Thanks, @adamrusted 👍🏻 Always a shame to lose a brand in this manner but nothing we can do about it.

PR #4365, opened 2020-12-15
Remove Brand.ai

fbernhart:
Closes #4252

As discussed in #4252 and documented [on their official blog](https://blog.brand.ai/were-joining-invision-and-building-the-future-of-design-systems-together-b0889abd3eef?gi=9930dc3b901c), Brand.ai doesn't exist anymore. They have joined the company InVision.

This PR removes the Brand.ai icon from simple-icons.

PR #5268, opened 2021-03-15
Remove Brand.ai

adamrusted:
As mentioned in #4002 - Brand.ai has been part of InVision since 2017. This PR is just to remove Brand.ai from our collection, as it seems even InVision have stopped using that name.


service-paradis:
Looks like @fbernhart already opened a PR for this: #4365

PR #5276, opened 2021-03-15
Letter O review

service-paradis:
**Issue:** #5251

### Checklist

### Question

- According to https://opensource.org/logo-usage-guidelines, should we add the trademark symbol in this logo?

### Description

- Update OCaml data
  - Add guidelines
  - Add license
- Update Octopus Deploy data
  - Add guidelines
- Update Oculus data
  - Update source
  - Add guidelines
- Update Okta data
  - Add guidelines
- Update OnePlus data
  - Update source
- Update OnlyFans data
  - Update source
  - Add guidelines
- Update Open Source Initiative data
  - Add guidelines
- Update OpenAPI Initiative data
  - Add guidelines
- Update OpenCV data
  - Add guidelines
- Update OpenStack data
  - Add guidelines
- Update OpenStreetMap data
  - Add guidelines
- Update OpenVPN data
  - Add guidelines
- Update Opera data
  - Update source
  - Add guidelines
- Update OPNSense data
  - Update source
  - Add guidelines
- Update Oracle data
  - Update source
  - Add guidelines
- Update ORCID data
  - Update source
  - Add guidelines
- Update Osano data
  - Update source
- Update Oshkosh data
  - Add license
- Update OVH data
  - Update source

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5898,7 +5898,7 @@
         {
             "title": "OVH",
             "hex": "123F6D",
-            "source": "https://www.ovh.com/fr/news/logo-ovh.xml"
+            "source": "https://www.ovh.com/ca/en/newsroom/"
         },
         {
             "title": "OWASP",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5875,7 +5875,10 @@
         {
             "title": "Oshkosh",
             "hex": "E6830F",
-            "source": "https://oshkoshdefense.com/media/photos/"
+            "source": "https://oshkoshdefense.com/media/photos/",
+            "license": {
+                "type": "CC-BY-SA-4.0"
+            }
         },
         {
             "title": "OSMC",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5658,8 +5658,7 @@
             "source": "http://ocaml.org/img/OCaml_Sticker.svg",
             "guidelines": "https://ocaml.org/docs/logos.html",
             "license": {
-                "type": "Unlicense",
-                "url": "https://github.com/ocaml/ocaml.org/blob/master/LICENSE.md"
+                "type": "Unlicense"
             }
         },
         {
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5825,7 +5825,8 @@
         {
             "title": "OpenVPN",
             "hex": "EA7E20",
-            "source": "https://openvpn.net/wp-content/themes/openvpn/assets/images/logo.svg"
+            "source": "https://openvpn.net/wp-content/themes/openvpn/assets/images/logo.svg",
+            "guidelines": "https://openvpn.net/terms/"
         },
         {
             "title": "Opera",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5870,7 +5870,7 @@
         {
             "title": "Osano",
             "hex": "7764FA",
-            "source": "https://www.osano.com/"
+            "source": "https://www.osano.com/company/assets"
         },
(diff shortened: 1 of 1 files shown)


adamrusted:
> ### Question
> * According to [opensource.org/logo-usage-guidelines](https://opensource.org/logo-usage-guidelines), should we add the trademark symbol in this logo?

I'm reading the below as a clear 'yep'.

> 1.2. Trademark Symbol
The 'TM' or ® trademark symbol as determined by OSI must always be visible and readable for both the OSI Logo in the placements shown in these guidelines. You may not remove or obfuscate either of the TM or ® symbols in the OSI Logo.

_data/simple-icons.json
```diff
@@ -5655,7 +5655,12 @@
         {
             "title": "OCaml",
             "hex": "EC6813",
-            "source": "http://ocaml.org/img/OCaml_Sticker.svg"
+            "source": "http://ocaml.org/img/OCaml_Sticker.svg",
+            "guidelines": "https://ocaml.org/docs/logos.html",
+            "license": {
+                "type": "CC-BY-SA-4.0",
```
adamrusted, line 5661: The icon is under 'Unlicense' as mentioned in that LICENSE.md file. We can also remove the link as it doesn't differ from the default for that license.

_data/simple-icons.json
```diff
@@ -5665,12 +5670,14 @@
         {
             "title": "Octopus Deploy",
             "hex": "2F93E0",
-            "source": "https://octopus.com/company/brand"
+            "source": "https://octopus.com/company/brand",
+            "guidelines": "https://octopus.com/company/brand"
```
adamrusted, line 5674: Is the 'guidelines' link necessary if it's the same location as the source? My understanding was it was only necessary if explicit guidelines regarding the usage of the icon were on a different page to where the icon was sourced. There's a fair few which simply seem to duplicate the existing source URL. I won't review every single instance, but we may be able to remove those.

adamrusted (changes requested):
Thanks for your help with #5251 @service-paradis! 🚀
There's a couple edits below, and I'd be interested to see what other @simple-icons/maintainers think on the `guidelines` entry in particular.

_data/simple-icons.json
service-paradis: > The guidelines may be used to specify the URL of the brand's guidelines/press kit/etc. This is useful if the SVG file was sourced from a different place, still if the SVG file was sourced from the guidelines, the URL should be duplicated here.

According to the current [guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md)

_data/simple-icons.json
service-paradis: Hum... You're right...
Should we accept the type `UNLICENSE` for the license? It would make clear that this icon is unlicensed and not just that we did not find any license to it.

service-paradis: In fact, we have!
I will update it :)

_data/simple-icons.json
adamrusted: That feels like unnecessary duplication on our part - though I guess if there are explicit guidelines and not just 'use this icon' we should include them. What do other @simple-icons/maintainers feel on this?

_data/simple-icons.json
adamrusted: Think we can also remove the URL - as it looks like the default Unlicense terms mentioned there.

service-paradis: Done 😃

adamrusted (approved):
Upon reviewing the guidelines, I'm happy all these duplications are necessary. Thanks for your work on the `O`s @service-paradis 🚀

PR #5281, opened 2021-03-16
Add webhint icon

service-paradis:
image

**Issue:** N/A
**Alexa rank:** [~166k](https://www.alexa.com/siteinfo/webhint.io)
  

### Checklist

### Description

- I took the logo from https://github.com/webhintio/webhint.io/blob/5c9f10a33a6d68e1f0d2b1eff0829685b9123433/src/webhint-theme/source/images/webhint-logo.svg
- I used Inkscape to remove unused path and to scale
  - I only kept the narwhal like the favicon
- I optimized it
- I used the main color used in their stylesheet


PeterShaggyNoble:
I think we'll need to include the circle here, @service-paradis; other than their favicon, they're pretty consistent in using it everywhere, such as their GH org and their social media sites.

Other than that, their are a few imperfections in the original (where the blowhole would be and a couple in the tusk) which I don't know if we should clean up - I can't tell if they're intentional or not!

service-paradis:
> I think we'll need to include the circle here, @service-paradis; other than their favicon, they're pretty consistent in using it everywhere, such as their GH org and their social media sites.

You're right, let me give this another try

> Other than that, their are a few imperfections in the original (where the blowhole would be and a couple in the tusk) which I don't know if we should clean up - I can't tell if they're intentional or not!

I asked myself the same question. Since it could be intentional, I did not fix them. Let me know if you think I should!

service-paradis:
Here is the new version with the circle
image

PeterShaggyNoble (approved):
Looks good to me, @service-paradis :+1: Let's assume those imperfections are intentional, particularly the blowhole one, and run with them.

The only other thing I spotted was a duplicate point at the top of the inner curve of the circle - can you see if you can remove that? Sometimes it's a result of the optimisation process so, if it can't be removed then this can be merged straight in.

service-paradis:
> The only other thing I spotted was a duplicate point at the top of the inner curve of the circle - can you see if you can remove that? Sometimes it's a result of the optimisation process so, if it can't be removed then this can be merged straight in.

Just gave it another try :)

PeterShaggyNoble:
Perfect 🙂 Nice work, @service-paradis.

PR #5314, opened 2021-03-19
Update Dior

adamrusted:
dior

**Alexa rank:** [1,760](https://www.alexa.com/siteinfo/dior.com)

### Checklist

### Description
Updated Source to Dior homepage, where capitalized logo is being used. Compared against socials, and this appears to be their current styling.


service-paradis:
I agree that we need to change it for the uppercase version.
The icon is looking good 💯 
There is a slight imperfection above the letter `I` in the official version. Maybe we should fix it since it does not look intentional?
image

Other than that, I was wondering if the precision of 4 was needed?

adamrusted:
Just generated a 3 decimal path [here](https://user-images.githubusercontent.com/1830867/112289644-61d83e00-8c86-11eb-9745-2a7fb2825770.png), with the source in green and 3 decimal in black. If we're happy with those differences around the curves, I'll submit. Otherwise will require a rebuild.

service-paradis:
> Just generated a 3 decimal path [here](https://user-images.githubusercontent.com/1830867/112289644-61d83e00-8c86-11eb-9745-2a7fb2825770.png), with the source in green and 3 decimal in black. If we're happy with those differences around the curves, I'll submit. Otherwise will require a rebuild.

Thank you! Looks difficult to keep it correct using 3 decimals. Might be better keeping at 4, what do you think?

adamrusted:
Sure - let me rebuild fixing that issue with the `I` and I'll push it here.

adamrusted:
Done, @service-paradis

service-paradis:
Thanks, looking better 🎉 

A small note. There are 2 points in top of each other in the top left corner of the `I` (where there was a glitch).
image
It creates a little bump:
image

I'm able to remove one using Inkscape. It is really minor, so we can go as-is if you want.

adamrusted:
Weird, I'm not seeing two points there when I edit the file in Illustrator. Think it's just a _slight_ curve to compensate for where two points used to be. Will ask for another @simple-icons/maintainers to check this though, just to be sure.

service-paradis (approved):
> Weird, I'm not seeing two points there when I edit the file in Illustrator. Think it's just a _slight_ curve to compensate for where two points used to be. Will ask for another @simple-icons/maintainers to check this though, just to be sure.

Looks good enough to be included. It is a micro-optimization and could be improved later. I'm merging this one. Thank you @adamrusted!

PR #5316, opened 2021-03-19
Letter N review

service-paradis:
**Issue:** #5251

### Question

I think we should update our NPM logo. The one we currently have is not really used by npm.
On their website, they use the filled wordmark in the header and the squared `n` in the footer.
I really think the following logo would fit better in our collection:
image

What do other @simple-icons/maintainers think about it?

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5566,7 +5566,8 @@
         {
             "title": "Node.js",
             "hex": "339933",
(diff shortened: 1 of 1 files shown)


adamrusted:
Yep, I think we should go with the boxed `n` for our version.

adamrusted (approved):
LGTM! Thanks @service-paradis 🚀

Issue #5340, opened 2021-03-23
Update npm icon


service-paradis:
**Brand Name:** npm
**Official resources for icon and color:**
  
- https://www.npmjs.com/

I think we should update our NPM logo. The one we currently have is not really used by npm.
On their website, they use the filled wordmark in the header and the squared `n` in the footer.
I really think the following logo would fit better in our collection:
![image](https://user-images.githubusercontent.com/12817388/111791718-db90b600-8899-11eb-9e72-217eae405e02.png)

PR #462, opened 2017-05-26
Add Sogou, Salesforce, Yandex, Pluralsight, NPM, Redux, Gerrit, AngelList, NGINX, Signal, Keep, Atom, Sublime, Visual Studio Code, WebStorm, Keybase, Pandora, Strava, SourceForge, SlickPic and AIM

birjolaxew:
Each commit has comments explaining various decisions for it.

Fixes #460, #455, #449, #450, #444, #441, #442, #452, #413, #393, #374, #351, #343, #230, #193, #138 and #421.

Issue #444, opened 2017-05-10
Request: NPM


greybax:
Website: https://www.npmjs.com
Logo: https://github.com/npm/logos
(https://upload.wikimedia.org/wikipedia/commons/d/db/Npm-logo.svg)

PR #5350, opened 2021-03-25
Update NPM icon

service-paradis:
image

**Issue:** #5340

### Checklist

### Description

- I took the squared `n` in the footer at https://www.npmjs.com/about
- I used Inkscape to scale
- I optimized it

**_I also think we should rename this brand in a future breaking release. They always use `npm`, not `NPM`._**


adamrusted (commented):
Icon looks spot on to me, and I may be wrong but I don't think a change in casing constitutes a breaking change, so we may get away with it here. Will ask another @simple-icons/maintainers to advise on that though.

I think we will need to keep the guidelines entry in, as there's a specific section about the logo, which states: 
`Our npm Logos are very recognizable and deserves special treatment. The npm Logos signify us, or a special relationship with us, and you may use them only with our permission. Since the goal is to avoid confusion about you being us, or your relationship with us, context counts. We will consider requests on a case-by-case basis.`

ericcornelissen:
In our npm package, searching by title is case insensitive, [see the `index.js` template](https://github.com/simple-icons/simple-icons/blob/develop/scripts/templates/index.js), and filenames are normalized to lowercase anyway so changing "NPM" to "npm" would not be a breaking change.

As for whether we should change it, it seems it was added way back in https://github.com/simple-icons/simple-icons/pull/462 (closing issue https://github.com/simple-icons/simple-icons/issues/444) without much thought or discussion regarding the name. I agree that the name is pretty much always stylized as "npm" so I'd be in favor of changing it :+1:

service-paradis:
Thank you @adamrusted and @ericcornelissen!

> I think we will need to keep the guidelines entry in, as there's a specific section about the logo, which states:

You're right, good catch! My mistake, dont know why I removed it 🤔 

> I don't think a change in casing constitutes a breaking change

I'm not sure why I was thinking it was breaking yesterday, but thank you both for the input 😄 

Just pushed the changes!!

adamrusted (approved):
Spot on, thanks @service-paradis!
Just a reminder for @simple-icons/maintainers to make sure this doesn't bump us to v5.0.0 next week!

PR #5351, opened 2021-03-26
Remove Mathworks icon

service-paradis:
According to https://www.mathworks.com/brand.html#logo, we cannot keep this logo.

They mention explicitly that we cannot use a monochrome version and that we cannot change the color.


adamrusted (approved):
This looks good to me, @service-paradis. Feel free to merge straight in when this is 'ready for review'.

Issue #5352, opened 2021-03-26
Add App Annie icon


LitoMore:
**Brand Name:** App Annie
**Website:** https://www.appannie.com
**Alexa rank:** 3,532
  

**Official resources for icon and color:**
https://www.appannie.com/en/about/press/

PR #5353, opened 2021-03-26
Add App Annie icon

LitoMore:
**Issue:** #5352
**Alexa rank:** 3,532
  

### Checklist

### Description

I pick the hex value from App Annie's primary button color.


adamrusted (changes requested):
Thanks for your contribution @LitoMore!
As you can see [here](https://user-images.githubusercontent.com/1830867/112718619-1d4ddc00-8eec-11eb-98cd-8e69d804090a.png) there are quite a few differences between your PR (red) and the source file (black). This is a particularly difficult one to work with, purely on the unnecessary paths in the clipping mask, but would you mind taking another look at it?
It also looks like their primary colour (from the UI screenshots) is actually the dark blue. Do you think that hex would better represent the brand?

Looks like there are officially provided EPS files on their [press page](https://www.appannie.com/en/about/press/) (including a monochrome treatment), so it may be worth using those - and updating the source link to that specific page.

LitoMore:
Hi @adamrusted, I updated the icon by using the EPS source file. But I got some problems with the icon size.

image

You can see that the height is not an integer, then the CI reported this:

```
x icon-centered <path> must be centered at (12, 12); the center is currently (12, 12.237)
```

Do you have any solution for this?

adamrusted:
> You can see that the height is not an integer, then the CI reported this:
> 
> ```
> x icon-centered <path> must be centered at (12, 12); the center is currently (12, 12.237)
> ```

Providing the icon is centered to a 24*24 canvas, the fact it's not an integer shouldn't be an issue. You may just need to raise the precision in both your SVG export and the SVGO config file. A precision of 4 across the board may fix that.

fbernhart:
@LitoMore The SVG has to have a size of maximum 24px*24px. Therefore you have to scale down the icon first, so that the larger size (the height in your case) is 24px. While keeping the aspect ration, the width should then be around 24/24.47*24 = 23.539px.

After it has the appropriate size, you can then center it.

LitoMore:
@adamrusted @fbernhart Thanks. Let me try.

PeterShaggyNoble:
Are you still working on this one, @LitoMore?

LitoMore:
@PeterShaggyNoble Ah yes, I'm using Sketch for editing this icon, but it supports only `0.00` precision. I'm looking for a new application to edit this.

PeterShaggyNoble:
[Inkscape](https://inkscape.org/) is a decent, free SVG editor if you want to give that a try.

PeterShaggyNoble:
Are you still working on this one, @LitoMore?

LitoMore:
Ah yes. Let me continue this work.

PeterShaggyNoble:
Any update, @LitoMore? We're going to try to work through some of our backlog this weekend so it would be good to get this one merged, if we can.

LitoMore:
Oh sorry, will finish it in this week.

LitoMore:
@PeterShaggyNoble Updated.

PeterShaggyNoble (changes requested):
Thanks, @LitoMore, that lines up perfectly now 👍🏻 However, there are a couple of excess points on the leftmost and rightmost corners that you should be able to remove. Also, could you update the source to https://www.appannie.com/en/about/press/, please?

LitoMore:
@PeterShaggyNoble Thanks! Updated!

PeterShaggyNoble (approved):
Nice work, @LitoMore, everything looks perfect to me 👍🏻 Thanks again for the contribution.

PR #5355, opened 2021-03-26
Letter L review

service-paradis:
**Issue:** #5251

### Notes

- I think we might need to remove https://letsencrypt.org/trademarks/
- I think we might need to update https://lgtm.com/, but I'm not sure if the logo we see is the Semml logo or not...
- I think we might need to rename https://www.webtoons.com/fr/, not sure why it is called `LINE WEBTOON`
- We need to update https://www.loom.com/press

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4665,7 +4665,11 @@
         {
             "title": "Liberapay",
             "hex": "F6C915",
-            "source": "https://liberapay.com/assets/liberapay/icon-v2_yellow-r.svg"
+            "source": "https://en.liberapay.com/about/logos",
+            "guidelines": "https://en.liberapay.com/about/logos",
+            "license": {
+                "type": "CC0-1.0"
+            }
         },
         {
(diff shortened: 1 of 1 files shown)


adamrusted:
> * I think we might need to remove [letsencrypt.org/trademarks](https://letsencrypt.org/trademarks/)

Yep. Specific line on not altering the colour would make me think the same. Worth getting in touch with them before removal?

> * I think we might need to update [lgtm.com](https://lgtm.com/), but I'm not sure if the logo we see is the Semml logo or not...

Looks like the icon we already have is for LGTM - as Semmle use a wordmark. Alternative we could use is the `lgtm` wordmark in the footer of the site.

> * I think we might need to rename [webtoons.com](https://www.webtoons.com/). `WEBTOON` instead of `LINE WEBTOON`?

Yep. Completely agree on this one. Looks like they've removed `LINE` from their branding across the board.

adamrusted (approved):
`L`GTM! Thanks @service-paradis 🎉

PR #5362, opened 2021-03-28
Letter E Review

adamrusted:

### Description
At time of review, [EverPlaces is offline](https://everplaces.com/). It is still [within scope](https://www.alexa.com/siteinfo/everplaces.com) at the moment, so we should keep an eye on it ahead of v5.0.0 just in case we need to remove it.


fbernhart:
LGTM 👍🏼  Great work.

PR #5367, opened 2021-03-28
Add Gravit Designer

stefanobartoletti:
image

New icon for [Gravit Designer](https://designer.gravit.io/), a design app similar to Sketch or Figma.

The icon was made by me, starting from their SVG logo, i merely added a small circular border to keep the round shape.

I took the hex value of the main brand color from the buttons and other UI elements inside the app, since the logo itself has a gradient and picking the right value from there was not easy.


adamrusted:
Thanks for your work on this @stefanobartoletti!
Couple of suggestions on my end - when we differentiate between colours, we try to use a `0.5px` stroke on most icons. The outer circle seems to be a bit wider than that, I'm not sure what other @simple-icons/maintainers would think, but I'd probably want to match that to our other icons in that regard.

It's worth also pointing out that the right-hand green section has been shaded, whereas the left has not.
Having a play with this, I'm not sure whether we should fill them both, outline them both, or ignore them completely! 😅

The other option would be to start with the base circle of the yellow, and [outline everything on top of it](https://user-images.githubusercontent.com/1830867/112967739-7b283100-9143-11eb-917f-7b7480f11556.png).

What're your thoughts on that kind of look?

stefanobartoletti:
Thank you for your feedback. Here are my reasoning about these points

- I set the width of the outer border with a small size icon in mind, but as you point out, in the larger versions it seems to wide. I can set this to be 0.5px if needed.

- for the green parts, I tried to preserve the most important shapes, so on the left I kept the pink strokes because I considered the green part to be  background, while on the right, I kept the green shape beacuse I considered the yellow part as a background. Let me know if this reasoning could be correct

-Outlining everything would work on larger size, but will the most recognizable parts of the logo be preserved when using small sizes?

PeterShaggyNoble:
Of the 2 proposed treatments my vote goes to @stefanobartoletti's but, for me, it's still not an entirely accurate representation of their branding.

stefanobartoletti:
The problem is that the original logo is a basic shape entirely filled, even if with different colors/gradients. There is no easy way to preserve this, especially for smaller sizes where some contrast is needed to prevent it to appear like a solid circle.

I can fix the border width, but for other changes I'll wait for suggestions.

PeterShaggyNoble:
Coming back to this with fresher eyes this morning, I just don't know that's it's going to work in monochrome, whatever we do with it. The design, to me, looks to be an abstract, entirely reliant on the colours - when the colours are removed the meaning of the icon changes. In your version, for example, I see a "Z" and in Adam's I see a tennis ball (although that might have as much to do with his choice of colour! 😆).

I've played around with it a bit myself this morning, trying various other things to you and Adam and these were the best I could come up with - as I worked on it more the darker green in the upper left looked more & more like it was supposed to be a shadow so I omitted it - but they still don't quite work for me; I see a bird in the first and a sunset in the second!

adamrusted:
Yeah, I'm feeling that unless we get a monochrome version that's approved by the brand themselves, that this is out of scope. None of our versions really do it justice, especially at the smaller sizes. @stefanobartoletti - did you want to get in touch and see if they have a monochrome version?

stefanobartoletti:
On their website there is not a "branding" page, so I doubt that there are guidelines and alternative logos (i.e. monochrome), and even contact info are not easily found.

At the moment I'm unable to investigate further about this.

PeterShaggyNoble:
They do have [_a_ presence](https://github.com/Gravit-io) on GitHub but they're not all that active and neither of the 2 listed teams members are active at all but, as a first step, let's ping @anna-ignatenkova & @teleranek anyway in the hopes that they do see this and can provide us with some feedback and suggestions.

stefanobartoletti:
I tried to contact them with a PM to their Twitter profile, this is what they answered:

> Unfortunately, we don't have a "press release" or public brand guidelines.
> I have talked to our designer team and the first option is the one that best represents the brand. Although we can't say this could be "official" as it should be evaluated by COREL and I'm not aware of the procedure to get their approval.
> If you are asking about our opinion, first one is the best but if you need an "official approval of usage", unfortunately we can't provide you with that.

So I think that we have two options, taking the original icon from this PR (maybe by fixing things like border width and the likes), or completely discard it and don't include this brand at all in the library. I would prefer the first option, but let me know how do you like to go on.

adamrusted:
Thanks for getting in touch with them @stefanobartoletti!
As I mentioned above, I unfortunately don't feel the monochrome versions presented can do the product justice - and as such I'd be inclined to say this is out of scope. I'll ping other @simple-icons/maintainers for their opinion though - before we take any action.

PeterShaggyNoble:
I'd be _slightly_ more in favour of adding it based on that response but the fact that they themselves would need the OK from COREL in order to be able to give official approve means that, unfortunately, I don't think we can add it at this time 🙁

Will leave this open for a spell, though, in case any of the other @simple-icons/maintainers want to make a counter argument for its inclusion.

service-paradis:
Corel have strict [trademark guidelines](https://www.corel.com/fr/trademarks/). Unfortunately, I think we wont be able to add this one.

stefanobartoletti:
I understand. If this is the case, it's ok if I close this PR?

PeterShaggyNoble:
Yep, unfortunately, with all that's transpired, I think we'll have to reject this one for now 🙁 We appreciate your work on it nonetheless, @stefanobartoletti, and thanks as well for reaching out to them on our behalf. Looking forward to future contributions from you.

PR #5369, opened 2021-03-28
Add aliases to the JSON schema

ericcornelissen:
**Issue:** #4797
**Alexa rank:** n/a
  

### Checklist

### Description

This adds the aliases as discussed in #4797 to the JSON schema - as far as I can tell nearly (except for licenses on duplicates) exactly as it is discussed there - and Contributing Guidelines. I omitted the `license` field from duplicates as it seems to me that the icon cannot be relicensed :thinking: If I'm mistaken I'm happy to update the PR. Also, while working to implement this still a few questions came up, so I listed those below.

I added some initial documentation regarding aliases, feedback on that is welcome :slightly_smiling_face: However, I propose we leave changes that touch things other than the aliases for a separate PR.

Also, I made some "unrelated" changes to the JSON scheme file to reduce duplication. It may be easier to review [the file directly](https://github.com/ericcornelissen/simple-icons/blob/jsonlint/aliases/.jsonschema.json).

#### Discussion points

<details>
<summary>[In discussion] `old` array or string?</summary>
Currently "old" is set to an array of string. I couldn't find any discussion on this topic in the issue, but I was wondering why it is an array and not just a string. The only argument for an array I can think of is a brand rebranding twice within the span of one major release of simple icons, which really shouldn't happen. However, that might be enough to justify it being an array, especially given that it is mainly intended for internal backwards compatibility support. Still, I wanted to raise this point in case someone disagrees.
</details>

<details>
<summary>[Implemented in f588f6f4680f4e5584b791cc42eea4f12062bddb] Localisation format</summary>

This PR currently implements the localisation as per the issue, that is an array of objects of the form `{ "title": "local title", "locale": "en-US" }`. I was wondering if we should perhaps change this to an object where the keys are locales and the values or the local title, as in: `"aliases": { "loc": { "en-US": "local name", "nl-NL": "lokale naam" }`. My reasoning being that users are most likely interested in a specific locale. In that case, my proposal is a bit easier to use as you can just do `icon?.aliases?.loc["en-US"]` to get the `en-US` locale if it exists. Otherwise you need to e.g. `.filter` the array and compare the "locale" value of each entry.

If we want this, that can be achieve by replacing the `locale` definition with

```json
"locale": {
  "$id": "#locale",
  "description": "A localized brand name",
  "type": "object",
  "patternProperties": {
    "^[a-z]{2}-[A-Z]{2}$": {
      "type": "string",
      "description": "The local name of the brand"
    }
  },
  "minProperties": 1,
  "additionalProperties": false
},
```

and the `loc` property of `aliases` with

```json
"loc": {
  "description": "Localized names of the brand",
  "$ref": "#/definitions/locale"
}
```
</details>

<details>
<summary>[Resolved] Name of `dup`</summary>

_Raised by @PeterShaggyNoble in [this comment](https://github.com/simple-icons/simple-icons/issues/4797#issuecomment-800214949)_

> Would `alt`(ernative) offer better clarity than `dup` for that type?

I personally disagree as `alt` seems to me to mean the same as `aka`. On that note, I would be okay with renaming `aka` to `alt`.
</details>

<details>
<summary>[Resolved] Name of `loc`</summary>

_Raised by @PeterShaggyNoble in [this comment](https://github.com/simple-icons/simple-icons/issues/4797#issuecomment-800214949)_

> And maybe `int`(ernational(ised)) instead of `loc` as well?

Again, I personally disagree because we aim to provide the local name of the brand, so "loc" is arguably more descriptive. I guess it depends on the exact definition of "localisation" and "internationalization"...
</details>

#### Post-merge task list

- [ ] Open issue to determine how long `old` values should be kept
- [ ] Restructure [`CONTRIBUTING.md`](https://github.com/simple-icons/simple-icons/blob/6798e69c5e3ad21a1233ba2e26887e334dd09be5/CONTRIBUTING.md)


PeterShaggyNoble:
Thanks for taking this one on, @ericcornelissen, I probably would have made a complete mess of it! 😆 To your points:

1. `old` should be an array to allow for the potential for a brand to change its name more than once through its lifetime in our collection.
2. I very much like your suggested change to the `loc` entry.
3. I agree with the counterpoints you made in your reply to the original issue in relation to the naming of `dup` & `loc` and that we should leave them as they are.

ericcornelissen:
> 1. `old` should be an array to allow for the potential for a brand to change its name more than once through its lifetime in our collection.

While I'm not necessarily against it (again, I would be fine changing it on the off chance a brand renames twice within 6 months), do we really want to keep support for a brand name change more than one major release ago?


> I very much like your suggested change to the `loc` entry.

Any further input from another @simple-icons/maintainers before I change this?


> I agree with the counterpoints you made in your reply to the original issue in relation to the naming of `dup` & `loc` and that we should leave them as they are.

:+1:

PeterShaggyNoble:
`old` aliases are going to be permanent so that people can _always_ find them under those names. This benefits people coming back looking for an icon they may have used that they don't realise has since had its name changed.

ericcornelissen:
If other @simple-icons/maintainers agree, I can life with that. That said, it does seem excessive to me, at some point the old name just shouldn't have to be supported anymore I think. Whether that's one of our release cycles or longer is a different question.

adamrusted:
> it does seem excessive to me, at some point the old name just shouldn't have to be supported anymore I think. Whether that's one of our release cycles or longer is a different question.

I think either one major release, or keeping the old name permanently if the brand still goes by the old name in some contexts sounds like a good idea. I agree that once the name is "no longer used" we can free up space by removing it, and the 6 (?) months between major releases should give people adequate time to transition their project.

ericcornelissen:
> if the brand still goes by the old name in some contexts

Depending on what you mean by that exactly, this sounds to me like either 1) it should be an "aka" or 2) that 6 months is too short a window for "old" values to be removed.

adamrusted:
I suppose the case I presented could be put into either `aka` or `old` - I can't think of an example off the top of my head, but either entry in the JSON would work. In that case then, we should perhaps keep the `old` entry until the brand name changes again, or until we can make a reasonable argument to remove it (guidelines explicitly prohibit use of old name, for example).

PeterShaggyNoble:
If we are to periodically remove `old` aliases I think 2 major releases, or a year, would be reasonable window. Although I'd still be in favour of keeping them permanently.

ericcornelissen:
> If we are to periodically remove old aliases I think 2 major releases, or a year, would be reasonable window. Although I'd still be in favour of keeping them permanently.

Whatever we choose, it's not blocking this Pull Request - `old` is currently set to an array of strings.

.jsonschema.json
```diff
@@ -145,11 +169,62 @@
...
+        "title": {
+          "$ref": "#/definitions/title"
+        },
+        "slug": {
+          "$ref": "#/definitions/slug"
+        },
+        "hex": {
+          "$ref": "#/definitions/hex"
+        },
+        "guidelines": {
```
PeterShaggyNoble, line 191: Do we want to allow for a `source` here as well, for the `hex`? Think we discussed that briefly in the issue.

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
 
 In general, make sure the URL does not contain any tracking identifiers.
 
+#### Aliases
```
PeterShaggyNoble, line 290: Our contributing guidelines are starting to get a bit cluttered to the point where new contributors are starting to be confused by them. As aliases are very much secondary information, should we document them in a separate file linked from `CONTRIBUTING.md`? Alternatively we could, separately to this PR, look into rewriting the guidelines completely to provide better clarity for new contributors.

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+    ]
+  }
+}
+```
+
+Where the string is **different** from the original title as well as all other strings in the list.
+
+##### Duplicates
+
+We collect the name of duplicates, brands that use the same logo but have a different name, to prevent duplicating an SVG while at the same time making the SVG available under the name of the duplicate. To add a duplicate you add the following to the icon data:
```
PeterShaggyNoble, line 313: Missing an "s" here and we should probably go with "icon" instead of "logo":
```suggestion
We collect the names of duplicates, brands that use the same icon but have a different name, to prevent duplicating an SVG while at the same time making the SVG available under the name of the duplicate. To add a duplicate you add the following to the icon data:
```

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+
+```json5
```
PeterShaggyNoble, line 315: Minor tyo:
```suggestion
```json
```

.jsonschema.json
```diff
@@ -29,6 +24,35 @@
           "description": "The brand guidelines",
           "$ref": "#/definitions/url"
         },
+        "aliases": {
```
PeterShaggyNoble, line 27: As it could, potentially, be a very large object and the information it contains is less important than the licensing info, I'd move the `aliases` entry to the end, after the `license` entry.

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
 
 In general, make sure the URL does not contain any tracking identifiers.
 
+#### Aliases
+
+Lastly, we aim to provide aliases of three types for various reasons. Each type of alias and their motivation can be found below. If you're unsure, you can mention an alias you're considering in your Pull Request so it can be discussed.
```
PeterShaggyNoble, line 292: Small linguistic fix:
```suggestion
Lastly, we aim to provide aliases of three types for various reasons. Each type of alias and its purpose can be found below. If you're unsure, you can mention an alias you're considering in your Pull Request so it can be discussed.
```

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+
+##### Also Known As
+
+We collect "also known as" names to make it easier to find brands that are known by different names or by their abbreviation/full name. To add an "also known as" name you add the following to the icon data:
```
PeterShaggyNoble, line 296: Should we make mention here that "different names" does not include localised names which are documented below?

PeterShaggyNoble (commented):
Oops, missed that you'd already made the updates to the `loc` type. Just a few suggested changes below.

CONTRIBUTING.md
mondeja: Note that the code block has comments, it is in JSON5 format.

.jsonschema.json
```diff
@@ -29,6 +24,35 @@
...
+            "dup": {
+              "description": "Different brands that use the exact same icon",
+              "type": "array",
+              "items": { "$ref": "#/definitions/duplicate" }
+            },
+            "loc": {
+              "description": "Localized names of the brand",
+              "$ref": "#/definitions/locale"
+            },
+            "old": {
```
mondeja, line 45: I feel like the renaming strategy should be documented, in contribution guidelines or in some internal document, right?

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+
+```json
+{
+  "title": "the original title",
+  "aliases": {
+    "aka": [
+      "tot"
```
mondeja, line 303: Maybe adding another example here to increase readability? This array looks to me more like a configuration file :sweat_smile: 

```suggestion
      "tot",
      "thetitle"
```

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+{
+  "title": "the original title",
+  "hex": "123456",
+  "aliases": {
+    "dup": [
+      {
+        "title": "the duplicate's title",
+        "hex": "654321", // Only if different from original's color
+        "guidelines": "..." // Only if different from original's guidelines
+      }
```
mondeja, line 325: Maybe the `license` field should be added here, if it is different from the original?

```suggestion
        "guidelines": "...", // Only if different from original's guidelines
        "license": {"..."} // Only if different from original's license
      }
```

CONTRIBUTING.md
```diff
@@ -287,6 +287,66 @@ If the SVG is sourced from:
...
+
+Where the string is **different** from the original title as well as all other strings in the list.
+
+##### Duplicates
+
+We collect the name of duplicates, brands that use the same logo but have a different name, to prevent duplicating an SVG while at the same time making the SVG available under the name of the duplicate. To add a duplicate you add the following to the icon data:
+
+```json5
+{
+  "title": "the original title",
```
mondeja, line 317: The data `_data/simple-icons.json` file is indented using 4 spaces, these new examples should use the same indentation of previous examples already defined inside Contribution Guidelines.

.jsonschema.json
PeterShaggyNoble: By and large, it's we maintainers that submit the PRs for breaking changes like a brand changing its name so, in the interests of minimising the information in the guidelines, I'd be in favour of omitting it. If an outside contributor does submit a PR to rename an icon, we can ask them to add the alias in the PR.

CONTRIBUTING.md
mondeja: After a bit of searching, I think that [`jsonc` (JSON with comments)](https://stackoverflow.com/a/55708553/9167585) fits better the needs here. Note that is supported by ]Github Flavored Markdown code blocks (linguist)](https://github.com/github/linguist/blob/e6860ab64b463e022294b0f71d468d5243687027/lib/linguist/languages.yml#L2569).

.jsonschema.json
mondeja: In that case, we should document this internally? I'm a bit loss with all the changes that are taking place lately.

CONTRIBUTING.md
ericcornelissen: If it's the same SVG I don't know how the license can be different :thinking:

.jsonschema.json
ericcornelissen: Definitely agree it should be documented somewhere (*) but to be frank we haven't really come to a conclusion yet as to what the renaming strategy is complete - specifically how long old names will stay around. So I think it's safe to omit in this PR - as indeed it is mostly just maintainers doing renames - and document it when we agree on a full strategy.

(*): my vote would be to keep it public, just because most everything can and should be public, but I agree that we can choose not to put it here as it is not relevant to most contributors.

.jsonschema.json
ericcornelissen: Can you elaborate? Do you mean the aliases should come after the license in the data file? If so, I agree but the ordering here does not enforce an order in the data file. I put it before the licenses here because this schema shorter than the licences schema.

CONTRIBUTING.md
ericcornelissen: Indeed, alternative could be to use `js` instead of `json5` but I would say that is even more confusing.

CONTRIBUTING.md
ericcornelissen: I'd be in favor of keeping one document, this keeps it easily searchable with <kbd>Ctrl</kbd>+<kbd>F</kbd>.

I'm fine with restructuring but as I said before I think that will create a lot of discussion so would prefer to do that seperately.

.jsonschema.json
ericcornelissen: [You brought up that we didn't need it](https://github.com/simple-icons/simple-icons/issues/4797#issuecomment-772705615)  and [I agreed](https://github.com/simple-icons/simple-icons/issues/4797#issuecomment-772829551).

Thinking about it again, though I think it is unlikely the duplicates hex is a different source, it's certainly possible so probably better to put it in - at the risk of having duplicate sources (which we _could_ lint for, but that seems very low priority to me :thinking:).

CONTRIBUTING.md
mondeja: Yes, it would be very strange. Lets keep as is and, if we find that edge case, we can document here later.

.jsonschema.json
PeterShaggyNoble: Ah, OK, I always thought this was where the ordering was enforced. Ignore that, so! 😆

PeterShaggyNoble:
Looks like we _should_ be ready to go, so, once the conflicts are resolved.

mondeja (approved):

ericcornelissen:
> Looks like we _should_ be ready to go, so, once the conflicts are resolved.

Updated :+1: Should be all good now, but just to be sure I'll let someone else look at it one more time and merge :slightly_smiling_face:

service-paradis (approved):

PeterShaggyNoble:
OK, let's go for it 🙂 Thanks again, Eric 👍🏻

PR #5370, opened 2021-03-28
Update NPM test suite

ericcornelissen:
**Issue:** n/a
**Alexa rank:** n/a
  

### Checklist

### Description

This PR updates our existing test suites, improving the testing conditions, and adding test cases for the guidelines and licenses.

We could further improve by 1) testing the SVG value which requires reading the `.svg` file during tests and 2) "abstracting" the common tests away. I opted not as 1 is not clearly beneficial and I'm not 100% sure what the best way to do 2 is. If anyone thinks either or both of these should be done in this PR, I'm open to doing so.


tests/icons.test.js
```diff
@@ -5,28 +5,54 @@ icons.forEach(icon => {
...
+  });
+
+  test(`${icon.title} has ${icon.license ? "the correct" : "no"} "license"`, () => {
+    if (icon.license) {
+      expect(typeof subject.license).toBe('object');
+      expect(subject.license).toHaveProperty('type', icon.license.type);
+      if (icon.license.type === "custom") {
+        expect(subject.license).toHaveProperty('url', icon.license.url);
+      } else {
+        expect(typeof subject.license.url).toBe('string');
```
mondeja, line 52: Why not adding equality comparison for predetermined SPDX license urls?

```suggestion
        expect(typeof subject.license.url).toBe('string');
        expect(subject.license.url).toEqual(`https://spdx.org/licenses/${icon.license.type}.html`);
```

tests/index.test.js
```diff
@@ -6,30 +6,55 @@ icons.forEach(icon => {
...
+  });
+
+  test(`${icon.title} has ${icon.license ? "the correct" : "no"} "license"`, () => {
+    if (icon.license) {
+      expect(typeof subject.license).toBe('object');
+      expect(subject.license).toHaveProperty('type', icon.license.type);
+      if (icon.license.type === "custom") {
+        expect(subject.license).toHaveProperty('url', icon.license.url);
+      } else {
+        expect(typeof subject.license.url).toBe('string');
```
mondeja, line 53: Also here?

```suggestion
        expect(typeof subject.license.url).toBe('string');
        expect(subject.license.url).toEqual(`https://spdx.org/licenses/${icon.license.type}.html`);
```

mondeja (commented):
Good work @ericcornelissen! I only have some minor suggestions...

tests/index.test.js
```diff
@@ -6,30 +6,55 @@ icons.forEach(icon => {
...
+  test(`${icon.title} has a valid "path" value`, () => {
     expect(typeof subject.path).toBe('string');
     expect(subject.path).toMatch(/^[MmZzLlHhVvCcSsQqTtAaEe0-9-,.\s]+$/g);
   });
 
-  test(`${icon.title} has a "slug"`, () => {
-    expect(typeof subject.slug).toBe('string');
+  test(`${icon.title} has ${icon.guidelines ? "the correct" : "no"} "guidelines"`, () => {
+    if (icon.guidelines) {
+      expect(typeof subject.guidelines).toBe('string');
```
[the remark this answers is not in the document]
ericcornelissen: Missed that, good catch :+1:

tests/icons.test.js
ericcornelissen: I'm fine adding it. I initially didn't because it is literally copying the source we're trying to test. That said it's not a bad idea to test it for regression.

A compromise could be to test if it is _a_ URL (i.e. matches a URL pattern) instead. This way we don't need to update the test if we/SPDX decides to change these URLs.

What do you think @mondeja?

mondeja: I agree, a regex would also avoid regressions here, so as you prefer @ericcornelissen

ericcornelissen: Added this in 9e71e9674dd45c328d2284fedf7ae9c32708d0b2 :+1: Should be all good now

mondeja (approved):
Perfect :tada:

PR #5388, opened 2021-03-31
Letter Q review

service-paradis:
**Issue:** #5251

### Notes

- [x] I think we might need to update https://www.qualtrics.com/brand-book/, looks like we need to use the full wordmark here - #5657
- [x] I think we might need to update https://www.quantcast.com/, looks like we need to use the full wordmark `QC` as in favicon
- [x] We need to remove https://www.quantopian.com/about - #5658
  - https://en.wikipedia.org/wiki/Quantopian
  - https://quantopian-archive.netlify.app/forum/threads/quantopians-community-services-are-closing.html

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -6737,7 +6737,8 @@
         {
             "title": "Qt",
             "hex": "41CD52",
-            "source": "https://qt-brandbook.webflow.io/design"
+            "source": "https://qt-brandbook.webflow.io/design",
+            "guidelines": "https://qt-brandbook.webflow.io/design"
         },
         {
             "title": "Qualcomm",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -6715,7 +6715,8 @@
         {
             "title": "Qi",
             "hex": "000000",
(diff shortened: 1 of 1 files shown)


adamrusted (approved):
> * I think we might need to update [qualtrics.com/brand-book](https://www.qualtrics.com/brand-book/), looks like we need to use the full wordmark here

There's a line on the brandbook page which states:
`Do not use the logo in any other color except the approved color version above`
While they do  provide a monochrome version (white in the ZIP download) - we may need to be careful with this one.

> * I think we might need to update [quantcast.com](https://www.quantcast.com/), looks like we need to use the full wordmark `QC` as in favicon

Yep. All their socials use the full wordmark. Not sure on the hex, they use a lot of blue shades on their website but the social pictures all use black text on white background.

> * We need to remove [quantopian.com/about](https://www.quantopian.com/about)

Yep - officially defunct from 2020 and the domain now doesn't work. I'd say that's good enough for removal.

I'll convert your points above into a checklist, so we can track who's working on them, and at least get PRs in before we merge this.

PeterShaggyNoble:
> While they do provide a monochrome version
I think that's good enough for us, so, as it comes from an official source and we're including the guidelines to cover ourselves further.

As for Quantcast, although not a strong argument, they do still use the "Q" as the favicon on [their support site](https://help.quantcast.com/).

PR #5396, opened 2021-04-01
Icon update, GNU Emacs: Remove circle from around icon

tecosaur:
Looking at the current icon for GNU Emacs in simple icons, and the actual icon, I find that the current design gives border of the icon undue visual emphasis. It also shrinks the size of the primary component of the design, the styled "E", making the icon harder to recognise at a glance at small sizes.

Reference icon:
image

Current icon:
image

I think a much fairer representation is obtained simply by not having the circle.

### Proposed

gnuemacs

### Current

gnuemacs


tecosaur:
Original PR to add the icon: #2248

adamrusted:
IMO, because the circle (with border) is used wherever the icon is displayed, we need to keep some form of that. Especially in the absence of any official brand guidelines that state we can remove the circle. Looking back at the original PR, the options were a full black circle with the `E` cut out, or the outline with the `E`. I think we either need to look at the full black circle, or an gap between the 'border' and the inner circle, with the `E` cut out (see below for example). Any other @simple-icons/maintainers want to chip in on this?

gnuemacs

tecosaur:
I made this PR because I thought it warranted some discussion, but I don't have particularly strong feelings on it.
However, going a bit beyond that single reference image, and see what's used to represent Emacs now and in the past, the focus is very much on the "E" not the circle.

As the logo has changed over the years the form has changed but the "E" has remained constant. For example see the Emacs 22 logo:

emacs22

Or the original logo the "E" was extracted from:

image

Or other Emacs-related material:

image

The impression I have from looking around at a lot of "Emacs stuff" is that the Emacs "E" is the single vital component of the icon / brand identity, and while the homepage has put a circle around it that is very much a secondary concern. As at such small sizes and when in an inline badge the circle reduces how much the "E" stands out, I'm thinking just the "E" would be best.

However, I am not the one who designed the icon. Perhaps @NicolasPetton might have something to add? He designed the most recent iteration of the Emacs logo.

PeterShaggyNoble:
Yeah, we definitely need to keep the circle in this one. Either filled or outlined works for me - I don't think a cut-out is necessary here - but if I were to pick one, I'd go with our current, outlined treatment on the basis of the difference in contrast between the circle and its border in the original. If the border were a few shades lighter then I'd probably opt for the filled circle.

service-paradis:
I agree that we need to keep the circle here. I'm voting to keep our current icon.

tecosaur:
Easy, looks like there's a strong leaning to keep thing as-is.
I can't say I'm a huge fan of as-is, but as I said before: this is something I raised because I think it warranted discussion, not because I have strong feelings about it.

PeterShaggyNoble:
Never any harm in a healthy discussion, @tecosaur 🙂 Thanks for raising it.

One thing we _do_ need to do is to add a `license` entry for this icon. From the source file, it's licensed under the GNU General Public License and from [this page](https://ftp.gnu.org/old-gnu/Manuals/emacs-20.7/html_chapter/emacs_3.html) it looks like it should be v2 or later so the value of the entry from [SPDX](https://spdx.org/licenses/) should be `GPL-2.0-or-later`. Do you want to revert the changes to the SVG and take of that in this PR or will we close this one and open a new PR for it?

tecosaur:
Given this PR is very much about the icon design, I think if we're settled on leaving that as is but updating the licence it would be best to close this PR and deal with that seperately.

PeterShaggyNoble:
👍🏻 Fair enough. I'll close this one so.

PR #5399, opened 2021-04-01
add storyblok icon

stefanobartoletti:
storyblok

**Issue:**
**Alexa rank:**  28,521 
  

### Checklist

### Description

Icon and color picked from the official logo from their website


_data/simple-icons.json
```diff
@@ -7996,6 +7996,11 @@
             "hex": "3A98D9",
             "source": "https://storify.com"
         },
+        {
+            "title": "Storyblok",
+            "hex": "0AB3AF",
+            "source": "https://www.storyblok.com/"
```
adamrusted, line 8002: The source should probably be https://www.storyblok.com/press. Within that ZIP I used `colored-standalone.svg` for comparison. I'd also suggest we add the link to the 'guidelines' entry as there's the line about requesting permission to use it.

_data/simple-icons.json
```diff
@@ -7996,6 +7996,11 @@
             "hex": "3A98D9",
             "source": "https://storify.com"
         },
+        {
+            "title": "Storyblok",
+            "hex": "0AB3AF",
```
adamrusted, line 8001: The hex in the `colored-standalone.svg` is `#09B3AF` - should we change to that?

adamrusted (changes requested):
Thanks @stefanobartoletti - I've made a couple of comments on your JSON file, and it appears there's some [differences](https://user-images.githubusercontent.com/1830867/113474190-50e9b280-9466-11eb-9477-2cbb86a0434f.png) between your PR in red, and the source icon in blue. Would you mind updating the source and using that one?

stefanobartoletti:
I've updated the icon according to your suggestions, I was not aware of that "Press" page.

_data/simple-icons.json
stefanobartoletti: Updated

_data/simple-icons.json
stefanobartoletti: Updated. Not sure about what you are exactly asking with the link to the guidelines entry. What link should I use as a source?

adamrusted: So you can add the press page as both the 'source' and 'guidelines' entry in the JSON, as below:

```
{
	"title": "Storyblok",
	"hex": "0AB3AF",
	"source": "https://www.storyblok.com/press",
	"guidelines": "https://www.storyblok.com/press"
}
```

It doesn't matter about the duplication - so don't worry about that. Just as I mentioned above, with them putting the line about requesting permission to use the icon - I think we're safest to include it.

stefanobartoletti: Ok, I understand. Fixed, BTW.

adamrusted (commented):
Looks good to me now @stefanobartoletti - thanks for working on this! 💪
There are a few points in the curves of the 'B' which you may be able to remove, if you fancy giving that a go? Otherwise I'd be happy to merge this.

stefanobartoletti:
Removing these points changes the shape of the curves, making the icon slightly different from the original.

adamrusted:
No worries if it alters the shape of the curves - just at a quick glance I thought they might be removable without too much hassle. I'll merge for now, and if anyone wants to try and optimize further they can give it a shot! Thanks for your work @stefanobartoletti 💪

PR #5400, opened 2021-04-01
Purgecss

stefanobartoletti:
purgecss

**Issue:**
**GitHub Stars:** 6k +

### Checklist

### Description

I vectorized the PNG logo from their website, where I took the color as well


adamrusted (changes requested):
Thanks for your work on this @stefanobartoletti!
I've been able to find [this SVG](https://github.com/FullHuman/purgecss/blob/89ece42929a3fd62b2a84fb0b57c0ac16c98908f/docs/.vuepress/public/safari-pinned-tab.svg) from within their repo, which seems to be our best shot at an SVG logo. Would you mind updating the source to that, and using the SVG as your base?

stefanobartoletti:
Done, the file itself is slightly bigger in size, because of some slightly different points from the one I vectorized myself

PeterShaggyNoble:
Thanks for the update, @stefanobartoletti. As this is based on a Potraced SVG, which butchers the source file to begin with, I think we can let the differences between your redrawn circle and the circle in the source slide - it's clearly meant to be a perfect circle. There is [a difference](https://user-images.githubusercontent.com/15157491/114727051-c9d5ed80-9d35-11eb-854b-41f554ff75b6.png) running up the bottom left of the "P" that we should probably try to fix, though. Could you also see if you could clean up some of the excess points.

We'll also need to update the source URL to the [GitHub commit URL](https://github.com/FullHuman/purgecss/blob/4e2bf58e218119cc9faf9faa615d62a059bf9d9a/docs/.vuepress/public/safari-pinned-tab.svg).

stefanobartoletti:
I tried to further optimize this icon according to your suggestions. This is the best I can do on this.

PeterShaggyNoble:
Nice one, @stefanobartoletti 👍🏻 Given the poor quality of the source SVG and original PNG, this looks good enough to me. How about you, @adamrusted?

The only other change needed is to update the source URL to https://github.com/FullHuman/purgecss/blob/4e2bf58e218119cc9faf9faa615d62a059bf9d9a/docs/.vuepress/public/safari-pinned-tab.svg

adamrusted:
Yep, looks good to me if @stefanobartoletti could update the source URL we can get this merged. Thanks for your work, and to @PeterShaggyNoble for looking at this also! 💪

adamrusted (approved):
Thanks for the update @stefanobartoletti - let's get this merged in ahead of this weekends release! 🚀

PR #5412, opened 2021-04-05
Remove '.html' extension for SPDX license urls

mondeja:
SPDX license pages can be accesed without `.html` extensions as you can see in [their repository `website/` folder](https://github.com/spdx/license-list-data/tree/master/website), so we can remove the extension building the URL.

### DIFF
--- a/scripts/build-package.js
+++ b/scripts/build-package.js
@@ -45,7 +45,7 @@ function licenseToObject(license) {
   }
 
   if (license.url === undefined) {
-    license.url = `https://spdx.org/licenses/${license.type}.html`;
+    license.url = `https://spdx.org/licenses/${license.type}`;
   }
   return license;
 }


ericcornelissen (commented):
I'm fine with this change if other @simple-icons/maintainers agree.

To clarify, I initially included the `.html` suffix just because that's what my browser showed me. That said, one argument in favour of keeping `.html` is to make it explicit that we're linking to a web page, given that the "same" URL can also be used to link to e.g. JSON data about the license. To me, that could be reason enough to keep `.html`, but as per my first sentence I'm fine removing it if the majority prefers it.

For the record, the only argument I can think of in favour of removing `.html` would be that it is unnecessary and removing it reduces the package size.

PeterShaggyNoble:
I think we should do this, too, as it will save us a good few bytes in the package as we add more & more licenses.

I'm not too concerned about explicitly informing people we're taking them to a web page through the extension. I think the primary concern any who's following the link has is what the license says, not what the file type is. Also, if it's _not_ a HTML file, it's much more likely to include the file extension, anyway.

PeterShaggyNoble:
Merging due to the lack of any other objections; we can always revert to the `.html` extension in the future if needs be.

PR #5463, opened 2021-04-15
Update Kaggle icon

### LINKED ISSUE #5335: Update Kaggle
image

**Brand Name:** Kaggle

- https://www.kaggle.com/brand-guidelines

Following the review of some icons we have while contributing to #5251 (#5279), I found the above guidelines. According to those guidelines, I think we need to use the full wordmark instead of the letter `k`.

service-paradis:
image

**Issue:** Close #5335

### Checklist

### Description

- I took the transparent SVG from https://www.kaggle.com/brand-guidelines
- I used Inkscape to scale


PeterShaggyNoble:
Could we get away without the "TM" here? They don't make explicit mention of its inclusion that I've seen.

I'd also like to make a case of us keeping just the "k"; they use it pretty extensively themselves in their favicons, within their software and on social media. Then again, the line below may make it a case of "do as we say, not as we do" although the counter argument we could make is that we're representing the icon, not the logo.

> Always use the above resources to represent the logo

service-paradis:
> Could we get away without the "TM" here? They don't make explicit mention of its inclusion that I've seen.

The following sentence make me think we have to, unfortunately, keep the TM. Looks like they provide 2 versions of the logo that we should use as-is.

> Always use the above resources to represent the logo as opposed to an image of the logo sourced elsewhere, and keep a minimum clearance of space between the logo and other visual elements akin to the images above.

Same for the "k". We would need to source the icon elsewhere.

If we want to use a modified version of the given logo, we would need to ask Google permission to do so according to https://about.google/brand-resource-center/brand-terms/.

adamrusted:
Before I review the path on this one, I _kind of_ agree with @PeterShaggyNoble on questioning the inclusion of the ™ symbol. I know we're having to be very careful around the inclusion of ® marks, but my understanding was that ™ could be omitted if there wasn't an express mention of it's inclusion, and while the terms listed on that source page do mention the 'trademark' a lot - there's nothing explicit saying we _can't_ remove the mark - especially given our native sizing of 24*24 whereby the trademark symbol would be nigh on impossible to see! Pinging other @simple-icons/maintainers for feedback on this before we proceed with a review.

PeterShaggyNoble:
Bsed on the 👍🏻s on Adm's comment above, it looks like we have a consensus. @service-paradis, could you update the SVG in the PR to remove the "TM"?

service-paradis:
Sorry for the past few weeks inactivity, I will try to get back in the game soon :)

I started from scratch, removed the ™ symbol, scaled and optimized! Let me know if there is something!

image

adamrusted:
Thanks for this, @service-paradis - though I'm seeing some small differences in the `g` shapes (see [here](https://user-images.githubusercontent.com/1830867/120763462-23c18b00-c50f-11eb-84d9-81d369a812fc.png)), which look to be just from optimization. Could you try with another decimal of precision?

service-paradis:
> Thanks for this, @service-paradis - though I'm seeing some small differences in the `g` shapes (see [here](https://user-images.githubusercontent.com/1830867/120763462-23c18b00-c50f-11eb-84d9-81d369a812fc.png)), which look to be just from optimization. Could you try with another decimal of precision?

Thank you for the review! Tried too much to optimize this one. Unfortunately, I need the precision of 4 and was not able to remove duplicate points inside the `g` letters without messing with the shape!

adamrusted (approved):
That's matching up perfectly for me now, thanks @service-paradis! 🚀

PR #5398, opened 2021-04-01
Letter P review

service-paradis:
**Issue:** #5251

### Notes

- [x] I think we might need to update https://www.paypal.com/, logo looks different than the current one
- [x] We need to remove https://www.periscope.tv/
  - https://periscope.medium.com/farewell-periscope-164db2742b7c
  - https://help.twitter.com/fr/using-twitter/periscope-faq
- [x] We need to remove http://pingup.com/
- [x] We need to update https://maps.google.com/pluscodes/
- [x] I think we might need to update https://brand.plex.tv/d/qxmJ3odkK0fj/style-guide, looks like we need to use the full wordmark - #5560
- [x] I think we might need to update https://wiki.postgresql.org/wiki/Logo to use the given monochrome version - #5561
- [x] I think we might need to update https://powerbi.microsoft.com/en-us/
  - https://powerbi.microsoft.com/pictures/application-logos/svg/powerbi.svg
  - https://en.wikipedia.org/wiki/Microsoft_Power_BI#/media/File:Power_BI_logo.svg
- [x] ~I think we might need to update https://publons.com/about/the-publons-logo~

PR #5467, opened 2021-04-16
Updated PayPal

adamrusted:
paypal

**Issue:** Contributes to #5398.
**Alexa rank:**
  

### Checklist

### Description
Updated source to PayPal homepage. Used SVG at source and cut gap between the two paths.


PeterShaggyNoble:
Treatment looks good to me, @adamrusted. [Just one small difference](https://user-images.githubusercontent.com/15157491/115383824-32a5e580-a1ce-11eb-9e33-ff1093522fbb.png) in the bottom "P", at the top of the cut-out which should be fixable by dragging the corner point in yours up to meet the curve of the original.

Also, looking at the products shots [here](https://www.paypal.com/uk/webapps/mpp/how-to-use-paypal-here/card-reader) it looks to me like the cut-out should be _outside_ rather than inside the bottom "P". What do you think?

adamrusted:
> [Just one small difference](https://user-images.githubusercontent.com/15157491/115383824-32a5e580-a1ce-11eb-9e33-ff1093522fbb.png) in the bottom "P", at the top of the cut-out

Will fix this now 👍

> Also, looking at the products shots [here](https://www.paypal.com/uk/webapps/mpp/how-to-use-paypal-here/card-reader) it looks to me like the cut-out should be _outside_ rather than inside the bottom "P". What do you think?

I think they've got a stylized monochrome logo that they're not sharing! Only way I could recreate that using the current SVG source would be to carry on the cut-out to the bottom of the bottom "P" - but that may look a bit odd.

PeterShaggyNoble (approved):
That looks perfect to me now, @adamrusted, nice one 👍🏻

I was so focused on the cut-out in the version one their card reader I hadn't noticed the other differences in it so, yeah, it looks like it's been stylised specifically for that purpose - just eyeballing it from the photos, it looks like the top "P" has been moved slightly further away from the bottom one.

PR #5483, opened 2021-04-18
Letter J review

service-paradis:
**Issue:** #5251

### Notes

- [ ] I think we might need to update https://www.jetlocal.co.uk/, logo looks different than the current one

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4251,7 +4251,8 @@
         {
             "title": "Jeep",
             "hex": "000000",
-            "source": "http://www.fcaci.com/x/JEEPv15"
+            "source": "http://www.fcaci.com/x/JEEPv15",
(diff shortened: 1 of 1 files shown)


adamrusted:
@service-paradis - our current icon matches up to https://www.jet-tankstellen.de/, which is the German one linked as our current source. There is _also_ JET Local, which is in the UK - though if we want to add that later down the line, we can add it as `jetlocal` as opposed to `JET`. Closing this one out for now - but feel free to open an issue for JET Local if you want.

service-paradis:
> @service-paradis - our current icon matches up to https://www.jet-tankstellen.de/, which is the German one linked as our current source. There is _also_ JET Local, which is in the UK - though if we want to add that later down the line, we can add it as `jetlocal` as opposed to `JET`. Closing this one out for now - but feel free to open an issue for JET Local if you want.

You're right, thanks!!

PR #5491, opened 2021-04-20
Update Zotero

### LINKED ISSUE #5485: On the Zotero icon
**Brand Name:** Zotero
**Official resources for icon and color:** https://www.zotero.org/support/brand

I have seen issue #2175 and PR #4561. First of all thanks for the work. I would like to ask for trying to reproduce the colours as in the original icon, so the `z` is red filled, and not just its edges, and white inside. As it is currently, it is almost invisible on small images (as using it for https://shields.io). Otherwise, I believe it would be preferable to use simply the `z` as already suggested in the comment https://github.com/simple-icons/simple-icons/issues/2175#issuecomment-570156747

adamrusted:
zotero

**Issue:** Fixes #5485
**Alexa rank:** [15,441](https://www.alexa.com/siteinfo/zotero.org)

### Checklist

### Description
Nice and easy edit. Took the `Z` from [here](https://www.zotero.org/support/lib/exe/fetch.php?tok=a60b55&media=https%3A%2F%2Fwww.zotero.org%2Fstatic%2Fimages%2Fpromote%2Fzotero-logo-512x123.svg) and resized to suit our needs. Hex code was already correct, as was the source URL.


PeterShaggyNoble (approved):
I was going to say that when used as an icon the "Z" should be thicker as seen [here](https://www.zotero.org/support/brand). However, that page is no longer listed in [the contents of their documentation](https://www.zotero.org/support/) so it may be outdated. Their use of the thinner "Z" in their favicon may back that up, too.

That said, this looks spot on to me. Thanks, @adamrusted 👍🏻

If anyone can find anything current that suggests we _should_ use the thicker version then we can address that at a later date.

PR #5496, opened 2021-04-21
Letter I review

service-paradis:
**Issue:** #5251

### Notes

- [x] We need to remove https://www.ideal.nl/cms/files/Manual_iDEAL_logo.pdf
- [ ] I think we need to update https://www.informatica.com/ to add gaps
- [x] I think we need to update https://inkscape.org/gallery/=inkscape-branding/inkscape-brand-assets/. The monochrome version available is different

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4084,7 +4084,7 @@
         {
             "title": "Informatica",
             "hex": "FF4D00",
-            "source": "https://www.informatica.com/content/dam/informatica-com/en/images/cc02v2/logo-informatica.svg"
+            "source": "https://www.informatica.com/"
         },
         {
             "title": "Infosys",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4155,7 +4155,7 @@
         {
             "title": "Internet Archive",
             "hex": "000000",
-            "source": "https://openlibrary.org/static/images/ia-logo.svg"
+            "source": "https://archive.org/"
         },
         {
             "title": "Internet Explorer",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4023,7 +4023,7 @@
         {
             "title": "iFixit",
             "hex": "0071CE",
-            "source": "https://www.ifixit.com/"
+            "source": "https://www.ifixit.com/Info/Media"
         },
         {
             "title": "iFood",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4099,7 +4099,10 @@
         {
             "title": "Inkscape",
             "hex": "000000",
-            "source": "https://commons.wikimedia.org/wiki/File:Inkscape_Logo.svg"
+            "source": "https://inkscape.org/gallery/=inkscape-branding/inkscape-brand-assets/",
+            "license": {
+                "type": "CC-BY-SA-3.0"
+            }
         },
         {
             "title": "Insomnia",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4023,7 +4023,8 @@
         {
             "title": "iFixit",
             "hex": "0071CE",
-            "source": "https://www.ifixit.com/Info/Media"
+            "source": "https://www.ifixit.com/",
+            "guidelines": "https://www.ifixit.com/Info/Media"
         },
         {
             "title": "iFood",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4171,7 +4171,7 @@
         {
             "title": "Invoice Ninja",
             "hex": "000000",
-            "source": "https://github.com/invoiceninja/invoiceninja/blob/aefe133ba02f28d9fd585ae731765ad5b7cdffe3/public/safari-pinned-tab.svg"
+            "source": "https://github.com/invoiceninja/invoiceninja/blob/2bdb26dd06123a0426cc7a8da77fc8fce7e5a222/public/images/round_logo.png"
         },
         {
             "title": "ioBroker",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4149,7 +4149,8 @@
         {
             "title": "Intercom",
             "hex": "6AFDEF",
-            "source": "https://www.intercom.com/press"
+            "source": "https://www.intercom.com/press",
+            "guidelines": "https://www.intercom.com/press"
         },
         {
             "title": "Internet Archive",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4171,7 +4171,7 @@
         {
             "title": "Invoice Ninja",
             "hex": "000000",
-            "source": "https://github.com/invoiceninja/invoiceninja"
+            "source": "https://github.com/invoiceninja/invoiceninja/blob/aefe133ba02f28d9fd585ae731765ad5b7cdffe3/public/safari-pinned-tab.svg"
         },
         {
             "title": "ioBroker",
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble:
> I think we need to update https://www.informatica.com/ to add gaps

I don't know that we'd need gaps between the various shades of orange but I do think we should do something to provide some separation between the purple shape and the rest

> I think we need to update https://inkscape.org/gallery/=inkscape-branding/inkscape-brand-assets/. The monochrome version available is different

Agreed.

To review the first few changes in this PR:
- I think we should go with https://www.ibm.com/design/language/ibm-logos/8-bar/ for the IBM guidelines as it includes details on the usage of the specific logo we've used?
- Looks like we need to update the IBM colour too; I don't see any mentioned of the blue we currently use on [this page](https://www.ibm.com/design/language/color/). _Which_ blue we should use, though, I'm not sure as they don't seem to state a clear preference and they themselves use a number of blues not present in their palette. Keeping it consistent with the shade birjolaxew picked in #1768, though, we could go with the current "Blue 70" of `#0043ce`.
- Do we need to update the IBM SVG to include the trademark symbol to be safe.
- Agree with that change to the ICON `source` 👍🏻
- We'll need to update our ICON icon, though, as their are [differences](https://user-images.githubusercontent.com/15157491/115686457-1e3e2600-a351-11eb-8c5c-06c7439d7ac2.png) between it (in red) and the vector files available for download.
- Agree with the addition of the iDEAL `guidelines`, even if it is scheduled for removal 👍🏻
- As the new page doesn't include any vector artwork, I'd leave the iFixit `source` as is and add the new page for the `guidelines` instead as it includes a line about requesting permission before using their logo.
- Agree with the addition of the `guidelines` entry for IFTTT & iHeartRadio 👍🏻
- Agree with the changes to IMDb 👍🏻
- Should we update our IMDb icon to use the outline version instead? If not then we should at least double check that our current version is consistent with what's available to download at the new source
- Agree with the changes to Imgur, even though the icon isn't directly available at the new source 👍🏻

I'll be back to review the rest later, if someone else doesn't beat me to it.

PeterShaggyNoble:
Continuing on ...
- Looks like we need to update the Indeed `hex` to `#003a9b`
- Agree with change to the Indeed `source` 👍🏻
- I don't think we need to add the `guidelines` entry for Indeed as it doesn't provide any information on _how_ the icon should be used
- Looks like we'll need to update our Indeed icon, some [major differences](https://user-images.githubusercontent.com/15157491/115703412-0327e200-a362-11eb-8795-ef17f26a3ef3.png) between it (in red) and the new source files
- Should we update the InfluxData `hex` to the "Deep Purple" seen [here](https://influxdata.github.io/branding/visual/colors/)? It looks to be pretty widely used on [their website](https://www.influxdata.com/).
- Agree with the change to the InfluxData 👍🏻
- Some [minor differences](https://user-images.githubusercontent.com/15157491/115705713-c5788880-a364-11eb-945e-512940648ec0.png) between our InfluxData icon (in red) and those available at the new source so we'll need to update that
- Agree with the changes to Informatica & Infosys :+1:
- We need to update our Infosys icon to fix the differences between it and the new source AI file ([example](https://user-images.githubusercontent.com/15157491/115706719-dd044100-a365-11eb-9209-78ef65c7765c.png))
-  Agree with the changes to Inkscape, Instagram, Intel, Intercom & Internet Archive 👍🏻 (Although Intel don't (yet) have vector files available to download, the logo in the header is still an SVG)
- Our Internet Archive icon needs to be updated based on the new source - [differences]https://user-images.githubusercontent.com/15157491/115707541-d629fe00-a366-11eb-867d-8cf6d8bc639e.png), with ours in red
- The `hex` for Internet Archive probably needs to be update to one of the greys

I'll get the last few done later on today.

service-paradis:
Thank you @PeterShaggyNoble for the review, really appreciated!

> I think we should go with https://www.ibm.com/design/language/ibm-logos/8-bar/ for the IBM guidelines as it includes details on the usage of the specific logo we've used?

You're right, done!

> Looks like we need to update the IBM colour too; I don't see any mentioned of the blue we currently use on this page. Which blue we should use, though, I'm not sure as they don't seem to state a clear preference and they themselves use a number of blues not present in their palette. Keeping it consistent with the shade birjolaxew picked in #1768, though, we could go with the current "Blue 70" of #0043ce.

You're right. I had it noted, but forgot to add it as a note in my PR. Thanks for catching it! `Blue 70` is probably a good choice. According to the IBM [developper guidelines](https://www.ibm.com/brand/systems/developer/brand/color/), `Blue 60` might be a good choice too.

`Blue 60` is also the second color used in their main carousel - `#0530ad` and `#0f62fe` (`Blue 60`):
Capture d’écran 2021-04-22 084807

> Do we need to update the IBM SVG to include the trademark symbol to be safe.

[This section](https://www.ibm.com/design/language/ibm-logos/8-bar/#legal-requirements) mention the trademark symbol, but I thought it is for "printed literature" only... Not too sure...

> As the new page doesn't include any vector artwork, I'd leave the iFixit source as is and add the new page for the guidelines instead as it includes a line about requesting permission before using their logo.

👍 

> - Looks like we need to update the Indeed hex to `#003a9b`
> - I don't think we need to add the guidelines entry for Indeed as it doesn't provide any information on how the icon should be used

Indeed, done!

> Should we update the InfluxData hex to the "Deep Purple" seen here? It looks to be pretty widely used on their website.

It is, for sure, the primary color of InfluxData. InfluxDB tend to use blue through their interface.
image

> The hex for Internet Archive probably needs to be update to one of the greys

Agreed, I updated it to use `#666666` used as search backgrond in header and for cards background. What do you think?

PeterShaggyNoble:
> > Do we need to update the IBM SVG to include the trademark symbol to be safe.

> This section mention the trademark symbol, but I thought it is for "printed literature" only... Not too sure...

They mention videos as well so it would seem to apply to physical _and_ digital media. And while it mentions it only being required in the US, I think the language for the ret of the world is ambiguous enough for us to invoke our own consensus on its inclusion.

Anyway, on to the last few - didn't realise there were so few left or I would have kept going! -  and then I'll take another pass over it:
- Our Invoice Ninja icon was actually manually vectorised from [this file](https://github.com/invoiceninja/invoiceninja/blob/2bdb26dd06123a0426cc7a8da77fc8fce7e5a222/public/images/round_logo.png) by phatbhoy67 in #1948 so we should probably use that for the source URL. The SVG source is poorly Potraced (although I've seen worse) from that PNG.
- Agree with the change to ioBroker; that's the very file I based our version off 👍🏻
- Agree with the changes to Issuu & Itch.io 👍🏻
- I think we should leave the IVECO source as-is as the new one doesn't include a vector logo. Although the file @runxel based our version off is no longer present so we should double check that ours still lines up with [this new SVG](https://www.iveco.com/Common/Style%20Library/RWDIveco/images/myiveco/myiveco.svg) found at the same source.

service-paradis:
> * Our Invoice Ninja icon was actually manually vectorised from [this file](https://github.com/invoiceninja/invoiceninja/blob/2bdb26dd06123a0426cc7a8da77fc8fce7e5a222/public/images/round_logo.png) by phatbhoy67 in #1948 so we should probably use that for the source URL. The SVG source is poorly Potraced (although I've seen worse) from that PNG.

Alright, done!

> * I think we should leave the IVECO source as-is as the new one doesn't include a vector logo. Although the file @runxel based our version off is no longer present so we should double check that ours still lines up with [this new SVG](https://www.iveco.com/Common/Style%20Library/RWDIveco/images/myiveco/myiveco.svg) found at the same source.

Weird, was not able to access this page 2 days ago 🤔 Reverted!

PeterShaggyNoble (approved):
All looks good to me, @service-paradis, nice work 👍🏻

If anyone wants to propose alternative colours for IBM or Internet Archive or feel we should update the colour for InfluxDB, we can address those separately.

PR #5520, opened 2021-04-22
Add IOTA icon

Mrcuve0:
**Issue:** 5516
**Alexa rank:** 48,669 (see https://www.alexa.com/siteinfo/)
  

### Checklist

### Description

The original logo can be downloaded https://files.iota.org/media/IOTA_Logo.zip
The main hex value from https://www.iota.org/connect/brand
I followed the guidelines, all the tests passed

iota


PeterShaggyNoble:
Welcome to Simple Icons, @Mrcuve0 and thanks for the contribution 🙂

I think we can get away with using just the swirl here as [they make it available](https://files.iota.org/media/IOTA_Mark.zip) in the second slide of the [source page](https://www.iota.org/connect/brand). Could you update to that, please?

For the source, you can use https://www.iota.org/connect/brand as well, rather than linking directly to the zip.

You can also omit the `url` entry from the `license` entry; that will be generated automatically on our end - the `url` entry is only needed for "custom" `type` licenses.

Colour's a bit tricky, based on their guidelines alone, I would agree with your choice of the green but, from their website, it looks to me like the "grays" are the primary colours and the greens the accent colours, what do others think?

PeterShaggyNoble:
I'd vote for "Midnight" myself so, given that, you can hold off on updating the colour until we get a few more opinions, if you like, or just run with your choice of "10pm" for now.

Mrcuve0:
Updates w/ the swirl and "midnight" theme. I also tested the preview with "10pm" but readability is a mess with that color.

* Midnight:
iota_midnight

* 10pm
iota_10pm

* Main Green
iota_main_green

* Dark Green
iota_dark_green

PeterShaggyNoble:
Thanks for the update, @Mrcuve0. However, I'm noticing some _hairline_ thin differences in the right edges of a few of the circles - see example below of the 3rd circle in the top row of the left-hand "wave" with your version in red.

There could be a couple of causes for this:
- There could be subtle differences between the 3 source files and I'm not comparing to to the same on you based your icon on - I'm using `IOTA Mark Black.svg`.
- The circles are shifting or distorting slightly when combining them all into a single path - [as happened to me](https://user-images.githubusercontent.com/15157491/115868288-2327d600-a434-11eb-9716-4559c0c2df57.png) when I tried unifying them in Inkscape, combining them produced better results.
- The precision need to be set a little higher during the optimisation process.

On the colour, we don't let how things look in our own use cases influence us*, we go with the primary brand colour or what we can best determine to be the primary brand colour, which isn't clear here. But let's hold off on any further updates to the colour for now, until we get the opinion of a few others. *We're actually in the process of building a new website, changing the layout and design so that icons are no longer sitting on coloured backgrounds for this very reason.

image

Mrcuve0:
Wow thanks for the feedback! Sadly I'm a real newbie for what concerns SVGs and Inkscape in general. For instance, I had to search for the steps to complete some of the required operations. So I don't know if I'll be able to find similar differences like the ones you reported for the next iterations. 

I used the `IOTA Mark White` file to produce the `iota.svg` file I pushed. I also used `Ctrl-K` (combine) to collect all the paths. Is there a better way to do that particular step? Finally, after having tried svgomg in action I'd be pretty sure it's just matter of increasing the precision over the level `3` as suggested, but as I said I'm definitely no expert 😆

I can try again with the same procedure I did and just changing the precision parameter on svgomg

PeterShaggyNoble:
> I used the IOTA Mark White file

OK, let me compare against that instead before you do anything else.

> I also used `Ctrl-K` (combine) to collect all the paths. Is there a better way to do that particular step?

"Combine" produced the best results for me too when I tried; there was a _lot_ of movement when I used "Union" (<kbd>Ctrl + +</kbd>) instead.

PeterShaggyNoble:
Seeing the same difference when comparing against the white version, @Mrcuve0, maybe even slightly more pronounced. The good news, though, is that it's nothing you did wrong; after recreating it myself I was seeing the same differences when I optimised with a precision of 3 but everything lined up perfectly when I increased it to 4 so could you give that a try?

PeterShaggyNoble (approved):
SVG looks spot on to me now, @Mrcuve0, nice work 👍🏻

Gonna merge this in as is with the "midnight" colour and, if someone feels strongly enough that we should be using one of the greens instead, we can address that in a separate PR.

Thanks again for the contribution 🙂

PR #5524, opened 2021-04-22
Add Lerna icon

service-paradis:
image

**Issue:** N/A
**Alexa rank:** [~27.8k Github ⭐](https://github.com/lerna/lerna)
  

### Checklist

### Description

That is a fun one, sorry reviewer 😄. ~~I even manage to broke the previewer as you can see above ⬆️~~

- I took `lerna.svg` from https://github.com/lerna/logo
- I used Inkscape to play with it and scale
- I chose the color between `#5e0074` and `#00045b` (gradient used in the https://lerna.js.org/ header logo and code block background)

###### Bonus to show how it is a fun one

image


PeterShaggyNoble:
Oh, @mondeja is going to love this one! 😆

More seriously, though, is now the time to have that discussion I've been wanting to have for a good while (since the addition of our original Elsevier icon) on where we draw the line between logos & icons? My stance on it is: To me the two keywords in what we do are "monochrome" and "icon". We already reject brands that don't have, or at least don't allow, a monochrome treatment of their icon and that, by the same logic, we should also reject brands that don't have, or at least don't allow, an iconified version of their logo. Of course, the tricky part would be defining what we consider to be an icon - something that's clearly legible and well-defined at our native `24*24` size, or some multiple thereof? That could be somewhat subjective. The larger issue of drawing a line is the retroactive effect it would have on existing brands that don't meet the criteria we set and it's on that basis alone - i.e., the precedent has been long since set and changing it now would result in us having to cull a large portion of our collection - that I'd be OK with us _not_ drawing a line at all. (And, with just my contributor's hat on, the more complex logos often make for a much more fun challenge than a lot of the simpler icons! 🙂)

PeterShaggyNoble (approved):
Forgot to actually leave my review of the icon itself! 🤦🏻‍♂️ Looks pretty darn good to me, couldn't spot any issues with it - nice work 👍🏻 Happy with the colour choice, too, although a pink might be another option based on the links at the end of their website, either the pink used in the links or the one from the gradient.

Given the complexity of the icon, though, I'd like one of the other @simple-icons/maintainers to throw another pair of eyes over it before merging, if we're happy to revisit the discussion of logos vs. icons at a later date.

ericcornelissen:
_Didn't review the SVG, just responding to @PeterShaggyNoble_


***


> [...] is now the time to have that discussion I've been wanting to have for a good while (since the addition of our original Elsevier icon) on where we draw the line between logos & icons? My stance on it is: To me the two keywords in what we do are "monochrome" and "icon". We already reject brands that don't have, or at least don't allow, a monochrome treatment of their icon and that, by the same logic, we should also reject brands that don't have, or at least don't allow, an iconified version of their logo. Of course, the tricky part would be defining what we consider to be an icon - something that's clearly legible and well-defined at our native `24*24` size, or some multiple thereof? That could be somewhat subjective. The larger issue of drawing a line is the retroactive effect it would have on existing brands that don't meet the criteria we set and it's on that basis alone - i.e., the precedent has been long since set and changing it now would result in us having to cull a large portion of our collection - that I'd be OK with us _not_ drawing a line at all. (And, with just my contributor's hat on, the more complex logos often make for a much more fun challenge than a lot of the simpler icons! slightly_smiling_face)

I'm open to the discussion, though I personally don't think the complexity of the icon is a reason to not include it. My stance has always been this: if the brand asks us to use an "icon" that doesn't look good (for whatever reason: trademark symbol, complexity, w/e) that is not our concern but something the brand should fix. 

If we do want to continue this discussion though, I'd suggest we open a dedicated issue for it :slightly_smiling_face:

LitoMore:
I think this is similar to https://simpleicons.org/?q=ferrari.

mondeja (approved):
Looks perfect! Another complex SVG that helps a lot refining the linter.

Thank you @service-paradis 🐉

PR #5527, opened 2021-04-23
Fix redirection when returning from new website

mondeja:
The current problem is: if an user comes from the new website (clicking in "Go to the old design" link) and has the automatic redirection active (saved at storage), will be redirected to the new website again. This happens because `document.referrer` [does not return the full URL of the page from the user comes](https://stackoverflow.com/q/23682718/9167585), the path is omitted, so the referrer comparation should be against the root domain.

### DIFF
--- a/site_script.js
+++ b/site_script.js
@@ -200,6 +200,7 @@
       $hideAlways = document.getElementById('hide-feedback-request');
 
   var redesignUrl = 'https://simple-icons.github.io/simple-icons-website/',
+      redesignRootDomain = 'simple-icons.github.io',
       hideBannerAlwaysIdentifier = 'hide-banner',
       redirectAutomaticallyIdentifier = 'redirect-to-redesign';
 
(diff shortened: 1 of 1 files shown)


PeterShaggyNoble (approved):
Looks good to me, @mondeja, nice catch 👍🏻 From [the MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer) on it, I too would have expected the full URL to be returned.

ericcornelissen:
Seems to be a difference of behaviour between browsers, as the current approach works (*) for me in Firefox - and if I check `document.referrer` in the console it shows the URL including path. That said, from my testing this solution also works in Firefox so thanks for fixing it @mondeja :tada: 

(*): Due to this issue I checked what the current behaviour is if you performed a search on the new website (i.e. the URL includes `?q=xxx`) and it actually seems this is included in `document.referrer` because I get redirected back to the new website... This actually seems like somewhat of a privacy concern to me :thinking:

mondeja:
I've tested in Firefox and Google Chrome, both are giving me only the domain. Are you double checked that you are clicking on a link pointing to another domain? Inside the same domain will give you the full path, but between different ones not.

ericcornelissen:
I checked with the life websites so yes. I'm on a somewhat old version of Firefox though, so this may be changed in more recent versions. It possibly relates to [this change](https://blog.mozilla.org/security/2021/03/22/firefox-87-trims-http-referrers-by-default-to-protect-user-privacy/) in Firefox, can't say for sure about Chrome.

PR #5545, opened 2021-04-25
Auto update CDN URLs in README.md in release PR

ericcornelissen:
**Issue:** Closes https://github.com/simple-icons/release-action/issues/1
**Alexa rank:** n/a
  

### Checklist

### Description

This Pull Request does three things:

1. ~~Add [semver](https://www.npmjs.com/package/semver) as a `devDependency`~~
2. Add a script (named `scripts/update-cdn-urls.js`) that automatically updates the major version number in CDN URLs in the README.md
3. Run the script of 2) when creating a version bump commit for release Pull Requests.


.github/workflows/create-release.yml
```diff
@@ -27,6 +27,8 @@ jobs:
           ref: develop
       - name: Bump version
         run: node ./scripts/bump-version.js "${{ needs.release-pr.outputs.new-version }}"
+      - name: Update CDN version in README.md
+        run: node ./scripts/bump-cdn-version.js
```
mondeja, line 31: The sripts `bump-version.js` and `bump-cdn-version.js` can be grouped in a `bump` directory after #5546. Additionally, I feel that the name of this script is really confusing, I would vote for:

```suggestion
        run: node ./scripts/bump/readme-cdn-urls-version.js
```

scripts/bump-cdn-version.js
```diff
@@ -0,0 +1,48 @@
...
+ * behaviour of this script should be consider undefined.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const getMajorVersion = require("semver/functions/major");
+
+const rootDir = path.resolve(__dirname, "..");
+const PACKAGE_JSON_FILE = path.resolve(rootDir, "package.json");
+const README_FILE = path.resolve(rootDir, "README.md");
```
mondeja, line 18: The style of these variable names should be consistent with the other scripts, basically the same suggestion of https://github.com/simple-icons/simple-icons/pull/5546#discussion_r619867415

 ```suggestion
const packageJsonFile = path.resolve(rootDir, "package.json");
const readmeFile = path.resolve(rootDir, "README.md");
```

**Review by mondeja — COMMENTED** [review]
I've some suggestions before perform a full review.

File: .github/workflows/create-release.yml (hunk 1 above)
Comment (ericcornelissen) [reply .github/workflows/create-release.yml:31]: > The sripts bump-version.js and bump-cdn-version.js can be grouped in a bump directory after #5546. 

Not sure if "bump" is a particularly descriptive folder name :thinking: 


> Additionally, I feel that the name of this script is really confusing, I would vote for:

I agree couldn't come up with something better when I created the file and forgot to check before submitting this PR. I like including "urls", but specifying the README seems a bit overkill (plus, who knows where else we might want to update these URLs in the future). How about `update-cdn-urls.js`?

Comment (mondeja) [reply .github/workflows/create-release.yml]: > Not sure if "bump" is a particularly descriptive folder name

What about `versioning` or just `version`?

> I like including "urls", but specifying the README seems a bit overkill (plus, who knows where else we might want to update these URLs in the future). How about `update-cdn-urls.js`?

I agree, that would be also descriptive!

Comment (ericcornelissen) [reply .github/workflows/create-release.yml]: > What about versioning or just version?

Hmm, I'm not convinced. I'm in favor of creating a folder for this but if the name is bad I'd rather hold off on it or now :thinking:

`release` is another option, what do you think about that?

In any case, I don't think this topic is blocking for this PR - or should even be solved in this PR. If anything, this PR can be merged and the folder name can be resolved in #5546.

I'd definitely say we should merge this in before the [v5 release](https://github.com/simple-icons/simple-icons/milestone/4), so to be sure we don't forget about this I added it to the milestone.

File: scripts/update-cdn-urls.js [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+ * package major version, or the package minor version minus 1. If not the
+ * behaviour of this script should be consider undefined.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const getMajorVersion = require("semver/functions/major");
+
+const rootDir = path.resolve(__dirname, "..");
+const packageFileJson = path.resolve(rootDir, "package.json");
```
mondeja, line 17: It seems that the name should be `packageJsonFile`, as has been named in #5546:

```suggestion
const packageJsonFile = path.resolve(rootDir, "package.json");
```

ericcornelissen:
@mondeja, do you agree this is good to merge?

scripts/update-cdn-urls.js
```diff
@@ -0,0 +1,48 @@
...
+ * nothing if the README.md is already up-to-date.
+ *
+ * It assumes the major version in the README is either equal to the current
+ * package major version, or the package minor version minus 1. If not the
+ * behaviour of this script should be consider undefined.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const getMajorVersion = require("semver/functions/major");
```
mondeja, line 14: Why use a third party library and not just to use `parseInt(manifest.version.split('.')[0])`? The replacement uses `majorVersion - 1` anyways.

scripts/update-cdn-urls.js
```diff
@@ -0,0 +1,48 @@
...
+
+const rootDir = path.resolve(__dirname, "..");
+const packageJsonFile = path.resolve(rootDir, "package.json");
+const readmeFile = path.resolve(rootDir, "README.md");
+
+function getManifest() {
+  const manifestRaw = fs.readFileSync(packageJsonFile).toString();
+  const manifestJson = JSON.parse(manifestRaw);
+  return manifestJson;
```
mondeja, line 23: Why not just return in the `JSON.parse` line? The function `getManifest` should be the readable logic wrapper itself.

scripts/update-cdn-urls.js
ericcornelissen: > Why are you using a third party library and not just to use `parseInt(manifest.version.split('.')[0])`?

Because why not, it should be more robust - though I agree there is not much that can go wrong here - and it's definitely easier to parse `getMajorVersion` than `parseInt(manifest.version.split('.')[0])`. If you have a particular concern about using a dependency (e.g. security, install speed, something else) I'd happily change it. For what it's worth we use [the same dependency in our release action](https://github.com/simple-icons/release-action/blob/4c952419b4374cce8d114104ac07679adf655a6c/package.json#L28).

If we do decide to change it, my vote would go towards having a local function named `getMajorVersion` that, in some form, does `parseInt(manifest.version.split('.')[0])`.


> The replacement uses `majorVersion - 1` anyways.

Not sure how this relates 🤔 Can you elaborate?

scripts/update-cdn-urls.js
ericcornelissen: No strong opinions on this one. I will wait with changing it until we decided on your other comment.

My reasoning for doing it this way: I have found that it makes future diffs easier to read as the `return` statement doesn't have to change or move if the logic of the function needs to change. You may think of it this way: the function must return a `manifestJson` variable, and I don't really care so much how it was obtained. That said, I'm not consistently using this style and this function is not a particular good example/use case of this principle.


> The function getManifest should be the readable logic wrapper itself.

I don't see how the extra line makes it less readable, unless you count longer as harder.

scripts/update-cdn-urls.js
mondeja: > Not sure how this relates thinking Can you elaborate?

That the update is just to increase an integer. I think it is not necessary to add a library for this if we are not going to generate more complex version names.

> Because why not, it should be more robust

How could be this more robust? I think that use a third party library for this simple task is just obfuscate the code needlessly.

> If we do decide to change it, my vote would go towards having a local function named getMajorVersion that, in some form, does `parseInt(manifest.version.split('.')[0])`.

I like that idea :+1:

scripts/update-cdn-urls.js
mondeja: > You may think of it this way: the function must return a `manifestJson` variable, and I don't really care so much how it was obtained.

Should not be named `getManifestJson` the function if that is always supposed to return that? Apart from that, this would really change in the future? It is just to read the `package.json` file, I don't think we are going to change the way to get the version.

scripts/update-cdn-urls.js
ericcornelissen: I will update this on the basis that indeed this is rather simple and using a dependency does have (other) drawbacks.


---


> That the update is just to increase an integer. I think it is not necessary to add a library for this if we are not going to generate more complex version names.

I don't agree that what you do with the date should determine how you get it. For example, we don't need to parse the JSON (from Node's standard library) in order to get the version number, we can also use the RegExp `/"version":\s*"([^"]*)"/` or go to the exact character-index we expect to get it. But parsing the JSON is easier to understand and work with.

> How could be this more robust? I think that use a third party library for this simple task is just obfuscate the code needlessly.

I completely agree that in this case there does not seem to be much to be worried about w.r.t. robustness. However, I don't see how using a third-party library obfuscates anything. The function name is descriptive, simple, and easy to understand provided you're familiar with SemVer (if you're not, library vs. local code is not going to make a difference). Moreover, this is JavaScript so you can just read the source code of the library (as opposed to e.g. a `.jar` file for Java).

scripts/update-cdn-urls.js
ericcornelissen: > Should not be named getManifestJson the function if that is always supposed to return that?

That is a good point, but at a different scope. My reasoning concerns only the function as it's scope. Don't get me wrong, I don't have any data or "official" guidelines to back me up, it's just something I've been playing with lately.


> Apart from that, this would really change in the future? It is just to read the `package.json` file, I don't think we are going to change the way to get the version.

Definitely agree, but you never know 😉 Even in my very short journey as a software engineer I've seen it happen many times that this was said and turned out not to be true.

scripts/update-cdn-urls.js
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
+/**
+ * @fileoverview
+ * Updates the CDN URLs in the README.md to the latest major version. Does
+ * nothing if the README.md is already up-to-date.
+ *
+ * It assumes the major version in the README is either equal to the current
+ * package major version, or the package minor version minus 1. If not the
+ * behaviour of this script should be consider undefined.
```
mondeja, line 9: I don't understand the statement "If not the behaviour of this script should be consider undefined.", seems confuse to me. Really, if the already defined major version in the README is not the one lower than the major version defined in the `package.json` file, the script does nothing, except for the case that the version defined in the README would be a number with different number of decimals, for example:

- `package.json`: "6.2.0"
- `README`: `v55.2.0`
- `README` after script: `v65.2.0`

It is a stupid case, but for consistency I would suggest to remove that statement and applies the other suggestion that I've mentioned below.

```suggestion
 * package major version, or the package minor version minus 1.
```

scripts/update-cdn-urls.js
```diff
@@ -0,0 +1,51 @@
...
+  const manifestRaw = fs.readFileSync(packageJsonFile).toString();
+  return JSON.parse(manifestRaw);
+}
+
+function updateVersionInReadmeIfNecessary(majorVersion) {
+  let content = fs.readFileSync(readmeFile).toString();
+
+  content = content.replace(
+    new RegExp(`simple-icons@v${majorVersion - 1}`, "g"),
+    `simple-icons@v${majorVersion}`,
```
mondeja, line 34: To prevent inconsistency described in other suggestion:

```suggestion
    new RegExp(`simple-icons@v${majorVersion - 1}/`, "g"),
    `simple-icons@v${majorVersion}/`,
```

mondeja (approved):
I'm seeing all good now, except for an edge case that could be improved. Anyways, I'm fine to go with the merge, feel free to do it without that change @ericcornelissen :+1:

scripts/update-cdn-urls.js
ericcornelissen: > I don't understand the statement "If not the behaviour of this script should be consider undefined.", seems confuse to me. Really, if the already defined major version in the README is not the one lower than the major version defined in the `package.json` file, the script does nothing, except for the edge case that the version defined in the README would be defined with different number of decimals, [...]

I believe the statement is self-explanatory, if you consider _why_ I wrote it. Of course, I _could_ describe the behaviour of the script in the scenario the input conditions are not met, but why would I? The input conditions should be met, and if they're not, something has already gone wrong. Moreover, if the behaviour in the invalid case changes, now I need to update the documentation for something that is irrelevant...  Why would I do that.

You can really read it just like this: "If you don't meet my preconditions, I cannot give any guarantees on what is going to happen. Of course something is going to happen, but I'm not going to worry about it." You can well argue that we should prevent that, maybe we should, or maybe we can fix the undefined behaviour. But when I wrote this I didn't feel like that's worth it: most of the time it's going to work as expected and when it doesn't it is easy to fix - so it's not worth the effort to solve the problem.

Also, of the two scenarios, consider which one is more likely? It must be the scenario that the version in the README is not exactly one less than the version in `package.json`, as the other is a subset of that. So which one should we fix? Probably the former.

So I did that, turns out it didn't take that much effort but when I put this Pull Request together I didn't think about it 🙃

mondeja (approved):
Much better now. Thank you @ericcornelissen :100:

PR #5555, opened 2021-04-27
Letter H review

service-paradis:
**Issue:** #5251

### Notes

- [x] I think we might need to update https://habr.com/, no more use of a squared "h" can be found
- [x] We need to update https://www.hackerrank.com/, logo is now H🟩
- [ ] Was not able to find a better source for https://www.pluralsight.com/hackhands
- [x] Should we add the circle around https://www.hackster.io/branding?
- [x] I think we might need to add the ®️ for https://www.hellyhansen.com/
- [x] We need to remove Highly (https://techcrunch.com/2019/04/17/twitter-highly/) - #4243
- [x] We need to remove HipChat (https://en.wikipedia.org/wiki/HipChat) - #5556
- [ ] Was not able to find a better source for https://www.homeadvisor.com/
- [x] I think we might need to use the full wordmark for https://press.hulu.com/branding/

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -3664,7 +3664,7 @@
         {
             "title": "HackHands",
             "hex": "00ACBD",
-            "source": "https://hackhands.com/"
+            "source": "https://www.pluralsight.com/hackhands"
         },
         {
             "title": "Hackster",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -3722,7 +3722,8 @@
         {
             "title": "Haxe",
             "hex": "EA8220",
-            "source": "https://haxe.org/foundation/branding.html"
+            "source": "https://haxe.org/foundation/branding.html",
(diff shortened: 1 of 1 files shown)


adamrusted (approved):
LGTM! Do you want to break the updates out into their own issues so we can get this merged in?

service-paradis:
> LGTM! Do you want to break the updates out into their own issues so we can get this merged in?

Done :)
Except for the 2 icons where I cannot find better source for now. Since they dont necessarily need to be updated, we might keep the new sources and update them if we find something in the future. What do you think?

adamrusted:
> Since they dont necessarily need to be updated, we might keep the new sources and update them if we find something in the future. What do you think?

I think we're okay to merge this without worrying about those for now. No doubt they'll get flagged again in the future, either when a rebrand takes effect (and hopefully, a new source is found at that point) or they re-vamp their website giving us the relevant files. Until then, if they're not outdated - I feel we're safe to ignore them.

PR #5573, opened 2021-04-29
Adds Bazel icon

dzintars:
bazel

### Checklist

### Description
- https://bazel.build/
- https://github.com/bazelbuild/bazel (16.6k stars)
- Modified the icon to add gaps.


PeterShaggyNoble:
Thanks for the contribution, @dzintars 🙂 Did you intend to close this PR, though?

Note that we would probably be best to follow the outlined monochrome treatment they themselves use in [their website logo](https://bazel.build/images/bazel-navbar.svg).

dzintars:
@PeterShaggyNoble Yes, i just messed up a bit in my previous PR, and so i closed this "incident" :)
I have prepared commit on my local. Will push it later this week. Hope it will get merged in the next release.
Their official logo does not match the Simple Icons guidelines, so i created gaps instead of using strokes.

PeterShaggyNoble:
> Yes, i just messed up a bit in my previous PR,

Ah, yes, I see now from the commits you just mixed up the PRs.

> Their official logo does not match the Simple Icons guidelines

What do you mean? It looks OK to me (see below). Although we may need to add some cut-outs in an attempt to maintain the illusion of depth.

dzintars:
I mean, their SVG is composed of many strokes, inline styles and so on.
On top of that in badge scale outlined variant looks messy (unrecognizable) on the non-retina screens.
In order to make it look better i made "filled" variant which in the current poll takes over outlined variant, thou there is some vacations or something and not a lot activity happening in the community right now. So i will wait for some time to get better feedback.
Until that i prepared 2 variants of the icon and asked in the SI Discussions is it OK to commit 2 variants of the same icon (by using modifier). That would be great. But not sure does it contradicts with the SI guidelines.
Would be grateful if you could explain what do you mean with "cut-outs".

dzintars:
> The strokes can be converted to paths by most all software an the inline styles are easily removed.

Yes i know that, and it also requires some cleanup to reduce the data size of the icon by removing redundant edges, etc.

> Understandable. But we do still have to abide by official branding and as they provide an (almost) monochrome version themselves, we'd have to run with that.

I totally understand that and agree with you, thou... it does not look that good on the small badges.

> Could you provide a link to that poll, so we can keep track of it ourselves? As it is currently under discussion, the 2 approaches we could take are to add the outlined version for now and update it to the filled version if/when they decide to change it or to hold off on adding it pending the outcome of that poll.

It's in their Slack channel. Not sure how "official" it is... but... at least something:
image
Its really strangely silent there today.

> See example below, where I've added cut-outs of the lines on the top of the shape to the lines deeper into it in an attempt to maintain the illusion of depth. It's only a rough mock-up, we may need to go a little thicker with the cut-outs and remove them from the corners.

Well... seems you are almost ready to push it. 😄  😄  So i can go home. 😃 
Thank you for clarification. TIL
But pay attention that in the original for "backside" lines they are using opacity/transparency probably to achieve that those lines visually appear thinner. If so... then probably we don't need those cutouts (thou i like the idea) and we can use thinner "lines" instead.
WIP:
image


I am not an designer by no means... 😃  i just use those tools daily and wanted to have native badges to use in my documentation because in-lining SVG feels kinda dirty. :) So if you want me to finish this task, i will go with your advise or i can just back-off.

I also did some progress on https://grpc.io/ This is little bit simpler, thou.. because icon is so wide, on the actual icon it appears to look bit strange.
image

Issue #5058, opened 2021-02-18
Odysee icon (LBRY frontend / rebrand?)


Emaleth:
**Brand Name: Odysee / LBRY**
**Website: https://odysee.com/**
**Alexa rank: 3,925**
**Official resources for icon and color: https://odysee.com/public/img/odysee_white.png**

PeterShaggyNoble:
Thanks for the request, @Emaleth

Would you like to have a read through [our contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and give this a try yourself, using the vector files available [here](https://odysee.com/@OdyseeHelp:b/odyseepresskit:b)?

PR #5577, opened 2021-04-30
Add Odysee

foo-dogsquared:
image

**Issue:** #5058
**Alexa rank:** ~3.8k (as of 2021-04-31)
  

### Checklist

### Description

The hex value is the first stop value for the gradient and similar colors can be found on the site so I picked that.


PeterShaggyNoble:
Nice work, @foo-dogsquared 👍🏻 You fared better with the stars than I did; took a few stabs at it and no matter what I did with them and in which order there was always at least one of them jumping all over the place. The is **a** [_tiny_ hairline](https://user-images.githubusercontent.com/15157491/117131331-370afa80-ad99-11eb-9057-34fc4eee2c68.png) down the left edge of the star under the arm on the left when comparing yours (in red) with the original. It's insignificant enough, though, that I'd be OK to run with it if it can't be cleaned up with too much hassle.

I agree with the choice of pink for the colour but would `#fa6165` be a better choice seeing as its variable name in their stylesheet is `--color-primary`?

foo-dogsquared:
> The is a tiny hairline down the left edge of the star under the arm on the left when comparing yours (in red) with the original. It's insignificant enough, though, that I'd be OK to run with it if it can't be cleaned up with too much hassle.

Thank goodness. It took me an embarrassingly long time than I initially expected just to get it right. :p

> I agree with the choice of pink for the colour but would #fa6165 be a better choice seeing as its variable name in their stylesheet is --color-primary?

The website does have two themes (light and dark) with different `--color-primary` values. Both of the `--color-primary` values are closer to the colors from the original icon vector file. I don't know what to pick from them. For now I'll be sticking to my choice since it is also of the color from the [favicon](https://odysee.com/public/favicon.png).

PeterShaggyNoble:
> It took me an embarrassingly long time than I initially expected just to get it right

Sounds like you had as much fun with it as I did! 😆

> The website does have two themes (light and dark) with different `--color-primary` values

Ah, in that case then it's definitely not teh primary _brand_ colour. Still not 100% convinced with our colour choice but as it _is_ used in the favicon and in the absence of anything else to go on, let's run with it for now.

Thanks, @foo-dogsquared 👍🏻

Issue #5582, opened 2021-05-01
Thingiverse (3D design platform)


foo-dogsquared:
**Brand Name:** Thingiverse
**Website:** https://www.thingiverse.com/
**Alexa rank:** [~1.3k as of 2021-05-01](https://www.alexa.com/siteinfo/thingiverse.com)
  

**Official resources for icon and color:**
- Found no media kits or something similar but the favicon is a good candidate. It appears on their social media consistently. There is an SVG on the website header.
- Chosen color would be #248BFB.

adamrusted:
The mobile header does have a Base64-encoded version of the SVG in it's header, which should serve as a good-enough source for us. Would you like to contribute this one @foo-dogsquared?

foo-dogsquared:
Sure, this seems simple enough.

PR #5591, opened 2021-05-02
Add Thingiverse

foo-dogsquared:
image

**Issue:** #5582
**Alexa rank:** ~1.3k as of 2021-05-02
  

### Checklist

### Description

The SVG icon in the website header is off-centered compared to the favicons so I had to make some adjustments.


adamrusted (approved):
LGTM! Thanks @foo-dogsquared! 🚀

Worth noting, that when I downloaded the Base64-encoded version, the 'T' is much higher up in the vector than it is in the PNGs used as the favicons. Comparing against the 32*32 favicon, yours matches up perfectly.

PR #5601, opened 2021-05-03
Letter G review

service-paradis:
**Issue:** #5251

### Notes

- [ ] We need to update https://g2acowebproddata.blob.core.windows.net/g2acowebproddata/2020/06/G2A_Brand_Book_Basic_September_2019.pdf since the logo is slightly different
  - We might be able to remove the `.com` as stated in their guidelines: "You can omit “.com” with the G2A logo in very small sizes"
- [ ] I think we might need to update https://get.gaug.es/
- [ ] We need to update https://www.geeksforgeeks.org/
- [ ] I think we might need to add the ® for https://www.geocaching.com/about/logousage.aspx
- [ ] We need to update https://ghost.org/docs/logos/ (or remove since it is hard to do a monochrome version?)
- [ ] We need to remove https://support.giphy.com/hc/en-us/articles/360022283772-GIPHY-Brand-Guidelines (no monochrome version allowed)
- [ ] We need to update https://github.com/logos
- [ ] We need to remove https://www.graph.cool/

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -3167,7 +3167,8 @@
         {
             "title": "Geocaching",
             "hex": "00874D",
-            "source": "https://www.geocaching.com/about/logousage.aspx"
+            "source": "https://www.geocaching.com/about/logousage.aspx",
+            "guidelines": "https://www.geocaching.com/about/logousage.aspx"
         },
         {
             "title": "Gerrit",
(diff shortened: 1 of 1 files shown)


service-paradis:
> @service-paradis as mentioned on other issues, could you break out the checklist into it's own issue(s) so that we can merge the updates already made here? 🙏

Done 😄

Issue #5599, opened 2021-05-03
Update Helly Hansen icon?


service-paradis:
**Brand Name:** Helly Hansen
**Website:** https://www.hellyhansen.com/
**Official resources for icon and color:**

Following #5555, I think we might need to add the ®️ symbol.
![image](https://user-images.githubusercontent.com/12817388/116910746-064d8880-ac14-11eb-9e92-c43940b4474d.png)

Every usage they make themselves have the symbol (website and social media). A SVG is available in the header of their website.

What other @simple-icons/maintainers thinks about it?

ericcornelissen:
I agree, seems to me to fall into the category of the third bullet of [our (registered) trademark icon guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#icon-guidelines) 👍

PR #5612, opened 2021-05-06
Update trademark guidelines in CONTRIBUTING.md

PeterShaggyNoble:
In light of #5599, #4804 and a few others, this PR updates our contributing guidelines on when to include and when to omit a (registered) trademark symbol by adding a line that requires it to be included if the brand itself includes it wherever they use their logo, even at icon sizes.

It also updates the example icon for logos that incorporate the symbol into the design of the logo, as Intel no longer does. Chupa Chups was the first one that sprang to mind for me but probably not the best choice as a showcase icon for our documentation - if anyone can think of a better example let me know.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,8 @@ Working with an SVG version of the logo is best. In the absence of an SVG versio
 If the icon includes a (registered) trademark icon we follow the guidelines below to decide whether to include the symbol or not:
 
 * If brand guidelines explicitly require including the symbol, it must be included.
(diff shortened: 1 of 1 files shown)


ericcornelissen (approved):
I think this is a good clarification, thanks @PeterShaggyNoble :+1: I'll leave it for at least one more @simple-icons/maintainers to review this before we merge it.

As for the example, I think it's fine like this and would probably fine with any other as well :slightly_smiling_face:

service-paradis (approved):
This is a good clarification! Thank you @PeterShaggyNoble :)

PR #5616, opened 2021-05-06
Remove "icon" suffix from SVG titles

### LINKED ISSUE #5478: Remove "icon" suffix from SVG `<title>`s
As per [this discussion](https://github.com/simple-icons/simple-icons/discussions/5186) I open this issue to remove the "icon" suffix from SVG `<title>`s. I.e., change for example `<title>SimpleIcons icon</title>` to `<title>SimpleIcons</title>` in [the Simple Icons SVG](https://github.com/simple-icons/simple-icons/blob/6c435fe7411182ea8e09c42d77ba827e37bb79f0/icons/simpleicons.svg). To avoid any potential problems for users this change will not happen until [the next major release](https://github.com/simple-icons/simple-icons/milestone/4).

Further comments regarding this issue are very welcome here.

As per [the discussion](https://github.com/simple-icons/simple-icons/discussions/5186), the pros are:

- Smaller SVGs by file size
- Keeping a minimum of built-in a11y while avoiding having to provide localized titles (*)

And the cons are:

- _Not clear if any_

---

(*): as per [the discussion](https://github.com/simple-icons/simple-icons/discussions/5186) another option would be to generate SVGs with localized names. Though feasible, unless this is done on the fly it would make the NPM published unnecessarily big.

PeterShaggyNoble:
**Note:** It would probably be best to merge all other v5 PRs before this one so that any conflicts that arise can be addressed here.

**Issue:** Resolves #5478

Although not a breaking change, it has been agreed that we hold off on this change until v5. However, I deliberately haven't submitted this PR as a draft in case we change our minds.

~~Is there a way to exclude this change from being picked up by the release action? Otherwise it's going to list all 1900+ icons as having been updated.~~


ericcornelissen:
> Is there a way to exclude this change from being picked up by the release action? Otherwise it's going to list all 1900+ icons as having been updated.

Good point, I will (try to) make it ignore this PR number 👍

ericcornelissen (approved):

ericcornelissen:
Minor request, can you remove the "icon" text from the examples in [the Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/233f9f605131958eb599a01fa700d62a105947af/CONTRIBUTING.md#4-annotate-the-icon) as well (subsection 4 and 5)?

PeterShaggyNoble:
Good catch, @ericcornelissen 👍🏻 Updated.

Do you think we still need the "Note the format" for the `title` in the checklist for annotating the icon?

ericcornelissen:
> Do you think we still need the "Note the format" for the title in the checklist for annotating the icon?

I thought about this, but wasn't sure what to suggest so I left it as I think it still _works_ like this - though, I agree it's not perfect...

Maybe something along the lines of "A title element with the brand name"? Potentially we can add something like "must be an exact match with the `"title"` in the JSON data", but I'm not sure about that as the JSON data is not yet mentioned at this point...

PR #5670, opened 2021-05-11
Add icon for Expensify

### LINKED ISSUE #5635: Expensify
**Brand Name:** Expensify
**Website:** https://www.expensify.com
**Alexa rank:** 6,099
  

[SVG on the site](https://d2k5nsl2zxldvw.cloudfront.net/images/logo/expensify-iconmark-reversed.svg) (contains a white E in the middle that may not be visible if background is white)

dhruvkb:
expensify

**Issue:** Closes #5635
**Alexa rank:** 6,099
  

### Checklist


PeterShaggyNoble (approved):
All looks good to me, @dhruvkb, nice work 👍🏻

I found a press kit at https://use.expensify.com/press-kit which includes some guidelines on the icon usage so could you update the JSON to add that as the `guidelines` entry. They also list the brand colours there, without explicitly stating which is the primary colour. In those case we _usually_ opt for the first listed colour, which would be the yellow, but in this case I do still think the blue is the correct choice as both shades of it seem to dominate their branding.

dhruvkb:
Thanks for discovering the press kit, I've updated the source URL in the JSON. I was slightly divided between the blue (used in large amount on the site) and yellow (which forms 50% of the logo's outer circle) but I agree with you that the former is more suitable.

_data/simple-icons.json
```diff
@@ -2742,6 +2742,11 @@
                 "type": "CC-BY-3.0"
             }
         },
+        {
+            "title": "Expensify",
+            "hex": "0185FF",
+            "source": "https://use.expensify.com/press-kit"
```
PeterShaggyNoble, line 2748: Thanks, @dhruvkb. If you can duplicate that for the `guidelines` entry then we'll be good to go
```suggestion
            "source": "https://use.expensify.com/press-kit",
            "guidelines": "https://use.expensify.com/press-kit"
```

dhruvkb: Done!

PeterShaggyNoble:
Thanks, again, @dhruvkb.

PR #5678, opened 2021-05-12
Update Fedora JSON

adamrusted:
### Notes
Following internal discussion with @mairin, we're allowed to continue using the current version of the Fedora icon for the time being, provided we link to the proper documents on their end. I've updated the JSON to include the guidelines and legal links for the existing icon. Awaiting feedback from the Fedora team as to whether we can rename this as 'Fedora Classic' once the new icon is in place. The question has been asked as to whether our users would _need_ a 'Fedora Classic' icon, or whether they can just use the new version. What are your thoughts @simple-icons/maintainers?

**Note:** This will not close the 'Update Fedora' issue, as the discussion around the new icon is still ongoing. This is just to update our JSON to be a bit more compliant until we're able to use the new icon.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -2872,7 +2872,12 @@
         {
             "title": "Fedora",
             "hex": "294172",
-            "source": "https://fedoraproject.org/wiki/Logo/UsageGuidelines"
+            "source": "https://fedoraproject.org/wiki/Logo/UsageGuidelines",
+            "guidelines": "https://fedoraproject.org/wiki/Logo/UsageGuidelines",
+            "license": {
+                "type": "custom",
+                "url": "https://fedoraproject.org/wiki/Legal/TrademarkGuidelines"
+            }
         },
         {
             "title": "FedRAMP",
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -2876,7 +2876,7 @@
             "guidelines": "https://fedoraproject.org/wiki/Logo/UsageGuidelines",
             "license": {
                 "type": "custom",
-                "url": "https://fedoraproject.org/wiki/Legal/TrademarkGuidelines"
+                "url": "https://fedoraproject.org/wiki/Legal:Trademark_guidelines"
             }
         },
         {


PeterShaggyNoble:
Unless specifically asked to include it by Fedora, I'd suggest removing the `license` entry as it's not, strictly speaking, a license and the page is linked from the usage guidelines which appears as both the `source` and `guidelines`. If we _were_ asked to include it by Fedora then it should probably be updated to https://fedoraproject.org/wiki/Legal:Trademark_guidelines at which stage this PR can be merged straight in.

On Fedora Classic: I'd prefer its removal but if Fedora envisage that there will be call for it based on their own usage of it going forward and are happy for us to include it then I think we should.

PeterShaggyNoble (approved):
OK, seeing as Fedora _did_ ask us to include their Trademark Guidelines as a license this can be merged in. Thanks, @adamrusted 👍🏻

ericcornelissen:
> I'd suggest removing the license entry as it's not, strictly speaking, a license and the page is linked from the usage guidelines which appears as both the source and guidelines. If we were asked to include it by Fedora then it should probably be updated to https://fedoraproject.org/wiki/Legal:Trademark_guidelines at which stage this PR can be merged straight in.

Though I agree `https://fedoraproject.org/wiki/Legal:Trademark_guidelines` feels a lot more like a license than the usage guidelines, and provides license information, it still doesn't (strictly) seem to me like a License.

We _could_ look into updating our approach towards capturing this kind of data - e.g. switching from a "license" field to a "legal" field. Not sure if that's worth the effort, especially for as long as Fedora is the only instance where we stretch the definition of license. I'd say we should keep an eye out for this kind of usage of the "license" field and consider updating it if necessary.

Also, another reason for me to not suggest immediately changing this is that we _can_ define "license" pretty strictly, whereas with a "legal" field the data could become complex quickly. E.g., we could do something like:

```json
{
  "title": "random example",
  "legal": {
    "kind": "license",
    "type": "custom",
    "url": "https://..."
  }
}
```

for licenses, and then something like

```json
{
  "title": "Fedora",
  "legal": {
    "kind": "guidelines",
    "url": "https://fedoraproject.org/wiki/Legal:Trademark_guidelines"
  }
}
```

for instances like Fedora. This may seem simple, until you actually want to parse it and need to consider all the various possibilities - which admittedly is still doable with two "kind"s.

adamrusted:
Yeah, I agree with your points above @ericcornelissen. I'm hoping when the new logo is finalized and sent through by the Fedora team we can get a more appropriate "license" page, or at least a definition of how the logo is licensed, as opposed to a generic trademarks page. Hoping this is one of the few where we have to make an exception, purely on the input of the team at Fedora themselves.

PR #5688, opened 2021-05-13
Add Weblate icon

### LINKED ISSUE #5085: Request: Weblate icon
**Brand Name: Weblate**
**Website: https://weblate.org/**
  

**Official resources for icon and color: https://github.com/WeblateOrg/weblate/**

service-paradis:
image

**Issue:** Close #5085
**Alexa rank:**
  

### Checklist

### Description

- I took the SVG from https://github.com/WeblateOrg/graphics/blob/669e4f910abd9ec36fda172d2ea6f2f424a32ace/logo/weblate-black.svg
- I used Inkscape to remove a lot of extraneous points, fix some imperfection (that looked unintentional) and scale


PeterShaggyNoble:
@service-paradis, I'm seeing [a hairline thin difference](https://user-images.githubusercontent.com/15157491/118622668-6f5e0000-b7bf-11eb-8890-833dac2b8011.png) (yours in red) in the inner curve of the left shape that's most likely due to optimisation; could you see if you can fix that up?

I'm also seeing some minor differences in the centre loop but it looks like they're all as a result of you cleaning up a couple of imperfections so I think we're OK to run with them.

service-paradis:
Thank you @PeterShaggyNoble for the review!

You're right, I missed this curve after the optimisation, I think it might be good now!

> I'm also seeing some minor differences in the centre loop but it looks like they're all as a result of you cleaning up a couple of imperfections so I think we're OK to run with them.

Yes, I though those bumps were not intentional, so I tried to make the curve more natural. Let me know if you would prefer to keep those bumps!

PeterShaggyNoble (approved):
Looks good now, @service-paradis, nice one 👍🏻

Issue #5721, opened 2021-05-17
Icon Request: Wwise


made-indrayana:
**Brand Name:** Wwise
**Website:** https://www.audiokinetic.com/products/wwise/
**Alexa rank:** [247,056](https://www.alexa.com/siteinfo/audiokinetic.com)

**Official resources for icon and color:**
Found [here](https://www.audiokinetic.com/resources/credits/)
  
Wwise is one of the most popular audio middleware for game engines.

I already prepped a pull request for this :)

PeterShaggyNoble:
Thanks for the request, @made-indrayana. For future reference, though, you don't necessarily need to create an issue for an icon you're already planning to contribute yourself unless you'd like to stake a claim on it while you work on it.

made-indrayana:
@PeterShaggyNoble ah, I see! Then the recommended way to contribute further is to just do a search first on the issues if somebody else is already doing work on that, if not, I'll just do a pull request directly, correct?

PeterShaggyNoble:
Yup 🙂 And if you trip over an unfulfilled request you'd like to take on in the process, all the better! 😉 

There's no _harm_ in opening the issue, it just saves you an extra step if you're already planning on submitting the PR anyway.

PR #5722, opened 2021-05-17
Added Wwise icon

made-indrayana:
wwise

**Issue:** #5721
**Alexa rank:** [247,056](https://www.alexa.com/siteinfo/audiokinetic.com)

### Checklist

### Description
I picked the hex value directly from the brand guideline. I did not respect the buffer zone since this is different than the in game credits, but let me know if I do have to do it.


PeterShaggyNoble (approved):
Everything looks perfect to me, @made-indrayana; nice work 👍🏻

For anyone questioning the Alexa rank: while it is for the parent company's website, Wwise would seem to be their only, or at least primary, product so I think it's safe to take it as an indication of the popularity of Wwise itself.

Issue #5661, opened 2021-05-11
Update Microsoft Azure?


PeterShaggyNoble:
**Brand Name:** Microsoft Azure
**Official resources for icon and color:** n/a

According to [Logopedia](https://logos.fandom.com/wiki/Microsoft_Azure), Microsoft are in the process of rolling out a new logo for Azure. They're not using it on [their website](https://azure.microsoft.com/) yet and they may end up keeping the current, more simplified version as an icon so this one is pending for now.

adamrusted:
Just FYI, I've been sent a Dynamics icon by a Microsoft Partner organisation, which will also need updating. Unfortunately the svg itself is just an embedded image file - so of no real use to us.

PeterShaggyNoble:
Yep, looks to be a part of a larger rebranding they're rolling out that we'll need to keep an eye on.

moodledan:
The latest Azure SVG data here. [azure.txt](https://github.com/simple-icons/simple-icons/files/6494422/azure.txt)

PeterShaggyNoble:
Thanks, @moodledan. However, could you let us know where you sourced that file?

moodledan:
@PeterShaggyNoble This SVG was captured from the Azure Portal https://portal.azure.com/, you need to be signed in to view it. The SVG is embedded in the portal as part of a 'loading' screen.

adamrusted:
> captured from the Azure Portal

Just logged in to check this myself. As @moodledan says it's available in the source code at that URL, and they're already using it as the favicon on that page too - so looks like they've already done the rebrand (just not on the public-facing side!

PeterShaggyNoble:
I think we can proceed with the update of Azure, at least, based on that. I'm having a dig around in there at the moment to see if I can find a monochrome treatment but the only one I've found so far is still using the old logo.

adamrusted:
> I think we can proceed with the update of Azure, at least, based on that. I'm having a dig around in there at the moment to see if I can find a monochrome treatment but the only one I've found so far is still using the old logo.

Given the 'Power Platform' icons I've contributed are in the Fluent Style also, it'd be worth trying to unify our approach to them. I've gone with a 0.5pt gap around each section - granted they're all different 'shapes' in those as opposed to one continuous letter.

PeterShaggyNoble:
We'll probably need to add a cut-out to the inner overlap, at least.

PeterShaggyNoble:
[...]

PR #5745, opened 2021-05-19
Update: Microsoft Azure Icon and Color

moodledan:
Microsoft Azure

**Issue:** #5661 
Alexa rank: N/A

### Checklist
  - [x] I optimized the icon with SVGOMG

### Description
Source, taken from the Microsoft Azure App Service extension for VS Code repo: https://github.com/microsoft/vscode-azureresourcegroups/blob/0a06362e82170fd7f8dc2496286825b1a69cc42b/resources/azure.svg

I've updated the HEX value for the icon, its the only color that forms part of the sourced icon that isn't a gradient start/stop value, it's also the primary accent/link colour used on the Microsoft Azure portal.


_data/simple-icons.json
```diff
@@ -5607,8 +5607,8 @@
         },
         {
             "title": "Microsoft Azure",
-            "hex": "0089D6",
-            "source": "https://developer.microsoft.com"
+            "hex": "0078D4",
+            "source": "https://portal.azure.com"
```
adamrusted, line 5611: While I'm reviewing the SVG - could you please update the source URL to the GitHub URL you used for this new version?

adamrusted (changes requested):
Thanks for the updates @moodledan - the icon matches up perfectly, though there are a few [excess points in the paths](https://user-images.githubusercontent.com/1830867/119240625-c01b8300-bb48-11eb-9226-8d2e2ea48c53.png) which could probably be tidied up before optimization for an even smaller file.

moodledan (commented):
Excess points removed and increased the floatPrecision value to 3 as per default for the project.

moodledan (commented):
Tried to further optimise but then lint check fails so this is the compromise

adamrusted (approved):
Looks good to me, thank you for your work on this @moodledan!
There are a couple of minor points that _may_ be improved at some point, but as you say they look like they're likely to mess with the geometry, so let's leave them for now. It's worth noting that the word 'icon' is no longer required in the `<title>` tag - so if you could merge develop into your branch, and then remove the word - this'll be good to go!

adamrusted:
Thanks for merging develop in, @moodledan - though it looks like the Azure icon itself has also reverted!
Would you mind resubmitting your edited SVG?

moodledan:
What a n00b, resubmitted now.

adamrusted (approved):
Spot on, thanks for your work on this @moodledan 🚀

PR #4717, opened 2021-01-11
Rename Open Nebula as OpenNebula

PeterShaggyNoble:
Their name is consistently stylised as "OpenNebula" (no space) throughout their site so we should update to much.

~~As spaces are ignore in filenames and slugs this shouldn't constitute a breaking change but could you confirm that's correct, @ericcornelissen?~~

PR #5769, opened 2021-05-24
Rename Open Nebula as OpenNebula

PeterShaggyNoble:
Their name is consistently stylised as "OpenNebula" (no space) throughout their site so we should update to much.

Recreation of #4717


adamrusted (approved):
LGTM! Thanks @PeterShaggyNoble 🚀

PR #5777, opened 2021-05-25
Rename 'get' to 'Get' on simpleIcons export

ericcornelissen:
**Issue:** Relates to #4920
**Alexa rank:** n/a
  

### Checklist

### Description

Not strictly necessary, and I would be okay pushing this away one more milestone, but from our side it's a quite simple change that just renames the `get` method on the `simpleIcons` object to `Get`. The reason for this is that, with #4920 merged, `Get` will never be a key occupied by a brand, but there _might_ be a brand whose slug is `get` causing an issue with the existing API.

It might be an idea to push the breaking change out to v6 and only add `Get` here, keeping `get` for backwards compatibility. Any opinions on this from other @simple-icons/maintainers?

EDIT: As per the discussion below, this Pull Request now adds `Get` and keeps `get` for backwards compatibility. This gives users some more time to switch. We will be removing `get` in v6 instead.


PeterShaggyNoble:
I think we should definitely add `Get` now and update all our documentation to use it instead of `get`. And then I'd be in favour of pushing the removal of `get` to v6 to give people a grace period to make the changeover. If we get a request for a brand named "Get" in the meantime we can get around it with the addition of a temporary custom slug.

PeterShaggyNoble (approved):
All looks good to me so, assuming we're all in agreement on the above, once `get` is added back in, this can be merged straight in.

tests/index.test.js
```diff
@@ -59,7 +59,15 @@ icons.forEach(icon => {
     }
   });
 
-  test(`${icon.title} can be found by it's slug`, () => {
+  test(`${icon.title} can be 'Get' by it's slug`, () => {
```
PeterShaggyNoble, line 62: Suggested change here for readbility. Other than that, everything looks good to me.
```suggestion
  test(`'Get' ${icon.title} by its slug`, () => {
```

tests/index.test.js
```diff
@@ -59,7 +59,15 @@ icons.forEach(icon => {
...
+    const found = simpleIcons.Get(slug);
+    expect(found).toBeDefined();
+    expect(found.title).toEqual(icon.title);
+    expect(found.hex).toEqual(icon.hex);
+    expect(found.source).toEqual(icon.source);
+  });
+
+  test(`${icon.title} can be 'get' by it's slug`, () => {
```
PeterShaggyNoble, line 70: And the same here:
```suggestion
  test(`'get'  ${icon.title} by its slug`, () => {
```

PeterShaggyNoble (changes requested):

Issue #1060, opened 2018-11-06
color not embedded in icons


aspiers:
### Kind of issue
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns
  - [ ] The NPM Package
  - [x] The website
  - [ ] Other, namely:

### Description

This is a great resource / website, and it's cool how the color of each icon is explicitly listed as metadata.  However I don't understand why each SVG available for download is monochrome.  I'm quite sure this was a deliberate design decision, but it makes it impossible to download an SVG and instantly import it into (say) an image/video editor and have it immediately be the right color.  An extra step is required to apply the color, and I'm confused why this extra work is being required.  I understand the need to avoid duplication of color values between each SVG and `simple-icons.json`, but surely the website could offer the option to download both monochrome and color versions?

Or even if that's not possible, it would be good to at least document some quick techniques for applying the colors to obtain a colored version of the SVG.  For example:

    sed 's/<path /<path stroke="#ff0000" fill="#ff0000" /' \
        youtube.svg > youtube-red.svg

ericcornelissen:
The primary reason the color is not included in the SVG is because not everyone wants to use the icon with that color. Although it may be common to use the icon in the specified color, other options (specifically `#000000`, `#FFFFFF` and grays) are also common.

Although we do not explicitly state the icons are monochrome, that is more or less the service we provide: monochrome version of brand logos. With regards to importing the icons into an image/video editor I would say that this is (at least currently) not our main focus, we focus mostly on web development (where styling should be done through CSS).

With regards to documenting techniques for applying colors, this would be an interesting challenge. As I said before in web development this should be done through CSS. Otherwise the most straightforward way is to just update the SVG which is a pain to do manually. A tool like [IconJar](https://geticonjar.com/) (see also #975) might help, but it is only available on MacOS. Another option would be to let user specify the color before downloading the icon as e.g. [FlatIcon](https://www.flaticon.com/) does, but the drawback here is that it now takes two clicks to download an icon...

In conclusion, I think we can and should look into this, especially with the redesign (#980) in mind. However, I hope you understand why we do things the way we do and that it might take a while to get your feature request out there 🙂

aspiers:
[...]

PR #5780, opened 2021-05-26
added colours to some files

vivian-dai:
most SVGs now have colours


vivian-dai:
wait, sorry to whoever is reviewing this but `script.py` is irrelevant

fbernhart:
@vivian-dai Thanks for your PR.

However, this is a huge change you're proposing. May I ask you why you're proposing that? How are you using the icons? Maybe there are other ways to achieve a colored version than adding the color to the SVGs. For example through CSS.

About colored icons: There already has been some discussion about this going on in https://github.com/simple-icons/simple-icons/issues/1060, and although the issue is still open, I think common ground is that adding the colors straight to the SVGs themselves is not something we're aiming for at the moment.

So unless the other @simple-icons/maintainers have a different opinion, we'll probably have to close this PR and label it as `won't add`.

If you just need a way to download the icons in their brand color, you might be interested in this proposed feature:

https://github.com/simple-icons/simple-icons-website/issues/32

vivian-dai:
I've been downloading the SVGs then manually adding in a fill for it so I thought it would be nice to just download with the fill. I'll take a look into the website proposal and see if I can add a download with colours option.

ericcornelissen:
I'm closing this as a precaution as this is definitely something we can just merge, this would likely break in many unexpected ways and even with V5 coming up I wouldn't want to risk this change.

For some additional context, we intentionally provide the SVGs without any styling so that users are free to style the SVGs as they see fit. Though there may be some/many users like yourself that want the icons coloured in the brand color we provide, many others use the icons in a different way and setting a fill on the SVG tag itself makes the latter more difficult.

That said, you're not the first to bring this issue to our attention and we would be glad to find an easy solution for users like you. Given that you use the website to get the SVGs the issue linked to by @fbernhart is probably your best bet, if you're really interested in this please leave some feedback on that issue so that we know what we can prioritize.

PR #5804, opened 2021-05-29
Add `simple-icons-pdf` to publish workflow

adamrusted:
As above, adds the PDF repo to our list or tasks during the release of the icons every week.
This shouldn't be merged before https://github.com/simple-icons/simple-icons-pdf/pull/2 is tested and merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -287,8 +287,8 @@ We use the source URL as a reference for the current SVG in our repository and a
 
 If the SVG is sourced from:
 
-- **Branding page**: For an SVG from a branding page the source URL should simply link to the branding page.
-- **Company website**: If the SVG is found on the company website (but there is no branding page) the source URL should link to a common page, such as the home page or about page, that includes the source material.
+- **Branding page**: For an SVG from a branding page the source URL should link to the branding page and not the image, PDF, or archive (such as `.zip`) file.
+- **Company website**: If the SVG is found on the company website (but there is no branding page) the source URL should link to a common page, such as the home page or about page, that includes the source image and not the image file itself.
 - **GitHub**: For an SVG from a GitHub (GitLab, BitBucket, etc.) repository the source URL should link to the file that was used as source material. If the color comes from another file in the repository the URL should link to the repository itself.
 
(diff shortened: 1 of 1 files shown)


ericcornelissen:
I'm currently thinking we should add the following step to [the `website` job in the publish workflow](https://github.com/simple-icons/simple-icons/blob/f0fb9f666a3745b309b946dd71b5718eec1c4c72/.github/workflows/publish.yml#L83-L93), giving the `pdf` trigger (plenty of) time to take effect before updating the website.

```yaml
- run: sleep 5m
```

The reason for this, over triggering the website workflow from the PDF repo, is that I'd like to keep the triggers (and by extension release schedule) centralized here. Admittedly, waiting 5 minutes is not ideal and there is a small chance for (silent) failures as a result.

As an aside, with https://github.com/simple-icons/simple-icons-website/pull/81 I'm looking into making it possible to get this and related Pull Requests merged :+1:

adamrusted:
> The reason for this, over triggering the website workflow from the PDF repo, is that I'd like to keep the triggers (and by extension release schedule) centralized here. Admittedly, waiting 5 minutes is not ideal and there is a small chance for (silent) failures as a result.

Is there any way we could return a status from the PDF repo, and then trigger the website build using that status?
Thinking then the release schedule could be:
- NPM Module
- Fonts / PDFs (if NPM succeeds)
- Website (if NPM and PDF succeeds)

ericcornelissen:
> Is there any way we could return a status from the PDF repo, and then trigger the website build using that status?
Thinking then the release schedule could be:
>
> - NPM Module
> - Fonts / PDFs (if NPM succeeds)
> - Website (if NPM and PDF succeeds)

The dispatch mechanism is asynchronous, i.e. the `curl` request we send receives a response immediately telling us if the workflow was triggered successfully, GitHub does not wait until the workflow finished to respond (which makes sense, given that workflows can take a long time). So, that's not possible.

The closest thing I can think of is implement the sleep and before triggering the website workflow checking if the Fonts/PDFs workflow(s) succeeded. But I don't think that's much better :confused:

adamrusted:
That's fine - I'll implement the `sleep` command for now, and we can improve on it at a later date.

ericcornelissen:
@adamrusted, why did you close this? It's been a while (😅) but I still intend to have a look at getting this set up

adamrusted:
Must've auto-closed when I updated my develop branch... I'll look to re-implement changes in a new branch.

PR #5842, opened 2021-06-03
Fix error in 'icon-precision' lint rule

mondeja:
I supose that has not been triggered until now.


PeterShaggyNoble (approved):
Thanks, @mondeja 👍🏻 What was it that triggered it?

mondeja:
A chained segment with a number with more than 5 decimals of precision, for example `M5 6 7.1234567 8`.

PR #5883, opened 2021-06-08
Remove FedRAMP from Ignored list

service-paradis:
I removed FedRAMP from `.svglint-ignored.json` since we no longer have this icon.


fbernhart:
Well spotted! 👍🏼

ericcornelissen:
I saw this:

image

and got a little worried. Luckily it's all fine, but just a small reminder to try and write descriptive commit messages :smile:

PR #5885, opened 2021-06-08
add denon icon

### LINKED ISSUE #5191: Request: Denon Icon
**Brand Name:** Denon
**Website:** https://www.denon.com/en-us/
**Alexa rank:** 51,955
  

https://en.wikipedia.org/wiki/File:Denon_logo.svg

raeffs:
denon

**Issue:** Close #5191
**Alexa rank:** 68,640

### Checklist

### Description
- I used the [SVG](https://www.denon.com/assets-denonmvp/dist/images/denon-logo.svg) from their website.


PeterShaggyNoble:
Thanks for taking this one on, @raeffs 👍🏻

SVG lines up perfectly for me but, before we merge it, do you think you would be up to redrawing the "O" to get rid of the excess points from the one in the source and maybe trying to clean up some of the ones in the "D", as well?

The only other thing is that we can just run with https://www.denon.com/ for the source here.

raeffs:
I will give it a try 👍🏻

PR #5912, opened 2021-06-10
Improve CI config: caching & refactoring

ericcornelissen:
**Issue:** n/a
**Alexa rank:** n/a
  

### Checklist

#### Motivation

As we recently started to require branches to be up-to-date with `develop` before merging - mainly because of the breaking changes introduced in v5 - so I wondered if we could speed up the CI using caches.

#### Jest Caching

First, I looked into using a cache for Jest. Locally, the cache seems to reduce the time the tests take from ~40 seconds to ~25 seconds on my machine. The GitHub action runs in this Pull Request are showing a reduction in the time tests take from ~55 seconds to ~50 seconds (+ an extra ~2 seconds needed for managing the cache).

Some details regarding the cache. In order to make it easy to set up in the CI, I [configured Jest's cache directory](https://jestjs.io/docs/configuration#cachedirectory-string) to be in the repository root and I ignored the cache folder in the `.gitignore`.

I set the cache key to include the date of when the CI runs as well as the `package-lock.json` file. This way, the CI will refresh the cache on a daily basis (or when the dependencies are updated). I think this is needed because it's okay if the cache is not completely up-to-date (tests will still fail for files that have changed since the cache was created), but if the cache gets too old it will hinder performance.

#### Dependency Caching

For completeness I also added caching of the dependencies, this currently doesn't seem to have much of an effect on Ci run times due to the low number of dependencies.

##### Update lint job

Instead of [running each linter separately](https://github.com/simple-icons/simple-icons/blob/18789ef9e69c3470502bf8241514bc93fcf9766f/.github/workflows/verify.yml#L31-L36), the linting job now uses `npm run lint`, which [runs each linter in series](https://github.com/simple-icons/simple-icons/blob/18789ef9e69c3470502bf8241514bc93fcf9766f/package.json#L36). The result of this change is that if we change anything related to the linters we only need to update this in one place.

##### Sanity check job in publish workflow

I extracted [the "Sanity check" part of the publish workflow](https://github.com/simple-icons/simple-icons/blob/18789ef9e69c3470502bf8241514bc93fcf9766f/.github/workflows/publish.yml#L40-L43) into a separate job as it was duplicated between the "Publish to NPM" and "Publish GitHub Release" jobs. This has made it easier to 1) use the Jest cache for the sanity check, and 2) add `npm run build` as an explicit sanity check.

* * *

_Ignore the note in the first commit that says the cache is invalided after 7 days, this is only if the cache has not been used._


raeffs:
@ericcornelissen just saw this PR and had a look at the workflows as I was wondering what you are doing during the CI runs. Have you considered other optimizations? For example you could do a git diff and just run the test for icons that were affected by changes. 🙂

ericcornelissen:
Nothing specific @raeffs, but suggestions are always welcome :slightly_smiling_face: 

How could we go about implementing that suggestion? If you have a robust and simple solution I definitely think we can do that :+1:

ericcornelissen:
In the spirit of further optimizations, I included dependency caching in this Pull Request. Again, it doesn't seem to help us much (given how relatively few dependencies we have) but it's something, about 1 to 2 seconds per job.

mondeja (approved):
Great work :+1:

PR #5918, opened 2021-06-11
Add Adidas Group

phongapp:
adidasgroup

**Issue:**
**Alexa rank:** [22,446](https://www.alexa.com/siteinfo/adidas-group.com)

### Checklist

### Description
- I got SVG code from View Code of [Adidas-group.com](https://www.adidas-group.com)


PeterShaggyNoble:

Before reviewing, I'd like to discuss whether or not we need to add this one separately to the main Adidas icon. If this as a follow-up to that PR then I don't see a need for it myself but if this is something you have a use for yourself then I think we can go ahead and add it.

huuphongnguyen:
> Thanks for the contribution, @huuphongnguyen 👍🏻
> 
> Before reviewing, I'd like to discuss whether or not we need to add this one separately to the main Adidas icon. If this as a follow-up to that PR then I don't see a need for it myself but if this is something you have a use for yourself then I think we can go ahead and add it.

I think we shoud add this one separately. Adidas Group have some other child brands. In the future, we could need to add the rest.

logo-adidas

ericcornelissen:
If it's a separate brand (which I agree it is) I agree that it _can_ be added, but I do also agree with Peter that I feel a bit hesitant adding this now if there's no real demand for it. @huuphongnguyen, do you have a need for this icon?

PR #5919, opened 2021-06-11
Add Puma

huuphongnguyen:
puma

**Issue:**
**Alexa rank:** [3,529](https://www.alexa.com/siteinfo/puma.com)

### Checklist

### Description
- I got SVG code from View Code of [puma.com](puma.com)


PeterShaggyNoble:

The SVG looks good to me but, again, it looks like you should be able to reduce the precision down to 3 when optimising without creating any issues.

On the colour: from the (PNG) logos available at the end of [this page](https://about.puma.com/en/newsroom/images-and-footage) as well as the colours used on the website, I'd expect a black or grey but I'm not _entirely_ sure on the correct choice.

PeterShaggyNoble:
SVG looks perfect to me now, @huuphongnguyen, thanks. We just need to decide on the colour before merging this. Any of the other @simple-icons/maintainers have any thoughts on that?

service-paradis:
I would expect a darker color for this brand. Maybe something like `#242b2f` which is the color used in the header and footer of their website?
It is also used in the background for multiple pages section (ex. https://about.puma.com/en/this-is-puma/reform).

PeterShaggyNoble:
Yep, I'd agree with that, @service-paradis 👍🏻 What do you think, @huuphongnguyen?

huuphongnguyen:
Yes. I agreed 👍👍.

PeterShaggyNoble:
OK, so, if you can update the PR with that colour then we can merge it straight in.

PeterShaggyNoble:
Hope you don't mind but, in the interests of getting some of our backlog cleared out, I went ahead and made that change for you.

PR #5920, opened 2021-06-11
Add Jordan

huuphongnguyen:
jordan

**Issue:**
**Alexa rank:** Inside Nike website

### Checklist

### Description
- I got SVG code from View Code of [Nike.com/jordan](https://www.nike.com/jordan)


PeterShaggyNoble:
Thanks for the contribution, @huuphongnguyen :+1:

All looks good to me but, as with the others, you should be able to get the precision down to 3 without any distortion of the icon. My own rule of thumb when optimising is to start with 3 and compare to the original. If it lines up perfectly then I'll try 2 (which _rarely_ works) and if it doesn't then I keep increasing the precision by 1 until it does.

PeterShaggyNoble (approved):
Perfect 👍🏻 Thanks again, @huuphongnguyen.

PR #5962, opened 2021-06-15
Add Pull&Bear

phongapp:
pullandbear

**Issue:**
**Alexa rank:** [9,302](https://www.alexa.com/siteinfo/pullandbear.com)

### Checklist

### Description
I got SVG code from View Code of [PullandBear.com](https://www.pullandbear.com/)
I used Sketchapp to adjust and export
I used https://jakearchibald.github.io/svgomg/ to optimize
I used Inscape to compare with source icons to make sure it fit.


PeterShaggyNoble:
Thanks for the contribution, @huuphongnguyen :+1:

I wonder if we could get away with using the "P&B" here as they do for [their app](https://play.google.com/store/apps/details?id=com.inditex.pullandbear)'s icon? If so then an SVG of that can be found on [this page](https://www.pullandbear.com/ie/store-locator.html) after searching for a location. If not then a stacked version of the "P&B" on top of the "PULL&BEAR" as seen in [their favicon](https://static.pullandbear.net/2/static2/itxwebstandard/images/icons/touch-icon-iphone-retina.png?t=20210624032002404318687) and social media sites among other places might be the best option as the "P&B", at least, would be more legible at smaller sizes. What do you and the other @simple-icons/maintainers think?

On the name, they do refer to them pretty consistently as "Pull&Bear" so I think we should go with that. They do capitilise it in a few places but they're not consistent in that so I think title case is fine for us. For the SVG title, the "&" will need to be replaced with "&amp;amp;".

For the colour, although I'm not overly familiar with the brand, their website would lead me to expect the `#313131` used for the logo in the header.

huuphongnguyen:
> Thanks for the contribution, @huuphongnguyen 👍
> 
> I wonder if we could get away with using the "P&B" here as they do for [their app](https://play.google.com/store/apps/details?id=com.inditex.pullandbear)'s icon? If so then an SVG of that can be found on [this page](https://www.pullandbear.com/ie/store-locator.html) after searching for a location. If not then a stacked version of the "P&B" on top of the "PULL&BEAR" as seen in [their favicon](https://static.pullandbear.net/2/static2/itxwebstandard/images/icons/touch-icon-iphone-retina.png?t=20210624032002404318687) and social media sites among other places might be the best option as the "P&B", at least, would be more legible at smaller sizes. What do you and the other @simple-icons/maintainers think?
> 
> On the name, they do refer to them pretty consistently as "Pull&Bear" so I think we should go with that. They do capitilise it in a few places but they're not consistent in that so I think title case is fine for us. For the SVG title, the "&" will need to be replaced with "&amp;".
> 
> For the colour, although I'm not overly familiar with the brand, their website would lead me to expect the `#313131` used for the logo in the header.

I think we should use "Pull&Bear" because it the main brand as we are calling it. "P&B" looks like the small version to help them using on small space, "P&B" made some confused I think.

ericcornelissen:
While I don't think "P&B" for the SVG would confuse people, from my experience with the brand I would be included to agree that we should go with the full wordmark version for the SVG. But, if a majority of the @simple-icons/maintainers thinks we should go with "P&B" I think that's fine as well.

If we do decide to continue with the full wordmark than the current SVG gets my approval 👍 

In the meantime, @huuphongnguyen, can you update the title from "Pull and Bear" to "Pull & Bear" following [Peter's comment](https://github.com/simple-icons/simple-icons/pull/5962#issuecomment-867576791).

PR #5965, opened 2021-06-15
Add Neuralink

phongapp:
neuralink

😄 Elon Musk is working on 3 great companies: Tesla, SpaceX, Neuralink. We got Tesla and SpaceX, so Neuralink as well. 

**Issue:**
**Alexa rank:** [59,613](https://www.alexa.com/siteinfo/neuralink.com)

### Checklist

### Description
- I got SVG code from View Code of [Neuralink.com](https://neuralink.com/)


PeterShaggyNoble:

It looks like there are a few excess points in the "N" that you should be able to remove, though and, as usual, a precision of 3 should suffice.

On the colour, I genuinely don't know what we should go with here! My first instinct on seeing their website was to go with one of the pastels but they're all used equally throughout the site with none standing out as the primary colour. A case could probably be made for `#b4bdcb` as it's used more prominently than the others on their homepage but I don't think it's a very strong case. SO, in the absence of any brand guidelines, the white's probably as good a choice as any. The only other viable choice might be black which is used in their favicon and also for the logo in their Wikipedia entry. What do you think? Do any of the other @simple-icons/maintainers have a preference?

ericcornelissen:
I'd personally be in favor of `#b4bdcb`, but would be okay with `#000000` as well. `#ffffff` does seem like an odd choice to me though :thinking:

Issue #5958, opened 2021-06-15
Request: Cytoscape.js Icon


SimonGolms:
**Brand Name:** Cytoscape.js
**Website:** https://js.cytoscape.org/
**Alexa rank:** 180,149 https://www.alexa.com/siteinfo/cytoscape.org
**GitHub stars:** 7.7k https://github.com/cytoscape/cytoscape.js

**Official resources for icon and color:** https://github.com/cytoscape/cytoscape.js/blob/master/documentation/img/cytoscape-logo.svg

PR #5968, opened 2021-06-15
feat: add cytoscape.js icon

SimonGolms:
cytoscapedotjs

**Issue:** #5958
**Alexa rank:** 180,149 https://www.alexa.com/siteinfo/cytoscape.org

### Checklist

### Description
I picked the hex value from [raw](https://github.com/cytoscape/cytoscape.js/blob/97c27700feefe2f7b79fca248763049e9a0b38c6/documentation/img/cytoscape-logo.svg) svg file and used Figma/Sketch and SVGOMG for further adjustments and optimizations


service-paradis:
I think I would keep the square around the logo. The current logo would fit more for [Cytoscape](https://cytoscape.org/logos.html), but not that much for Cytoscape.js.
I would expect the following for Cytoscape.js which is a ply of the JS logo:
image

_In fact, we can even find the path for the `j` and the `s` when taking the logo in the header at https://js.cytoscape.org/_

SimonGolms:
Thanks for checking, I wasn't sure about this one, sounds very reasonable that way! 
I'll be happy to make the change, unless you're already on it?

service-paradis:
> Thanks for checking, I wasn't sure about this one, sounds very reasonable that way!
> I'll be happy to make the change, unless you're already on it?

Great, I'll let you go if you're ok with it! :)

SimonGolms:
simpleicons

Icon updated

PeterShaggyNoble:
You timed that update perfectly, @SimonGolms; I got the notification at the exact same moment this PR popped up next on my list to review!

I'm seeing a couple of small differences in some of the circles - with [this one in the leftmost circle](https://user-images.githubusercontent.com/15157491/122581618-2663ca80-d04f-11eb-80b4-f7604d48a974.png) being the most pronounced - between your version (in red) and the source. That might simply, but not necessarily, be a result of needing to set the precision a little higher when optimising. Would you mind taking another look, please?

SimonGolms:
cytoscapedotjs (1)

@PeterShaggyNoble here it is - I used now a precision of 4 on SVGOMG :)

PeterShaggyNoble (approved):
That looks spot-on to me now, @SimonGolms, nice work 👍🏻 And welcome to Simple Icons 🙂

Issue #5959, opened 2021-06-15
Request: React Table Icon


SimonGolms:
**Brand Name:** React Table
**Website:** https://react-table.tanstack.com/
**Alexa rank:** 42,379 https://www.alexa.com/siteinfo/tanstack.com
**GitHub Stars:** 14.4k https://github.com/tannerlinsley/react-table
**Official resources for icon and color:** https://github.com/tannerlinsley/react-table/blob/master/docs/src/images/emblem-light.svg

PR #5979, opened 2021-06-16
feat: add react table icon

SimonGolms:
reacttable

**Issue:** #5959 
**Alexa rank:** 42,155 https://www.alexa.com/siteinfo/tanstack.com

https://www.alexa.com/siteinfo/tanstack.com

### Checklist

### Description
I picked the most prominent color value from the [raw](https://github.com/tannerlinsley/react-table/blob/8c77b4ad97353a0b1f0746be5b919868862a9dcc/docs/src/images/emblem-light.svg) svg file and compared it to the primary color on the [web page](https://react-table.tanstack.com/), which are identical.
I used Sketch and SVGOMG for further adjustments and optimizations


PeterShaggyNoble:
Thanks for the contribution, @SimonGolms 👍🏻

Upon seeing the source SVG, my first instinct was to give it the same treatment you did here but then I saw [this image](https://github.com/tannerlinsley/react-table/blob/master/docs/src/images/og-dark.png) and wondered if we shouldn't do the reverse, as below? Has the stroke been removed there or has it simply blended into the background? What do you and the other @simple-icons/maintainers think? I definitely prefer your outlined treatment.

ericcornelissen:
I'd be inclined to go with the version currently in this Pull Request. I feel like the lines are "missing" from the version @PeterShaggyNoble linked (called "og-dark.png") because the background is the same color as the lines. That said, I do think both are valid.

Given your original instinct was to go with the current version as well, @PeterShaggyNoble, I'd say we continue with that. After all, it can always be changed later.

PeterShaggyNoble (approved):
In that case, everything looks perfect here 👍🏻 Nice work, @SimonGolms, and thanks again for the contribution.

PR #5989, opened 2021-06-17
Update Foursquare & Add Foursquare City Guide

### LINKED ISSUE #5977: Update Foursquare
Looks like Foursquare have a new logo now
image

**Brand Name:** Foursquare

- https://foursquare.com/brand/

service-paradis:
image

**Issue:** Close #5977
**Alexa rank:** [~3.5k](https://www.alexa.com/siteinfo/foursquare.com)
  

### Checklist

### Description

- I took the SVG found in the footer at https://foursquare.com/
- I used Inkscape to scale


service-paradis:
After looking more into this one, I think the old logo is now for their product called [Foursquare City Guide](https://fr.foursquare.com/city-guide)
image

So instead of updating the current brand, I'm wondering if we should rename the current brand to "Foursquare City Guide" and add another one for Foursquare with the logo in this PR...

If so, it would be a breaking change...

ericcornelissen:
> So instead of updating the current brand, I'm wondering if we should rename the current brand to "Foursquare City Guide" and add another one for Foursquare with the logo in this PR...
>
> If so, it would be a breaking change... What other @simple-icons/maintainers think of this one?

I'd say we can just update the logo for Foursquare and add a new one for Foursquare City Guide. If I'm understanding this correctly that would have correct end result and require no breaking changes.

service-paradis:
It is not a _hard_ breaking change, for sure, but I assume most people using the current icon want to display the icon for "Foursquare City Guide". So the icon for these people will be wrong and they will need to update the slug if they want the original icon.

ericcornelissen:
> It is not a _hard_ breaking change, for sure, but I assume most people using the current icon want to display the icon for "Foursquare City Guide". So the icon for these people will be wrong and they will need to update the slug if they want the original icon.

Hmm, I guess that's a fair point. Technically they're using the wrong icon if they're using the "Foursquare" icon for "Foursquare City Guide", but the confusion is understandable and not entirely their fault.

On the other hand, people that are actually using it for Foursquare, _correctly_, expect to have the correct logo. So from that point of view it should be updated.

I'm personally of the opinion that the people using the icon for the "wrong" brand should have to deal with this rather than the people using the icon for the "right" brand, but I would love to hear some more opinions.

If we do decide to go ahead and merge this in as "Foursquare" I do think we should try to release "Foursquare City Guide" in the same release and add a note about it to the release notes.

PeterShaggyNoble:
Interesting conundrum; I agree with both arguments being out forth and for me it _would_ be a breaking change. Ultimately, though, I think I land on @ericcornelissen's side of things: Foursquare is Foursquare, that they've separated the branding of the company from the City Guide is on them not us and we should update our Foursquare icon now for those using it to represent the company and, with it, add back in the old icon as Foursquare City Guide. To me, it would be similar to when they separated check-ins out into Swarm and rebranded Foursquare - at the time, most people would have been using the Foursquare icon to link people to checking in.

That said the icon and colour look perfect to me 👍🏻 I think, though, we should use the brand page for the `source` as well as the `guidelines` - there's an SVG in the footer of that page. And, to _ensure_ it goes out in the same release, let's add the old icon back in in this PR as Foursquare City Guide.

service-paradis:
Thank you both for your feedback.
So I just pushed the mentionned changes:

- Uses the brand page as source
- Add Foursquare City Guide entry with the previous Foursquare icon

It gives the following icons:
image
image

PeterShaggyNoble (approved):
Thanks, @service-paradis. Given the 👍🏻s on Eric's & my comments, it looks like we're in agreement in making this change now.

Issue #5936, opened 2021-06-12
Libera Chat


MrTimscampi:
**Brand Name:** Libera Chat
**Website:** https://libera.chat/
**Alexa rank:** 253,622
  

**Official resources for icon and color:**
  
https://libera.chat/static/img/libera-color.svg

foo-dogsquared:
I'm surprised it gained quite some traction that fast.

On the other hand, the source code of the site is also open. You can find the logo [here](https://github.com/Libera-Chat/libera-chat.github.io/tree/main/static/img).

PR #6009, opened 2021-06-22
Add Libera.Chat logo

foo-dogsquared:
image

**Issue:** #5936 
**Alexa rank:** ~240k as of 2021-06-22
  

### Checklist

### Description

Don't know whether the dot is present or not but I put it there since the logo seems to insist it.


PeterShaggyNoble:
Thanks for this one, @foo-dogsquared 👍🏻

SVG looks good to me but there are some excess points in the ends of some of the rays that you might be able to clean up, and you should be able to get the precision down to 4 when optimising.

On the name, outside of the logo I've so far only found one occurrence where they refer to themselves as "Libera.Chat", other than that, they're pretty consistent in calling themselves "Libera Chat" so I think that's what we should use.

Not convinced of the colour but it's not clear from their website what it _should_ be so I can't offer a better suggestion; let's run with the pink.

Finally, could you update the source URL to point to the website's homepage rather than the logo, please?

adamrusted (commented):
Controversial opinion: Do we add a gap between the pink/purple such as I've done in the below preview @simple-icons/icon-review? Happy for this path to be used if so, can contribute if someone can review.

liberachat

LitoMore:
They are on GitHub, we can ask them for advise.

Here is their GitHub organization: https://github.com/Libera-Chat

adamrusted:
@swantzter - would you or another member of the Libera Chat team be able to take a quick look over this PR and provide feedback? Would be great to be able to include the icon!

libera-allie:
Hiya, Allie from Libera.Chat here: Thanks for doing this! Clearing up some points from above: 


1. "Libera.Chat" is the IRC network, "Libera Chat" is the organisation that runs the IRC network. The logo is for the IRC network, so "Libera.Chat" is preferred. 

2. Colour-wise, for monochrome versions of the logo, take the pink from our [full-colour logo](https://libera.chat/static/img/libera-color.svg) - I was sure we'd written a style guide for this, but apparently not yet. One for the to-do list!

3. If possible, it'd be good to stick to the original logo form if possible, for consistency's sake. 

Again, this is very neat and will no doubt help folk make charming little Libera.Chat channel badges for their READMEs!

Cheers,
Allie @ Libera.Chat

libera-allie:
> Got it! Is there a logo we could/should use for the organisation? No stress if not, we can add the IRC network as a 'product' of the company.

Nope, the organisation doesn't really matter outside of the context of the IRC network - you can conflate the two because "Libera Chat" the org will never do anything other than "Libera.Chat" the IRC network.

> Just to confirm, by this you mean the original version presented in this PR? More than happy to stick with that if that works better for you!

Your understanding is correct. 

Chers,
Allie

adamrusted (approved):
Merging based on feedback from @libera-allie

Issue #5628, opened 2021-05-07
Restructure and/or improve the Contributing Guidelines


ericcornelissen:
This topic has come up in several recent Pull Request making changes/adding to [the Contributing Guidelines](https://github.com/simple-icons/simple-icons/blob/8a38721ed7239168539846db1ece01bb759dad59/CONTRIBUTING.md), and it seems to many of us that it can use some improvements, making it more concise and easier for first time contributors to follow.

This issue is just a reminder to tackle this problem, I have no concrete suggestions on how we should do this.

jorgeamadosoria:
reading https://github.com/simple-icons/simple-icons/issues/5385, I think adding StackOverflow queries as a metric for technology / programming languages is a good one to add to the guidelines.
Something like, say, more than 15k queries in Stack Overflow for a given technology or technical subject should be a good threshold, going by the discussion on that issue.

service-paradis:
The guidelines have been recently updated. I think we can close this issue for now. We could reopen a new one with more precise improvement ideas when needed.

PR #6023, opened 2021-06-24
Create VERSIONING.md

ericcornelissen:
**Issue:** Relates to #5628
**Alexa rank:** n/a
  

### Checklist

### Description

This removes [the "Versioning" section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#versioning) from the Contributing Guidelines and moves it into a new file called `VERSIONING.md` (suggestions for a different name are welcome, but I feel like I've seen this around). The new file contains approximately the same text as what was originally in the Contributing Guidelines, just reworded some things. On top of that I added a section on our release schedule.

Suggestions for the new file, whether improvements or other content, are very welcome!


VERSIONING.md
```diff
@@ -0,0 +1,17 @@
+# Versioning
+
+We use [Semantic Versioning](https://semver.org/) to version Simple Icons. In short, this means that version number are structured as `MAJOR.MINOR.PATCH`. For example, version `3.1.4` is major version `3`, minor version `1`, and patch `4`. Increasing each of this numbers implies certain kinds of changes.
```
PeterShaggyNoble, line 3:
```suggestion
We use [Semantic Versioning](https://semver.org/) to version Simple Icons. In short, this means that version numbers are structured as `MAJOR.MINOR.PATCH`. For example, version `3.1.4` is major version `3`, minor version `1`, and patch `4`. Increasing each of these numbers implies certain kinds of changes.
```

VERSIONING.md
```diff
@@ -0,0 +1,17 @@
...
+
+| Version number increase | Kinds of changes |
+| :---- | :---- |
+| _Major_ | Removed icons; Renamed icons; Breaking API changes |
+| _Minor_ | New icons; API changes |
+| _Patch_ | Updated SVGs; Updated metadata |
+
+## Release Schedule
+
+_Minor_ releases and _patches_ are scheduled on a weekly basis and are generally release on Sundays.
```
PeterShaggyNoble, line 15:
```suggestion
_Minor_ releases and _patches_ are scheduled on a weekly basis and are generally released on a Sunday.
```

VERSIONING.md
```diff
@@ -0,0 +1,17 @@
...
+
+_Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs. That is, approximately every 6 months a normal Sunday release is a major release instead of a _minor_ release or _patch_.
```
PeterShaggyNoble, line 17:
```suggestion
_Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs. That is, approximately every 6 months a normal Sunday release is a _major_ release instead of a _minor_ release or _patch_.
```

PeterShaggyNoble (changes requested):
Looks good to me, @ericcornelissen. Just a few typos and inconsistencies to clean up and this should be good to go.

ericcornelissen:
I was thinking, should we add a section on when old versions are deprecated?

I believe we said we would deprecated the previous major version roughly halfway through a major release's lifecycle - i.e. ~3 months after a new major release. The details for this are up for debate, but any opinions on adding this @simple-icons/maintainers?

ericcornelissen:
To ease the review of this Pull Request if we decide to go ahead and include a section on when old version will be deprecated, I propose we add the following:

```diff
  # Versioning

  We use [Semantic Versioning](https://semver.org/) to version Simple Icons. In short, this means that version numbers are structured as `MAJOR.MINOR.PATCH`. For example, version `3.1.4` is major version `3`, minor version `1`, and patch `4`. Increasing each of these numbers implies certain kinds of changes.

  For Simple Icons, given a change to the version number you can expect the following kinds of changes:

  | Version number increase | Kinds of changes |
  | :---- | :---- |
  | _Major_ | Removed icons; Renamed icons; Breaking API changes |
  | _Minor_ | New icons; API changes |
  | _Patch_ | Updated SVGs; Updated metadata |

  ## Release Schedule

  _Minor_ releases and _patches_ are scheduled on a weekly basis and are generally released on a Sunday.

- _Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs. That is, approximately every 6 months a normal Sunday release is a _major_ release instead of a _minor_ release or _patch_.
+ _Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs. That is, approximately every 6 months a normal Sunday release is a _major_ release instead of a _minor_ release or _patch_. Approximately 3 months after a _major_ release, the _major_ release preceding it will be deprecated.
```

jorgeamadosoria:
> LGTM @ericcornelissen 🎉
> Will await some feedback on other @simple-icons/maintainers before merging though.

looks good to me. LGTM

adamrusted:
> What does it mean for the previous major release to not be deprecated (what do we do in those 3 months after a new major release)?

So for example when v5 came out, v4 remains available, but not updated. After 3 months v4 is deprecated on npm, so anybody installing that version will get a warning that the version is no longer supported.

It just gives our users some notice of change-over without us having to "support" too many versions. Ideally, users should always use the up to date version - but we should allow 3 months after a major release for people to adopt the new version.

ericcornelissen:
That is actually a very fair questions @sachinraja, and it could be worth being explicit about. Typically "support" in OSS is limited to releasing bugfixes and (security) patches. However, while I think we should support in that way, those are not too relevant for us as we rarely have bugs or security issue due to the simple nature of our JS package.

A more interesting question for us would perhaps be: do we provide "support" for updates the icons in the previous major release? I would say, only on request if there's a legitimate reason the user can't upgrade to the latest major version right now.

sachinraja:
> Typically "support" in OSS is limited to releasing bugfixes and (security) patches.

I think this is the intent for a lot of people, but most often libraries do not do this and only update previous versions when requested.

> I would say, only on request if there's a legitimate reason the user can't upgrade to the latest major version right now.

That sounds good to me.

ericcornelissen:
> I think this is the intent for a lot of people, but most often libraries do not do this and only update previous versions when requested.

I agree, and to be explicit about it it's basically what I'm proposing we do with SVGs (and also with our JavaScript code, should it need a patch).

Also, to share my opinion on the matter, for your average OSS project I think this is a very fair approach. There's limited bandwidth to fix things, so testing old versions for problems and fixing them isn't a priority unless it actually affects one of your users.

ericcornelissen:
I added a new section called "Deprecation" that explains the deprecation process and what we mean by "supporting" an old major version. Let me know what you think :slightly_smiling_face:

sachinraja (approved):

adamrusted (approved):

service-paradis (approved):

Issue #5942, opened 2021-06-13
RxJs


PitPietro:
Brand Name: RxJs
 Website: [rxjs.dev](https://rxjs.dev)

service-paradis:
We already have the same icon under the name [ReactiveX](https://simpleicons.org/?q=reactivex). We might add RxJs as an alias.

Issue #3263, opened 2020-06-26
Request: RxJS


Ninad99:
**Name:** RxJS
**Website:** https://rxjs.dev/
**Alexa rank:** 260,308

**Official resources for icon and color:** This is the [logo](https://rxjs.dev/assets/images/logos/logo.png) used in the website.

PeterShaggyNoble:
Thanks for the request, @Ninad99 

There's a monochrome SVG in one of their GitHub repos that we could use: https://github.com/ReactiveX/rxjs/blob/master/resources/CI-CD/logo/svg/RxJs_Logo_Black.svg

Not sure on the colour, black seems pretty dominant on their site (but that may be temporary while showing their support for BLM) but I'd actually lean towards the pink: `#d81b60`.

Would you like to have a look through [our contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and give this a try yourself?

Ninad99:
Sure!

> Thanks for the request, @Ninad99
> 
> There's a monochrome SVG in one of their GitHub repos that we could use: https://github.com/ReactiveX/rxjs/blob/master/resources/CI-CD/logo/svg/RxJs_Logo_Black.svg
> 
> Not sure on the colour, black seems pretty dominant on their site (but that may be temporary while showing their support for BLM) but I'd actually lean towards the pink: `#d81b60`.
> 
> Would you like to have a look through [our contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and give this a try yourself?

NovaGL:
Duplicate of #3354 

If you think this is in error. Please reopen the issue

PR #6029, opened 2021-06-25
Add RxJS alias to ReactiveX

service-paradis:
image

**Issue:** #3263, #5942
**Alexa rank:** [~178k](https://www.alexa.com/siteinfo/rxjs.dev)
  

### Checklist


PeterShaggyNoble:
Looks good to me, @service-paradis  but what would you reckon to `#d81b60` for the colour instead? Looks like that shade is used everywhere _but_ the header.

service-paradis:
I agree, it is a better choice! I think the class name `mat-primary` convince me too fast 😅

PR #6114, opened 2021-07-18
Add `Douban`, `Douban Read` license

LitoMore:
### Checklist

### Description

Add `license` field according to https://github.com/simple-icons/simple-icons/pull/6084#pullrequestreview-708965455.

I also added the license for the Douban icon.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -2434,13 +2434,19 @@
             "title": "Douban",
             "hex": "007722",
             "source": "https://zh.wikipedia.org/wiki/Douban",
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Thanks for this @LitoMore - just one question: judging by the URL it seems like `"license"` might make more sense than `"guidelines"` :thinking: But as none of the maintainers reads mandarin, I'll leave the decision between `"guidelines"` and `"license"` up to you. Just want to confirm that you considered both.

LitoMore:
Good point. Use `license` would be better. Let me update this to `license`.

ericcornelissen (approved):

Issue #6091, opened 2021-07-08
Request: Zenodo icon


janniklasrose:
**Brand Name:** Zenodo
**Website:** https://zenodo.org
**Alexa rank:** 26,379
  

**Official resources for icon and color:**
  
  
SVG files and other information available at https://about.zenodo.org/

service-paradis:
Thank you @janniklasrose for the request. Do you think you would be up trying to do a pull request?

You can take a look at our contributing guidelines if you want to give this a

janniklasrose:
Thanks! I have read the guidelines and it seems straightforward. I will do a PR soon.

PR #6162, opened 2021-07-24
Zenodo

janniklasrose:
**Issue:** #6091
**Alexa rank:** 28,954
  

### Checklist

### Description

I set the hex value to `000000` because the primary colour of the logo is white on a linear gradient of `0047A8` to `2BBCFF` . Should I put either of these colours, or the "average", or something else?

zenodo


_data/simple-icons.json
```diff
@@ -10413,6 +10413,12 @@
             "hex": "3EA8FF",
             "source": "https://zenn.dev/mediakit"
         },
+        {
+            "title": "Zenodo",
+            "hex": "000000",
+            "source": "https://about.zenodo.org/static/img/logos/zenodo-black.svg",
```
ericcornelissen, line 10419: Linking to a specific file can be a bit fragile as files tend to be moved around - hence we use the page on which the SVG is found as a source.

```suggestion
            "source": "https://about.zenodo.org",
```

_data/simple-icons.json
```diff
@@ -10413,6 +10413,12 @@
             "hex": "3EA8FF",
             "source": "https://zenn.dev/mediakit"
         },
+        {
+            "title": "Zenodo",
+            "hex": "000000",
```
ericcornelissen, line 10418: > I set the hex value to `000000` because the primary colour of the logo is white on a linear gradient of `0047A8` to `2BBCFF`. Should I put either of these colours, or the "average", or something else?

We usually go with a calculated midpoint in case of gradients, in this case [that would be `#1682D4`](https://meyerweb.com/eric/tools/color-blend/#0047A8:2BBCFF:1:hex). As I'm not seeing a particularly prominent blue on the website, nor is any specific hex mentioned, I think we should go with that for now.

ericcornelissen (changes requested):
The SVG looks spot on, nice work! Got two requests for changes to the metadata but after that this should be good to go :smile:

janniklasrose:
I have updated the metadata as requested. With the new blended colour, the preview looks better.

zenodo

janniklasrose:
I am struggling to see how to address the failed `Lint` workflow reporting that the horizontal size is not exactly 24px.
In Inkscape, I am scaling the image to a width of 24px (and centre the logo horizontally & vertically). Saving from Inkscape creates a much longer path resulting in a larger (7kb) file. Compressing in SVGOMG then causes a width of 24.001px, ...

ericcornelissen:
This could be either due to a quirk in Inkscape or due to SVGO(MG). You can try to run the SVG linter (`npm run svglint`) directly on the SVG you get from Inkscape and see if it reports an issue about the size. If it doesn't, it's likely that SVGO(MG) messes up the path ever so slightly, in that case try to increase the precision of SVGO(MG) by 1.

janniklasrose:
Increasing precision in SVGO(MG) seems to have done it. Thanks for the help!

ericcornelissen (approved):

ericcornelissen:
Great, thanks for the contribution @janniklasrose :tada:

PR #6179, opened 2021-07-29
Drop `package-lock.json`

LitoMore:
### Overview

- Remove `pakcage-lock.json`
- Drop lockfile related code from GitHub Actions scripts
- Drop lockfile maintenance configurations from `renovate.json5`
- Drop lockfile logic from `bump-version.js`


package-lock.json
```diff
@@ -1,8 +1,5709 @@
 {
```
mondeja, line 1: What are the benefits for this change? It seems that `v2` is not totally backwards compatible with previous versions of NPM 🤔

.github/workflows/verify.yml
```diff
@@ -8,10 +8,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Use Node.js 12.x
```
mondeja, line 11: I would go with 14 because [is the current active branch in Node releases lifecyle](https://nodejs.org/en/about/releases/#releases). It would be also awesome to test in different versions in master or periodically, but not for all PRs.

package-lock.json
LitoMore: ~~npm updated this file automatically when I run `npm install`.  If we decide to use Node.js 16 we should let its default behavior not produce any unstaged changes.~~

Good point. I tested with Node.js 14 and it will rollback to `lockfileVersion: 1`.

This change is not friendly with npm 6 (Node.js 12, 14) users.

.github/workflows/verify.yml
LitoMore: Good point. I am going to close this pull request and take more time to research a better solution for this.

Thank you so much.

ericcornelissen: To be more precise, v14 is the active LTS whereas v16 is the active "branch". The relevant difference between v14 and v16 for us, as mentioned in the other comment, is the change from NPM v6 to NPM v7 by default (though you can upgrade to NPM v7 before Node v16) which comes with a change to the `package-lock.json` file format.

To avoid confusion like this, I think it might be best to upgrade to Node v16/NPM v7. Even though prior NPM versions don't fully support the new format, it's still able to install the dependencies in most cases (I'm not sure on the details but I'm pretty sure it won't matter much to this project). Importantly, NPM prior to v7 won't downgrade the `package-lock.json` to v1, whereas NPM v7 will upgrade it to v2.

In any case, this transition period from NPM v6 to NPM v7 is a bit of a mess and I'm also not sure what the best way to deal with it is. So in that regard, thanks for bringing this up for dicussion @LitoMore :smile: 


> Consistent with the Node.js version in Docker.

To this point, I don't think it's necessary to align the versions "just because". The reason the Dockerfile has v16 is because it got updated automatically by Renovate and since it's just intended for running SVGO in a Docker container I didn't bother telling Renovate to stop updating it. So, unless someone actually wants to use the Dockerfile for development (if that's you @LitoMore, please say so) I don't see a reason spend time aligning things with the Dockerfile.

LitoMore: @ericcornelissen Agree! Thank you for your answer!

> I don't think it's necessary to align the versions "just because".

I just noticed the Docker is using Node.js 16, but I didn't realize the Docker is not for the development.

ericcornelissen: > I just noticed the Docker is using Node.js 16, but I didn't realize the Docker is not for the development.

The confusion is understandable :slightly_smiling_face:  just wanted to clarify.

package-lock.json
ericcornelissen: > Good point. I tested with Node.js 14 and it will rollback to `lockfileVersion: 1`.
>
> This change is not friendly with npm 6 (Node.js 12, 14) users.

Good thing I read this and tested it myself :sweat_smile:  For some reason [I thought npm 6 does not rollback to `lockfileVersion: 1`](https://github.com/simple-icons/simple-icons/pull/6179#discussion_r679402311) but as you said it does do that.

This complicates things a bit as it doesn't make the decision as clearcut as I'd hoped. Staying with `lockfileVersion: 1` is annoying for people with npm v7. Switching to `lockfileVersion: 2` is annoying for people with npm 6 (or lower). There's a couple of solution but none of them actually solve the problem. One is to use a `Dockerfile` for development, but given the low bar for entry we aim for I don't like that. A second is to use an `.nvmrc` file to specify the Node (and by extension NPM) version, which I don't like for the same reason. A third solution is to add  `"npm": ">=7.0.0"` to [the engines in `package.json`](https://github.com/simple-icons/simple-icons/blob/0ccb7cbfaf55b7151c5fbeec7075c53fbbc18a95/package.json#L49-L51), but that's only a solution if the contributes has set the `engine-strict` flag manually which, again, I don't think we should expect from our contributors.

Clearly, at some point we need to make the switch, given my experience reviewing Pull Requests for this project, a majority of contributors is still using NPM v6 (presumably because they installed Node quite some time ago and ignore the notice to update NPM) so given the above I actually think we should stick with `lockfileVersion: 1` until we see more Pull Requests that accidentally update `package-lock.json`.

In the meantime, we could implement all of the above in an attempt to make it easier on those contributors that use the relevant tools/configurations. @LitoMore (and @mondeja), what do you think? If you agree with my (new) opinion, do you want to implement it here or should I create a (new) Pull Request for that?

ericcornelissen: @LitoMore, did you see this comment?

LitoMore: @ericcornelissen Sorry I missed it.

Agrees. Let's stick with `lockfileVersion: 1`.

LitoMore: Or just drop the lock file? Add `package-lock=false` to `.npmrc`. As [Sindre Sorhus does](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514).

ericcornelissen: > I didn't find a way to stick `lockfileVersion: 1`. Any thought?

Not sure if I follow, seems like your PR doesn't change `package-lock.json` anymore - did you manage to resolve the problem yourself?

> Or just drop the lock file? Add `package-lock=false` to `.npmrc`

Interesting suggestion, didn't know about that. I'm not totally against it, but the reasons lined out in the thread you linked do not apply here. This project [only has `devDependencies`](https://github.com/simple-icons/simple-icons/blob/96c32669d2d450bc8c85c7b748b3bb9f2876c312/package.json#L20-L32) and I, as pointed out by others in the thread, a `package-lock.json` provides is with a deterministic dev environment.

What do you think about that @LitoMore?

LitoMore: > Did you manage to resolve the problem yourself?

There are too many conflict in the file and hard to solve. We could handle it after determining the solution.

> This project only has devDependencies.

I think it's OK to drop lock file. Even we can face problems in the development environment earlier. If there is any dependency problem and hard to solve, just pin it.

ericcornelissen: > There are too many conflicts in the file and hard to solve. We could handle it after determining the solution.

Sure

> I think it's OK to drop lock file. Even we can face problems in the development environment earlier. If there is any dependency problem and hard to solve, just pin it.

We can pin our `devDependency`s - which we already do - but not their dependencies, that's what the lockfile provides us with.

But again, I'm not against it. I'm definitely open to at least trying it out. In that case, could you add `.npmrc` file and revert all changes so that we're still on Node v12 and npm v6 in this Pull Request? I propose we take the discussion/work for upgrading to Node v16 in the CI, adding `.nvmrc` and updating the `Dockerfile` to a different Pull Request/issue - the sooner this is merged the sooner other contributors benefit from the change :slightly_smiling_face:

LitoMore: Updated.

ericcornelissen (changes requested):
Two further changes are required before this can be merged:

- Remove [line 7 of `.gitattributes`](https://github.com/simple-icons/simple-icons/blob/0a3621df0e5cb43a7a818e5c29cb07494ada5c39/.gitattributes#L7).
- Remove the usage of `package-lock.json` from [the `bump-version.js` script](https://github.com/simple-icons/simple-icons/blob/0a3621df0e5cb43a7a818e5c29cb07494ada5c39/scripts/release/bump-version.js). To test if the updated script works you can use `node ./scripts/release/bump-version.js v5.42.0`.

package-lock.json
mondeja: I agree overall with the solution, but just for curiosity: how [renovate security updates](https://github.com/simple-icons/simple-icons/pull/6199) are handled after this? Are just ignored? Additionally, if we do this, we must remove [this part of the renovate configuration](https://github.com/simple-icons/simple-icons/blob/0a3621df0e5cb43a7a818e5c29cb07494ada5c39/.github/renovate.json5#L9-L20)?

sachinraja:
If we want to lock the npm version, we can specify `engine-strict=true` in the .npmrc file and contributors will not have to set anything. They will not be able to install the project without an npm version that matches the range specified in `engines.npm`.

<details>
<summary>Also my thoughts on deleting the lockfile</summary>
That does not seem reasonable considering that we have <code>devDependencies</code> (as @ericcornelissen said) and we can never ensure that we are using the same version as our users. Our dependency can publish a patch version and users will be using that new version. They will definitely catch problems before a contributor does. Here's a great article on this: <a href="https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/">https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/</a>.
</details>

ericcornelissen:
> If we want to lock the npm version, we can specify `engine-strict=true` in the .npmrc file and contributors will not have to set anything. They will not be able to install the project without an npm version that matches the range specified in `engines.npm`.

That is an interesting suggestion as well. I tested it and it does work (after changing the incorrectly named ["engine" field](https://github.com/simple-icons/simple-icons/blob/d2bd9906d8a0a2dd2aef855dbd223bc3937f4ce6/package.json#L49) to "engines", oops :sweat_smile:). However, for this project I'm not quite sure if we really want to prevent people with `npm` v7 or greater from being unable to install dependencies. I think it's quite common for our contributor base to not use a tool like `nvm` (if you don't switching to npm v6 can be a real pain).

> Also my thoughts on deleting the lockfile:
> That does not seem reasonable considering that we have `devDependencies` (as @ericcornelissen said) and we can never ensure that we are using the same version as our users. Our dependency can publish a patch version and users will be using that new version. They will definitely catch problems before a contributor does. Here's a great article on this: https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all.

I still share this concern, but I don't expect this to be a big problem for use as we 1) don't have a lot of dependencies and 2) don't do anything fancy with them. For this reason I'm open to **trailing** the no `package-lock.json` approach. But I completely agree that as soon as something goes wrong with that, we add it back in. What do you say @sachinraja?

An alternative to fix the problem we're having: we can add a pre-commit hook that reverts any changes to `package-lock.json`. New dependencies are rarely added and if you do you can easily circumvent the hook using `--no-verify`.

sachinraja:
> I think it's quite common for our contributor base to not use a tool like nvm (if you don't switching to npm v6 can be a real pain).

Unfortunately, I think you're right as many contributors here are new.

> I still share this concern, but I don't expect this to be a big problem for use as we 1) don't have a lot of dependencies and 2) don't do anything fancy with them. 

I think you're right, so I'm alright with experimenting with removing it. 
Though there are a few things we should be aware of:
- We get a lot of new contributors, and it could definitely put them off if something went wrong after just an install
- We will lose out on installation optimizations: "Optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages." (https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#description)
- We will lose out on performance gains *for npm v7*: "As of npm v7, lockfiles include enough information to gain a complete picture of the package tree, reducing the need to read package.json files, and allowing for significant performance improvements."

> An alternative to fix the problem we're having: we can add a pre-commit hook that reverts any changes to package-lock.json.

This would be better handled in CI anyway (looks like we are correctly using `npm ci`). Our lockfile should not change after a CI installation because otherwise someone could manually edit it to add some malicious package and we wouldn't know.

ericcornelissen:
> I think you're right, so I'm alright with experimenting with removing it.

Cool, in that case I think we can move ahead. @LitoMore can you resolve the conflicts for us?


> We will lose out on installation optimizations

That is a fair point, but again we don't have a lot of dependencies so I don't think our gains are big here. Not to mention the fact that it takes way more time to comment on unwanted changes and reverting said changes.

> We will lose out on performance gains for npm v7

If we don't have a lockfile you do get those benefits, whereas now you don't because you can't run `npm install` with npm v7. If we want that we would need to update the lockfile version first.


> This would be better handled in CI anyway (looks like we are correctly using npm ci). Our lockfile should not change after a CI installation because otherwise someone could manually edit it to add some malicious package and we wouldn't know.

I disagree. The CI can only check if a change was made to the lockfile that shouldn't be there, not revert it (at least for PRs from forks, this is difficult).

I'm talking about reverting changes to the lockfile (such as upgrading to v2 or downgrading to v1) in a precommit hook to prevent the change from showing up in any commit. If we use [husky](https://www.npmjs.com/package/husky) for this, pretty much everyone who runs `npm install` (and therefore possibly changes the lockfile inadvertedly) is going to have this hook. And anyone that has husky disabled should understand Node/NPM/git enough not to accidentally commit changes to the lockfile.

sachinraja:
> I'm talking about reverting changes to the lockfile (such as upgrading to v2 or downgrading to v1) in a precommit hook to prevent the change from showing up in any commit.

Oh, that makes sense, thanks for clarifying that for me.

I think this is ok to merge then.

mondeja:
Just for curiosity: how [renovate security updates](https://github.com/simple-icons/simple-icons/pull/6199) are handled after this? Are just ignored? Additionally, if we do this, we must remove [this part of the renovate configuration](https://github.com/simple-icons/simple-icons/blob/0a3621df0e5cb43a7a818e5c29cb07494ada5c39/.github/renovate.json5#L9-L20)?

ericcornelissen:
> Just for curiosity: how [renovate security updates](https://github.com/simple-icons/simple-icons/pull/6199) are handled after this? Are just ignored?

I'm not 100% and my idea was to just merge this in an see how it goes, if we encounter problems we bring back `package-lock.json`.

What I'm pretty sure about is that Renovate will (should) no longer address "deep" dependencies (i.e. dependencies of [our dependencies](https://github.com/simple-icons/simple-icons/blob/d2bd9906d8a0a2dd2aef855dbd223bc3937f4ce6/package.json#L20-L32)). This isn't of much concern because:

- If a patch is available for a "deep" dependency **and** Renovate can update it, than it will now be installed by default as `npm install` will always install the most up-to-date version of a package it can.
- If a patch is available for a "deep" dependency that is not compatible with our dependencies, than Renovate couldn't open a PR to fix it to begin with.


> Additionally, if we do this, we must remove [this part of the renovate configuration](https://github.com/simple-icons/simple-icons/blob/0a3621df0e5cb43a7a818e5c29cb07494ada5c39/.github/renovate.json5#L9-L20)?

Good point. I don't think we _must_  remove it, but we should as it is no longer relevant. @LitoMore can you take care of that?

ericcornelissen (approved):
Awesome, with 3 more-or-less approvals let's give this change a try :+1: 

Thank you very much for sticking with us through the discussion @LitoMore :confetti_ball:

PR #6185, opened 2021-07-31
Remove executable flag from file mode

janniklasrose:
I noticed that 5 svg files (`dgraph`, `django`, `homebrew`, `pagekit`, and `uikit`) have their mode set to `755` instead of `644`. This can lead to git annoyingly reporting changes depending on OS/gitconfig.

It is a very minor problem but also a very easy fix, so I made a PR for it.


ericcornelissen (commented):
This is great! Thank you very much for updating this :smiley: 

I would merge it straight in but I'm wondering if we can catch this in the CI - and if so include that in this Pull Request. Any ideas @janniklasrose?

janniklasrose:
> I'm wondering if we can catch this in the CI

I had the same thought and including it in this PR would make sense.

Probably the most intuitive check is to use `find` with the `-executable` flag?

verify.yml could be changed to include the following:

```yml
# ...
jobs:
  # ...
  lint:
    steps:
      # ...
      - name: File permissions
        run: find icons/ -type f -executable
```

janniklasrose:
I don't have too much experience with what the commands in GitHub Actions need to do for it to fail, so perhaps this requires a check that `find` returns nothing?

.github/workflows/verify.yml
```diff
@@ -44,6 +44,14 @@ jobs:
         run: npm ci
       - name: Run linter
         run: npm run lint
+      - name: Verify file permissions
+        run: |
+          EXE_FILES=$(find icons/ -type f -executable)
+          if test -n "${EXE_FILES-}"
+          then
+            echo "${EXE_FILES}"
```
ericcornelissen, line 52: I think we should add some additional message here to help people understand what's wrong and how to potentially fix this, e.g.:

```suggestion
            echo "Some files were detected to have their executable bit set, see the list below."
            echo "To fix this you can use 'chmod -x icon/FILE_NAME'\n"
            echo "${EXE_FILES}"
```

ericcornelissen (changes requested):
Thank you very much for looking into this :smiley: Your solution looks great, just have one suggestion to help contributors out when this causes their build to fail.


> verify.yml could be changed to include the following:

I think it makes sense to include it in the lint step of `verify.yml`, good choice :+1: 


> I don't have too much experience with what the commands in GitHub Actions need to do for it to fail, so perhaps this requires a check that find returns nothing?

Indeed, the current solution _should_ work but just to be sure you could include a commit with an executable bit set to verify the behavior (followed by a commit that revert that change, of course).

janniklasrose:
I have made the action more verbose as you suggested, including a message if everything goes as it should. I also added the `_data` directory to the `CHECK_DIRS` variable to include it in the test.

Tests for 58e5df5 and 804a182 fail and succeed as expected.

ericcornelissen (approved):
Cool, this is awesome @janniklasrose, thank you very much for this contribution 🎉 


> I have made the action more verbose as you suggested, including a message if everything goes as it should.

Nice, I like your approach better 😄 


> I also added the `_data` directory to the `CHECK_DIRS` variable to include it in the test.

That makes sense, thank you


> Tests for 58e5df5 and 804a182 fail and succeed as expected.

Awesome

jankeromnes:
> Cool, this is awesome @jankeromnes, thank you very much for this contribution 🎉

Agree that this is awesome! 🎉 Forwarding all the praise to @janniklasrose 😁

ericcornelissen:
Apologies for the incorrect at-mention 😅

PR #6189, opened 2021-08-01
export all icons from a single file

### LINKED ISSUE #6186: Export all icons from one file
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### This issue concerns 
  - [x] The NPM Package
  - [ ] The Packagist Package

### Description

Currently you have to import icons one by one (using the global `.Get` won't allow tree shaking):
```js
import csharpIcon from 'simple-icons/icons/csharp'
import cssIcon from 'simple-icons/icons/css3'
import elixirIcon from 'simple-icons/icons/elixir'
import htmlIcon from 'simple-icons/icons/html5'
import javascriptIcon from 'simple-icons/icons/javascript'
```
I propose all icons be exported from a single package so we can import like this:
```js
import {
  csharpIcon,
  cssIcon,
  elixirIcon,
  htmlIcon,
  javascriptIcon,
} from 'simple-icons/icons'
```
Bundlers should be able to tree shake this. I'm willing to submit a PR for this.

sachinraja:
**Issue:** closes #6186
### Description
Exports all icons from a barrel file (`index.mjs`). A `.mjs` file is necessary because bundlers cannot tree shake commonjs, but they can tree shake esm (see [webpack](https://webpack.js.org/guides/tree-shaking/#conclusion)). In order for this to happen, it was necessary for the generated icon files to have the `.cjs` extension. Otherwise, the module file would not have recognized them as commonjs and would have been unable to import it. 

I'm not sure what to do about the potential **breaking change** here if someone was importing using the `.js` extension. Though we could generate both `.js` and `.cjs` files for icons, this would double the package's size.

A `.d.ts` file is also generated for the file.

I don't really see the need to generate a file for each icon. If you're deploying a web app, you should be using a bundler (you can't import commonjs in the browser anyway). I think a better solution would be to just generate one commonjs and one esm file for all the icons.

Note: Looks like I upgraded the lockfileVersion too. Not sure if that's acceptable.


sachinraja:
Not sure what to do about the breaking change.

scripts/build/package.js
```diff
@@ -68,19 +71,38 @@ function minifyAndWrite(filepath, rawJavaScript) {
...
 
   // write the static .js file for the icon
-  const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
+  const jsFilepath = path.resolve(iconsDir, `${filename}.cjs`);
   minifyAndWrite(jsFilepath, `module.exports=${iconToObject(icon)};`);
+
+  // add name to the barrel file
+  const firstLetter = icon.slug[0].toUpperCase();
+  const rest = icon.slug.slice(1);
+  const capitalizedSlug = `icon${firstLetter}${rest}`;
```
sachinraja, line 90: The `icon` prefix is necessary here because some icons start with numbers (1Password, for example).

ericcornelissen:
Before I review the code changes I want to agree on the approach we will be taking, I hope you understand :slightly_smiling_face: Also, before I will review the code changes I want to ask you to revert all unrelated (stylistic) changes in the code. If you want to discuss changing the code style and/or add a formatter to the project I encourage you to do that in a separate Pull Request so we can keep the two discussion separate. This will simplify reviewing your work and also make it easier to get each individual thing merged.

> In order for this to happen, it was necessary for the generated icon files to have the .cjs extension. Otherwise, the module file would not have recognized them as commonjs and would have been unable to import it.

Can we not make it recognize those as CommonJS files by adding `"type": "commonjs"` to `package.json`?


> I'm not sure what to do about the potential breaking change here if someone was importing using the .js extension. Though we could generate both .js and .cjs files for icons, this would double the package's size.

This is a breaking change - I tested this to be sure. While I'm not completely against doing an intermittent major release to support this I would like to avoid that if possible :slightly_smiling_face: 


> A .d.ts file is also generated for the file.

I would prefer to keep the types in DefinitelyTyped so they can more easily be updated by the community (this avoids the requirement of maintainers that knows TypeScript). You can open a Pull Request updating the types there after this is released.

Also, I'm not sure the `.d.ts` file that is generated now will work as we want but I didn't verify that yet.


> I don't really see the need to generate a file for each icon. If you're deploying a web app, you should be using a bundler (you can't import commonjs in the browser anyway). I think a better solution would be to just generate one commonjs and one esm file for all the icons.

I'm not sure if I follow, if you're not using ESM and want to reduce your bundle size with CJS you should import individual icons. This is backed by your own statement that "[...] bundlers cannot tree shake commonjs." Hence, the individual files have a clear and useful purpose. Moreover, and I'm not super familiar with the ins-and-outs of tree shaking so this might be stupid, don't we need separate files for each icon to make tree shaking possible?


> Note: Looks like I upgraded the lockfileVersion too. Not sure if that's acceptable.

That should be reverted, it's being discussed in #6179 but should definitely not happen in this Pull Request.

scripts/build/package.js
ericcornelissen: Thanks for clarifying - that is a bit unfortunate but I think this makes sense. Going to leave this unresolved to allow others to weigh in though.

sachinraja:
> Also, before I will review the code changes I want to ask you to revert all unrelated (stylistic) changes in the code.

Ah that's my bad, forgot to turn off my formatter, sorry about that.

> I would prefer to keep the types in DefinitelyTyped so they can more easily be updated by the community (this avoids the requirement of maintainers that knows TypeScript). You can open a Pull Request updating the types there after this is released.

> I would prefer to keep the types in DefinitelyTyped so they can more easily be updated by the community (this avoids the requirement of maintainers that knows TypeScript). You can open a Pull Request updating the types there after this is released.

This would be quite difficult to maintain because you'd need a type for each icon that is exported. Though I guess the types could be generated in DefinitelyTyped.

> I'm not sure if I follow, if you're not using ESM and want to reduce your bundle size with CJS you should import individual icons.

You're right, we should keep those.

> Moreover, and I'm not super familiar with the ins-and-outs of tree shaking so this might be stupid, don't we need separate files for each icon to make tree shaking possible?

No, you do not. Webpack or a similar bundler will automatically drop the "dead code" (the icons that haven't been imported).

### Proposal
Instead of importing each icon into the `simple-icons/icons` file, I suggest placing all icon objects inside the file too. This would allow us to keep the current file extensions as the index file would not have to import from anywhere.

ericcornelissen:
> Ah that's my bad, forgot to turn off my formatter, sorry about that.

No problem :slightly_smiling_face: 


> This would be quite difficult to maintain because you'd need a type for each icon that is exported. Though I guess the types could be generated in DefinitelyTyped.

That doesn't sound ideal, and I don't think DefinitelyTyped would be super happy about it. I investigated a bit further and it does indeed seem to be impossible to declare all exported values as being the same type :disappointed: 


> You're right, we should keep those.

:+1: 


> No, you do not. Webpack or a similar bundler will automatically drop the "dead code" (the icons that haven't been imported).

Thanks for clarifying, I did no know it's that smart :smiley: 


> Instead of importing each icon into the simple-icons/icons file, I suggest placing all icon objects inside the file too. This would allow us to keep the current file extensions as the index file would not have to import from anywhere.

That sounds reasonable :thinking: Would that file look something like the following?

```js
export const icon1password = { ... };
// ...
export const iconSimpleIcons = { ... };
// ...
```

And is it correct that something like the following (similar to the existing `index.js`) would not work? (because it can't determine which parts of the top-level object to leave out)

```js
export const icons = {
  icon1password: { ... },
  // ...
  iconSimpleIcons = { ... },
  // ...
};
```

sachinraja:
> That sounds reasonable thinking Would that file look something like the following?

Yes

> And is it correct that something like the following (similar to the existing index.js) would not work? (because it can't determine which parts of the top-level object to leave out)

Yes, I've tested with webpack and it does not work.

It seems like I was wrong before, and module files (`.mjs`) can actually import commonjs files without changing the extension. Here's a demo of the bundling with webpack: https://github.com/sachinraja/simple-icons-esm-webpack-test. Note that the `node_modules` folder being committed is intentional as it uses my local version of `simple-icons`. You can also look at the `dist` folder for the output.

ericcornelissen:
> Yes

Good, then we're on the same page :+1: 


> Yes, I've tested with webpack and it does not work.

Thank you for testing this :smiley: 


> It seems like I was wrong before, and module files (.mjs) can actually import commonjs files without changing the extension.

Not sure how I missed that either but indeed: ESM can import CommonJS without any problem, however CommonJS cannot just `require` ESM files.


---


With that I think we're closing in on a solution :smile: First things first though: I think we should have this file as `./icons.mjs` (so a file at the same level as `index.js`) instead of `./icons/index.js`. This is because if we ever want to add a brand named "Index" (or any other name with the slug "index") it would result in a name clash. That way, importing `simple-icons/icons` should still work, and import `simple-icons/icons/index` also if it ever gets added (though this may require my last point in this comment).

Second, I think we should include a `.js` fallback for CommonJS, just to simplify usage and documentation.

Third, I think we should consider adding some other fields to `package.json` as part of this Pull Request, both for completeness and clarity (and possibly because otherwise it doesn't work :sweat_smile:). I'm thinking something like the following, but I'm not 100% sure if this is correct and/or valid - you can read more about this in [the NodeJS packages docs](https://nodejs.org/api/packages.html). Could you have a look at this as well @sachinraja? If not, I can give it a try myself.

```json
{
  "type": "commonjs",
  "main": "./index.js",
  "exports": {
    ".": {
      "import": "./index.js",
      "require": "./index.js",
    },
    "./icons": {
      "import": "./icons.mjs",
      "require": "./icons.js",
    }
  }
}
```

sachinraja:
> That way, importing simple-icons/icons should still work, and import simple-icons/icons/index also if it ever gets added (though this may require my last point in this comment).

Interesting, tested it and you're right. I didn't realize that `simple-icons/icons` had precedence over `simple-icons/icons/index`.

I've implemented the changes, but I think we still need to address the problem of types. Do you want to generate the `icons.d.ts` file?

package.json
```diff
@@ -8,6 +8,19 @@
     "icons"
   ],
   "main": "index.js",
+  "exports": {
+    ".": {
+      "module": "./index.js",
```
sachinraja, line 13: Module condition is documented [here](https://webpack.js.org/guides/package-exports/#providing-commonjs-and-esm-version-stateless) (webpack docs).
> The module condition is used as optimization when preprocessing node-targeted code with a tool that supports ESM for require() (like a bundler, when bundling for Node.js). For such a tool the exception is skipped. This is technically optional, but bundlers would include the package source code twice otherwise.

I'm not exactly sure what this does, but it seems to be an optimization nonetheless.

scripts/build/package.js
```diff
@@ -79,8 +83,24 @@ data.icons.forEach(icon => {
...
 // write our generic index.js
 const rawIndexJs = util.format(indexTemplate, icons.map(iconToKeyValue).join(','));
 minifyAndWrite(indexFile, rawIndexJs);
+
+// write to our file containing the exports of all icons
+const iconsJs = `module.exports={${iconsBarrelJs.join('')}};`;
+fs.writeFileSync(iconsJsFile, iconsJs);
+
+const iconsMjs = iconsBarrelMjs.join('');
+fs.writeFileSync(iconsMjsFile, iconsMjs);
```
ericcornelissen, line 106: For the sake of consistency, and to be absolutely 100% sure this is as compact as possible, I think we should be using `minifyAndWrite` here. (also some minor other suggestions)

```suggestion
// write our file containing the exports of all icons in CommonJS ...
const rawIconsJs = `module.exports={${iconsBarrelJs.join("")}};`;
minifyAndWrite(rawIconsJs, iconsJs);
// and ESM
const rawIconsMjs = iconsBarrelMjs.join("");
minifyAndWrite(iconsMjsFile, rawIconsMjs);
```

package.json
```diff
@@ -8,6 +8,19 @@
     "icons"
   ],
   "main": "index.js",
```
ericcornelissen, line 10: Did you try to add this one as well?

```suggestion
  "type": "commonjs",
  "main": "index.js",
```

package.json
```diff
@@ -8,6 +8,19 @@
...
+      "import": "./index.js",
+      "default": "./index.js"
+    },
+    "./icons": {
+      "module": "./icons.mjs",
+      "import": "./icons.mjs",
+      "default": "./icons.js"
+    }
+  },
+  "sideEffects": false,
```
ericcornelissen, line 23: Did you test the behaviour if you omit this field? Is it still as expected?

package.json
```diff
@@ -8,6 +8,19 @@
     "icons"
   ],
   "main": "index.js",
+  "exports": {
+    ".": {
+      "module": "./index.js",
+      "import": "./index.js",
+      "default": "./index.js"
+    },
```
ericcornelissen, line 16: I realize now this should probably just be:

```suggestion
    ".": "./index.js",
```

package.json
ericcornelissen: Thank you for sharing the docs, that's very helpful. One question though, based on both this and Node's docs, shouldn't the keys and values be like the following:

```json
{
"module": "./icons.mjs",
"import": "./icons.mjs",
"require": "./icons.js",
"default": "./icons.js"
}
```

Or potentially the following if we want something shorter (based on what's in the webpack docs)

```json
{
  "module": "./icons.mjs",
  "require": "./icons.js",
}
```

Though my preference goes to the first, as I'm not sure we can omit the "import" field since the "module" fields isn't mentioned in the Node docs :thinking:

ericcornelissen (commented):
Below are some minor requests for changes w.r.t. the current implementation. Indeed the types is something we should still think about (forgot about those last time around).

Another thing that I was thinking of while reviewing this, how does this affect performance if you're not bundling your code? I think for both `icons.mjs` and `icons.js` this will trigger file reads for almost 2000 files, which isn't great :disappointed: : I wonder if we should (and can) only export this file if you're bundling your code :thinking:

sachinraja:
> shouldn't the keys and values be like the following:
{
  "module": "./icons.mjs",
  "import": "./icons.mjs",
  "require": "./icons.js",
  "default": "./icons.js"
}

You can `omit` the "require" field here because everything that doesn't use `import` (or `module` if you're using a bundler) will fallback to `default`.

package.json
sachinraja: This is the default, so it shouldn't be necessary, though we can add it if you want to be explicit about it.

package.json
sachinraja: Yes, I tested and the tree shaking does not happen without it, webpack requires it. Read more [here](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)

package.json
sachinraja: "require" is not necessary with "default," but we can include it if you want to be more explicit (see https://github.com/simple-icons/simple-icons/pull/6189#issuecomment-898907491). We cannot omit the "import" field, Node does not resolve from the "module" field.

sachinraja:
> Another thing that I was thinking of while reviewing this, how does this affect performance if you're not bundling your code? I think for both icons.mjs and icons.js this will trigger file reads for almost 2000 files, which isn't great 😞 : I wonder if we should (and can) only export this file if you're bundling your code 🤔

Good point on the performance. The only alternative I see is placing the objects inside the barrel files, rather than the imports (to support all cases). I tested it and it seems that webpack will resolve the "module" field, but Node will not, so we may be able to restrict entrypoints to bundlers that way. Note that the bundlers would still have to read 2000 files at build time. Splintering the entrypoints between users with and without bundlers does not seem like a good idea to me though.

ericcornelissen:
> Good point on the performance. The only alternative I see is placing the objects inside the barrel files, rather than the imports (to support all cases)

So, instead of importing all the individual icons, we can have all the icon objects (individually) in one file, and this will still "treeshake" all the unused objects? So we would have:

```js
// icons.mjs

export const icon1password = { ... };
//...
```

and 

```js
// icons.js

module.exports.icon1password = { ... };
//...
```

correct? If so, that seems fine to me.


What do you mean by "(to support all cases)" though?


> I tested it and it seems that webpack will resolve the "module" field, but Node will not, so we may be able to restrict entrypoints to bundlers that way. 

That sounds correct as Node doesn't know about the "module" field, thanks for testing :+1: 


> Splintering the entrypoints between users with and without bundlers does not seem like a good idea to me though.

I agree it's not great, but if it's the only option that allows us to provide treeshaking for bundlers and avoids poor performance for others I think that might be worth it - though it needs to be well documented.

package.json
ericcornelissen: True. I don't think it hurts to be explicit about it though :thinking:

sachinraja:
> So, instead of importing all the individual icons, we can have all the icon objects (individually) in one file, and this will still "treeshake" all the unused objects

Yes, and I think we actually discussed this before in https://github.com/simple-icons/simple-icons/pull/6189#issuecomment-894679762

> What do you mean by "(to support all cases)" though?
 
To support users with and without bundlers because it won't trigger 2000 file reads.

> I personally like to think ahead: when/if we change the default to an ESM file the package will unexpectedly break for CommonJS users and it would be easy to miss that when changing it or when reviewing the change.

Alright, we can go ahead and add that then.

So I think the best case here is to place the icon objects inside each index file, though this would definitely increase the size of the library. Do you agree that this is an acceptable solution?

ericcornelissen:
> Yes, and I think we actually discussed this before in https://github.com/simple-icons/simple-icons/pull/6189#issuecomment-894679762

My apologies, in my memory the examples I gave there were using `import`/`export`/`require` :stuck_out_tongue: 


> So I think the best case here is to place the icon objects inside each index file, though this would definitely increase the size of the library. Do you agree that this is an acceptable solution?

Not sure if I follow what you're saying (specifically "place the icon objects inside each index file"). Do you mean, have `icons.(m)js` specify all icons as individual object? If so, I think that's fine w.r.t. the package size on the premise that it would allow for automated tree-shaking.

Note that the individual icon files (e.g. `icons/1password.js`) were originally added to support treeshaking so with this solution we could consider removing those individual files in v6. That would actually net us a reduction in the package size. Not saying we _should_ do this, just something to consider w.r.t. the package size.

sachinraja:
> Not sure if I follow what you're saying (specifically "place the icon objects inside each index file"). Do you mean, have icons.(m)js specify all icons as individual object? If so, I think that's fine w.r.t. the package size on the premise that it would allow for automated tree-shaking.

Yes, so instead of importing each object, we would just directly use the icon object.

> Note that the individual icon files (e.g. icons/1password.js) were originally added to support treeshaking so with this solution we could consider removing those individual files in v6.

Definitely something that I believe should be added (rather, removed :laughing:) to the release.

scripts/build/package.js
```diff
@@ -68,6 +70,8 @@ function minifyAndWrite(filepath, rawJavaScript) {
 }
 
 // 'main'
+const iconsBarrelMjs = [];
+const iconsBarrelJs = [];
```
ericcornelissen, line 74: I guess technically "Barrel" is not the correct term anymore. That said, unless you have any suggestions for names I don't think this has to change (as I can't think of an alternative).

scripts/build/package.js
```diff
@@ -76,11 +80,28 @@ data.icons.forEach(icon => {
   icon.slug = filename;
   icons.push(icon);
 
+  const iconObject = iconToObject(icon)
```
ericcornelissen, line 83: Minor, but for consistency this should have a semicolon (we should probably just add prettier to this repo, but that's for a separate PR)

```suggestion
  const iconObject = iconToObject(icon);
```

scripts/build/package.js
```diff
@@ -76,11 +80,28 @@ data.icons.forEach(icon => {
...
-  minifyAndWrite(jsFilepath, `module.exports=${iconToObject(icon)};`);
+  minifyAndWrite(jsFilepath, `module.exports=${iconObject};`);
+
+  // add export to the barrel file
+  const firstLetter = icon.slug[0].toUpperCase();
+  const rest = icon.slug.slice(1);
+  const iconExportName = `icon${firstLetter}${rest}`;
+
+  iconsBarrelJs.push(`${iconExportName}: ${iconObject},`);
+  iconsBarrelMjs.push(`export const ${iconExportName} = ${iconObject}`);
```
ericcornelissen, line 95: Minor but do you think you can refactor this to something that looks like

```suggestion
  const iconExportName = slugToIconExportName(icon.slug);
  iconsBarrelJs.push(`${iconExportName}:${iconObject},`);
  iconsBarrelMjs.push(`export const ${iconExportName}=${iconObject}`);
```

Where `slugToIconExportName` is something like:

```js
function slugToIconExportName(slug) {
  const slugFirstLetter = icon.slug[0].toUpperCase();
  const slugRest = icon.slug.slice(1);
  const iconExportName = `icon${slugFirstLetter}${slugRest}`;
}
```

ericcornelissen, line 95: For consistency we can consider already writing this minified (note that I also included this change in the comment above):

```suggestion
  iconsBarrelJs.push(`${iconExportName}:${iconObject},`);
  iconsBarrelMjs.push(`export const ${iconExportName}=${iconObject}`);
```

ericcornelissen (changes requested):
Alright, this is looking good. This round is some comments regarding the implementation. Mostly some nits. After this, let's discuss the types - and we should probably also get some feedback on the naming of these variables from the community/other maintainers 🤔 

> Yes, so instead of importing each object, we would just directly use the icon object.

Great 👍 


> Definitely something that I believe should be added (rather, removed 😆) to the release.

Not sure if I necessarily agree with this, my only real concern at this point being that the `icons.js` file forces certain variable names (whether that's good or bad depends on who you ask 🙃). I personally would be fine removing them, but I'd like more input on that before we make the change. In any case, removing the `icon/*.js` files would require a major release and I don't think it's necessary to do an early major release just to remove this, so once this is merged (or already now) we should open an issue for removing them.

scripts/build/package.js
ericcornelissen: @simple-icons/maintainers does anyone have any opinions in this regard?

sachinraja:
> my only real concern at this point being that the icons.js file forces certain variable names

There's really nothing we can do about it other than changing the export name, which would cause a mismatch between it and the slug. FontAwesome actually prefixes all their icons with fa or fas (see [docs](https://fontawesome.com/v5.9/how-to-use/with-the-api/setup/importing-icons#icon-names) (granted, they also use it to denote each category [solid, brands, etc]).

>  but I'd like more input on that before we make the change

I agree, we should verify that current use cases are still covered.

> I don't think it's necessary to do an early major release just to remove this

I agree

scripts/build/package.js
ericcornelissen: From [this comment](https://github.com/simple-icons/simple-icons/pull/6189#issuecomment-903273698):

> There's really nothing we can do about it other than changing the export name, which would cause a mismatch between it and the slug. FontAwesome actually prefixes all their icons with fa or fas (see [docs](https://fontawesome.com/v5.9/how-to-use/with-the-api/setup/importing-icons#icon-names) (granted, they also use it to denote each category [solid, brands, etc]). So I think it would be fine to prefix with "si," like [the react-icons pack for simple-icons](https://react-icons.github.io/react-icons/icons?name=si) does.

As do we in [the font package](https://github.com/simple-icons/simple-icons-font#usage)! Good thing I brought it up I guess. Given the "precedence" that has been set, I think we should use "si" as a prefix rather than "icon". The added benefit is that it prevents any issues if an other icon library decides to use "icon" as a prefix and someone wants to use both.

scripts/build/package.js
sachinraja: Technically it works in the exact same way, so I'm fine with that name too.

scripts/build/package.js
```diff
@@ -27,7 +29,7 @@ const indexTemplate = fs.readFileSync(indexTemplateFile, UTF8);
 const iconObjectTemplate = fs.readFileSync(iconObjectTemplateFile, UTF8);
 
 const data = require(dataFile);
-const { getIconSlug, titleToSlug } = require("../utils.js");
+const { getIconSlug, slugToVariableName } = require("../utils.js");
```
ericcornelissen, line 32: It's not really clear what should go where, but I think it would make more sense to put this (`slugToVariableName`) as a [local helper function](https://github.com/simple-icons/simple-icons/blob/d2bd9906d8a0a2dd2aef855dbd223bc3937f4ce6/scripts/build/package.js#L32). Provided you agree, could you make that change?

sachinraja: Done

ericcornelissen (commented):
While we wait for input for others regarding the naming of variables, let's turn our attention to the type definitions.

I would still love to find a suitable solution through DefinitelyTyped so that we can avoid the requirement of having a maintainers that is (very) familiar with TypeScript. Unfortunately, it does still seem impossible (at least very hard) to find a complete solution that way. The closest to a solution I was able to get is adding the following to the type definitions in DefinitelyTyped:

```typescript
declare module 'simple-icons/icons' {
    interface AllIcons {
        [key: string]: icons.SimpleIcon;
    }

    const allIcons: AllIcons;
    export = allIcons;
}
```

This will provide type definitions for icons, but it won't tell you if the icon you have actually exists. I agree it's not an ideal solution :disappointed: Any thoughts @sachinraja?

sachinraja:
This is definitely not a great solution. I would prefer to host the types here, but I understand your concerns. Also, am I setting it up wrong, or does the suggested solution not work?

sachinraja:
Mentioning @simple-icons/maintainers again to make sure they see this. I don't think we can make a decision here without more input.

jorgeamadosoria:
> Mentioning @simple-icons/maintainers again to make sure they see this. I don't think we can make a decision here without more input.
> 
> We're considering whether to host the typescript types in this project or at DefinitelyTyped (`@types/simple-icons`). If we host them here, we can generate the proper types for each icon that is exported thorough the barrel file.

I did see this, unfortunately I do not have enough interest (and much more importantly, knowledge) to meaningfully contribute on this.
I will have to abstain and trust that you guys have this one.

sachinraja:
@ericcornelissen Not sure how to proceed with this. I think it would be quite difficult to implement in DefinitelyTyped because it requires a weekly release to stay updated.

ericcornelissen:
If the only way to support type definitions here is to have them generated and you're comfortable having the type definitions included in the project, the feel free to include them here. And don't forget to update the README accordingly.

sachinraja:
I'll do that then. I think I'll create a different branch on repo for this and get this merged first though. Then I can work on the definitions in a separate PR.

ericcornelissen:
I support putting them in a separate PR, but I think we should at least try to merge them together to make sure they're released as one change.

sachinraja:
Changed target branch from `simple-icons:develop` to `simple-icons:barrel-exports`.

ericcornelissen (approved):
This PR looks good to me. Ideally we get some more input [on the variable names](https://github.com/simple-icons/simple-icons/pull/6189#discussion_r693508671) and as per [my previous comment](https://github.com/simple-icons/simple-icons/pull/6189#issuecomment-921148872) I don't think we should merge this just yet.

Also, @sachinraja, can you investigate what we should do to the DefinitelyTyped types once we start publishing our own types.

sachinraja:
> merge them together to make sure they're released as one change

I don't think I was clear enough, I meant I would change the base branch of this PR, so it's now targeting `simple-icons:barrel-exports` instead `simple-icons:develop`.

> can you investigate

Already looked into it, looks like all the information is [here](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package). We just have to run a command and open a PR.

ericcornelissen:
> I don't think I was clear enough, I meant I would change the base branch of this PR, so it's now targeting simple-icons:barrel-exports instead simple-icons:develop.

I see now - in that case if you want to merge this as is, go right ahead.

Issue #6170, opened 2021-07-26
Sonatype


johannroux:
**Brand Name:** Sonatype - Will also apply to their products, like Sonatype Nexus, an artifact repository
**Website:** https://sonatype.com/
**Similarweb rank:** [95,093](https://www.similarweb.com/website/sonatype.com/#overview)

**Official resources for icon and color:**
https://sonatype.com/company/press-kit

service-paradis:
Thank you @johannroux for the request. Sonatype is reaching our Alexa Rank requirement. Do you think you would be up trying to do a pull request?

You can take a look at our [contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) if you want to give this a try.

If you also want to add their products, we will need to find a way to assess their popularity before.

jorgeamadosoria:
I took the liberty of addressing this one: https://github.com/simple-icons/simple-icons/pull/6211

jorgeamadosoria:
@service-paradis is there an additional action I need to take to move the PR for this ticket along? First time contributor, so I don't really know how the workflow works for this organization.

service-paradis:
Hello @jorgeamadosoria,

Thank you for your multiple contributions so far! 💯 

You're doing exactly what is expected, from what I can see so far. When a new PR is opened, I (or another maintainer) will link the issue and assign the PR to our priorisation project. We've all been quite busy in the past few weeks, so the review process might take some time right now. Once a review have been made, we'll let you know if there are some expected changes. If everything is fine, we'll merge the PR.

jorgeamadosoria:
> Hello @jorgeamadosoria,
> 
> Thank you for your multiple contributions so far! 💯
> 
> You're doing exactly what is expected, from what I can see so far. When a new PR is opened, I (or another maintainer) will link the issue and assign the PR to our priorisation project. We've all been quite busy in the past few weeks, so the review process might take some time right now. Once a review have been made, we'll let you know if there are some expected changes. If everything is fine, we'll merge the PR.

Ah, that puts me at ease.
Thank you. I'll keep scouring the issues to see what else can be addressed. I'm thinking Google to also help with the Google icon review.
Excellent collection you have going here, looking forward to help some more while I still can.

PR #6211, opened 2021-08-08
adding Sonatype icon

jorgeamadosoria:
sonatype

In response to:
**Issue:** https://github.com/simple-icons/simple-icons/issues/6170


johannroux:
Very nice! Thanks for taking care of it. Maybe we don't need the name Sonatype under the logo, so that the hexagram takes all the place?

jorgeamadosoria:
> Very nice! Thanks for taking care of it. Maybe we don't need the name Sonatype under the logo, so that the hexagram takes all the place?

Yes, I think so too, but the guidelines are very clear that the logo always include the text. I'm happy to remove it if the hexagon alone is accepted in the library, but since I'm new here I wanted to do this by the book

service-paradis:
> By downloading our logos you agree not to change, edit, or manipulate them in any way.

According to the above sentence from their press kit, I tend to agree that we should probably keep the wordmark on this one.

jorgeamadosoria:
> I think the icon should be unstacked (the hexagon and the wordmark are on the same line)

I don't understand what you mean.
You want to use the horizontal version of the logo?

jorgeamadosoria:
> > You want to use the horizontal version of the logo?
> 
> yes, though that is just my preference. Not sure if a standard has already been set

no, there is no standard. I'm using this one because I think it would scale better since it's more square.
Tell you what, I'm going to process the horizontal variante and upload the preview so you can compare by yourself.
I don't really care about one over the other, so we'll go with the one you consider better.

sachinraja:
Still not sure. I see your point on the scaling. @simple-icons/maintainers was a previous standard set for this?

jorgeamadosoria:
> I would say for slightly better clarity on the text, we should use the vertically stacked version. Other than that, it'd be worth dropping their press email a line to see if there's _any_ way we can use just the icon - while linking to official guidelines they write. The favicon on their site is just the shape - and so they may allow us to use just the shape if guidelines are present?

that's true, but personally I would prefer to keep the wordmark. The icon alone is just... an hexagon. It would be contextless, I don't know. it lacks identity, in my opinion.

adamrusted:
Apart from a couple of [incredibly minor](https://github.com/simple-icons/simple-icons/assets/1830867/d7ded504-80b8-4c79-b174-73ee5754191d) differences between the PR (black) and source (red) - I'm happy to merge the vertically stacked version currently in this PR. @jorgeamadosoria - would you mind taking another quick look at this, perhaps upping the decimals of the precision? We can close this one off then!

PR #6219, opened 2021-08-09
updated sat1 icon by removing ineffective segment

jorgeamadosoria:
sat1

**Issue:** https://github.com/simple-icons/simple-icons/issues/4590

Updated this icon for the 2021 icon review:
- removed segment "m0 0l-.003.002". the ineffective segment reported was "m0 0" but I don't see the logic being the "l-.003.002" part. It does nothing visible AFAIK. 
- removed the icon from the ignored.json

Also, nitpick: shouldn't this be named satdot1?


service-paradis:
@jorgeamadosoria I understand you put those one as priority 1 since it is an update of an already existing icon. Personally, since it does not really change the icon in itself and it is only to optimize the path, I would consider those priority 3 (or even 4). I would prefer making review of actual icon update before, and even try to add other contributor new icon.

We can ask other @simple-icons/maintainers opinion on this particular case.

jorgeamadosoria:
> > @jorgeamadosoria I understand you put those one as priority 1 since it is an update of an already existing icon. Personally, since it does not really change the icon in itself and it is only to optimize the path, I would consider those priority 3 (or even 4). I would prefer making review of actual icon update before, and even try to add other contributor new icon.
> > We can ask other @simple-icons/maintainers opinion on this particular case.
> 
> I was going by the rules, I thought all icon updates where supposed to be P1.
> If there are different kinds of outdated icons, I'm sorry about changing the priority.
> 
> I would still think that any kind of oudated icon should be prioritized for update over pretty much everything else, except other outdated icons. Even in cases such as this where the icon is largely the same visually: it should be fast to review since the changes are minimal and invisible, right?
> 
> After all, just because this is a P1 now doesn't mean that other outdated icons are not P1 as well, right?
> You could always prioritize the actual icon updates before this one, But I wouldn't prioritize creating a new icon over reviewing this one, for example.

jorgeamadosoria:
image

The result of inkscape's comparison of the new and old versions of Sat1.
I'm 100% sure this is as close as it can be, so imo, this is ready to merge.
Just need a second mantainer to confirm, please.

ericcornelissen:
> The result of inkscape's comparison of the new and old versions of Sat1.
> I'm 100% sure this is as close as it can be, so imo, this is ready to merge.
> Just need a second mantainer to confirm, please.

Comparing myself, and looking at your comparison I'm seeing some slight difference on the segment to the right of the smallest segment - the line on the top there is slightly ticker.


> Also, nitpick: shouldn't this be named satdot1?

[The slug is explicitly stated in the data file](https://github.com/simple-icons/simple-icons/blob/0c52e63e606580c2d8ac2b63b1caecfc581afadb/_data/simple-icons.json#L8003), presumably because the phonetic name is "sat 1" not "sat dot 1" (though I'm not familiar with the brand so I can't speak to that).

jorgeamadosoria:
> > The result of inkscape's comparison of the new and old versions of Sat1.
> > I'm 100% sure this is as close as it can be, so imo, this is ready to merge.
> > Just need a second mantainer to confirm, please.
> 
> Comparing myself, and looking at your comparison I'm seeing some slight difference on the segment to the right of the smallest segment - the line on the top there is slightly ticker.
> 
> > Also, nitpick: shouldn't this be named satdot1?
> 
> [The slug is explicitly stated in the data file](https://github.com/simple-icons/simple-icons/blob/0c52e63e606580c2d8ac2b63b1caecfc581afadb/_data/simple-icons.json#L8003), presumably because the phonetic name is "sat 1" not "sat dot 1" (though I'm not familiar with the brand so I can't speak to that).

Try again when you can, I uploaded a version with precision 5 (I think precision 4 also works). The comparison shows a correction in those thicker segments.
The ineffective segment is "m0 0", so that's not what's introducing the problem, is the lack of precision.

ericcornelissen (approved):
Looking good now, thanks for taking care of this one @jorgeamadosoria :+1:

PR #6222, opened 2021-08-09
removed collinear segment from matternet

jorgeamadosoria:
matternet

**Issue:** https://github.com/simple-icons/simple-icons/issues/4590

- removed collinear segment "v7.458-.011"
- removed matternet from ignored.json


jorgeamadosoria:
> h @jorgeamadosoria ,
> 
> how do you remove collinear segments? I have the problem in the Tesco icon... I use Inkspace.. Thanks

I'm sorry to say, I only open it on Visual Studio Code and just remove the segment manually :(
Problem with that is that sometimes there are some deformations introduced that require the svgo precision to be upped from 3 to 4 or 5 🤷 
I also use inkscape. I can take a look at the Tesco icon to help you out tomorrow if you want, although, no promises: I suck at this.

Issue #6143, opened 2021-07-22
Update Starling Bank icon?


service-paradis:
The icon currently in use looks different from what we have.
![image](https://user-images.githubusercontent.com/12817388/126700028-c954fad1-325e-4d55-824a-ddb2d96b9168.png)

**Brand Name:** Starling Bank
**Official resources for icon and color:**
  
- https://www.starlingbank.com/media/
- https://www.starlingbank.com/docs/brand/starling-bank-brand-guidelines.pdf

jorgeamadosoria:
provided an option for this update: https://github.com/simple-icons/simple-icons/pull/6229

PR #6229, opened 2021-08-10
updating starling icon

jorgeamadosoria:
starlingbank

**Issue:** https://github.com/simple-icons/simple-icons/issues/6143

Using square icon for this, since it is more similar to the currently existing one.
Hex and sources are the same as existing ones.


jorgeamadosoria:
I think this one is ready for review by someone from @simple-icons/maintainers.

service-paradis:
I wonder if we could get away with the icon only without wordmark here as they use for their favicon and social media. I found no indication stating we cannot.

I did not yet search a lot for this, but are able to find a page that links to their guidelines?
https://www.starlingbank.com/docs/brand/starling-bank-brand-guidelines.pdf

jorgeamadosoria:
> https://www.starlingbank.com/docs/brand/starling-bank-brand-guidelines.pdf

honestly, I don't know.
Their twitter account and favicon use only the logo, but everywhere else in the google image search they (and everyone else) use one of the two approved logos.
So... I don't know what to do about it.
If you make an executive decision to only stick with the logo, I'll back it up. Reducing wordmarks is in general something that works in our favor, I think.

jorgeamadosoria:
> https://www.starlingbank.com/docs/brand/starling-bank-brand-guidelines.pdf

I was not able to find a page with the link for this file. There are a couple pages of people that worked on the branding for this entity. Maybe if they could be authenticated to have worked on it, they could serve as source, but all in all, I prefer this one rather than a lengthy looking into people's online profiles.

service-paradis:
> I prefer this one rather than a lengthy looking into people's online profiles.

So, you will add this link for the guidelines entry?

> Reducing wordmarks is in general something that works in our favor, I think.

I'm in favor of using only the icon, but I'll ask other @simple-icons/maintainers to weight in before we make a decision.

jorgeamadosoria:
> > I prefer this one rather than a lengthy looking into people's online profiles.
> 
> So, you will add this link for the guidelines entry?

I don't see any other choice, unfortunately

jorgeamadosoria:
removed the wordmark, added the guidelines.

icons/starlingbank.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Starling Bank</title><path d="M12.257.631a15.683 15.702 0 0 0-1.242.021C7.693.926 4.671 2.455 2.621 4.895A10.886 10.899 0 0 0 .397 14.903c1.132 4.11 4.681 7.326 9.071 8.223.925.188 1.572.251 2.626.247a12.232 12.247 0 0 0 6.19-1.66 11.419 11.434 0 0 0 5.636-8.102 14.876 14.895 0 0 0-.042-3.444 11.079 11.094 0 0 0-3.39-6.22 12.228 12.243 0 0 0-6.994-3.24A15.683 15.702 0 0 0 12.257.63zm3.122 3.6.005 1.283.008 1.287h-.26c-1.525.013-2.724.43-3.665 1.284a4.184 4.184 0 0 0-.995 1.287c-.332.651-.438 1.067-.498 1.915l-.05.723-1.35.004-1.359.009.02-.828c.016-.6.052-.933.118-1.241.65-2.928 3.113-5.182 6.195-5.648a9.696 9.696 0 0 1 1.117-.071zm1.455 7.75-.02.677a13.315 13.33 0 0 1-.072 1c-.458 2.964-2.794 5.329-5.92 5.982a6.88 6.888 0 0 1-1.316.117l-.874.025-.004-1.284-.008-1.283.635-.033c.721-.035 1.256-.136 1.785-.33a4.773 4.782 0 0 0 2.454-2.095c.387-.715.528-1.256.56-2.157l.021-.606 1.38-.009z"/></svg>
```
service-paradis, line 1: I used `Starling Bank_Logo_Vertical.eps` from their ZIP media kit and I'm seeing a major difference. The icon looks stretched horizontally:
image

Which file did you use to create the icon? Can you take another look at it?

service-paradis (changes requested):
Thank you @jorgeamadosoria for your contribution! Can you take another look at it according to my comment?

icons/starlingbank.svg
jorgeamadosoria: I htink I used the EPS, that's the vector one.
Maybe I converted it wrong.

jorgeamadosoria: I've tried again. The icon was indeed squashed, it was 23x22 for some reason. You can check again now.

service-paradis (approved):
It is a perfect match now, thank you @jorgeamadosoria for your work on it 💯 🏦

Issue #6141, opened 2021-07-22
Update Sourcegraph icon


service-paradis:
According to their brand guidelines, the 1-color version should have gaps
![image](https://user-images.githubusercontent.com/12817388/126694872-6628aa2d-c350-427c-a3be-56ba0ceea3a2.png)

**Brand Name:**
**Official resources for icon and color:**
  
- https://about.sourcegraph.com/handbook/marketing/brand/brand_guidelines

jorgeamadosoria:
updated with the recommended star logo with gaps: https://github.com/simple-icons/simple-icons/pull/6231

PR #6231, opened 2021-08-10
updating sourcegraph icon

jorgeamadosoria:
sourcegraph
**Issue:** https://github.com/simple-icons/simple-icons/issues/6141

hex and sources remain the same.


service-paradis (changes requested):
Thank you @jorgeamadosoria for your contribution!

Which file did you use as source? I used `Sourcegraph Brand Kit 2.2 - May 2021/Logos/Star Logomark/SVG/Sourcegraph_Logomark_BLK.svg` from the ZIP file to compare.
There are some differences. For example, the top left bar:
image

There are also some extra segments that could be removed. Ex.
image

By using `Union` instead of `Combine`, it will probably produce less of those.

Can you gave this another try?

jorgeamadosoria:
> Thank you @jorgeamadosoria for your contribution!
> 
> Which file did you use as source? I used `Sourcegraph Brand Kit 2.2 - May 2021/Logos/Star Logomark/SVG/Sourcegraph_Logomark_BLK.svg` from the ZIP file to compare.
> There are some differences. For example, the top left bar:
> image
> 
> There are also some extra segments that could be removed. Ex.
> image
> 
> By using `Union` instead of `Combine`, it will probably produce less of those.
> 
> Can you gave this another try?

source remained the same.

service-paradis:
> source remained the same.

Just wanted to know which file from the Zip file did you use. To be sure I compare your icon with the same version you used. As mentioned above, I used `Sourcegraph Brand Kit 2.2 - May 2021/Logos/Star Logomark/SVG/Sourcegraph_Logomark_BLK.svg` to compare and I got some differences.

service-paradis:
@jorgeamadosoria Do you still want to work on this?

PR #6233, opened 2021-08-10
Update Kotlin hex color and guideline

wiryadev:
kotlin

**Issue: #6235**

### Description
I pick this hex value according to `Kotlin One Color Logo Mark RGB` file included in the source.
The new guideline is from their [blog post](https://blog.jetbrains.com/kotlin/2021/07/adding-volume-to-the-kotlin-identity/).

### Additional Context
I can not update the logo therefore i just can help with hex color and guideline url.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4945,7 +4945,7 @@
         },
         {
             "title": "Kotlin",
-            "hex": "0095D5",
+            "hex": "7F52FF",
             "source": "https://resources.jetbrains.com/storage/products/kotlin/docs/kotlin_logos.zip",
             "guidelines": "https://www.jetbrains.com/company/brand/"
         },
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -4947,7 +4947,7 @@
             "title": "Kotlin",
             "hex": "7F52FF",
             "source": "https://resources.jetbrains.com/storage/products/kotlin/docs/kotlin_logos.zip",
-            "guidelines": "https://www.jetbrains.com/company/brand/"
+            "guidelines": "https://blog.jetbrains.com/kotlin/2021/07/adding-volume-to-the-kotlin-identity/"
         },
         {
             "title": "Krita",


wiryadev:
> This is an interesting Pull Request 🤔 I don't think it makes sense to do a partial update in this way and I would prefer to update the color and SVG in one Pull Request (or one release with multiple Pull Requests). What is your motivation for doing this Pull Request with just the color @wiryadev?
> 
> Regarding the guidelines, I don't think it makes sense to link to the blog post, it doesn't seem to add much more information on _how to use the logo_ and is less likely to be updated if the guidelines change in the future.

i honestly cant do the SVG one, so i think it might be good if someone would do SVG update without worrying about color, thats why i update only the color. And the recent color is too different with what Kotlin has now, so at least the color is right even tho the SVG is still outdated.

In terms of guidelines, yes that is only a blogpost, but it is Kotlin's compared to current url which i believe is Jetbrains`s instead of Kotlin's. Even the zip file in resource one can be accessed from that blogpost.

ericcornelissen:
> i honestly cant do the SVG one, so i think it might be good if someone would do SVG update without worrying about color, thats why i update only the color. And the recent color is too different with what Kotlin has now, so at least the color is right even tho the SVG is still outdated.

You can always give the SVG a try :smile: If you can't, I don't think it's too much trouble for someone updating the SVG to update the color as well. From my perspective having the old SVG with the new (wrong) color is incorrect, whereas having the old SVG with the old color is correct (though outdated).


> In terms of guidelines, yes that is only a blogpost, but it is Kotlin's compared to current url which i believe is Jetbrains`s instead of Kotlin's. Even the zip file in resource one can be accessed from that blogpost.

Correct me if I'm wrong but Kotlin is a programming language by Jetbrains, so I don't see a problem with linking to a Jetbrains page if that's the best link available.

wiryadev:
Its okay tho, since we still can change the icon color manually (at least for shields io). I will close this PR.

PR #6238, opened 2021-08-10
Added "-D" for dev dependency

max-programming:
Simple change in docs to mark type definitions as dev dependency as they should not be installed as a dependency

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ for (const iconSlug in simpleIcons) {
 There are also TypeScript type definitions for the Node package. To use them, simply run:
 
 ```shell
-npm install -D @types/simple-icons
+npm install --save-dev @types/simple-icons
 ```
 
 ### PHP Usage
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,7 +113,7 @@ for (const iconSlug in simpleIcons) {
 There are also TypeScript type definitions for the Node package. To use them, simply run:
 
 ```shell
-npm install @types/simple-icons
+npm install -D @types/simple-icons
```
ericcornelissen, line 116: Minor but I think we should go with `--save-dev` for clarity to those that are not as familiar with the npm CLI.

Most people will copy this anyway so it doesn't affect them. If you're going to type it manually than using `-D` or `--save-dev` is up to you so it doesn't matter to much what this says.

```suggestion
npm install --save-dev @types/simple-icons
```

ericcornelissen (changes requested):
Good idea, completely agree this makes more sense :+1: Just one minor suggestion.

README.md
max-programming: I agree!

ericcornelissen (approved):

Issue #805, opened 2018-04-01
Request: File Hosting Sites


Adamanthus:
Would be nice if some of the more common file hosting services could be added. GOG seem to have some suitable SVG's. I think these 5 are very well known.

- [x] Mediafire (https://www.onlinewebfonts.com/icon/432848)
- [x] MEGA (https://commons.wikimedia.org/wiki/File:01_mega_logo.svg)
- [x] Uploaded (https://goggames.goodolddownloads.com/static/img/hoster_logos/uploaded.svg}
- [x] UpToBox (https://goggames.goodolddownloads.com/static/img/hoster_logos/uptobox.svg)
- [x] Zippyshare (https://goggames.goodolddownloads.com/static/img/hoster_logos/zippyshare.svg)

birjolaxew:
Official logo sources:

- [Mediafire](https://www.mediafire.com/press/)
- [MEGA](https://mega.nz/corporate/page-2) (PNG only; use the Wikimedia link for SVG)
- Uploaded - ? Link is dead, googling gives nothing, never heard of the service.
- UpToBox - no high quality logo available. Poor PNG at [their site](http://uptobox.com/)
- ZippyShare - no high quality logo available. Poor PNG at [their site](http://www.zippyshare.com/)

ZippyShare is going to be difficult to turn monochrome, and since there isn't a high quality logo available, I'd say it isn't really worth the hassle (unless someone feels like putting in the work of course :wink:). Although UpToBox is easy to turn monochrome, the logo is so poor quality that the same goes there.

Adamanthus:
Uploaded has been around a long time. The site is https://uploaded.net/
Zippyshare announced a new logo on their blog but it's just a png @ http://blog.zippyshare.com/z.png and they haven't applied their redesign yet so I guess it's not 'current'.

Are the GOG SVG's not suitable?

birjolaxew:
I suppose the name just gave too many irrelevant hits (also the logo link is 404).

We prefer to use official sources for our logos, with the option of falling back onto a reputable neutral source (usually Wikimedia) if one isn't available. Third parties (e.g. GOG) aren't usually used, unless the logo is checked carefully to match official sources exactly.

Adamanthus:
https://goggames.goodolddownloads.com/static/img/hoster_logos/uploaded.svg works for me and fair enough if these requests cannot be met :)

Adorable-SkullMaster:
[...]

PR #6244, opened 2021-08-11
added uptobox icon

jorgeamadosoria:
uptobox

**Issue:** https://github.com/simple-icons/simple-icons/issues/805
Using the logo on the page since there are no hi def pngs or raster files of any kind,anywhere.


jorgeamadosoria:
> This does require a bit of creativity, but we should try to keep the letters as close to the original as possible, so I think the "B" should be aligned here image

I don't understand what do you want me to align. I suppose that the black letters are mine, but what are the white ones, where are they coming from?

jorgeamadosoria:
> I got the white one from https://uptoboxpremium.org/. I was just saying that I think there should be an small indent in the middle right of the B.

you mean modifying the B, ok.

PR #6245, opened 2021-08-11
adding zippyshare icon

jorgeamadosoria:
zippyshare

**Issue:** https://github.com/simple-icons/simple-icons/issues/805
**Alexa rank:** 611 (!)

Had to trace this one from a facebook page and do some manual editing. There are no high definition PNG or raster files of any kind available. There are no official guidelines or brand kit (it's a file hosting service...).

For the hex I choose the red color from the source, I think it's the more representative.
I don't really like this one, but it does have a pretty high Alexa rank, so here we are.

Can be improved A LOT by someone with actual knowledge of SVG. If anyone wants to take it from me, I'm happy to let it go.


adamrusted:
Zippyshare has now died a death - do we need to continue adding this icon?

PR #6281, opened 2021-08-13
Update Fandom

PeterShaggyNoble:
image

**Issue:** n/a
**Alexa rank:** [~81](https://www.alexa.com/siteinfo/fandom.com)

### Checklist

### Description
Fandom [rolled out a rebrand](https://community.fandom.com/wiki/The_Future_of_Fandom) last week; they haven't updated [their design system](https://fandomdesignsystem.com/) as yet so the icon is taken from the side bar on [their website](https://www.fandom.com/). The new branding includes a lot of colours with no clear indication of which is the primary so I've gone with the yellow because, of the 2 colours used in the new logo, it looks to be the one used the most throughout their website. When they update their design system to the new branding, that might give us a bit more clarity and we can update then if necessary.


jorgeamadosoria:
> image
> 
> **Issue:** n/a
> **Alexa rank:** [~81](https://www.alexa.com/siteinfo/fandom.com)
> 
> ### Checklist
> * [x]  I updated the JSON data in `_data/simple-icons.json`
> * [x]  I optimized the icon with SVGO or SVGOMG
> * [x]  The SVG `viewbox` is `0 0 24 24`
> 
> ### Description
> Fandom [rolled out a rebrand](https://community.fandom.com/wiki/The_Future_of_Fandom) last week; they haven't updated [their design system](https://fandomdesignsystem.com/) as yet so the icon is taken from the side bar on [their website](https://www.fandom.com/). The new branding includes a lot of colours with no clear indication of which is the primary so I've gone with the yellow because, of the 2 colours used in the new logo, it looks to be the one used the most throughout their website. When they update their design system to the new branding, that might give us a bit more clarity and we can update then if necessary.

I would use `#f9025b` rather than yellow. Notice that the icon in its monochromatic version has a transparent heart and a filled blaze. The blaze is red, not yellow in the logo. The users that would add this icon to their projects don't necessary know that the fandom website is primarily yellow, so they would just see that the icon is yellow instead of red without knowing the rationale behind it.

I would go for red instead of yellow.

sachinraja:
@jorgeamadosoria considering discussion in previous PRs about the color being related to the brand and not the logo, do you think we can merge this?

sachinraja:
If you look [here](https://community.fandom.com/wiki/Community_Central) (an actual wiki page) yellow appears to be used more. However, I could see the case for both and I think both are equally as relevant.

service-paradis:
Now that they have updated their design system (https://fandomdesignsystem.com/?path=/story/base-colors--colors), I agree that we can use the first brand color from there: `#fa005a` (even if yellow is about equally good).
We might also use https://fandomdesignsystem.com/?path=/story/assets-brand--brand for the source now since SVGs are available there.

adamrusted:
So the URL you've been referencing isn't available for me. I would've said (based on the homepage) that yellow is their primary colour too, and without any official guidelines being available - would suggest merging the yellow in.

jorgeamadosoria:
> So the URL you've been referencing isn't available for me. I would've said (based on the homepage) that yellow is their primary colour too, and without any official guidelines being available - would suggest merging the yellow in.

the pages are not showing right now, I don't know why.
Regardless, I still think red is better. The icon is red, everything in the home page is red but the yellow sidebar, the 6 icons listed in the (now inaccesible ) page were red, the listed brand colors started with red... why wouldn't the icon be red if the part that is colored in the monochrome version is the red part? I think it's just logical.

adamrusted:
I've made the relevant changes to this PR, could another @simple-icons/maintainers review the icon?

sachinraja (approved):

Issue #6293, opened 2021-08-14
adding Cubana icon


jorgeamadosoria:
![cubana](https://user-images.githubusercontent.com/15719479/129432750-5c559fa8-9afa-4724-a453-614eba0f7877.png)

**Brand Name:** Cubana de Aviación
**Website:** https://www.cubana.cu/
hex: `#0d3c82`
Cubana de Aviación is the flag carrier of the Republic of Cuba, and the only airline of import for the country

gizmecano:
I guess I missed a previous discussion, but can someone explain to me the changes in criteria that mean that an [Alexa rank close to 3,000,000](https://www.alexa.com/siteinfo/cubana.cu) now allows an icon to be added? Maybe a simple link to the exchange that introduced these changes would be enough for my knowledge? Thanks.

ericcornelissen:
Thank you for bringing this up @gizmecano, the Pull Request was merged accidentally and this has already been reverted in 20f2a1ef4e65d27bed8dd3fca7967988868d6f25 (hence the icon won't be included in the next release). But the Alexa rank should indeed have been discussed.

jorgeamadosoria:
There is no change in criteria in regards with Alexa rank. 
There was a previous discussion about the criteria for inclusion of particularly airlines, trains and other such companies. The conclusion, at least as I took it, was that the company in question should have a good argument in their favor for inclusion. In this case, this is the official flag carrier for Cuba, hence it has a special place as officially sanctioned carrier.

Regardless, it was merged in error, and has been reverted. This is not representative of the criteria for selection for the library.

ericcornelissen:
I'm guessing you're referring to #6260 @jorgeamadosoria? If so, the only thing that was said there is that airlines are _allowed_ (i.e. they won't be rejected _because of_ being an airline). That does not mean, any and all airlines will necessarily be added. First and foremost are still the Alexa rank (and GitHub stars, but those are (probably) irrelevant here). As that criteria is not met here we can either look for an alternative metric by which we can include this; or, as you seem to suggest, close this as being [out of scope](https://github.com/simple-icons/simple-icons/labels/out%20of%20scope).

gizmecano:
> (...) the only thing that was said there is that airlines are _allowed_ (...). That does not mean, any and all airlines will necessarily be added. First and foremost are still the Alexa rank (...)

[...]

PR #6296, opened 2021-08-14
adding Cubana de Aviacion

jorgeamadosoria:
image

**Issue:** #6293 

Brand Name: Cubana de Aviación
Website: https://www.cubana.cu/
hex: #0d3c82
Cubana de Aviación is the flag carrier of the Republic of Cuba, and the only airline of import for the country


sachinraja:
Build is failing because your committed SVG file is called `cubana.svg`, but it's expecting `cubanadeaviacion.svg` based off the title ("Cubana de Aviación"). It seems like that is the right brand name, though you can still put "Cubana" as an alias.

ericcornelissen:
Reverted this in 20f2a1ef4e65d27bed8dd3fca7967988868d6f25.

The source material ([the PNG in the header](https://www.cubana.cu/css/images/logocubana.png)) is very low quality, but I think it's safe to say that at least the bottom line of the triangle should be straight, which it isn't in the SVG being added here. Also, from the CSS I would say the blue should be `#022F7C`.

Issue #6265, opened 2021-08-11
add Habanos brand icon


jorgeamadosoria:
**Brand Name:** several (Cohiba, Montecristo, Romeo y Julieta, Partagás, H. Upmann)
**Website:** http://www.habanos.com/es/unicos-desde-1492/
**Alexa rank:** 379,794

hex `#c9aa4c`

You can assign this one to me. I will take a look at the icons (they are very complex) and create different tickets, one per brand

service-paradis:
Thank you for the request. We will need to asses the popularity for each of these brands. Since they do not have individual websites, which metrics do you propose to evaluate their popularity?

jorgeamadosoria:
> Thank you for the request @jorgeamadosoria. We will need to asses the popularity for each of these brands. Since they do not have individual websites, which metrics do you propose to evaluate their popularity?

I'm sorry, this may have been a bit of cultural arrogance in my part. I apologize.

I guess that I could make a case for the brands chosen, but it's all subjective (smoking is all about personal taste, or lack thereof, I suppose )
For example, we can start with Habanos being the actual word used in the Spanish speaking world to refer to cigars in general. This is a case of generized trademark, like Kleenex for tissues or Band-Aid for bandages.

We have this, hm, "news" report from Business Insider:
https://www.youtube.com/watch?v=vA_j6rjsBec
TL;DR; there are very good Dominican and Nicaraguan cigars, but Cuban cigars are widely considered some of the best, if not straight up the best. The video includes close ups of Cohiba, Romeo y Julieta, Bolivar and Partagas brand cigars, all Habana branded habanos.

We could go by [Cigar Aficionado](https://en.wikipedia.org/wiki/Cigar_Aficionado), a specialized magazine on the matter. If we accept this as an authoritative source (I do), we can see that:
- **Romeo y Julieta** is included in 2020's best 25 cigars: https://www.cigaraficionado.com/top25/2020 (other Habanos brands include Punch, Bolivar and La Aroma de Cuba. I'm not intending to include these as icons, but it's a supporting argument for the overall collection of Habanos products).
- **Partagas** was coincidentally named cigar of the week, with 91/100 points in their evaluation:
![image](https://user-images.githubusercontent.com/15719479/129308219-6abfc72d-2560-4ca5-8298-e84bc9655468.png)
[...]

PR #6298, opened 2021-08-14
adding Habanos icon

jorgeamadosoria:
habanos

**Issue:** #6265


jorgeamadosoria:
I think this is ready for merging. Can we get someone from @simple-icons/maintainers to sign off / review, please?

ericcornelissen:
Reverted this pro-actively on the bases that the source used here (`https://habanos.com/`) is misleading and should have been `http://www.habanos.com/` (or rather `http://www.habanos.com/en/`) instead - Habanos does not offer a secure HTTP connection it would seem :disappointed:

jorgeamadosoria:
> Habanos does not offer a secure HTTP connection it would seem 😞

it does, but it is self signed - most likely for political reasons related to the embargo

ericcornelissen:
Whether that's secure is debatable, but this is not the place to discuss that :sweat_smile: 

Anyway, if you visit the source used here in any modern browser with default settings you'll be hit with a warning, I think that's a reasonable argument against the source.

Furthermore, if we do want to continue this discussion than it should happen before the change is included in `develop` and included in the next release.

Issue #3305, opened 2020-07-09
Update freeCodeCamp


gizmecano:
**Name:** freeCodeCamp
**Official resources for icon and color:** [freeCodeCamp - Style Guide](https://design-style-guide.freecodecamp.org/)

***

Since #830, the color associated with freeCodeCamp icon is set to `#006400`. But, if you check the  [freeCodeCamp - Style Guide](https://design-style-guide.freecodecamp.org/), this color no longer appears anywhere.

This document states in [Logo Colors](https://design-style-guide.freecodecamp.org/#logo-colors) section:

> The freeCodeCamp logo is designed for use primarily on dark backgrounds. For instances in which the full color logo is not practical, the following color combinations are approved alternatives. 

The [Layout colors](https://design-style-guide.freecodecamp.org/#layout-colors) section gives various values, but `#0a0a23` (called _gray 90_) seems to be used for displaying the [main logo](https://design-style-guide.freecodecamp.org/#logo).

In addition, the main glyph no longer seems to have a horizontal line under the flame representation.

PS: While the (former) _green logo_ seems to have disappeared from the [main site](https://www.freecodecamp.org/), it is paradoxically still used... for the favicon of the [Style Guide page](https://design-style-guide.freecodecamp.org/) (which seems strange by the way: cf. https://github.com/freeCodeCamp/design-style-guide/issues/40).

ericcornelissen:
Thanks for bringing this to our attention @gizmecano, would you be interested in submitting a PR to update the _freeCodeCamp_ logo and color?

gizmecano:
> (...) would you be interested in submitting a PR

Yes, as it seems to not present difficulties, I think I could do this.

But first, I wanted to be sure that this brand has really changed as I described (its's why I also opened an issue on its own repository: https://github.com/freeCodeCamp/design-style-guide/issues/40)... 

Furthermore, as I have already two other open PR (#3303  and #3304), I prefer to not too multiplicate branches on my fork at this time...

ericcornelissen:
> Yes, as it seems to not present difficulties, I think I could do this.
>
> But first, I wanted to be sure that this brand has really changed as I described (its's why I also opened an issue on its own repository: freeCodeCamp/design-style-guide#40)...

To me, from [their landing page](https://www.freecodecamp.org/), it seems that they already fully updated their branding, and the Favicon on the style guide page is just an accident. But to be fair, it might be worth waiting for their response 👍 


[...]

PR #6322, opened 2021-08-17
Update data for freecodecamp logo

gizmecano:
* Issue: #3305 and #3344
* Alexa rank: _circa_ [1'400](https://www.alexa.com/siteinfo/freecodecamp.org) (August 2021)

### Checklist

### Description

With the [introduction of optional data](https://github.com/simple-icons/simple-icons/blob/e62894f1578951f76bc6b1c2e181b35b3f51f020/CONTRIBUTING.md#optional-data), it may seem more appropriate to revamp data for the freecodecamp logo (by explicitly splitting source file and guidelines reference).


jorgeamadosoria:
@gizmecano We try not to be overly specific with the urls for sources, just in case they change and the source gets invalidated.
In this case, it's fine to put the same url in both guidelines and source, since the download for the source is within the guideline pages (this is not always the case).

jorgeamadosoria (approved):

Issue #6309, opened 2021-08-16
redmine icon request


sleepymalc:
**Brand Name:redmine**
**Website:https://www.redmine.org/**
**Alexa rank:74,301**
  

**Official resources for icon and color:**

PR #6326, opened 2021-08-17
Adding Redmine icon

forsgren:
redmine

**Issue:** https://github.com/simple-icons/simple-icons/issues/6309
**Alexa rank:** 76,256 
  

### Checklist

### Description
- Color picked from official logo resource on https://github.com/edavis10/redmine_logo/blob/2afe855c4e9cd955b648972d09cc20d76dabbf4c/Redmine-Logo-Invers.eps
-


jorgeamadosoria:
This is the match I get when comparing yours (white) with the icon as extracted from the SVG file in the Redmine folder.
Perhaps it would be best if you extract the logo from the svg file instead of the EPS one. You can still use the color from the EPS even if you use the svg as a source.

image

jorgeamadosoria:
Any developments on this @forsgren ?

forsgren:
@jorgeamadosoria I would argue that the EPS should be used as a source, since Redmine are referring to it on their own wiki under "Vector based EPS versions": https://www.redmine.org/projects/redmine/wiki/logo

jorgeamadosoria:
> @jorgeamadosoria I would argue that the EPS should be used as a source, since Redmine are referring to it on their own wiki under "Vector based EPS versions": https://www.redmine.org/projects/redmine/wiki/logo

That's a good point. I've changed the source to that page, and will be merging this PR momentarily.

Issue #6361, opened 2021-08-24
Request: Malt


gurvancampion:
**Brand Name:** Malt
**Website:** https://www.malt.fr/
**Alexa rank:** 131,910

**Official resources for icon and color:** https://newsroom.malt.com/media-kit

PR #6362, opened 2021-08-24
Malt

jorgeamadosoria:
malt

issue: #6361


sisp (changes requested):
I think only the icon without the wordmark should be used. Also the favicon of the website uses just the icon. I'm not sure though whether adding just the icon complies with their brand guidelines.

dirien:
Agree on @sisp, what do you think @jorgeamadosoria?

Issue #6235, opened 2021-08-10
Update Kotlin icon


wiryadev:
New version doesnt have any strip or gap and just gradient or one flat color.
![image](https://user-images.githubusercontent.com/26060382/128810864-21a4aeac-e85e-4415-bbca-ce76faa12da0.png)

**Brand Name:**  Kotlin
**Official resources for icon and color:** Link [here](https://resources.jetbrains.com/storage/products/kotlin/docs/kotlin_logos.zip) or attached below
**Alexa Rank:** [10,391](https://www.alexa.com/siteinfo/kotlinlang.org)
**Revamped Guideline**: [https://blog.jetbrains.com/kotlin/2021/07/adding-volume-to-the-kotlin-identity/](https://blog.jetbrains.com/kotlin/2021/07/adding-volume-to-the-kotlin-identity/)

  
![Kotlin One Color Logo Mark RGB](https://user-images.githubusercontent.com/26060382/128810127-86f35629-5938-49db-97ef-17e6a4f370f1.png)

PR #6364, opened 2021-08-24
Proposal for clarification about links to be included for icon source data

gizmecano:
### Description

After learning about the possibility of adding [optional data](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#optional-data), I have recently decided to update some of my previous contributions. In two cases, changes were requested by maintainers: https://github.com/simple-icons/simple-icons/pull/6323#pullrequestreview-733968026 by @service-paradis and https://github.com/simple-icons/simple-icons/pull/6320#pullrequestreview-736484627 by @jorgeamadosoria.

In both cases, it was because the modified source link pointed toward a file instead of a page.

Currently, the [source guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#source-guidelines) seem not be explicit enough to indicate that the source link **should** be a page and not a file (with a notable exception about GitHub repositories).

This pull request is a proposal for clarification about links to be included for icon source data, adapted from the comment posted by @service-paradis (mentioned above):

> In both cases above (**Branding page** and **Company website**), pay attention to link to a **page** that includes file instead of link to direct files. This way, if the file name or path changes, the link will still be working.

---

To conclude, regarding this question, at least two sources indicated in the data file are currently pointing to files and not pages: #5497 and #1000 (which could be updated according to #6235). As these are indeed "Branding page" or "Company Website" types, they probably have to be corrected in the future according to this point.


ericcornelissen (commented):
While I'm not necessarily against including a note like this (after all, the past gives us proof the current guidelines are insufficient) I can't help but wonder _why_ the current guidelines are sufficient. For example, take the string for Branding page: "For an SVG from a branding page the source URL **should simply link to the branding page**." To me, that can only mean one thing, namely that you should link to the branding page.


> with a notable exception about GitHub repositories

Just to clarify here, in the case of GitHub we can be nearly 100% certain (provided it's a default branch) the file will remain available there. Of course, this also goes for other git hosting websites.


> To conclude, regarding this question, at least two sources indicated in the data file are currently pointing to files and not pages: #5497 and #1000 (which could be updated according to #6235). As these are indeed "Branding page" or "Company Website" types, they probably have to be corrected in the future according to this point.

Thanks for bringing this up. #1000 definitely happened before our current guidelines so that's no surprise. As for #5497, sometimes these things just slip through the review process. But yes, ideally both of these are fixed.

jorgeamadosoria:
> In both cases above (**Branding page** and **Company website**), pay attention to link to a **page** that includes file instead of link to direct files. This way, if the file name or path changes, the link will still be working.

I think that including the excerpt would be redundant, but in the light of this being an issue that shows up fairly frequently, maybe we should add it just to spell it out for future contributors.

I'm ok adding it if we can get a consensus. 
A text similar to the one @gizmecano proposes (or that same text) could work just fine: 

`In both cases above (Branding page and Company website), pay attention to link to a page that includes file instead of link to direct files. This way, if the file name or path changes, the link will still be working.`

Thoughts? @simple-icons/maintainers ?

ericcornelissen:
My thoughts exactly @jorgeamadosoria, but before we merge this I would like to explore if we cannot just improve the existing text.

As for the text, I think there are a couple of improvements for grammar, brevity, and correctness.

`In both cases above (Branding page and Company website), pay attention to link to a page that includes the source image instead of the source image itself. This is because such a link is less likely to become invalid than a link to the image itself.`

sachinraja:
A few things that (I think) would be better:
`For Branding page and Company website, link to a page that includes the source image, and not the source image itself. Such a link is less likely to become invalid.`

jorgeamadosoria:
> My thoughts exactly @jorgeamadosoria, but before we merge this I would like to explore if we cannot just improve the existing text.
> 
> As for the text, I think there are a couple of improvements for grammar, brevity, and correctness.
> 
> `In both cases above (Branding page and Company website), pay attention to link to a page that includes the source image instead of the source image itself. This is because such a link is less likely to become invalid than a link to the image itself.`

I' ok with your version or @sachinraja 's. They spell the meaning rather nicely.

gizmecano:
> Hey, don't worry about keeping this updated. 

Sorry for the inconvenience: I've made this only for updating my local branch. 😕

gizmecano:
> Okay, in that case I would personally suggest to make the individual points more explicit (...)

So, _if I understand correctly_, you are proposing to also alter texts of both preceding sections. I have included these amendments in a new pull request: https://github.com/simple-icons/simple-icons/pull/6364/commits/2a3d3e2e23cd005520fcc407ad5a7922533a3276.

But would it be _in addition_ to the additional paragraph (as I think I understood) or _in replacement_ of this paragraph (because I have doubts about my understanding)? 

As you wrote _redundant_, I supposed the first hypothesis is the right one, but let me know if you want I update the pull request by removing the added block.

ericcornelissen:
My suggestion was to _only_ update the existing sentences, not add the new sentence.


> Sorry for the inconvenience: I've made this only for updating my local branch

If it can be avoided that is ideal. Such commits may trigger a notification for everyone subscribed to this Pull Request, which is a bit pointless if nothing really changed.

ericcornelissen (approved):
I would be happy to merge this as is, what do you think @sachinraja & @jorgeamadosoria?

sachinraja:
Looks good to me

jorgeamadosoria:
> I would be happy to merge this as is, what do you think @sachinraja & @jorgeamadosoria?

I think the text makes the issue clearer. LGTM

ericcornelissen:
Thanks @gizmecano for improving the Contributing Guidelines! :tada:

PR #6417, opened 2021-08-28
Update graphql icon and source

sachinraja:
graphql

### Checklist

### Description
GraphQL added a branding page, so I updated the source and recreated the icon using the svgs provided there. Hex value is `#E10098` from https://graphql.org/brand.

Note: `fill-rule` really threw me for a loop here, but I was able to remove it using `Break Apart` + `Difference` in Inkscape.


service-paradis (approved):
Good catch for the guidelines, thank you @sachinraja!

The icon is looking good 💯

PR #6424, opened 2021-08-28
Upgrade git-message-action from v1 to v1.2.0

ericcornelissen:
Update the publishing workflow, simplify getting the release name and release body using the new features added in [kceb/git-message-action v1.2.0](https://github.com/kceb/git-message-action/releases/tag/v1.2.0)


sachinraja:
Wasn't it already on the latest version? It seems as though this is pinning the version, rather than upgrading it.

ericcornelissen:
Indeed @sachinraja, the versioning is a bit confusing when compared to the "standard" approach of versioning actions. I'll see if we can get the tags/versions of that action updated :+1: 

If you don't see anything that's wrong, @sachinraja, I guess you can go ahead and merge this, agreed?

sachinraja:
> I'll see if we can get the tags/versions of that action updated

I think [actions-tagger](https://github.com/Actions-R-Us/actions-tagger) is a good solution for this.

Everything looks good me :+1:

ericcornelissen:
> I think [actions-tagger](https://github.com/Actions-R-Us/actions-tagger) is a good solution for this.

You mean for [kceb/git-message-action](https://github.com/kceb/git-message-action) to use? Maybe, but I don't think it's necessarily worth the effort of adding to provided the action is rarely updated, not to mention that it's not up to me.

sachinraja:
Oh makes sense, there's not much of a point in using it then.

Issue #6418, opened 2021-08-28
Popular icons from unofficial sources


sachinraja:
It seems as though there is no official JavaScript icon, which makes sense as there is no official governing body for it. The current source:
https://github.com/simple-icons/simple-icons/blob/198c42d6f8cc91c8e5551adfe0353d06c3828c34/_data/simple-icons.json#L4555-L4557

This uses the source https://github.com/voodootikigod/logo.js, which is unofficial. I realize that it may be too late to remove this icon as I'm sure it is a very popular one, but I thought I'd open this for discussion anyway.

gizmecano:
I understand this appreciation, but however, in this case, it is probable that other icons could be considered in the same case.

For example, Markdown icon has become a _de facto_ standard (notoriously used by [Primer Octicons](https://primer.style/octicons/markdown-16)), but it does not seem to me that it is any formal brand (by the way, source in data file refers to the work of the designer, not to [the language page](https://daringfireball.net/projects/markdown/)).

https://github.com/simple-icons/simple-icons/blob/198c42d6f8cc91c8e5551adfe0353d06c3828c34/_data/simple-icons.json#L5437-L5440

jorgeamadosoria:
I think this one is so established as a de facto standard that it even has other icons styled after it, like [Typescript](https://www.typescriptlang.org/branding/) for example.

I feel good keeping it in, even if there is no official source. Javascript certainly is very popular and the icon is well established and necessary for technological lists and such (speaking from experience).
I vote to keep it in, with the current source (after a fashion, see #6399 and it's linked PR).

sachinraja:
I disagree that "de facto standard" icons should be included. The [contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors) say to "identify official logos and colors." 

Icons are supposed to represent a brand, not the community. If this icon was brought up and used by JSConf (as said in the source), it should be labeled as such. Also note that the [JS Wikipedia page](https://wikipedia.org/wiki/JavaScript) does not use the icon.

> For example, Markdown icon has become a de facto standard

I guess this issue also raises discussion on those icons.

gizmecano:
As I wrote before, I can understand this opinion: it may sound perhaps rigorous, but reading the current guidelines, it seems founded and logical... 😕

[...]

PR #6428, opened 2021-08-29
Updating guidelines unofficial source

jorgeamadosoria:
**Issue:** #6418 

Adding a textual amendment to contribution guidelines to address de facto standard icons to be added to the library via consensus of maintainers.
The text in the change is merely a proposal to get the ball rolling, I don't mind changing it completely as long as the spirit of it remains.


ericcornelissen (commented):
I think the current approach is not correct.
- First the bullet list forms a set of conditions, all of which have to be met in order for the icon to be accepted. The bullet you added is an or-condition with the bullet preceding it, which is confusing.
- This section should probably not cover how the logo should be chosen, none of the other bullets and the title doesn't indicate the section will talk about logos.

I think it makes more sense to add it in [the "1. Identify Official Logos and Colors" section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors).

jorgeamadosoria:
> I think the current approach is not correct.
> 
> * First the bullet list forms a set of conditions, all of which have to be met in order for the icon to be accepted. The bullet you added is an or-condition with the bullet preceding it, which is confusing.
> * This section should probably not cover how the logo should be chosen, none of the other bullets and the title doesn't indicate the section will talk about logos.
> 
> I think it makes more sense to add it in [the "1. Identify Official Logos and Colors" section](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors).

changed taking this input into account

CONTRIBUTING.md
```diff
@@ -86,6 +87,9 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Wikimedia (which should provide a source)
 1. GitHub repositories
 
+It may be the case that official sources simply do not exist, but an unofficial icon has gained widespread acceptance and popularity. These icons can also be included, after being reviewed and accepted by maintainers on a case-by-base basis. Javascript is the prime example of this situation.
```
mondeja, line 90: The linting is failing due to this trailing whitespace, could you remove it?

```suggestion
It may be the case that official sources simply do not exist, but an unofficial icon has gained widespread acceptance and popularity. These icons can also be included, after being reviewed and accepted by maintainers on a case-by-base basis. Javascript is the prime example of this situation.
```

mondeja (changes requested):

CONTRIBUTING.md
jorgeamadosoria: Committed suggestion

CONTRIBUTING.md
```diff
@@ -45,6 +45,7 @@ We welcome icon requests. Before you submit a new issue please make sure the ico
     - For websites, the [Alexa rank](https://www.alexa.com/siteinfo) should be less than 500k.
         - Note that for brands that have already been added the threshold for continued inclusion rises to 750k.
     - For GitHub projects, the amount of "stars" should be above 5k.
+    - For a well known or widely accepted de facto standard logo (where official sources don't exist), popularity will be assesed on a case-by-case basis by maintainers. For example, Javascript falls in this category.
     - For anything else, popularity will be judged on a case-by-case basis.
```
ericcornelissen, line 49: This feels to me like we're repeating ourselves, both bullets say "popularity will be judged on a case-by-case basis". I don't think the separate bullet is necessary. Especially as the extra bullet is (still) not primarily about the popularity but rather about the logo.

CONTRIBUTING.md
```diff
@@ -86,6 +87,9 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Wikimedia (which should provide a source)
 1. GitHub repositories
 
+It may be the case that official sources simply do not exist, but an unofficial icon has gained widespread acceptance and popularity. These icons can also be included, after being reviewed and accepted by maintainers on a case-by-base basis. Javascript is the prime example of this situation.
```
ericcornelissen, line 90: A couple suggestions, some for brevity and some for clarity:

```suggestion
It may be the case that no official source exist, but an unofficial icon has gained widespread acceptance and popularity. In such cases the unofficial icon can be included, but the details will be judged on a case-by-base basis. The JavaScript icon is an example of this
```

CONTRIBUTING.md
```diff
@@ -86,6 +87,9 @@ Official high quality brand logos and brand colors can usually be found in the f
...
+Notice that an unofficial icon won't replace or supersede an official one, even if it more popular. An unofficial icon will only be accepted if official options do not exist.
```
ericcornelissen, line 91: A couple suggestions for brevity:

```suggestion
Notice that an unofficial source will never supersede an official one, even if it is more popular. An unofficial icon will only be accepted if no official option exists.
```

CONTRIBUTING.md
```diff
@@ -45,6 +45,7 @@ We welcome icon requests. Before you submit a new issue please make sure the ico
     - For websites, the [Alexa rank](https://www.alexa.com/siteinfo) should be less than 500k.
         - Note that for brands that have already been added the threshold for continued inclusion rises to 750k.
     - For GitHub projects, the amount of "stars" should be above 5k.
+    - For a well known or widely accepted de facto standard logo (where official sources don't exist), popularity will be assesed on a case-by-case basis by maintainers. For example, Javascript falls in this category.
```
sachinraja, line 48: > For a well known or widely accepted de facto standard logo (where official sources don't exist)

I think we should pick one from "well known," "widely accepted," "de facto," and "standard," instead of all four, to make the language simpler.

sachinraja, line 48:
```suggestion
    - For a well known or widely accepted de facto standard logo where official sources don't exist, popularity will be assesed on a case-by-case basis by maintainers. For example, Javascript falls in this category.
```
This information is necessary, so I don't think the parentheses should be here.

CONTRIBUTING.md
sachinraja: "exist" would also be changed to "exists"

CONTRIBUTING.md
sachinraja: "exist" would also be changed to "exists"

CONTRIBUTING.md
jorgeamadosoria: it is indeed the same info. I did it on purpose to make the message extra clear.

CONTRIBUTING.md
jorgeamadosoria: make the choice @sachinraja. I don't really prefer any over the others

CONTRIBUTING.md
ericcornelissen: I personally disagree that it makes things extra clear, but I understand it's highly subjective. The reason I disagree is 1) now there is more text to read, 2) the added sentence isn't easy to read (at least compared to the existing sentence), and 3) again, this makes the list inconsistent (the focus changes from popularity to where the logo comes from).

In fact, I would argue it may cause confusion. The repetitions reads, to me, as if either 1) not much care was put into creating a coherent document making me less inclined to read it fully, or 2) one of the bullets is incorrect and does not apply.

CONTRIBUTING.md
sachinraja: "widely accepted" makes the most sense to me.

CONTRIBUTING.md
jorgeamadosoria: I disagree.
point 3) the list is not inconsistent. we are still talking about popularity. The previous bullets were also talking about the metrics of popularity and their sources. I think we are staying on topic. At least as much as the last bullet is.

point 2) I welcome rewrites. I've already committed yours and @sachinraja suggestions for more concise text.

point 1 and second paragraph:  Not at all. You are assuming people are going to read the full document top to bottom. While this is going to be true of some people, most people will just refer to the information they need, be it accepted sources and guidelines, or popularity metrics.
In any case, reiterating a point is not necessarily bad writing. Making the point clear in the places an user may jump into is useful. When the user don't have to scan the whole text, it's more likely that they will contribute and not be overwhelmed.

An alternative way of doing this would be to link one text to the other. Perhaps removing the bullet in the popularity list and adding a link at the end of the last one:
`- For anything else, popularity will be judged on a case-by-case basis. (for example, an unofficial de facto standard logo)` where the text in parenthesis would point to the sources text:`It may be the case that no official source exist...`

sachinraja: I don't think this should be mentioned in the popularity section as the last point already covers the case. The statement at the end (It may be the case...) can explain the policy for unofficial icons.

ericcornelissen: Again, highly subjective. I don't mind it being there but from my point of view I would just recommend not putting it there. I think @sachinraja put it best in their comment.

<details>
<summary>Just a few responses to @jorgeamadosoria</summary>

> we are still talking about popularity

Yes, but the **focus** of the sentence is on the icon being unofficial.

> I welcome rewrites

I don't think any amount of rewrite can fix my problem with this bullet.

> You are assuming people are going to read the full document top to bottom

That was not my assumption - though I must admit my comment wasn't well written. I should've said something along the lines of: "less inclined to follow the guidelines".
</details>

CONTRIBUTING.md
mondeja: > "widely accepted" makes the most sense to me

To me too.

jorgeamadosoria: widely accepted it is, then.

CONTRIBUTING.md
jorgeamadosoria: Removed bullet point.

jorgeamadosoria:
Is this ready for merging now @ericcornelissen @sachinraja ?

CONTRIBUTING.md
```diff
@@ -86,6 +86,9 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Wikimedia (which should provide a source)
 1. GitHub repositories
 
+It may be the case that no official source exist, but an unofficial icon has gained widespread acceptance and popularity. In such cases the unofficial icon can be included, but the details will be judged on a case-by-base basis. The JavaScript icon is an example of this
```
mondeja, line 89: Just a typo:

```suggestion
It may be the case that no official source exist, but an unofficial icon has gained widespread acceptance and popularity. In such cases the unofficial icon can be included, but the details will be judged on a case-by-base basis. The JavaScript icon is an example of this.
```

mondeja (changes requested):

_data/simple-icons.json
```diff
@@ -1898,7 +1898,7 @@
         {
             "title": "Conda-Forge",
             "hex": "000000",
-            "source": "https://github.com/conda-forge/conda-forge.github.io/"
+            "source": "https://github.com/conda-forge/conda-forge.github.io/blob/34adb68298ca266af13c3d615f7af8b6c232f6fb/img/anvil.svg"
```
mondeja, line 1901: It seems that you slipped this change from another PR 🙃

mondeja (changes requested):

ericcornelissen (approved):

_data/simple-icons.json
jorgeamadosoria: yes. reverted this already.

CONTRIBUTING.md
```diff
@@ -86,6 +86,9 @@ Official high quality brand logos and brand colors can usually be found in the f
 1. Wikimedia (which should provide a source)
 1. GitHub repositories
 
+It may be the case that no official source exist, but an unofficial icon has gained widespread acceptance and popularity. In such cases the unofficial icon can be included, but the details will be judged on a case-by-base basis. The JavaScript icon is an example of this.
```
sachinraja, line 89:
```suggestion
It may be the case that no official source exists, but an unofficial icon has gained widespread acceptance and popularity. In such cases the unofficial icon can be included, but the details will be judged on a case-by-case basis. The JavaScript icon is an example of this.
```

Just two grammar fixes

jorgeamadosoria: good catch! Committed the changes already.

jorgeamadosoria:
Thanks @mondeja and @sachinraja for the fixes. What do you think is pending yet?

sachinraja (approved):

Issue #6437, opened 2021-08-29
Update:  FNAC Icon


dirien:
### Brand Name

FNAC

### Official Resources for Icon and Color

From their website: https://www.fnac.com/

```
[...]

PR #6438, opened 2021-08-29
Update Fnac icon

dirien:
Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

image

**Issue:** #6437 
**Alexa rank:** 3,058
  

### Checklist


jorgeamadosoria:
The svg has slight discrepancies with the source, due to the svgo optimization.
I tried with `precision:5` and the match is 100%.

dirien:
@jorgeamadosoria: how to i set the precision in inkscape?

jorgeamadosoria:
> @jorgeamadosoria: how to i set the precision in Inkscape?

what I do is look in the `svgo.config.js` file for the main repository, and set the precision parameter there. If you optimize the SVG from inkscape, it will show a dialog window asking for the precision you want to set (this is available in `Save As... -> Optimized SVG` file type).
I think there is also a command line parameter if you want to execute SVGO outside Inkscape or simple-icons, but I haven't used it.

If you are going to change the precision parameter in `svgo.config.js`, please remember to revert it back to 3 before committing. 

As I understand it precision 3 is the desired value for all the icons, but in the cases in which you can't make it work with that, precision 4 or 5 may be accepted. @service-paradis, can you please confirm or correct this interpretation, please?

service-paradis:
> As I understand it precision 3 is the desired value for all the icons, but in the cases in which you can't make it work with that, precision 4 or 5 may be accepted. @service-paradis, can you please confirm or correct this interpretation, please?

Yes, we really try to keep precision at 3. For some icons, it does not work. In those cases, a precision of 4 can be accepted. If it still does not work, we accept a precision of 5 (it is rarely necessary to increase to 5).

jorgeamadosoria:
> > As I understand it precision 3 is the desired value for all the icons, but in the cases in which you can't make it work with that, precision 4 or 5 may be accepted. @service-paradis, can you please confirm or correct this interpretation, please?
> 
> Yes, we really try to keep precision at 3. For some icons, it does not work. In those cases, a precision of 4 can be accepted. If it still does not work, we accept a precision of 5 (it is rarely necessary to increase to 5).

For some reason I'm running into these cases quite often :(

jorgeamadosoria:
> hi @service-paradis and @jorgeamadosoria,
> 
> changed the percision to 5. Looking for your feedback 😄

I changed it to precision 3 and it worked.
@dirien, plz check on your side. if  you get a match, LGTM

Issue #1436, opened 2019-05-27
Request: Pokémon GO


KovuTheHusky:
**Name:** Pokémon GO
**Website:** https://www.pokemongo.com (The Pokémon Company) & https://pokemongolive.com (Niantic)
**Official resources for icon and color:** The cleanest vector is probably the omni notification icon from their app's APK. I have included it here, but it is only served as a PNG.

Normally I would open a PR myself, but for this one I think It's probably better for me to leave the rest to somebody more experienced than I am.

![omni_notification_icon](https://user-images.githubusercontent.com/361739/58436828-51a8a500-8095-11e9-8d7e-9acda1057344.png)

davidklebanoff:
The Pokemon Go brand is certainly big enough to warrant having an icon in Simple Icons, but I don't believe the notification icon within the app is the appropriate icon to represent the brand.

KovuTheHusky:
I mean, it's basically the app icon but single color. I'm not sure how you get closer to the app icon without adding additional colors. There is a reverse style icon as well if you think that fits better. As in the color and transparency swapped.

![app_icon](https://user-images.githubusercontent.com/361739/58500541-68600200-8150-11e9-8f76-2ab0bf0b7f32.png)
![notification_icon](https://user-images.githubusercontent.com/361739/58500599-7e6dc280-8150-11e9-8d96-33a79b84cc95.png)
![notification_icon](https://user-images.githubusercontent.com/361739/58500582-7877e180-8150-11e9-8f40-e5f532e3a51f.png)
![omni_notification_icon](https://user-images.githubusercontent.com/361739/58500610-8463a380-8150-11e9-9a53-45b164956876.png)

adamrusted:
Looks like Pokémon GO is outside the threshold for inclusion in Simple Icons at the moment. Happy for another @simple-icons/maintainers to say we should keep this open, but for now I'll look to close this as Out of Scope.

adamrusted:
Ignore me, got a little trigger happy before checking the Niantic version of the website, which is ranking a lot higher. 😅

adamrusted:
Closing, see [this page](https://nianticlabs.com/terms) on Niantic's website.
![image](https://github.com/simple-icons/simple-icons/assets/1830867/cae7bb6c-f6e1-4d71-943b-49694963c81f)

PeterShaggyNoble:
Does Niantic own the trademark on this logo, though, @adamrusted, or does it belong to The Pokémon Company?

[...]

PR #6463, opened 2021-09-03
Adding Pokemon Go icon

jorgeamadosoria:
pokemongo

**Issue:** #1436

### Description
Re-taking #2340 . Used the same source and hex, got the icon to work with precision 4. It seems to be a 100% match. There may be details that need to be worked out still.


jorgeamadosoria:

Issue #6476, opened 2021-09-05
Request: Argos


dirien:
### Brand Name

Argos

### Website

https://www.argos.co.uk/

### Popularity Metric

Alexa: 5,229

### Official Resources for Icon and Color

svg: https://argos.careers/wp-content/themes/argos/assets/images/argos-logo.svg
color: `#DA291C`

### Additional Comments

Took the svg from their careers page as the main page had no svg only image

### Contributing

Yes

PR #6477, opened 2021-09-05
Add Argos icon

dirien:
Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

image

**Issue:** #6476 
**Alexa rank:** 5,229
  

### Checklist

### Description

see issue


sachinraja (changes requested):
I think this should be based off the smaller icon (from the footer):
image

dirien:
@sachinraja, tbh I only now the Argos logo with the swoosh/smile! I would never think the A with swoosh/smile is Argos?

sachinraja:
Do you mean you only see the icon currently implemented in this PR used everywhere? If so then we can move forward with that

dirien:
@sachinraja: Yes.

sachinraja (approved):

PR #6505, opened 2021-09-10
Use getter for `svg` instead of `path` in packaged javascript version of icons

### LINKED ISSUE #6481: inline `path` for js files
### Kind of Issue

Performance

### This issue concerns the...

NPM package

### Package Version

All

### Other Software

not needed

### Description

I think `icon.path` should be inlined [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/build/templates/icon-object.js) instead of being computed as a getter. The majority of use cases are going to need `path`, and not the stringified `svg`. For example, a React component could like something like the following:
```js
const SIIcon = ({ icon }) => <svg><path d={icon.path} /></svg>
```
Instead of running the (relatively slow) regex on render to get the path, it could already be available. The `svg` property is not used here and I'm not sure what the use case for that is.

### Proposal
We should just remove the `svg` property and inline `path` as an actual property. If we want to avoid the breaking change, `svg` can be the computed value instead of `path`. This makes more sense than the current way because it does not require a regex.

mondeja:
Closes #6481

Instead of computing a regular expression getting the `path` attribute of icons, use string concatenation with `+` operator. 

This is a great performance improvement for `path` attribute (more than 20 times faster) and a very small performance decrease for `svg` attribute. See issue for performance tests. Also this reduces a bit the size for each produced javascript file, exactly 25 characters minus for each.


scripts/build/package.js
```diff
@@ -27,7 +27,7 @@ const indexTemplate = fs.readFileSync(indexTemplateFile, UTF8);
 const iconObjectTemplate = fs.readFileSync(iconObjectTemplateFile, UTF8);
 
 const data = require(dataFile);
-const { getIconSlug, titleToSlug } = require("../utils.js");
+const { getIconSlug, svgToPath, titleToSlug } = require("../utils.js");
```
sachinraja, line 30: Looks good, but I think this should be a [local helper function](https://github.com/simple-icons/simple-icons/blob/b9cc1788d1efa38de69a17e1c8cb9a86eeed4954/scripts/build/package.js#L32) based on https://github.com/simple-icons/simple-icons/pull/6189#discussion_r693525489.

scripts/build/templates/icon-object.js
```diff
@@ -1,10 +1,10 @@
 {
   title: '%s',
   slug: '%s',
-  svg: '%s',
-  get path() {
-    return this.svg.match(/<path\s+d="([^"]*)/)[1];
+  get svg() {
+    return '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>%s</title><path d="' + this.path + '"/></svg>';
```
sachinraja, line 5: Can we use `this.title` here instead of escaping the title and duplicating it?

mondeja: > Can we use `this.title` here instead of escaping the title and duplicating it? 

I've tested it and it is slower, so I prefer to maintain the string as it is even if we could decrease the size of files using `this.title`.

> You may want to use a template literal to make it look better.

Template literals are an ES6 feature right? It seems that [is not universally supported](https://caniuse.com/?search=string%20template%20literals), so I prefer to maintain the compatibility for all platforms and NodeJS versions. This would be a breaking change.

scripts/build/package.js
mondeja: 🤔  I'm not really sure, perhaps `svgToPath` makes more sense to declare it an utility as it is more general than `slugToVariableName` because is converting from an attribute to another like `titleToSlug`. Maybe @ericcornelissen could clarify here what he relies on to declare functions as utilities or locals ones.

scripts/build/templates/icon-object.js
mondeja: >>  Can we use `this.title` here instead of escaping the title and duplicating it? 
> I've tested it and it is slower, so I prefer to maintain the string as it is even if we could decrease the size of files using `this.title`.

Just seeing now [@ericcornelissen comment in issue](https://github.com/simple-icons/simple-icons/issues/6481#issuecomment-917197995) regardling this. I want to share here the performance and file size tests that I've done because to take the current approach of duplicate the title:

```sh
$ time node -e "for (let i=0; i<500000; i++) { require('./icons/elsevier').svg }"

real	0m0,152s
user	0m0,149s
sys	0m0,015s
$ cat icons/elsevier.js | wc --bytes
54338
$ sed -i "s/<title>Elsevier<\/title>/<title>\'+this.title+\'<\/title>/g" icons/elsevier.js
$ time node -e "for (let i=0; i<500000; i++) { require('./icons/elsevier').svg }"

real	0m0,172s
user	0m0,170s
sys	0m0,013s
$ cat icons/elsevier.js | wc --bytes
54344
$ sed -i "s/<title>'+this.title+'<\/title>/<title>Elsevier<\/title>/g" icons/elsevier.js
```

I'm on Ubuntu 20.04. I suppose that it is replicable in other machines as well. If, even knowing this performance decreasement and that the file size would be greater for icons whose title has minus than 14 characters ("Elsevier" has 8 and the difference is +6 changing by `'+this.title+'`) do you want to go with that change, I'm open to it, but IMHO I think it is unnecessary in this case.

scripts/build/package.js
ericcornelissen: As I said in the other comment, it's not very clear where these kinds of functions should go at the moment. It's mostly a matter of "what feels right". I tend to ask myself the question: "could this be useful outside of this script?"

I would agree with @mondeja here that it's similar to `titleToSlug` in that converting between these properties may be useful elsewhere as well, and therefor can stay in the utilities file.

scripts/build/templates/icon-object.js
sachinraja: I don't think saving a few bytes will really matter when we're already in the tens of thousands. I do see a slight performance hit (also on Ubuntu 20.04), so I'm okay with inlining `this.title`. I think we should wait for a third opinion though.

ericcornelissen: > Template literals are an ES6 feature right? It seems that is not universally supported, so I prefer to maintain the compatibility for all platforms and NodeJS versions. This would be a breaking change.

We should indeed avoid using template literals until we're ready to introduce a breaking change that bumps up the minimum version of JS significantly. We _could_ use it to make the source look cleaner if we can configure, e.g., UglifyJS to convert template literals to string concatenation. But not sure if that's worth the effort.


> I want to share here the performance and file size tests that I've done before to take the current title duplication approach: [...]

Thanks for sharing @mondeja, that's very useful. I think I agree that embedding the title value directly is the better option here as it improves performance. Duplication is not a concern here as we generate the code and when considering all brands I'm 1) not even sure using `this.title` would decrease the package size and 2) of the opinion that the performance increase is more significant.

scripts/build/package.js
sachinraja: > could this be useful outside of this script?

Oh okay, it makes sense to put it in the utilities file then.

scripts/build/templates/icon-object.js
sachinraja: I think we should save the escaped title to a variable though (in `iconToObject`) just to avoid escaping it twice.

ericcornelissen: Actually, the title value in the `<title>` probably shouldn't be escaped

mondeja: Done in https://github.com/simple-icons/simple-icons/pull/6505/commits/4e375963525b9795b1f2b5f193f538c2429829bf :+1:

scripts/build/package.js
```diff
@@ -47,10 +47,12 @@ function licenseToObject(license) {
   return license;
 }
 function iconToObject(icon) {
+  const escapedIconTitle = escape(icon.title);
   return util.format(iconObjectTemplate,
-    escape(icon.title),
+    escapedIconTitle,
     escape(icon.slug),
-    escape(icon.svg),
+    escapedIconTitle,
```
ericcornelissen, line 54: This one should also have HTML character escaped. A reverse version of [`htmlFriendlyToTitle`](https://github.com/simple-icons/simple-icons/blob/b9cc1788d1efa38de69a17e1c8cb9a86eeed4954/scripts/utils.js#L39-L43) should probably suffice.

mondeja: Good point, done in https://github.com/simple-icons/simple-icons/pull/6505/commits/7b061f62a7cb6751dc30c15033c4bb3d97039978

sachinraja (approved):

ericcornelissen (commented):
The implementation looks good to me now, good work everyone :+1: 

I'm however wondering if, now that we construct the `svg` value, we should update the test for the `svg` property to actually verify that it matches the content of the related `.svg` file (instead of just that it's a string: [ref 1](https://github.com/simple-icons/simple-icons/blob/b9cc1788d1efa38de69a17e1c8cb9a86eeed4954/tests/index.test.js#L30), [ref 2](https://github.com/simple-icons/simple-icons/blob/b9cc1788d1efa38de69a17e1c8cb9a86eeed4954/tests/icons.test.js#L29)) :thinking:

mondeja:
Are you thinking in something like next or too much literal?

```js
expect(subject.svg).toEqual(
  '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">'
  + `<title>${titleToHtmlFriendly(subject.title)}</title>`
  + `<path d="${subject.path}"/></svg>`
);
```

sachinraja:
I think @ericcornelissen means to read the SVG file and check if `subject.svg` is equal to that value. Not sure if this is a good idea though as every icon file will be read so the tests may be slow.

ericcornelissen:
I was indeed suggesting to read the SVG files and compare that way. The reason I didn't do it like that before was indeed performance, however, with f6416401f1a125eef964d051b7c4846662af42eb I feel like we're testing whether our implementation works by, well, re-implementing it in the test.

sachinraja:
This also feels like re-implementing it. I think we should just read the files (asynchronously).

mondeja:
IHMO, I think that it is too slow. Takes ~150 seconds in my machine. Are you sure that you want to test it using that approach? @ericcornelissen @sachinraja

sachinraja:
All 36,103 tests, including the file reads, only take 7 seconds for me. I did this to test:
```js
const fs = require("fs");

test(`${icon.title} has an "svg" value`, () => {
  const svgFileContents = fs.readFileSync(path.join(process.cwd(), 'icons', `${slug}.svg`), 'utf8')
  expect(subject.svg).toEqual(svgFileContents);
});
  ```
  I noticed something else though, why do we have two test files iterating over the icons and testing the same things (see [here](https://github.com/simple-icons/simple-icons/tree/develop/tests))?

**Comment by mondeja** [comment]
[It takes more than 3 minutes in Github Actions](https://github.com/simple-icons/simple-icons/pull/6505/checks?check_run_id=3580657691#step:7:1) @sachinraja @ericcornelissen

**Comment by sachinraja** [comment]
Can confirm that it takes very long locally with the `--ci` flag.

File: tests/icons.test.js [hunk 4]
```diff
@@ -25,8 +30,10 @@ icons.forEach(icon => {
     expect(subject.source).toEqual(icon.source);
   });
 
-  test(`${icon.title} has an "svg" value`, () => {
+  test(`${icon.title} has an "svg" value`, (done) => {
```
sachinraja, line 33: It takes a long time because the `done` callback argument here. Jest is waiting for it to be called to finish the test: https://jestjs.io/docs/asynchronous#callbacks.

mondeja: You're right 👍🏼

tests/icons.test.js
```diff
@@ -27,6 +32,17 @@ icons.forEach(icon => {
...
+    const svgFileContents = fs.readFileSync(svgPath, 'utf8')
+      .replace('" />', '"/>')
+      .replace(/\r?\n/, '')
+      .replace('></path>', '/>')
+      .replace(
+        'xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 24 24\">',
+        'viewBox=\"0 0 24 24\" xmlns=\"http://www.w3.org/2000/svg\">')
+      .replace(
+        /<title>([^<]+)<\/title>/,
+        `<title>${titleToHtmlFriendly(subject.title)}</title>`);
```
mondeja, line 44: To prevent these replacements the icons contents must be checked before, maybe creating some new SVG lint rules.

mondeja: > maybe creating some new SVG lint rules

Opened a PR for this on #6513

sachinraja: I'm ok with merging this and removing them in the follow up PR.

sachinraja (approved):

sachinraja:
Not saying we have to do it in this PR, but can we address that there are two test files testing the same thing?

mondeja:
> But can we address that there are two test files testing the same thing?

Yes, we should refactor the tests so we don't have to use the same resources twice.

ericcornelissen:
> Not saying we have to do it in this PR, but can we address that there are two test files testing the same thing?

They don't quite test the same thing, they test the same thing for both (the generated) `index.js` (in `index.test.js`) file and for all (generated) `icon/*.js` files (in `icons.test.js`).

To avoid the fact that this adds `2 * n` tests per brand, we could switch to testing the code that generates these files rather then the generated files. Also, after #6189 we could consider removing the individual `icon/*.js` files as their purpose will be fulfilled (assuming a bundler) by the added file (but this requires a breaking change) - which should improve the performance of `icons.test.js` as it doesn't have to load >2000 individual `.js` files.

tests/index.test.js
```diff
@@ -1,10 +1,15 @@
+const path = require('path');
+const fs = require('fs');
 const { icons } = require('../_data/simple-icons.json');
 const simpleIcons = require('../index.js');
-const { getIconSlug } = require("../scripts/utils.js");
+const { titleToHtmlFriendly, getIconSlug, iconSvgTemplate } = require("../scripts/utils.js");
```
ericcornelissen, line 5:
```suggestion
const { titleToHtmlFriendly, getIconSlug } = require("../scripts/utils.js");
```

ericcornelissen (changes requested):

tests/index.test.js
```diff
@@ -28,6 +33,17 @@ icons.forEach(icon => {
...
+    const svgFileContents = fs.readFileSync(svgPath, 'utf8')
+      .replace('" />', '"/>')
+      .replace(/\r?\n/, '')
+      .replace('></path>', '/>')
+      .replace(
+        'xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 24 24\">',
+        'viewBox=\"0 0 24 24\" xmlns=\"http://www.w3.org/2000/svg\">')
+      .replace(
+        /<title>([^<]+)<\/title>/,
+        `<title>${titleToHtmlFriendly(subject.title)}</title>`);
```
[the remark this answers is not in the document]
ericcornelissen: From the Jest output you shared:

```
[...]
Expected: [...]<title>Let&apos;s Encrypt</title>[...]
Received: [...]<title>Let&amp;apos;s Encrypt</title>[...]
[...]
```

Looks like there's bug in `titleToHtmlFriendly` (ampersand should be replaced first), good thing we tested it this way :upside_down_face:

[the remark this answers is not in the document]
mondeja: It is a problem of unifying the data. We want that these `title`s must be html-encoded in _icons/*.svg_ SVG contents or rather we should accept the names html-decoded in _icons/*.svg_ and encode them in the `svg` getter building the package?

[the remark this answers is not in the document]
sachinraja: I think we should accept only encoded names and fix the ones that are not encoded.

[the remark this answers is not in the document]
mondeja: > I think we should accept only encoded names and fix the ones that are not encoded.

I agree and suggest to create another lint rule for this. What do you think @ericcornelissen?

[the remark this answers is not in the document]
ericcornelissen: I'm not sure if I follow @mondeja. Here's what I can say:

- The `titleToHtmlFriendly` currently in this Pull Request is incorrect, as can be seen from the "Let’s encrypt" example ("Let’s encrypt" -> "Let&apos;s encrypt" -> "Let&amp;apos;s encrypt"). This should definitely be fixed.
- The `.svg` file _should_ all already have properly encoded `<title>` values. If not, that should be fixed. If we can lint for that, we probably should.
- I think the `svg` value in all `js` files should be an exact match of the contents of the corresponding SVG so that it can be embedded in HTML as is. This is the way it has always been, and changing that could definitely be considered a breaking change from my point of view. If you want to add a non-HTML safe version of the SVG in the JavaScript (whether hardcoded or through a helper function), I guess we can do that but I would have no clue as to why anyone would want that.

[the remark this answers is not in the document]
mondeja: Opened a PR for this in #6522

scripts/utils.js
```diff
@@ -41,4 +47,15 @@ module.exports = {
...
+
+  /**
+   * Converts a brand title (as it is seen in simple-icons.json) into a brand
+   * title in HTML/SVG friendly format.
+   * @param {String} brandTitle The title to convert
+   */
+  titleToHtmlFriendly: brandTitle => (
+    brandTitle
+      .replace(/’/g, "&apos;")
+      .replace(/&/g, "&amp;")
```
ericcornelissen, line 59: > The `titleToHtmlFriendly` currently in this Pull Request is incorrect, as can be seen from the "Let’s encrypt" example (`Let’s encrypt` -> `Let&apos;s encrypt` -> `Let&amp;apos;s encrypt`). This should definitely be fixed.

```suggestion
      .replace(/&/g, "&amp;")
      .replace(/’/g, "&apos;")
```

mondeja: As I stated in #6522, `letsencrypt` has a wrong title. The source in `data/simple-icons.json` uses the character `’` (`&#8217;`) but the SVG file uses `'` (`&apos;`). You are converting here the character  `’` (`&#8217;`) to `&apos;` (`'`), which is not true.

ericcornelissen: @mondeja, this suggestion is just about switching these two statements around, it's not my suggestion that replaces `’` by `&apos;`, that is how this function works as of 58a95c42f41d37c85e0fdc25009d328d749d9515. I agree that we shouldn't replace `’` (or `'` for that matter) by `&apos` but I just wanted to have this separate comment that `/&/g` always needs to be replaced first. (This may or may not be rendered obsolete by #6522, but I wanted to make sure we don't forget about it if it isn't rendered obsolete.)

```suggestion
      .replace(/&/g, "&amp;")
      .replace(/’/g, "&#8217;")
```

And as I already [said here](https://github.com/simple-icons/simple-icons/pull/6522#issuecomment-920704634) - at least at the time of adding `letsencrypt` - it's title should use `’`/`&#8217;`, so the SVG `<title>` has to be changed.

mondeja: OK, but anyways it seems to me that this implementation is wrong because the `&` characters of an already encoded input would be also replaced, regardless of the order of replacements. I think that this should be addressed in #6522.

ericcornelissen: But this function is intended to convert not-encoded titles (i.e the ones in the data file) to html-encoded (`<title>`) values

mondeja: Yes, but it could be safer. I want to replace this function by a safer converter like:

```js
const newHtmlFriendlyToTitle = (htmlFriendlyTitle) => {
  return htmlFriendlyTitle.replace(
    /&(amp|quot|#039|apos|lt|gt);/g,
    (_, rep) => {
      return {amp: '&', quot: '"', '#039': "'", apos: "'", lt: '<', gt: '>'}[rep]
    }
  );
};

console.log(newHtmlFriendlyToTitle("Hello &amp; world"));  // Hello & world
console.log(newHtmlFriendlyToTitle("Hello & world"));      // Hello & world
console.log(newHtmlFriendlyToTitle("Hello world &"));      // Hello world &

// Actual implementation
const htmlFriendlyToTitle = (htmlFriendlyTitle) => {
  return htmlFriendlyTitle
    .replace(/&/g, "&amp;")
    .replace(/’/g, "&#8217;");
}

console.log(htmlFriendlyToTitle("Hello &amp; world"));  // Hello &amp;amp; world
console.log(htmlFriendlyToTitle("Hello & world"));      // Hello &amp; world
console.log(htmlFriendlyToTitle("Hello world &"));      // Hello world &amp;
```

You know, it would be necessary also to handle the cases where there were decimal or hexadecimal representations. But anyways, I suggest to discuss this in #6505.

mondeja: Are we going to include every single unicode character in this function?

```
.replace(/’/g, "&#8217;")
``` 

This implementation makes no sense, Please, could we discuss this in #6522?

ericcornelissen: To reiterate, I want to avoid a situation where we decide not to merge #6522 and forget about this error in this Pull Request. I agree the suggestion should change in the context of #6522, the current suggestion is in the context of the current state of this Pull Request.

mondeja: > I want to avoid a situation where we decide not to merge #6522 and forget about this error in this Pull Request.

In that case, I'm going to leave this conversation unresolved just to remember about it :+1:

scripts/build/package.js
```diff
@@ -49,18 +49,20 @@ const licenseToObject = (license) => {
...
-const iconToObject = (icon) => {
+function iconToObject(icon) {
+  const escapedIconTitle = escape(icon.title);
   return util.format(
     iconObjectTemplate,
-    escape(icon.title),
+    escapedIconTitle,
     escape(icon.slug),
-    escape(icon.svg),
+    titleToHtmlFriendly(escapedIconTitle),
```
ericcornelissen, line 58: I think this should be

```suggestion
    escape(titleToHtmlFriendly(icon.title)),
```

because `escape(x)` ensures any single quotes (`'`) in the string being embedded are escaped properly when the string is being embedded into a JavaScript template in the string [^1]. So, the escaping should happen after the title is transformed to a HTML friendly string to 1) avoid accidentally encoding what `escape` added and 2) ensure any single quote added by the encoding is escaped. (even if neither of these _should_ happen, and so it doesn't really matter, this suggestion is the _correct_ implementation I think so it's better to change for clarity).

[^1]: Example, `util.format("let x = '%s'", escape("foo'bar"))` results in `let x = 'foo\'bar'`.

scripts/utils.js
```diff
@@ -36,11 +42,28 @@ module.exports = {
...
+
+  /**
+   * Converts a brand title (as it is seen in simple-icons.json) into a brand
+   * title in HTML/SVG friendly format.
+   * @param {String} brandTitle The title to convert
+   */
+  titleToHtmlFriendly: (brandTitle) =>
+    brandTitle
+      .replace(/"/g, '&quot;')
+      .replace(/&/g, '&amp;')
```
ericcornelissen, line 62: The current replacements will change `"` into `&amp;quot;` [^1], it should be the following instead.

```suggestion
      .replace(/&/g, '&amp;')
      .replace(/"/g, '&quot;')
```

[^1]: You can run `"\"".replace(/"/g, '&quot;').replace(/&/g, '&amp;')` in Node to verify this.

tests/index.test.js
```diff
@@ -26,8 +31,14 @@ icons.forEach((icon) => {
     expect(subject.source).toEqual(icon.source);
   });
 
-  test(`${icon.title} has an "svg" value`, () => {
+  test(`${icon.title} has a valid "svg" value`, () => {
     expect(typeof subject.svg).toBe('string');
+    const svgFileContents = fs
+      .readFileSync(svgPath, 'utf8')
+      .replace(/\r?\n/, '');
```
ericcornelissen, line 38: Not necessarily in scope for this Pull Request, but given that we also do this transformation here:

https://github.com/simple-icons/simple-icons/blob/3fdcb11536ef0356f3f84594fbee835531bcd05b/scripts/build/package.js#L90

and now in both test suites, it might make sense to extract the functionality into a helper function.

tests/icons.test.js
```diff
@@ -25,8 +30,14 @@ icons.forEach((icon) => {
...
 
-  test(`${icon.title} has an "svg" value`, () => {
+  test(`${icon.title} has a valid "svg" value`, () => {
     expect(typeof subject.svg).toBe('string');
+    const svgFileContents = fs
+      .readFileSync(svgPath, 'utf8')
+      .replace(/\r?\n/, '');
+    expect(subject.svg.substring(subject.svg.indexOf('<title>'))).toEqual(
+      svgFileContents.substring(svgFileContents.indexOf('<title>')),
+    );
```
ericcornelissen, line 40: Shouldn't the `.svg` content and the `subject.svg` file not just equal eachother?

tests/index.test.js
mondeja: I think makes more sense to unify newlines in SVGs contents, but yes in another PR.

tests/icons.test.js
mondeja: The attributes `role`, `viewBox` and `xmlns` are not ordered in all SVGs, but in the template now there are.

ericcornelissen: :unamused:

Wonder if that makes it worth it for us to enforce the ordering through SVGLint, so that we can update all the SVGs and then these tests :thinking: Definitely outside the scope of this Pull Request, but it could be worth opening a ticket if you agree.

tests/index.test.js
ericcornelissen: Unless the unification is _no newlines_ (see also  #1682) you'll need a utility function like this (at least[^1]) to be able to build the package correctly.

[^1]: In case of all `.svg`s having a newline, I guess you can test against `${subject.svg}\n`, but that results in duplication between the two test files.

scripts/build/package.js
```diff
@@ -49,18 +49,19 @@ const licenseToObject = (license) => {
   }
   return license;
 };
-const iconToObject = (icon) => {
+function iconToObject(icon) {
```
ericcornelissen, line 52: Arrow function? :thinking: 

```js
const iconToObject = (icon) => {
```

(Don't forget to add the semicolon back at the end of the declaration.)

scripts/utils.js
```diff
@@ -36,11 +42,28 @@ module.exports = {
    * it is seen in simple-icons.json)
    * @param {String} htmlFriendlyTitle The title to convert
    */
+
```
ericcornelissen, line 45:
```suggestion
```

tests/icons.test.js
```diff
@@ -25,8 +30,14 @@ icons.forEach((icon) => {
     expect(subject.source).toEqual(icon.source);
   });
 
-  test(`${icon.title} has an "svg" value`, () => {
+  test(`${icon.title} has a valid "svg" value`, () => {
     expect(typeof subject.svg).toBe('string');
+    const svgFileContents = fs
+      .readFileSync(svgPath, 'utf8')
+      .replace(/\r?\n/, '');
```
ericcornelissen, line 37: Not sure about this, but would making the test asynchronous more efficient?

(didn't test if this suggestion works as is)

```suggestion
  test(`${icon.title} has a valid "svg" value`, async () => {
    expect(typeof subject.svg).toBe('string');
    const _svgFileContents = await fs.readFile(svgPath, 'utf8');
    const svgFileContents = _svgFileContents.replace(/\r?\n/, '');
```

ericcornelissen (approved):
Looks good to me; just added two comments for style and one to potentially improve the performance of the tests.

tests/icons.test.js
mondeja: Opened #6826 to track it 👍🏼

tests/icons.test.js
mondeja: It doesn't changes performance

sachinraja (approved):
Feel free to merge

PR #6512, opened 2021-09-12
Updating hexes for Chromecast and Google Classroom

jorgeamadosoria:
googlecast

googleclassroom

colors from the icons on https://about.google/products/#all-products, translated to the closest colors in Google's color palette as depicted here: https://www.behance.net/gallery/9028077/Google-Visual-Assets-Guidelines-Part-1


adamrusted:
# Google Cast
So with this one, I was unsure on whether it'd be a breaking change. There are now two 'cast' icons, one called `Chromecast` which uses this icon, but `Google Cast` uses a filled-in icon (and also redirects to a Chromecast page). As such, I would suggest we rename `Google Cast` to `Chromecast`, adding `Google Cast` in as an alias. Could another @simple-icons/maintainers confirm this wouldn't break anything?

# Google Classroom
Is there any documentation on which of the two colours (green or yellow) is preferred when referring to the brand? I would've said personally that the green is more prominent - but without an official source, an argument could be made for either.

ericcornelissen:
Aliases don't do anything yet so your suggestion regarding cast would be a breaking change

adamrusted:
> Aliases don't do anything yet so your suggestion regarding cast would be a breaking change

Thought as much. For the sake of this PR then, should we add the alias of 'Chromecast', so that it can be found by that name (eventually) - and create a new issue to rename it properly as of v6?

jorgeamadosoria:
> > Aliases don't do anything yet so your suggestion regarding cast would be a breaking change
> 
> Thought as much. For the sake of this PR then, should we add the alias of 'Chromecast', so that it can be found by that name (eventually) - and create a new issue to rename it properly as of v6?

That's done now

sachinraja:
> create a new issue to rename it properly as of v6

Done in #6901

sachinraja:
> # Google Classroom
> Is there any documentation on which of the two colours (green or yellow) is preferred when referring to the brand? I would've said personally that the green is more prominent - but without an official source, an argument could be made for either.

I would also say that the green is more prominent. Otherwise, this looks good.

jorgeamadosoria:
> > # Google Classroom
> > Is there any documentation on which of the two colours (green or yellow) is preferred when referring to the brand? I would've said personally that the green is more prominent - but without an official source, an argument could be made for either.
> 
> I would also say that the green is more prominent. Otherwise, this looks good.

Changed from yellow to green.
If you think it's fine, @sachinraja can we merge?

sachinraja (approved):

PR #6513, opened 2021-09-13
Add lint rule to unify self-closing tags format

mondeja:
Testing SVG icons content in #6505 I've found that some have next inconsistencies:

- `cytoscapedotjs.svg` has an explicit closing `path` tag (`<path id="..."></path>`) while the rest have a `<path d="..."/>` opening-closing tag.
- 3 icons have an space before a self-closing tag (` />`) and this character can be removed.

So this pull request adds an `svg-format` rule which triggers ~~these two...~~ the first inconsistency and fixes `extraneous` rule to make trigger the second. The unification simplifies testing in #6505 and decreases repository/package size.


.svglintrc.js
```diff
@@ -569,6 +569,31 @@ module.exports = {
...
+            const explicitClosingPathIndex = ast.source.indexOf('</path>');
+            if (explicitClosingPathIndex > -1) {
+              const reason = `found explicit closing "path" tag at index ${explicitClosingPathIndex}.`
+                          + ' The syntax \'></path>\' should be replaced by \'/>\'.';
+              reporter.error(`Invalid SVG content format: ${reason}`);
+            }
+
+            // Don't allow spaces before implicit closing tags
+            const spacedImplicitClosingTagIndexes = [];
+            let index = ast.source.indexOf(' />', -1);
```
ericcornelissen, line 586: what if an SVG has `../>` (i.e. with two spaces)? Looks like the reported index would be off by one 🤔

.svglintrc.js
```diff
@@ -569,6 +569,31 @@ module.exports = {
...
+          function(reporter, $, ast) {
+            reporter.name = 'svg-format';
+
+            // Don't allow explicit '</path>' closing tag
+            const explicitClosingPathIndex = ast.source.indexOf('</path>');
+            if (explicitClosingPathIndex > -1) {
+              const reason = `found explicit closing "path" tag at index ${explicitClosingPathIndex}.`
+                          + ' The syntax \'></path>\' should be replaced by \'/>\'.';
+              reporter.error(`Invalid SVG content format: ${reason}`);
+            }
```
ericcornelissen, line 582: I've never heard of the term "explicit closing tag" and couldn't really find it being used in a _credible_ way in my quick 2 minutes Google search. Maybe we can instead use the term "self-closing" tag, e.g. rewrite it along the lines of

```suggestion
            // Don't allow explicit '</path>' closing tag
            const pathClosingTagIndex = ast.source.indexOf('</path>');
            if (pathClosingTagIndex > -1) {
              const reason = `found a closing "path" tag at index ${explicitClosingPathIndex}.`
                          + ' The path should be self-closing, use  \'/>\' instead of \'></path>\'.';
              reporter.error(`Invalid SVG content format: ${reason}`);
            }
```

Also wondering if we can use [`String.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) here, but I'm not sure if `ast.source` is a string.

ericcornelissen (changes requested):
I'm kinda surprised these are not being picked up by [the `extraneous` rule](https://github.com/simple-icons/simple-icons/blob/e64b98144fe9130373c102d3c17d80f9cfbcc671/.svglintrc.js#L485-L491) and I'm wondering we should try to include it there. I guess the advantage of this approach is that we can be more specific about where the error occurs? In that case, I still wonder if we can't just leverage regular expressions - in particular to avoid the (somewhat dangerous) `while` loop.

.svglintrc.js
mondeja: In that case, the next time the linter will also trigger this rule.

.svglintrc.js
mondeja: > Also wondering if we can use [`String.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) here, but I'm not sure if `ast.source` is a string.

Done in https://github.com/simple-icons/simple-icons/pull/6513/commits/bb7847863e624e0f6e208fad555a915b5c9fef57

.svglintrc.js
ericcornelissen: I see. I personally think a linter should not have to run multiple times to tell you all the things that are currently wrong with the code but if I'm the only one that thinks so we can leave it like this.

mondeja: >  I personally think a linter should not have to run multiple times to tell you all the things that are currently wrong with the code

This also happen with other rules as well. As far as I know we are not linting the possible future changes that we are requiring to the user to make pass the current SVG.

mondeja:
> I'm kinda surprised these are not being picked up by the extraneous rule and I'm wondering we should try to include it there.

It seems because that rule is not taking the real source as input, let me fix it and we can remove `Don't allow spaces before implicit closing tags` rule.

.svglintrc.js
mondeja: I've removed this part of the rule because is now triggered by the `extraneous` rule.

ericcornelissen (approved):

ericcornelissen:
Awesome, thanks @mondeja! :confetti_ball:

Issue #6542, opened 2021-09-15
Update: Toggl icon


jorgeamadosoria:
### Brand Name

Toggl

### Official Resources for Icon and Color

https://toggl.com/media-toolkit

### Additional Comments

change hex from `#E01B22` to `#E57CD8`

### Contributing

Yes

service-paradis:
In a future major release, I wonder if we should rename this icon as "Toggle Track"...

jorgeamadosoria:
> In a future major release, I wonder if we should rename this icon as "Toggle Track"...

they do seem to name their product as such, so I think that wouldn't hurt.

service-paradis:
The icon we currently have is for "Toggle Track" and the parent company named "Toggle" have another logo. It makes me think of another case we had: #5989

jorgeamadosoria:
> The icon we currently have is for "Toggle Track" and the parent company named "Toggle" have another logo. It makes me think of another case we had: #5989

apparently, there was a renaming at some point in the past. From the wiki: 

_Toggl Track (formerly Toggl) is a time tracking software operated by Toggl OÜ, headquartered in Tallinn, Estonia,[3] that offers online time tracking and reporting services through their website along with mobile and desktop applications._

So I guess an aka should suffice for this. Toggl aka Toggle track, or viceversa, whichever is less breaking.

service-paradis:
Since they use different icon, I would not use `aka` for this. I would do the same as Foursquare #5989 (similar case), but we can ask other @simple-icons/maintainers opinion on it.

jorgeamadosoria:
> Since they use different icon, I would not use `aka` for this. I would do the same as Foursquare #5989 (similar case), but we can ask other @simple-icons/maintainers opinion on it.

Oh? I don't think they use different icons. The media toolkit lists all of these as options. It names the full wordmark ones as "logos" and the compact ones as "icons", but they all represent the same product, the "toggl track"

adamrusted:
See  [here](https://user-images.githubusercontent.com/1830867/133510521-ce82767d-e371-4c44-badf-1a119d0b7e2c.png) for parent company logo on that page.

jorgeamadosoria:
> See [here](https://user-images.githubusercontent.com/1830867/133510521-ce82767d-e371-4c44-badf-1a119d0b7e2c.png) for parent company logo on that page 😅

see options in my last comment

jorgeamadosoria:
[...]

PR #6548, opened 2021-09-15
Updating Toggl icon

jorgeamadosoria:
toggl (1)

toggl (2)

issue: #6542


sachinraja:
Haven't checked the icons, but I think the title should be "Toggl Track" with a capital T in track because that's how they use it in their website.

sachinraja:
A bit of an inconsistency on the Toggl Track icon:
image

Also I'm not sure the Toggl icon should be yellow when their website seems to be very much purple.

jorgeamadosoria:
> A bit of an inconsistency on the Toggl Track icon (red is the one from this PR, black is the one from the media kit):
> image
> 
> Also I'm not sure the Toggl icon should be yellow when their website seems to be very much purple.

you are looking at the Toggle Track website, which is pink. 
Toggle is yellow. The page has a download with both icons

I will have to update the icon with the one in black. The one in this PR is the existing icon, but the one in the media kit is the official one, so it supersedes what we have. I will update that.

jorgeamadosoria:
updated the Toggl Track icon

_data/simple-icons.json
```diff
@@ -9407,7 +9407,12 @@
...
         {
             "title": "Toggl",
-            "hex": "E01B22",
+            "hex": "FFDE91",
+            "source": "https://toggl.com/media-toolkit"
+        },
+        {
+            "title": "Toggl Track",
+            "hex": "E57CD8",
             "source": "https://toggl.com/media-toolkit"
```
service-paradis, line 9416: Did not review the icons, but I think the link should be https://toggl.com/track/media-toolkit/ since the current one is redirecting there.

jorgeamadosoria: done

service-paradis: I was more talking about the `/track` part of the URL

jorgeamadosoria: done

adamrusted:
Returning to this @jorgeamadosoria - should we keep the full text mark, or a circled 't' as is their favicon for 'Toggl' - and add 'Toggl Track' as a new icon?

adamrusted (approved):
LGTM! Thanks @jorgeamadosoria 🎉

adamrusted:
Closing PR as head branch no longer exists.

PR #6552, opened 2021-09-15
add removal form

### LINKED ISSUE #6350: Add issue template/form to report icons for removal
I think it makes sense to add an issue template dedicated to reporting icons that should be removed, may two main reasons being:

- It may encourage contributors to report icons for removal, either by making them aware this is a thing or because it makes it simpler.
- It makes more sense as a link in [the disclaimer](https://github.com/simple-icons/simple-icons/blob/12ca93cf27680e0a35dd0457f8a9d1ebe5ca2660/DISCLAIMER.md#removal-of-brands).

The only argument against I can think of is that having _too many_ options isn't nice either. Personally, I think having 6 (or 5, depending on how you count) is still fine.

Any thoughts @simple-icons/maintainers ?

sachinraja:
**Issue:** closes #6350
View form to report icons for removal at https://github.com/sachinraja/simple-icons/blob/icon-removal-form/.github/ISSUE_TEMPLATE/icon_removal.yml.

I've created the label `removal request` just to open this, but feel free to tweak it.


jorgeamadosoria:
I want this template so I can link it from the website from the report flag, and also so it can be used to report an icon for removal by users or maintaners. 
Maybe we can change the template not to be for icon removal but rather for icon reporting? That's more general and better suited for integration in the website.

sachinraja:
That's already how I've set it up though. Is there anything specific that should be changed?

jorgeamadosoria:
> That's already how I've set it up though. Is there anything specific that should be changed?

the mentions to icon removal. Maybe the reported icon is reported because it's not being used properly, or the icon is obsolete, or the hex is wrong, or the icon is offensive for some reason, etc. 
Point being, 99% of the time a reported icon is reported with the intention of being removed, but I'm not sure if they are the same thing due to that unknown 1%.

sachinraja:
I think the icon update and icon removal covers both these use cases. Just "report" seems a bit too general for me. I think we should just link to both forms on the website.

.github/ISSUE_TEMPLATE/icon_removal.yml
```diff
@@ -0,0 +1,30 @@
...
+description: Report an icon for removal
+title: "Remove: "
+labels: [removal request]
+
+body:
+  - type: markdown
+    attributes:
+      value: >-
+        Before opening a new issue, make sure it isn't covered by an existing issue.
+        Please search for [issues with the `removal request` label][removal-request-issues]
```
ericcornelissen, line 11: I would expect the `removal request` label to be removed once the removal has been confirmed. From that perspective this search should probably include the `breaking change` label. What do you think about that @sachinraja?

.github/ISSUE_TEMPLATE/icon_removal.yml
```diff
@@ -0,0 +1,30 @@
...
+  - type: input
+    attributes:
+      label: Brand Name
+      placeholder: "Example: Simple Icons"
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Reason for Removal
```
ericcornelissen, line 27: I think we should add a description here that asks the reporter to include a link as well if there's one, which should usually be the case.

E.g. we would probably want the reports to look something like "look, the website says the company was closed in XXXX" or "this blogpost announced the brand has been discontinued"; rather than "the website is offline" or "there was a blog post".

ericcornelissen (commented):
Looking good already @sachinraja, just have two suggestions for improvements.

---

@jorgeamadosoria, "the icon is obsolete" and (presumably) "the icon is offensive for some reason" would both fall under _Icon removal_ (the form being added here) and "not being used properly" and "the hex is wrong" would both fall under [_Icon update_](https://github.com/simple-icons/simple-icons/issues/new?assignees=&labels=icon+outdated&template=icon_update.yml&title=Update%3A+). So in that sense everything is covered - there's not much more we can do with an existing icon than changing it and removing it.

The advantage of having separate templates is that each requires different fields and different labels and that can be achieved without any intervention from a maintainer. The advantage of mixing these two in "Icon report" is _just_ that it's easier to reference from the website, but after that it's more work for us and it also makes it more difficult for people reporting directly through the GitHub UI. So, I think we either need two links (update and remove) or just link to [the choose an issue page](https://github.com/simple-icons/simple-icons/issues/new/choose).

jorgeamadosoria:
> I think the icon update and icon removal forms cover these cases. Just "report" seems a bit too general for me. I think we should just link to both forms on the website.

linking to both is problematic, from an UI standpoint. unless we do as @ericcornelissen says and link to the page to choose the issue template, which would be fine.

.github/ISSUE_TEMPLATE/icon_removal.yml
sachinraja: I agree

jorgeamadosoria (approved):

.github/ISSUE_TEMPLATE/icon_removal.yml
ericcornelissen: Apologies for the confusion, I wasn't suggesting to [add the breaking change label](https://github.com/simple-icons/simple-icons/pull/6552/commits/360ef0b90f40d5b8f9291a898bc68c91b2b42438) to this template, but rather that this link should include `breaking change` labels in the search result.

I personally wouldn't even add `breaking change` by default. I'd add it if the removal request has been verified, rather than having to remove it when the removal request turns out to be invalid. But it can go either way so whatever you think makes the most sense works for me.

.github/ISSUE_TEMPLATE/icon_removal.yml
```diff
@@ -0,0 +1,36 @@
...
+      description: |
+        Include all relevant links, for example:
+        - The website announcing the company was closed
+        - A blogpost announcing the brand has been discontinued
```
ericcornelissen, line 31: I feel like providing specific examples for how a brand might have announced it's discontinued here isn't very productive. Also, a discontinued brand is not the only reason for a removal request (see e.g. [the disclaimer](https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md#removal-of-brands)).

```suggestion
        - If the brand no longer exists, include a link supporting that claim.
        - If the brand should be removed for another reason, explain why.
```

ericcornelissen (changes requested):
Can you update the `submit an issue` link for icon removals in the disclaimer to point to this form as part of this PR? (related #6560)

.github/ISSUE_TEMPLATE/icon_removal.yml
sachinraja: That makes more sense to me. It's more clear to people that it's just a request and no changes are being made yet.

jorgeamadosoria:
what's missing to move this issue forward?

ericcornelissen:
I think this is still missing:

> Can you update the submit an issue link for icon removals in the disclaimer to point to this form as part of this PR?

sachinraja:
This PR now closes #6560

ericcornelissen (approved):

Issue #6541, opened 2021-09-15
Update: Tailwind CSS icon


jorgeamadosoria:
### Brand Name

Tailwind CSS

### Official Resources for Icon and Color

https://tailwindcss.com/

### Additional Comments

change hex from `#38B2AC` to `#06B6D4`

### Contributing

Yes

sachinraja:
Should the source also be updated to https://tailwindcss.com/brand? Not really sure how the review process works for existing icons.

jorgeamadosoria:
> https://tailwindcss.com/brand

I think you are the review process, @sachinraja.
And yes, the source should be that one, thanks for noticing.

Issue #6530, opened 2021-09-15
Update: TUI


jorgeamadosoria:
### Brand Name

TUI

### Official Resources for Icon and Color

https://www.design.tui/brand/logos/

### Additional Comments

change source from https://www.tuiholidays.ie/ to https://www.design.tui/brand/logos/
also change hex as depicted in the new source

### Contributing

Yes

PR #6553, opened 2021-09-16
Change hexes for TUI and TailwindCSS

jorgeamadosoria:
tui

tailwindcss

issues: #6541 #6530


sachinraja (approved):

sachinraja:
I think we can also add the brand url as guidelines

jorgeamadosoria:
> I think we can also add the brand url as guidelines

done.

sachinraja (approved):

sachinraja:
I would remove the trailing slash on the TUI url just to match the other ones.

jorgeamadosoria:
> I would remove the trailing slash on the TUI url just to match the other ones.

the TUI urls without the trailing slash redirect to the ones with the slash. 
Besides, there is no format for them: TailwindCSS does not use a trailing slash, but TinyLetter does not work without it 🤷

sachinraja:
> Besides, there is no format for them

Ok, I think this can be merged now then.

PR #6580, opened 2021-09-22
feat: add types

sachinraja:
Add TypeScript types to this repository. See discussion in #6189.

Once this is merged we will need to deprecate `@types/simple-icons`. Instructions are [here](https://github.com/definitelytyped/definitelytyped#removing-a-package).


scripts/build/package.js
```diff
@@ -91,10 +93,17 @@ data.icons.forEach(icon => {
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
   minifyAndWrite(jsFilepath, `module.exports=${iconObject};`);
 
+  const dtsFilepath = path.resolve(iconsDir, `${filename}.d.ts`);
+  fs.writeFileSync(dtsFilepath, `
+  import { SimpleIcon } from "../types";
+  declare const icon: SimpleIcon
+  export default icon;`);
+
```
sachinraja, line 101: It is necessary to produce an declaration file for each icon, I could not get TS to work with the `declare module` glob in `types/index.d.ts`. The package size increased by 7 MB (30 to 37). I see this a minor problem as we plan to remove support for the `icons/*` files anyway.

package.json
```diff
@@ -9,14 +9,16 @@
...
   "exports": {
     ".": "./index.js",
     "./icons": {
       "module": "./icons.mjs",
       "import": "./icons.mjs",
       "require": "./icons.js",
       "default": "./icons.js"
-    }
+    },
+    "./icons/*": "./icons/*.js"
```
sachinraja, line 21: I missed this in the last PR, this is necessary for Node >= 12 to import those files.

PR #5198, opened 2021-03-07
Auto assign to project workflow

service-paradis:
### Take 3 (see #4863 and https://github.com/simple-icons/simple-icons/commit/7181ea0b31adce9364fc6e131471be8e17b0781d)

Add new PR to the Organisation Priorisation Project:

- New PRs to Unprioritised column
- PRs labeled as icon outdated to Priority 1

Now using `pull_request_target`. 🤞

PR #6589, opened 2021-09-23
Draft: Auto assign to project workflow - Take 4 😅

service-paradis:
### Take 4 (see #4863 and #5198)

Add new PR to the Organisation Priorisation Project:

- New PRs to `Unprioritised` column
- PRs labeled as icon outdated to `Priority 1`

🤞

#### Notes

Be cautious when merging this one. This needs to be tested soon after and we should revert if something is wrong again...

@ericcornelissen Is there any `AUTO_ASSIGN_WORKFLOW_TOKEN` secrets still configured?


sachinraja:
I believe you can just use the [GITHUB_TOKEN secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret), which is automatically added to each workflow.

ericcornelissen:
Based on https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token, it seems @sachinraja is right. Can you update this PR accordingly @service-paradis ?

service-paradis:
> I believe you can just use the [GITHUB_TOKEN secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret), which is automatically added to each workflow.

I think you're right, great! Updated!

sachinraja (approved):

sachinraja:
I'm actually unsure as to whether it will work as the project is on the organization level.

service-paradis:
> I'm actually unsure as to whether it will work as the project is on the organization level.

Hum... Maybe that is why we used another token previously...

service-paradis:
According to https://github.com/srggrs/assign-one-project-github-action/issues/64#issuecomment-792225751, we need a PAT with org permission since the project is not in the repo itself.

ericcornelissen:
> We need a PAT with org permission since the project is not in the repo itself.

That makes sense, guess 945cbdef58d35790ae89ca527483f35a69036fda should be reverted then. @mondeja, I believe you're an admin as well - can you add the token?

mondeja:
Sure, I've added the PAT `AUTO_ASSIGN_WORKFLOW_TOKEN` as secret to this repository with `write:org` and `read:org` permissions:

image

mondeja (approved):

sachinraja (approved):

mondeja:
I'm going to merge this removing the file after if doesn't work.

mondeja:
- It works adding pull requests to "Unprioritised".
- The [labeler](https://github.com/ericcornelissen/labeler) action takes a bit long than the assigner, so the `icon outdated` label is created after of the assignation. I suggest to use [`needs`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idneeds) for the assigner. The counterpart is that both configurations must be in the same file.
- It doesn't work manually re-adding the `icon outdated` label, you can see [the failure here](https://github.com/simple-icons/simple-icons/pull/6606/checks?check_run_id=3721952929).

service-paradis:
Thank you @mondeja for testing this!

> It works adding pull requests to "Unprioritised".

Good news (kind of). At least, it prevents forgetting PRs and saves one manual action

> The labeler action takes a bit long than the assigner, so the icon outdated label is created after of the assignation. I suggest to use needs for the assigner. The counterpart is that both configurations must be in the same file.
> It doesn't work manually re-adding the icon outdated label, you can see the failure here.

Oh, it make sense to use `needs` in this case. Project and labels are quite related, so it might be correct to use the same workflow file for both.

Since the current plugin only adds card to a project and dont move them depending on conditions, we might have a problem anyway with the current configuration. Right now, the first condition to put card in "Unprioritised" column will always be fired.

Right now, this bring a slight improvement since it adds every PR in the project without any manual action.

mondeja:
> Since the current plugin only adds card to a project and dont move them depending on conditions, we might have a problem anyway with the current configuration. Right now, the first condition to put card in "Unprioritised" column will always be fired.

I've opened #6607 to fix this problem

service-paradis:
> > Since the current plugin only adds card to a project and dont move them depending on conditions, we might have a problem anyway with the current configuration. Right now, the first condition to put card in "Unprioritised" column will always be fired.
> 
> I've opened #6607 to fix this problem

That was fast ⚡

Issue #2170, opened 2019-12-19
FL Studio (DAW) and Image-Line (company behind FL Studio)


foo-dogsquared:
**Name:** FL Studio 
**Website:** https://www.image-line.com/flstudio/
**Official resources for icon and color:** 
* https://www.image-line.com/press/presskitblog.php?category=High%20res%20images - [The outlined logo](https://www.image-line.com/innovaeditor/assets/FLStudio20Logo_Outline.png) has the logo in monochrome. 
* There is also a (better, IMO) monochromatic version that [appears as a loader](https://www.image-line.com/flstudio/img/loader.gif). 
* Suggested brand color is `#ff7629` extracted both from the loader GIF and the orange color from their stylesheet. 
  

**Name:** Image-Line
**Website:** https://www.image-line.com
**Official resources for icon and color:**
* https://www.image-line.com/press/presskitblog.php?category=High%20res%20images

Alexa ranking: [~6.4k](https://www.alexa.com/siteinfo/image-line.com).

mondeja:
Image Line logo images and brand guidelines are in https://www.image-line.com/high-resolution-images/

demberto:
Yes please

adamrusted:
As mentioned [here](https://github.com/simple-icons/simple-icons/pull/6593#issuecomment-1627723401) - Image Line don't want their images transforming in any way. Unless we can get an official SVG source from them, we will probably have to reject this.

adamrusted:
Closing this one, based on https://github.com/simple-icons/simple-icons/pull/6593#issuecomment-1627723401.
If anyone reaches out to Image-Line and they're happy for us to proceed, feel free to re-open.

PR #6593, opened 2021-09-24
Add FL Studio icon

jorgeamadosoria:
flstudio

**Issue:** #2170


demberto:
Wow, I need this like right now

adamrusted:
Based on the comments [here](https://www.image-line.com/high-resolution-images/) stating:

> You cannot alter, transform or build upon these logos and artwork and you must respect Image Line's [moral rights](https://en.wikipedia.org/wiki/Moral_rights_(copyright_law)).

I would be inclined to decline this one, both on the grounds that they explicitly ask you not to transform the artwork, but also agreeing with @demberto that the bottom half is barely visible. What do other @simple-icons/icon-review folk think?

jorgeamadosoria:

Issue #6594, opened 2021-09-24
Request: SolidJS


tuhanayim:
### Brand Name

SolidJS

### Website

https://www.solidjs.com

### Popularity Metric

164,175 - https://www.alexa.com/siteinfo/solidjs.com

### Official Resources for Icon and Color

https://www.solidjs.com/media

### Contributing

No

PR #6595, opened 2021-09-24
adding SolidJS icon

jorgeamadosoria:
solidjs

**Issue:** #6594 

EDIT:  Added the gaps
solidjs


sachinraja:
Not sure about this. It may need extra detail to preserve the shape.

sachinraja (dismissed):
Not sure about this. It may need extra detail to preserve the shape.

service-paradis:
> Not sure about this. It may need extra detail to preserve the shape.

I agree! Maybe we can try adding 0.5 gaps to keep the shape

jorgeamadosoria:
> > Not sure about this. It may need extra detail to preserve the shape.
> 
> I agree! Maybe we can try adding 0.5 gaps to keep the shape

I agree with both of you. I just don't know if we should add the gaps around each of the paths or only in the middle between the top and bottom wedges.

jorgeamadosoria:
I've added gaps as requested (look at the beginning of the PR).

sachinraja (commented):
This looks good me, but I haven't reviewed anything with gaps before, so I'll leave that to someone else.

sachinraja (approved):

PR #6596, opened 2021-09-25
update Google Drive icon

jorgeamadosoria:
googledrive

**Issue:** #4205


service-paradis:
Which source did you use for this?
The new Google Drive icon looks like this:
image

jorgeamadosoria:
> Which source did you use for this? The new Google Drive icon looks like this: image

they have a dedicated monochrome icon: 
https://developers.google.com/drive/api/v3/branding

dirien (changes requested):
hi @jorgeamadosoria,

can you have a look again? I cant get a perfect fit on the left icon part
image

jorgeamadosoria:
@dirien 
image

I get 100% match. 
I took the PNG and convert it to PNG to use as base and comparison, so that comparison is SVG without optimization vs SVG optimized.

What are you comparing against?

PR #6599, opened 2021-09-25
adding Google Marketing Platform, updating Google Keep, Google Meet a…

jorgeamadosoria:
googlekeep

googlemeet (1)

googlemessages

googlemarketingplatform (1)

**Issue:** #4205


service-paradis:
@jorgeamadosoria No need to request review on PRs. There are more than 200 to review and they all need to be reviewed.

jorgeamadosoria:
> @jorgeamadosoria No need to request review on PRs. There are more than 200 to review and they all need to be reviewed.

this is a pinned icon, and apparently time sensitive (since it seems to be something that is done regularly / has been done more than once). I would think this one is priority 0, thus I requested reviews.

adamrusted:
The pinned issues are just there so we can monitor much larger projects than one or two updates. Doesn't make them priority 0, which I've always used for "non-icon updates that need attention". This PR contains both priority 1 and priority 3 icons, and but I guess would be placed in priority 1 because of the update - but should be reviewed (roughly) in order along with the other PRs in that priority band.

service-paradis:
> Hard disagree, but, as you wish.

There are rules right now for priorisation: https://github.com/orgs/simple-icons/teams/maintainers/discussions/9

If you disagree with those rules, you can open an issue to discuss about this, but this PR is not the place to do so.

sachinraja (approved):
All changes look good, great job!

PR #6607, opened 2021-09-27
Fix unsync between 'add-labels' and 'auto-assign-to-project' actions

mondeja:
- This PR makes the `auto-assign-to-project` workflow depends on `add-labels` workflow.
- Also, I've disabled the execution at `labeled` event because the action doesn't support to move between columns, so when is manually labeled and the PR is in the "Unprioritised" column already, it fails silently with a message "Project already has the associated issue" (see [here](https://github.com/simple-icons/simple-icons/pull/6606/checks?check_run_id=3722161416)).


service-paradis (approved):
LGTM 💯 Thanks for your help on this!

PR #6610, opened 2021-09-27
Test (DON'T MERGE)

mondeja:


service-paradis:
Hello @mondeja !

I think we might need to check if there are no labels before negating on labels name. Maybe something like:
```yml
if: join(github.event.issue.labels) == '' || !contains(github.event.pull_request.labels.*.name, 'icon outdated')
```

PR #6612, opened 2021-09-27
Get labels at runtime in `assign-to-project` workflow

mondeja:
The labels in `github.event.pull_request.labels.*.name` are defined at PR open time, so the `contains(github.event.pull_request.labels.*.name, 'icon outdated') == false` conditional is always executed, even if the previous step has added a new `icon outdated` label.

This pull introduces a step to get the updated labels from the PR with cURL, using them in conditionals. [See it working here](https://github.com/mondeja/simple-icons/pull/9/checks?check_run_id=3723663912#step:3:10).

### DIFF
--- a/.github/workflows/add-labels-priority.yml
+++ b/.github/workflows/add-labels-priority.yml
@@ -18,16 +18,22 @@ jobs:
     env:
       MY_GITHUB_TOKEN: ${{ secrets.AUTO_ASSIGN_WORKFLOW_TOKEN }}
     steps:
+    - id: get-labels
+      name: Get labels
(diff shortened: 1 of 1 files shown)


service-paradis (approved):
Great 🎉 
The default Github behavior is weird to me, but this looks to be a good workaround! LGTM 💯 
Thanks again @mondeja!

mondeja:
[It seems that is working](https://github.com/simple-icons/simple-icons/pull/6613) :tada:

service-paradis:
Oh, yeah, good job @mondeja!

PR #6615, opened 2021-09-28
Auto assign pull requests to 'Priority 2' column

mondeja:
This is the next step for automatic priorization of pull requests. Implements the Priority 2 [group of rules for priorization](https://github.com/orgs/simple-icons/teams/maintainers/discussions/9) (si maintainers link only):

> ## Priority 2
>
> This is for all "[new icon](https://github.com/simple-icons/simple-icons/pulls?q=is%3Aopen+is%3Apr+label%3A%22new+icon%22)" PRs submitted by outside contributors that fulfil a request. If a new member joins the team then any PRs they submitted prior to accepting the invitation should remain at this priority and not be re-priortised.

The implementation follows that. The `if` conditional for `Priority 2` means "if the pull request has a 'new icon' label, has a linked issue and the user which has opened the pull request is not a member of simple-icons organization":

```
contains(steps.get-labels.outputs.labels, 'new icon') &&
join(steps.get-linked-issues.outputs.issues) != '' &&
contains(steps.get-si-members.outputs.members, github.event.actor.login) == false
```

The "Unprioritised" conditional has been changed by "if priority conditionals has not been executed", so it works like an else for previous rules (see `conclusion` in [`steps` context](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context)):

```
steps.priority-1.conclusion == 'skipped' &&
steps.priority-2.conclusion == 'skipped'
```

To get the linked issues for pull requests I'm using [mondeja/pr-linked-issues-action](https://github.com/mondeja/pr-linked-issues-action). ~~This action works parsing the HTML of the pull request page because I've not found a way to get them using Github APIs. I'm testing this every month in [a CI workflow](https://github.com/mondeja/pr-linked-issues-action/blob/fa2264e6d9e447315ab5ef608e754fd3817971a5/.github/workflows/ci.yml) without mocking, so I think it will not be difficult to debug if an error starts to occur.~~

I've tested it in [mondeja/simple-icons#10](https://github.com/mondeja/simple-icons/pull/10) all the 3 possible strategies and works as expected.

### DIFF
--- a/.github/workflows/add-labels-priority.yml
+++ b/.github/workflows/add-labels-priority.yml
@@ -33,7 +33,9 @@ jobs:
 
       - id: get-linked-issues
         name: Get linked issue numbers
-        uses: mondeja/pr-linked-issues-action@v1
+        uses: mondeja/pr-linked-issues-action@v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - id: priority-1
         name: Assign `icon outdated` pull requests to "Priority 1"
--- a/.github/workflows/add-labels-priority.yml
+++ b/.github/workflows/add-labels-priority.yml
@@ -18,22 +18,49 @@ jobs:
     env:
       MY_GITHUB_TOKEN: ${{ secrets.AUTO_ASSIGN_WORKFLOW_TOKEN }}
     steps:
-    - id: get-labels
-      name: Get labels
-      run: |
-        labels="$(curl --retry 5 -s https://api.github.com/repos/simple-icons/simple-icons/pulls/${{ github.event.pull_request.number }} | jq '.labels[].name' | tr '\n' ',' | sed 's/"//g' | sed 's/,$//')"
-        echo "::set-output name=labels::$labels"
+    steps:
+      - id: get-labels
+        name: Get labels
+        run: |
+          labels="$(curl --retry 5 -s https://api.github.com/repos/simple-icons/simple-icons/pulls/${{ github.event.pull_request.number }} | jq '.labels[].name' | tr '\n' ',' | sed -e 's/"//g' -e 's/,$//')"
+          echo "::set-output name=labels::$labels"
 
-    - name: Assign pull requests to "Unprioritised"
-      uses: srggrs/assign-one-project-github-action@1.2.1
-      if: contains(steps.get-labels.outputs.labels, 'icon outdated') == false
-      with:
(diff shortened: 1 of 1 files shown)


service-paradis:
Wow, good job @mondeja! It is great to see this coming!!

Just a note about linked issues, it looks like we can get them now using GraphQL. Something like this, for example:

```graphql
query { 
  repository(name: "simple-icons", owner: "simple-icons") { 
    pullRequest(number: 5978) {
      closingIssuesReferences(first: 1) {
        nodes {
          number
        }
      }
    }
  }
}
```

Will give the following result:

```json
{
  "data": {
    "repository": {
      "pullRequest": {
        "closingIssuesReferences": {
          "nodes": [
            {
              "number": 6072
            }
          ]
        }
      }
    }
  }
}
```

mondeja:
Wow, that s great! Let me update the action...

mondeja:
Now it is using the GraphQL API, thank you @service-paradis!

service-paradis (approved):
Oh yeah, I love it, thank you @mondeja! 💯

service-paradis:
There is a little problem with this auto-assignment. Linked issues are not often made directly by the contributor. #6624 is an example, but there are many.

Maybe, we do not have to check for a linked issue. If a contributor (not an org member) creates a PR, even if there are no issues for it, we could consider it as a Priority 2?

Or we may change the PR template from `**Issues: **` to `Close: `

mondeja:
> Or we may change the PR template from `**Issues: **` to `Close: `

I vote to give this solution a try as we don't have so much external contributors working in pull requests without linked issues, like `tracking` label ones. I remember that this was proposed a year ago, but we were working in `tracking` issues so it was not convenient to make the change then.

service-paradis:
There some contributor that creates PR directly without creating an issue first (ex. #6518). So, maybe, we'll cover more cases by ignoring linked issues?

mondeja:
But this would break the rule, not? Are not PRs with an opened issue more prioritary? I don't really know because I'm not currently contributing nor reviewing icons.

service-paradis:
I'm not sure... If someone open an issue and create a PR right after, it would be priority 2... I think I would consider the PR by an outsider with the same priority. We always can reprioritize the manually if needed, but most of the times, the priority would be correct, I think.

mondeja:
Feel free to open a PR if you feel that is more useful in that way. You only need to remove lines 33 to 38 and 54 from the workflow file.

PR #6623, opened 2021-09-30
Updated number of icons in the readme

DenverCoder1:
The number in the readme is "Over 1500."

It is more accurate to say "Over 2000" since Simple-icons has 2020+ icons now! :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 </a>
 <h3 align="center">Simple Icons</h3>
 <p align="center">
-Over 1500 Free SVG icons for popular brands. See them all on one page at <a href="https://simpleicons.org">SimpleIcons.org</a>. Contributions, corrections & requests can be made on GitHub.</p>
+Over 2000 Free SVG icons for popular brands. See them all on one page at <a href="https://simpleicons.org">SimpleIcons.org</a>. Contributions, corrections & requests can be made on GitHub.</p>
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,7 +138,7 @@ echo file_get_contents('path/to/package/icons/simpleicons.svg');
 ?>
 ```
 
-## Third Party Extensions
+## Third-Party Extensions
```
adamrusted, line 141: I don't think this hyphen needs adding, though there are uses of both variations across the internet. Thoughts, @simple-icons/maintainers?

adamrusted:
Wow, that 2k came around quick! 🤯

README.md
DenverCoder1: I believe it's meant to be hyphenated when used as an adjective, but can be with a space when used as a noun.

So grammatically speaking, this usage would be better with a hyphen. If you'd rather I not change it, I can put it back though.

adamrusted: Figured it'd be something like that, though the use cases of the phrase I find seem to mix and match between the two, hence why I've pinged the other maintainers to gain their thoughts on it. If there's no objections by the end of the day, I'll look to merge this one in.

mondeja (approved):

README.md
mondeja: I can see a lot of references about hyphenation of compound words when used as adjetves. Altough I'm not a native english speaker, this change seems good to me.

sachinraja (approved):

README.md
adamrusted: Marking this as resolved, and merging this PR due to multiple other maintainers approving. Thanks for your work @DenverCoder1

Issue #6549, opened 2021-09-15
Request: Mautic


dmroch:
### Brand Name

Mautic

### Website

https://www.mautic.org/

### Popularity Metric

The Alexa Rank is 31,763. See https://www.alexa.com/siteinfo/mautic.org

### Official Resources for Icon and Color

svg: https://www.mautic.org/sites/default/files/2020-03/Mautic_Logo_Monochrome_LB.svg
color: '#4e5e9e'

### Contributing

No

PR #6627, opened 2021-10-01
Adding Mautic icon

jorgeamadosoria:
mautic

**Issue:** #6549


_data/simple-icons.json
```diff
@@ -5575,6 +5575,11 @@
             "hex": "261C29",
             "source": "http://mttr.net"
         },
+        {
+            "title": "Mautic",
+            "hex": "FDB933",
```
adamrusted, line 5580:
```suggestion
            "hex": "4E5E9E",
```

They seem to be using the purple more prominently. Not knowing the brand, I would assume that is the colour we should be using.

adamrusted (commented):
Icon itself is perfect, thanks @jorgeamadosoria 🎉
@simple-icons/icon-review - what do we think on the brand colour? If we're happy with the yellow, this can be merged as-is. Otherwise, the change will need to be made.

_data/simple-icons.json
service-paradis: The first color mentioned in https://www.mautic.org/about/logos-and-graphics is blue (purplish). I agree this should be the color.

@adamrusted Feel free to commit your suggestion to merge this one!

PR #6629, opened 2021-10-01
feat: camelcase names

krishguptadev:
**Issue:**
**Alexa rank:**
  

### Checklist

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -1,12 +1,12 @@
 {
     "icons": [
         {
-            "title": ".NET",
+            "title": "net",
             "hex": "512BD4",
             "source": "https://github.com/dotnet/brand/"
         },
         {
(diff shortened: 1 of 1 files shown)


adamrusted:
Thanks for the contribution, @krishguptadev - though I don't see the need for this PR. The titles on all icons should always be in the format the name of the brand is used (for the most part, not camelcase). I believe you're trying to implement something more along the lines of the 'slug' for icons, which is a much bigger task than just changing the JSON - and would probably require a new issue and severe @simple-icons/maintainers oversight on the work undertaken. Happy to have that discussion if you think there's a legitimate use-case for camelcase file names, though as hinted at above - this would require significant re-working to our linting and display processes, not to mention annoying users that have just gotten used to the current scheme! I'll close this for now, but feel free to discuss this further in an issue.

PR #6640, opened 2021-10-02
Force the usage of implicit closepath commands in paths

### LINKED ISSUE #5070: Unify implicit/explicit closepath commands
Currently in simple-icons, there are 107 icons that ends without a [closepath](https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand) command (`z` or `Z`). As you can see in [this comment in StackOverflow](https://stackoverflow.com/a/10213649/9167585), as [SVG specification points](https://www.w3.org/TR/SVG/painting.html#SpecifyingFillPaint), if a path ends without a closepath command, will be treated as if a closepath command is added filling its geometry.

So, in order to unify all icon paths, we can:

- Force addition of explicit `z` or `Z` command at the end of the paths (explicit notation).
- Force removal of `z` or `Z` command at the end (smaller file size).

And this could be done in two ways:

- Add `[Zz]$` regex check to tests or linting process (the user must edit it manually).
- Configure some way of do this automatically, so as not to worry the user about these little optimizations, for example, doing one commit before release. I'm thinking also in the lint rule `negative-zeros`, that could be fixed in this way. There is a precedent of this kind of commit before release in #5002.

mondeja:
Don't allow to end SVG paths with `Z` or `z` characters in `path-format` SVGLint rule. I've removed them from SVG icons paths.

[According to the SVG specification](https://www.w3.org/TR/SVG/painting.html#SpecifyingFillPaint), paths which don't end in a closepath command (`z/Z`) are treated as if they end in one of these characters.

Closes #5070


mondeja:
I'm no sure if this could potentially break stroke paths, so better let left as is.

ericcornelissen:
> I'm no sure if this could potentially break stroke paths, so better let left as is.

Can you elaborate? If the spec says it's fine it _should_ be fine. If we don't want to risk it, should we instead force the trailing `Z`/`z` (an alternative you also suggested in #5070)?

mondeja:
The specification [says here](https://www.w3.org/TR/SVG/paths.html#PathDataClosePathCommand): 

> A closed subpath must be closed with a "closepath" command, this "joins" the first and last path segments. Any other path is an open subpath.

> A closed subpath differs in behavior from an open subpath whose final coordinate is the initial point of the subpath. The first and last path segments of an open subpath will not be joined, even when the final coordinate of the last path segment is the initial point of the subpath. This will result in the first and last path segments being capped using the current value of stroke-linecap rather than joined using the current value of stroke-linejoin.

So to select between one solution could break the geometry.

ericcornelissen:
I see now, thanks for clarifying!

Issue #3601, opened 2020-09-26
Request: Google Campaign Manager Icon


scxttd:
**Name:** Google Campaign Manager
**Website:** https://www.google.com/dfa/trafficking
**Alexa rank:** #1 as it is part of the Google marketing platform
  

**Official resources for icon and color:** https://www.gstatic.com/analytics-suite/header/suite/v2/ic_campaign_manager.svg

PR #6648, opened 2021-10-03
adding Google Campaign Manager

jorgeamadosoria:
googlecampaignmanager

**Issue:** #3601

Had to use precision 5 because the icon was change sizes after running npm run svgo every time. Precision 5 didn't had that issue. If someone makes it work with precision 3, then by all means.
Not happy with the white gaps, but I did what I can.


svgo.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
           // customize options
           convertPathData: {
             // 3 decimals of precision in floating point numbers
-            floatPrecision: 3,
+            floatPrecision: 5,
```
adamrusted, line 11:
```suggestion
            floatPrecision: 3,
```

_data/simple-icons.json
```diff
@@ -3648,6 +3648,11 @@
             "hex": "4285F4",
             "source": "https://fonts.gstatic.com/s/i/productlogos/calendar_2020q4/v8/192px.svg"
         },
+        {
+            "title": "Google Campaign Manager",
+            "hex": "5BB974",
+            "source": "https://www.gstatic.com/analytics-suite/header/suite/v2/ic_campaign_manager.svg"
```
adamrusted, line 3654: I would suggest we need to find a better source than a static SVG. Where is this SVG visible?

adamrusted (changes requested):

jorgeamadosoria:

Issue #6665, opened 2021-10-07
Request: Ask Different


alagos:
### Brand Name

Ask Different

### Website

https://apple.stackexchange.com/

### Popularity Metric

# 135
https://www.alexa.com/siteinfo/stackexchange.com

### Official Resources for Icon and Color

svg: https://cdn.sstatic.net/Sites/apple/Img/logo.svg
color: #74A5D2 from the logo

### Contributing

Yes

sachinraja:
@simple-icons/maintainers has there been a decision made on https://stackexchange.com websites?

PR #6666, opened 2021-10-07
Add Ask Different

alagos:
askdifferent

**Issue:** #6665 
**Alexa rank:** 135

### Checklist

### Description
I picked the hex value from the logo itself, as it was the predominant color in the icon.
Also, I've added the Stack Overflow guidelines, which should apply for the rest of Stack Exchange sites.


sachinraja (changes requested):
I would invert this svg so only the central logo can be seen and colored. Currently, the central logo is cut out and the background can be colored.
image
So only this part remains.

adamrusted:
Marking as pending, awaiting outcome of #9294. That may determine this PR is out of scope.

LitoMore:
I think this symbol belongs to Apple. It's the command key symbol.

LitoMore:
Closing according to #9294.

Issue #6651, opened 2021-10-04
Request: Ruby Sinatra


LeslieCayas:
### Brand Name

Ruby Sinatra

### Website

http://sinatrarb.com/

### Popularity Metric

The Alexa Rank is #638,264. See https://www.alexa.com/siteinfo/sinatrarb.com

### Official Resources for Icon and Color

http://sinatrarb.com/
from website header

### Contributing

No

adamrusted:
Thanks for submitting this request, @LeslieCayas. Looking at the Alexa rank this would currently be out of scope, and given the sudden rise in it's popularity between 6-13 Sept, would want to keep an eye on how stable the popularity is anyway. Have you got another metric we could use to determine whether or not to include this brand? (GitHub stars, etc)

service-paradis:
With [~11k](https://github.com/sinatra/sinatra) Github stars, I think we're ok with this one!

service-paradis:
If @adsingh14 did not start working on it, I think you can create a PR for this.

adsingh14:
@service-paradis Done!

PR #6749, opened 2021-10-18
Ruby Sinatra icon from a scratch

adsingh14:
**Issue:** #6651
**Alexa rank:** 612,196
  

### Checklist

### Description
- This one is a bit challenging. I created this icon from a scratch.
- I used 'sinatra-classic-hd' file as reference

rubysinatra


dirien (commented):
hi @adsingh14,

thanks for the PR. IMHO it is not a close fit with the png of the site.

image

I would like to hear the opinion of @jorgeamadosoria or @sachinraja

sachinraja:
It does seem to be missing the "bump" in the hat, but I think it looks alright.

jorgeamadosoria:
> It does seem to be missing the "bump" in the hat (not sure how that would be represented), but I think it looks alright.

I honestly don't know how to represent that neither, but I don't think we would be losing a lot. The lack of 3D in monochrome icons makes that kind of things hard to represent.
Let's focus on seeing if the shape of the hat works

jorgeamadosoria:
> It does seem to be missing the "bump" in the hat (not sure how that would be represented), but I think it looks alright.

Before I start working on this in earnest, what do you guys think of this look @sachinraja @dirien :

image

sachinraja:
I don't think changing the color would be the way to go. The bump is too subtle for that, so I'd be ok with accepting the flat version.

jorgeamadosoria:
> I don't think changing the color would be the way to go. The indent is too subtle for that, so I'd be ok with accepting the flat version.

ok, how about this one

image

sachinraja:
I don't think the top should be a different color. It just doesn't look right.

jorgeamadosoria:
> I don't think the top should be a different color. It just doesn't look right.

flat, then. I don't like the one taken from PNG, the lines are wavy because of the PNG problem.

jorgeamadosoria:
rubysinatra

Updated to correct wavy lines due to PNG low res.

adsingh14:
I used `.eps` instead of `.png` file to generate the icon. I removed bumps to keep it minimal  and flat.

jorgeamadosoria:
> I used `.eps` instead of `.png` file to generate the icon. I removed bumps to keep it minimal and flat.

Did you flatten the lines of the hat? t
he preview shows wavy lines, and the icon in the PR has too many nodes in the paths that produce these bumps.

adsingh14:
I used inkscape to merge the groups into three sections. Then, I removed black group and minimized using SVGO.

jorgeamadosoria:
> I used inkscape to merge the groups into three sections. Then, I removed black group and minimized using SVGO.

you did fine. THe only missing thing is to smooth the lines. Please compare the previews and you'll see what I mean.

adsingh14:
>THe only missing thing is to smooth the lines 

@jorgeamadosoria I tried to keep it original.

jorgeamadosoria:
> > THe only missing thing is to smooth the lines
> 
> @jorgeamadosoria I tried to keep it original.

in cases like this when you are working with a faulty source (usually a low resolution PNG) it's fine to make common sense adjustments.
For example, we can assume that the top of the hat is straight. Same with the black band top and bottom.
The conversion from PNG to SVG when the PNG is low res somethings does that. 
In those cases the result of adjusting these conversion errors will make the icon closer to the original's intent. 
It happens more often that you would think. It's amazing the amount of popular brands that only offer PNGs or even JPGs as their official resources!

adsingh14:
> in cases like this when you are working with a faulty source (usually a low resolution PNG) it's fine to make common sense adjustments. For example, we can assume that the top of the hat is straight. Same with the black band top and bottom. The conversion from PNG to SVG when the PNG is low res somethings does that. In those cases the result of adjusting these conversion errors will make the icon closer to the original's intent. It happens more often that you would think. It's amazing the amount of popular brands that only offer PNGs or even JPGs as their official resources!

Pardon! I didn't use PNG but vector format `.eps` file and converted it into `.svg` via Adobe Illustrator for my recent submission. I created a GIF using original PNG and SVG (eps) and SVG generated icon files .

example

jorgeamadosoria:
if the source is an eps, then I guess the wavy curves are part of the original icon, then.
I think @dirien can merge this one for you.

dirien (approved):
ok @jorgeamadosoria

PR #6758, opened 2021-10-20
Setup prettier

ericcornelissen:
Adding Prettier and running it with the same configuration as other simple-icons projects.

~~At the time of opening this I did not yet add this as a pre-commit hook - like in other simple-icons projects - mainly because this project doesn't have any git hooks yet. If the @simple-icons/maintainers all agree we should add that here as well, then I'll add Husky and a similar git hook setup to other simple-icons projects as well.~~ EDIT: Added a pre-commit hook following [this comment](https://github.com/simple-icons/simple-icons/pull/6758#issuecomment-947663221).

Notably, I opted to ignore the `_data/simple-icons.json` file. While it could benefit from being automatically formatted, I'm afraid it will interfere too much with existing Pull Requests.


sachinraja:
I'm in favor of adding a pre-commit hook

PR #6769, opened 2021-10-24
adding comsol icon

jorgeamadosoria:
comsol

Had to create the full logo with wordmark because their guide seems to imply that using only the logo part is not allowed (see the pdf in the source)


dirien (changes requested):
Hi @jorgeamadosoria,

Something is not right at the rigth leg of the M.

image

jorgeamadosoria:
> Hi @jorgeamadosoria,
> 
> Something is not right at the rigth leg of the M.
> 
> image

I'll take a look, thanks for reviewing.

jorgeamadosoria:
I never saw that problem with the M.

regardless, I see it is right now after configuring precision 5.

image

Issue #5368, opened 2021-03-28
Support iterating over all icons


ericcornelissen:
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely:

### This issue concerns 
  - [x] The NPM Package
  - [ ] The Packagist Package

### Description

Currently, the `simpleIcons = require("simple-icons")` variable is only enumerable, meaning you can go over the slugs of the brand names. This is because under to hood it is just an object that maps slugs to icon data. It would be nice if we can somehow make it easier for users to iterate over the icons as well, rather than having to do `Object.values(simpleIcons)`, which may not be know to everyone.

#### Option 1

One option is to add a `Symbol.iterator ` value to the exported object like:

```javascript
Object.defineProperty(icons, Symbol.iterator, {
  enumerable: false,
  value: function*() { // Maybe we should shy away from generators, not sure how their support is
    for (const icon of Object.values(icons)) {
      yield icon;
    }
  }
});
```

Which would allow for `for-of` iteration like:

```javascript
for (const icon of simpleIcons) {
  console.log(icon.title);
}
```

The big question here is whether we `Object.values(icons)` once, which is "bad" for people that don't want to iterate, or every time we want to iterate, which is bad (worse) for people that iterate more than once.

The main disadvantage of this approach is that it doesn't (directly) support `.forEach`, `.map`, `.filter` etc. Of course these function could be added manually.

#### Option 2

Another option is to add a new key to the exported object that is just a list of the icons, e.g.:

```javascript
Object.defineProperty(icons, "list", {
  enumerable: false,
  configurable: false,
  value: Object.values(icons)
  // OR
  get() { return Object.values(icons); }
});
```

Which would allow for `for-of` iteration like:

```javascript
for (const icon of simpleIcons.list) {
  console.log(icon.title);
}
```

This has the same question of how often we want to call `Object.values`.

The main advantage of this approach is that it does support `.forEach`, `.map`, `.filter` etc. The disadvantage is that `simpleIcons` is not itself iterable which is arguably a bit nicer.

Thinking more long term, it could be argued that this option is better than option 1 - provided we call it `List` instead of `list` (and rename the `get` method to `Get`). This is because _if_ a brand's slug is "map" or "filter" (or "list" or "get), option 1 cannot support mapping or filtering of the icons. 

#### Option 3

[...]

PR #6789, opened 2021-10-29
replace index.js entry point

sachinraja:
Opening this just for discussion, this change is not approved by any other maintainer.

Replace the `index.js` entry point (`simple-icons`) with the current icons entry point (`simple-icons/icons`). My reasoning for this is that the majority of users will import the icon objects from `simple-icons/icons` and not import the `simpleIcons` object with all the icon data in `simple-icons`. It is already possible to perform a computation on each icon like so:
```js
import * as icons from 'simple-icons/icons'

for (const icon in icons) {
    // do stuff
}
```
The only use case that this removes support for is using `simpleIcons.Get`. I think users should solve this on their end, which is easily possible with a loop over each icon.

I did not update the readme.


ericcornelissen:
> The only use case that this removes support for is using `simpleIcons.Get`. I think users should solve this on their end, which is easily possible with a loop over each icon.

While it's easy, it's a much worse solution. The implementation of `simpleIcons.Get` is much more efficient as it looks up the icon in an object.


While I would be fine with updating the "default" entrypoint (i.e. `import from "simple-icons"`), I don't think we should remove the functionality to look up an icon using the `.Get` function (or similar) for the above reason.

My only concern with changing the default export is that it breaks backwards compatibility kinda unnecessarily. I think that _could_ be addressed by adding `.Get` as a non-iterable eleemnt on the default (i.e. `export default x`) from the default entrypoint, which I believe would still allow for importing individual icons with tree-shaking.

sachinraja:
> While it's easy, it's a much worse solution. The implementation of simpleIcons.Get is much more efficient as it looks up the icon in an object.

Users can create an object from the `import * as icons`

> I think that could be addressed by adding .Get as a non-iterable eleemnt on the default

But this would require us to create an object on top of all the exported icons. Not sure if that is tree shakable but it also seems no different than the current solution. My opinion is that not enough people are using `.Get` to justify this, but could be wrong here.

ericcornelissen:
> Users can create an object from the import * as icons

That is a fair point, though (probably?) still less efficient in terms of runtime than having the object hardcoded available. Of course, depending on your point of view/use-case that may be preferred over the increased file size that the hardcoded object requires.


> But this would require us to create an object on top of all the exported icons. Not sure if that is tree shakable [...]

To elaborate, my idea was that we provide the following options when importing:

```js
import { siSimpleicons } from "simple-icons"; // option 1
import * as icons from "simple-icons"; // option 2
import simpleIcons from "simple-icons"; // option 3
simpleIcons.Get("simpleicons");
```

I don't see why that would make the first option not-tree-shakable. But I'm not sure if we can have both 2 and 3 working. If not, for this solution we could look into making option 3 iterable (see also #5368).


> [...] but it also seems no different than the current solution.

Indeed! It is almost no different from the current solution because I was trying to avoid breaking backwards compatibility.


> My opinion is that not enough people are using `.Get` to justify this, but could be wrong here.

We have no idea whether that is true or not. For front-end I indeed hope it's not used. But it might be used on back-end systems or at built time.


---

To clarify, my primary concern is backwards compatibility.

sachinraja:
> That is a fair point, though (probably?) still less efficient in terms of runtime than having the object hardcoded available. Of course, depending on your point of view/use-case that may be preferred over the increased file size that the hardcoded object requires.

We could also create it at runtime to reduce the file size. I'm ok with leaving this for now if no one else thinks it should be removed.

mondeja:
I agree with @sachinraja here. I think that the `get/Get` property was there just to satisfy the requirement of gettting an icon by slug and title. In fact, in #4920, where this functionality was replaced by `simpleIcons[slug]`, Eric points in the PR description:

> This does beg the question whether we should convert back to the old API:
>
> ```javascript
> const simpleIcons = require('simple-icons');
> 
> console.log(simpleIcons['simpleicons']); // Use the filename of the SVG
> ```

So I would replace it in the two next releases, adding a deprecation warning in v7 (the next release) as was done in #6900 with `simple-icons/icons/[slug].js` imports and removing it completely in v8.

mondeja:
Hi @sachinraja. The next major version would be released ideally the next week. Could you update this pull to include it? Thanks in advance.

sachinraja:
updated at #8058

PR #6818, opened 2021-11-03
Fix typo in linter

service-paradis:
### Description

In some circumstances with ineffective segments, we have the following error:

> ReferenceError: chain is not defined


sachinraja (approved):
Good catch :+1:

jorgeamadosoria (approved):

jorgeamadosoria:
looks good, more readable messages. LGTM

PR #6819, opened 2021-11-03
Draft - Improve ineffective segments detection

service-paradis:
**Issue:** Follow up of #4257 and #4262

### Description

We already have a linter rule to determine if there are segments that does nothing. It works as expected and saves multiple bytes on multiple icons.

In some icons, there are nodes that are really close of each other and removing them does not change anything visually. This improvements to the linter try to detect nodes that are too close depending on a defined threshold (right now, I used 0.005). 

There are a lot of icons (~360) that could be improved according to the version in this PR. I need to do more verification to be sure there are no detection that could break the icon. I might need to adjust the threshold accordingly (we could also potentially increases it cince many close nodes are not even detected yet).

This is only a proposal for now. I will wait some feedback before going any further with this. If this is accepted, I will add current detected icons to the ignore list (and will create a `The 2022 Icon Review` 😅 maybe).

### DIFF
--- a/.svglintrc.js
+++ b/.svglintrc.js
@@ -15,6 +15,7 @@ const iconSize = 24;
 const iconFloatPrecision = 3;
 const iconMaxFloatPrecision = 5;
 const iconTolerance = 0.001;
+const segmentProximityThreshold = 0.005;
 
 // set env SI_UPDATE_IGNORE to recreate the ignore file
 const updateIgnoreFile = process.env.SI_UPDATE_IGNORE === 'true';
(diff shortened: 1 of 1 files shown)


dirien:
I love and hate ineffective segments 😅. But its good we get more feedback during linting. +1

mondeja:
Great idea! It seems to me a good threshold because 0.005 is a 0.0012% of the width and height, so surely nothing would change visually.

To be sure and to illustrate this PR, could you add here a comparison between the old and new shape of the icon with more failures (`composer.svg`) @service-paradis?

mondeja (commented):
Great idea! It seems to me a good threshold because 0.005 is a 0.0012% of the width and height, so surely nothing would change visually.

To be sure and to illustrate this PR, could you add here a comparison between the old and new shape of the icon with more failures (`composer.svg`) @service-paradis?

service-paradis:
> To be sure and to illustrate this PR, could you add here a comparison between the old and new shape of the icon with more failures (composer.svg) @service-paradis?

I should have tried this one! While `.005` is a really small value, the addition of them can create some discernible differences. Just blindly remove every nodes mentioned threw the icon off-center.

The preview looks pretty much similar:
image
image

But, comparing using Inkscape, we can see some "major" differences:
image

Thank you @mondeja for bringing this icon 😅, tried a bunch and it works pretty well! Might need to lower the threshold a bit to account for complicated icon as this one!

mondeja:
Interesting, so you plan to decrease the threshold?

service-paradis:
I will try to make more tests to be sure I'm not missing something. Most of the nodes removed from the composer icon were really unnecessary. I will think about this a little more!

service-paradis:
Well... It's been a long time! Closing this one for now after 2 years 😅. Tried with a threshold as minimal as `.001`. While it does a good job for most icons, some of the icons have too many little movements that create too much differences when removing them.

PR #6829, opened 2021-11-06
refactor: reorganize tests, reduce repetition

sachinraja:
* test ESM build
* test commonjs `icons/*.js`
* extract icon tests to separate function to avoid code duplication


package.json
```diff
@@ -61,7 +61,7 @@
     "prepare": "is-ci || husky install",
     "prepublishOnly": "npm run build",
     "postpublish": "npm run clean",
-    "test": "jest",
+    "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js",
```
sachinraja, line 64: Necessary for esm

jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  testMatch: ['**/tests/**/*.test.?(m)js'],
+  moduleFileExtensions: ['js', 'mjs', 'json'],
```
sachinraja, line 3: Necessary for esm

mondeja:
Since #6505 has been merged you need to fix conflicts 🔧

sachinraja:
Should be good now

package.json
ericcornelissen: Alternatively we could [ask contributors](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#testing-package-locally) to use at least v14, but this is fine as well. Though I'd say it is ideally removed when LTS for v12 ends ([on 2022-04-30](https://nodejs.org/en/about/releases/))

tests/icons.test.mjs
```diff
@@ -0,0 +1,30 @@
...
+
+  test(`${icon.title} has a valid svg value`, () => {
+    const svgFileContents = fs
+      .readFileSync(svgPath, 'utf8')
+      .replace(/\r?\n/, '');
+
+    expect(subject.svg.substring(subject.svg.indexOf('<title>'))).toEqual(
+      svgFileContents.substring(svgFileContents.indexOf('<title>')),
+    );
+  });
```
ericcornelissen, line 29: Shouldn't this be tested in `testIcon()`?

tests/test-icon.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {String} slug Icon data slug
+ */
+const testIcon = (icon, subject, slug) => {
+  describe(icon.title, () => {
+    it('has the correct "title"', () => {
+      expect(subject.title).toStrictEqual(icon.title);
+    });
+
+    it('has the correct "slug"', () => {
+      expect(subject.slug).toStrictEqual(slug);
```
ericcornelissen, line 14: There might be a good reason for this, but why not:

```suggestion
      expect(subject.slug).toStrictEqual(icon.slug);
```

sachinraja: `icon.slug` won't be defined in all cases though. `icon` is directly from `_data/simple-icons.json`

package.json
sachinraja: This is necessary even if you're on the latest Node version ([jest docs](https://jestjs.io/docs/ecmascript-modules)).

ericcornelissen: Thanks for pointing that out, didn't realize it's a Jest requirement :slightly_smiling_face:

tests/test-icon.js
ericcornelissen: Ah indeed, my bad :sweat_smile: Thank you for pointing that out

(I knew there was a reason but couldn't think of it, felt it was better to comment than to leave it)

ericcornelissen (approved):

mondeja (approved):

PR #6880, opened 2021-11-19
Avoid CRLF in SVG

LitoMore:
**Issue:** Fix #6869 

### Description

FYI: https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreeol

We should always use LF instead of CRLF in our code. Otherwise some developers will get a unstaged change when they pull the latest code.

image

### Changes

- ~~Added `end_of_line = lf` in our `.editorconfig`.~~
- ~~Changed our `wslint` to `editorconfig-checker -ignore-defaults` for linting SVG files.~~
- ~~Add `eol=lf` to `.gitattributes` to make sure developers are using the correct configuration with Git~~
- Updated regex pattern for `svglint`
- Removed unnecessary args from `wslint`

cc @ericcornelissen

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "jslint": "prettier --check .",
     "jsonlint": "node scripts/lint/jsonlint.js",
     "svglint": "svglint icons/*.svg --ci",
-    "wslint": "editorconfig-checker -exclude \\.svg$",
+    "wslint": "editorconfig-checker",
     "prepare": "is-ci || husky install",
     "prepublishOnly": "npm run build",
     "postpublish": "npm run clean",
--- a/.svglintrc.js
+++ b/.svglintrc.js
@@ -8,7 +8,7 @@ const svgPathBbox = require('svg-path-bbox');
 const parsePath = require('svg-path-segments');
 
 const svgRegexp =
-  /^<svg( [^\s]*=".*"){3}><title>.*<\/title><path d=".*"\/><\/svg>\r?\n?$/;
+  /^<svg( [^\s]*=".*"){3}><title>.*<\/title><path d=".*"\/><\/svg>\n?$/;
 const negativeZerosRegexp = /-0(?=[^\.]|[\s\d\w]|$)/g;
 
 const iconSize = 24;


.gitattributes
```diff
@@ -1,4 +1,4 @@
-*                  text=auto
+*                  text=auto eol=lf
```
ericcornelissen, line 1: Is this necessary? I believe `*   text=auto` already normalizes newlines the way `eol=lf` does (ref. https://git-scm.com/docs/gitattributes#_end_of_line_conversion)

package.json
```diff
@@ -57,7 +57,7 @@
     "jslint": "prettier --check .",
     "jsonlint": "node scripts/lint/jsonlint.js",
     "svglint": "svglint icons/*.svg --ci",
-    "wslint": "editorconfig-checker -exclude \\.svg$",
+    "wslint": "editorconfig-checker -ignore-defaults",
```
ericcornelissen, line 60: Is this change the reason a bunch of `.svg`s have their final newline removed? If so see #1682 for context on why they were excluded in the first place.

If we want to use editorconfig-checker here to check the type of newlines then I'm not sure what the best approach is. Forcing no final newlines has proven to be a real pain for some contributors with very little benefit. On the other hand, switching to requiring final newlines on SVGs would probably mean nearly 2000 file changes in a PR, causing a bunch of conflicts in the process... :thinking:

icons/anta.svg
```diff
@@ -1 +1 @@
...
+<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Anta</title><path d="M24 15.2372a20.744 20.744 0 0 0-9.86-7.5087 31.2657 31.2657 0 0 1 6.9097 1.0135l-1.5536-1.3192A29.1614 29.1614 0 0 0 9.0497 5.509a29.0797 29.0797 0 0 0-6.4051.7036L0 8.032c.335 0 .8376-.021 1.1747-.021a25.1537 25.1537 0 0 1 20.4571 10.48ZM9.1963 12.9758h3.3334l-.3329 1.183h-1.0532L9.9333 18.491H8.7692l1.206-4.3322H8.8655zm-3.771 0H6.468l.4376 2.9544.8229-2.9544h1.1977l-1.537 5.5152H6.221l-.4041-2.743-.7643 2.743H3.8841ZM0 18.491l2.8225-5.5131h1.181L3.769 18.491H2.5838l.0545-.7391H1.5264l-.3601.7391zm2.0206-1.8844h.6889l.2094-1.9474zm8.2122 1.8844 2.8288-5.5131h1.1768l-.2346 5.5131h-1.181l.0524-.7391h-1.1076l-.3644.7391zm2.0247-1.8844h.689l.2093-1.9474z"/></svg>
```
ericcornelissen, line 1: I'd recommend we fix this particular SVG in a separate PR so that the fix is out sooner than the solution to prevent it from happening again in the future - as evidenced by the commit history of this project, it doesn't happen frequently that an SVG with the wrong line-ending is added.

.gitattributes
LitoMore: Yes, it's necessary. Windows guys often commit their CRLF characters to the open-source world by mistake. Their editors all use CRLF by default. As I know, some people even have `eol=crlf` and `autocrlf=true` in their global config.

We don't use Windows-related tools so we can promise we don't need any CRLF in all files. It's OK to set all files to `eol=lf`.

package.json
LitoMore: Then we can use `.svglintrc.js` for linting SVG instead. Let's lint newline characters there.

BTW, the `-exclude  \\.svg$` won't take effect. Because `editorconfig-checker` already ignored them by default.

icons/anta.svg
LitoMore: Sure. PR created: https://github.com/simple-icons/simple-icons/pull/6893

.gitattributes
ericcornelissen: I'm not disagreeing we should normalize line endings and prevent CRLF characters from ending up in this project @LitoMore, I'm asking whether your change actually makes sense as, from the documentation, it would seem to me that `* text=auto` already normalizes newlines to LF. From the reference I provided:

> If you want to ensure that text files that any contributor introduces to the repository have their line endings normalized, you can set the text attribute to "auto" for all files.
>
> ```
> *	text=auto
> ```

package.json
ericcornelissen: > Then we can use .svglintrc.js for linting SVG instead. Let's lint newline characters there.

Please update the Pull Request accordingly, preferably including reverting the removal of final newlines from SVGs with an eye towards avoiding merge conflicts.

> BTW, the `-exclude \\.svg$` won't take effect. Because `editorconfig-checker` already ignored them by default.

I see thanks for pointing that out.

.gitattributes
LitoMore: OK. I reverted this change.

package.json
LitoMore: Done.

ericcornelissen (approved):
Looks good to me, this _should_ prevent CRLFs from creeping into the project in SVGs in the future (with a surprisingly simple fix :smiley:).

Pinging @mondeja or @sachinraja to merge this if they agree with the changes :slightly_smiling_face:

sachinraja (approved):

mondeja (approved):
Great, thanks!

PR #6896, opened 2021-11-23
add jsdoc preview images

sachinraja:
This idea was inspired from tailwindlabs/heroicons#286.

It converts the SVG to base64 so jsdoc renders it. I did have to do undo some minification so TS could read the jsdoc comments.

I am marking this as "in discussion" because the package size increased from 18.6 MB to 23.0 MB, so I'm not sure if this change is worth it. Here are a few nice previews that this makes possible though:
image
image
image

### DIFF
--- a/scripts/build/package.js
+++ b/scripts/build/package.js
@@ -100,24 +100,26 @@ data.icons.forEach((icon) => {
   writeJs(jsFilepath, `module.exports=${iconObject};`);
 
   const dtsFilepath = path.resolve(iconsDir, `${filename}.d.ts`);
+
+  writeTs(
+    dtsFilepath,
+    'declare const i:import("../alias").I;export default i;',
+  );
+
+  // jsdoc preview image (encoded in base64)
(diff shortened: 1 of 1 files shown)


jorgeamadosoria:
the idea is that a person adding an icon can preview it when importing within the IDE, and the trade off is that it adds 5MB more to the NPM package?

If correct, I don't think it's worth it. Still, I'm not adamantly against, I just wouldn't add it if it was solely up to me.
Let's see what the @simple-icons/maintainers say.

mondeja:
I'm against this for the reasons that @jorgeamadosoria has posted. It feels really IDE-based, not suitable for production environments.

sachinraja:
I agree, this isn't worth the size hit, so I'll close this. Just thought I'd try it for fun.

jorgeamadosoria:
> I agree, this isn't worth the size hit, so I'll close this. Just thought I'd try it for fun.

trying things is good. Keep at it!

PR #6898, opened 2021-11-23
make building async and minify with esbuild

sachinraja:
* Rewrite `scripts/build/package.js` to be async
* Switch `uglify-js` to `esbuild`

Time before changes:
```
❯ hyperfine 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):      6.833 s ±  0.206 s    [User: 10.050 s, System: 0.493 s]
  Range (min … max):    6.441 s …  7.046 s    10 runs
  ```
  
  Time after changes:
  ```
  ❯ hyperfine 'npm run build'
Benchmark 1: npm run build
  Time (mean ± σ):      1.516 s ±  0.092 s    [User: 1.176 s, System: 0.400 s]
  Range (min … max):    1.419 s …  1.656 s    10 runs
  ```
  I used https://github.com/sharkdp/hyperfine for more accurate benchmarks.

### CODE REVIEW

**Comment by mondeja** [comment]
What a great improvement 👏🏼 Could you solve the conflicts @sachinraja?

**Comment by sachinraja** [comment]
Should be good now

File: tests/icons.test.mjs [hunk 1]
```diff
@@ -1,14 +1,30 @@
...
+
+  test(`${icon.title} has a valid svg value`, () => {
+    const svgFileContents = fs
+      .readFileSync(svgPath, 'utf8')
+      .replace(/\r?\n/, '');
+
+    expect(subject.svg.substring(subject.svg.indexOf('<title>'))).toEqual(
+      svgFileContents.substring(svgFileContents.indexOf('<title>')),
+    );
+  });
```
mondeja, line 29: What is the purpose of these lines? Is already tested [here](https://github.com/simple-icons/simple-icons/blob/7c8146e475ffaf3a0c77d1c5c88a42686ef2399b/tests/test-icon.js#L61-L71), right?

sachinraja: I guess my conflict fixing skills were not as good as I thought they were :laughing:

mondeja (approved):

PR #6900, opened 2021-11-23
add deprecation warnings for icons/* imports

sachinraja:
See #6788 for context

* warn users when they `require('simple-icons/icons/[ICON_NAME]')`
* suggest they change to `const {si[ICON_NAME]} = require('simple-icons/icons')` instead

Not sure if I should suggest cjs or esm (current suggestion is cjs).

Tested this without a bundler, with webpack, and with esbuild and all log warnings as expected.
Example from my personal site:
image


scripts/build/package.js
```diff
@@ -94,10 +96,15 @@ data.icons.forEach((icon) => {
   icons.push(icon);
 
   const iconObject = iconToObject(icon);
+  const iconExportName = slugToVariableName(icon.slug);
 
   // write the static .js file for the icon
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
-  writeJs(jsFilepath, `module.exports=${iconObject};`);
+  const message = `This import is deprecated, use \\"const {${iconExportName}} = require('simple-icons/icons')\\" instead`;
```
mondeja, line 103: Maybe the warning should be more descriptive?

```suggestion
  const message = `This way of importing icons has been deprecated in v6.0.0 and will no longer work from v7.0.0, use \\"const {${iconExportName}} = require('simple-icons/icons')\\" instead`;
```

scripts/build/package.js
```diff
@@ -13,6 +13,8 @@ const util = require('util');
 const { minify } = require('uglify-js');
 
 const UTF8 = 'utf8';
+// yellow and bold
+const WARN = '\\x1b[33m\\x1b[1mwarn\\x1b[22m\\x1b[39m';
```
mondeja, line 17: I'm not sure about this. Is it supposed to work within the browser also? What happens with terminals without colors (see [`NO_COLOR`](https://no-color.org/)? Also, it works under Windows CMD? I think that is better to not include color nor formatting for the warning.

mondeja (changes requested):

scripts/build/package.js
sachinraja: I think it'll help people notice that it's a warning.
My browser (Chromium) renders it fine: image

It does break on Firefox though, but I don't think that's a big deal:
image

For the `NO_COLOR` part, we could use [`picocolors`](https://github.com/alexeyraspopov/picocolors) (which is what I looked at to get those ansi codes), which detects that. I'm not sure about Windows cmd, but I'm guessing ansi codes don't work there.

mondeja: I think it's not worth to add a dependency to the package for a warning, Firefox needs HTML for formatting and in Windows CMD doesn't work.

sachinraja: Actually I just tested in Windows CMD and it works fine. I guess Node handles it somehow.

mondeja: Note that `console.warn` itself renders with yellow background color in browsers.

sachinraja: I don't think it's worth it either, and I'll remove the formatting, but that's interesting to know.

mondeja: 🤔 Wow that's new, I would swear it hadn't worked long time ago...

sachinraja:
I added the deprecation warnings to the types too (jsdoc).

scripts/build/package.js
```diff
@@ -94,19 +94,26 @@ data.icons.forEach((icon) => {
   icons.push(icon);
 
   const iconObject = iconToObject(icon);
+  const iconExportName = slugToVariableName(icon.slug);
 
   // write the static .js file for the icon
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
-  writeJs(jsFilepath, `module.exports=${iconObject};`);
+  const newImportMessage = `use "const {${iconExportName}} = require('simple-icons/icons')" instead`;
```
mondeja, line 101: Prettier adds spaces next to `{` `}` characters. Just a minor change but seems reasonable:

```suggestion
  const newImportMessage = `use "const { ${iconExportName} } = require('simple-icons/icons')" instead`;
```

sachinraja: I originally had it like that to make the message smaller to reduce the impact on the package size, but it doesn't make much of a difference.

scripts/build/package.js
```diff
@@ -94,19 +94,26 @@ data.icons.forEach((icon) => {
   icons.push(icon);
 
   const iconObject = iconToObject(icon);
+  const iconExportName = slugToVariableName(icon.slug);
 
   // write the static .js file for the icon
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
-  writeJs(jsFilepath, `module.exports=${iconObject};`);
+  const newImportMessage = `use "const { ${iconExportName} } = require('simple-icons/icons')" instead`;
```
mondeja, line 101: Another stupid but newbie-friendly change:

```suggestion
  const newImportMessage = `use "const { ${iconExportName} } = require('simple-icons/icons');" instead`;
```

scripts/build/package.js
mondeja: It doesn't matter I guess, as this is temporal in v6

scripts/build/package.js
```diff
@@ -94,19 +94,26 @@ data.icons.forEach((icon) => {
...
 
   const iconObject = iconToObject(icon);
+  const iconExportName = slugToVariableName(icon.slug);
 
   // write the static .js file for the icon
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
-  writeJs(jsFilepath, `module.exports=${iconObject};`);
+  const newImportMessage = `use "const { ${iconExportName} } = require('simple-icons/icons');" instead`;
+  const message = JSON.stringify(
+    `This way of importing icons has been deprecated in v6.0.0 and will no longer work from v7.0.0, use ${newImportMessage}`,
```
mondeja, line 103: I'm thinking now that it could be really weird for users that could see this message without knowing where is being triggered, especially in big codebases. Maybe we can add an even more descriptive message?

```suggestion
    `Imports like "const ${icon.slug} = require('simple-icons/icons/${icon.slug}');" have been deprecated in v6.0.0 and will no longer work from v7.0.0, use ${newImportMessage}`,
```

This would warn:

```
warn - Imports like "const simpleicons = require('simple-icons/icons/simpleicons');" have been deprecated in v6.0.0 and will no longer work from v7.0.0, use "const { siSimpleicons } = require('simple-icons/icons');" instead
```

scripts/build/package.js
```diff
@@ -94,19 +94,26 @@ data.icons.forEach((icon) => {
   icons.push(icon);
 
   const iconObject = iconToObject(icon);
+  const iconExportName = slugToVariableName(icon.slug);
 
   // write the static .js file for the icon
   const jsFilepath = path.resolve(iconsDir, `${filename}.js`);
-  writeJs(jsFilepath, `module.exports=${iconObject};`);
+  const newImportMessage = `use "const { ${iconExportName} } = require('simple-icons/icons');" instead`;
```
mondeja, line 101:
```suggestion
  const newImportMessage = `"const { ${iconExportName} } = require('simple-icons/icons');" instead`;
```

mondeja (changes requested):

scripts/build/package.js
sachinraja: I was thinking that too, but wasn't sure how to solve it. This looks good!

mondeja (approved):
Looks perfect :100:
Feel free to merge if you don't have additional concerns

sachinraja:
Thanks for reviewing so quickly on this one! I think it's great that we can get this out in the v6 release instead of a random release between v6-v7.

Issue #6889, opened 2021-11-22
Request: AWS Lambda


itamarc:
### Brand Name

AWS Lambda

### Website

https://aws.amazon.com/lambda/

### Popularity Metric

"AWS Lambda is the most mature and widely used function-as-a-service (FaaS) offering (...)"
Source: https://www.datadoghq.com/state-of-serverless/

### Official Resources for Icon and Color

https://aws.amazon.com/trademark-guidelines/

https://d1.awsstatic.com/product-marketing/Lambda/Diagrams/product-page-diagram_Lambda-RealTimeFileProcessing.a59577de4b6471674a540b878b0b684e0249a18c.png

https://worldvectorlogo.com/pt/download/aws-lambda-1.svg

### Contributing

No

adsingh14:
I will contribute to this issue.

adsingh14:
I want to contribute but unable to decide which icon is correct one?
According to official [website](https://aws.amazon.com/lambda/), the following is the correct one (~~https://worldvectorlogo.com/pt/download/aws-lambda-1.svg~~).

![image](https://user-images.githubusercontent.com/35252877/143176993-aed90908-4ea5-4332-9a44-7fefc9f7a9b3.png)

PR #6907, opened 2021-11-24
AWS lambda icon

adsingh14:
**Issue:** #6889
**Alexa rank:** 11
  

### Checklist

### Description

Color used from official brand guidelines
Icon used '32-bit svg icon' as reference from [this link](https://aws.amazon.com/architecture/icons/)
awslambda


dirien (commented):
Hi @adsingh14, I am not sure if we can make the stroke bigger.


Asking one ohter @simple-icons/maintainers for advice?

adamrusted:
> I am not sure if we can make the stroke bigger.

I would say no, given the entire icon is the outline. We would be altering the core logo too much to just make it bolder.

adsingh14:
> > I am not sure if we can make the stroke bigger.
> 
> I would say no, given the entire icon is the outline. We would be altering the core logo too much to just make it bolder.

No worries, I will match the stroke-width and update this commit.

LitoMore:
> No worries, I will match the stroke-width and update this commit.

No. You should find out the AWS Lambda logo from here: https://aws.amazon.com/architecture/icons/

adsingh14:
> No. We should always respect their artwork. The generated SVG from PNG/JPG definitely has some differences. You should find out the AWS Lambda logo from here: https://aws.amazon.com/architecture/icons/
> 
> Next time please spend some time on search logos with simple words: https://www.google.com.hk/search?q=aws+lambda+logo

I'm designer too and respect the official work. I kept this on hold for some reasons with keeping in mind that they'd ask to curate it. I was unable to find  https://aws.amazon.com/architecture/icons back then.

adamrusted:
> Edit: Even their provided .ppt file contains the png image. I've found another icon library and found the exact icon. Will update shortly.

The 'Asset Pack' below contains SVG versions of all the icons. Please use those.

LitoMore:
I think we should drop its background and make the `λ` filled.

@simple-icons/maintainers What do you think?

sachinraja:
I agree, so I think we can move forward with that @adsingh14

adsingh14:
> I agree, so I think we can move forward with that @adsingh14 

What about official icon then if we invert the fill? There are other icons, eg. MSFT translator, Adobe's, etc. , which we used as it is.

sachinraja:
That's actually a fair point and on second thought, after reading through the presentation a bit more, I think it makes sense to keep the current icon (even though I think it looks better filled).

LitoMore:
> I think we should drop its background and make the `λ` filled.

What I mean is to fill the stroke. Not to fill them all. Like this below:

CleanShot 2022-01-10 at 14 24 47@2x

jorgeamadosoria:
I think the icon as presented by @LitoMore will work just fine: no background and stroke filled only.

Background could work too, but just to move this along, I would go with Lito's version.

dirien:
Hi @adsingh14 did you updated the logo with the feedback of @LitoMore?

adsingh14:
> Hi @adsingh14 did you updated the logo with the feedback of @LitoMore?

I updated this logo on the basis of sachinraja jorgeamadosoria and LitoMore, suggestion. 

I've decided to keep it original like I mentioned (https://github.com/simple-icons/simple-icons/pull/6907#issuecomment-1008553035) but other members were not okay with the original icon.

sachinraja:
Are you sure you pushed your changes @adsingh14? I see the updated icon in the icon preview but I don't see any new commits.

adsingh14:
> Are you sure you pushed your changes @adsingh14? I see the updated icon in the icon preview but I don't see any new commits.

..

sachinraja:
@adsingh14 did you see my comment above?

adsingh14:
> @adsingh14 did you see my comment above?

Yes, I did. I checked that too.

sachinraja (changes requested):
image
Some differences here (I believe in stroke size)
I used this icon from the asset pack:
Res_AWS-Lambda_Lambda-Function_48_Dark
and removed the circle around it.

sachinraja (approved):
Nevermind this icon matches with this (also from the asset pack):
Arch_AWS-Lambda_48
 
 Good work on this, thanks for sticking with it!

PR #6926, opened 2021-11-29
Update SVGO config

### LINKED ISSUE #6899: Update SVGO config
SVGO configuration was not properly documented when v2 was released, but now it is, so we must update `svgo.config.js` file accordingly.

mondeja:
Closes #6899

Don't uses default preset, instead explicitly enables plugin by plugin.

### DIFF
--- a/svgo.config.js
+++ b/svgo.config.js
@@ -1,99 +1,100 @@
 module.exports = {
   multipass: true,
+  eol: 'lf',
   plugins: [
+    'cleanupAttrs',
+    'inlineStyles',
+    'removeDoctype',
+    'removeXMLProcInst',
+    'removeComments',
+    'removeMetadata',
+    'removeDesc',
+    'removeUselessDefs',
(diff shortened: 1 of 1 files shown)


sachinraja:
I think it would be good to use the default preset because that would opt us into optimizations made in new plugins without explicitly enabling them.

mondeja:
I'm not against it, but it has some drawbacks experienced in the past that I would like to avoid:

- For the same reason it could also break the optimization. Has been a common problem that SVGO has changed configuration many times without specify it in documentation. Also, breaking changes have occurred without major release changes, even for plugins included in the default preset.
- It seems, at least to me, really counterintuitive to know what plugins are the defaults and disable them. Even more is the syntax to disable them.
- With that approach we need to change the default plugins when upstream changes occurred in SVGO, so it doesn't really solves anything.

 So, for me, it's better to enable plugins explicitly.

sachinraja:
Ok, I agree then.

PR #6946, opened 2021-12-04
convert scripts to esm

sachinraja:
Convert the script files to esm. This includes all files under the `scripts` and `tests` dirs.


scripts/html-friendly-to-title.cjs
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  /**
+   * Converts a brand title in HTML/SVG friendly format into a brand title (as
+   * it is seen in simple-icons.json)
+   * @param {String} htmlFriendlyTitle The title to convert
+   */
+  htmlFriendlyToTitle: (htmlFriendlyTitle) =>
```
sachinraja, line 7: This has to be moved to a separate file and written in cjs because `.svglintrc.js` imports it and it does not allow for esm config files.

scripts/build/package.js
```diff
@@ -7,15 +7,25 @@
  * tree-shakeable
  */
 
-const fs = require('fs').promises;
-const path = require('path');
-const util = require('util');
-const { transform: esbuildTransform } = require('esbuild');
+import fsSync from 'fs';
```
LitoMore, line 10:
```suggestion
import {promises as fs} from 'fs';
```

scripts/html-friendly-to-title.cjs
LitoMore: Yes. @ericcornelissen is working on it: https://github.com/birjolaxew/svglint/pull/41.

scripts/build/package.js
sachinraja: Looks better :+1:

scripts/lint/jsonlint.js
```diff
@@ -4,23 +4,33 @@
...
 
-const path = require('path');
-const { Validator } = require('jsonschema');
+import fsSync from 'fs';
+import { fileURLToPath } from 'url';
+import path from 'path';
+import { Validator } from 'jsonschema';
+import { getIconData } from '../utils.js';
+
+const fs = fsSync.promises;
```
LitoMore, line 13: This line needs to be removed. Please check all files in this PR.

sachinraja:
Personally, I don't mind mixing the two, especially since it's only for two files, but I'm ok with waiting.

.svglintrc.mjs
```diff
@@ -1,11 +1,26 @@
...
+  __dirname,
+  'node_modules',
+  'named-html-entities-json',
+  'index.json',
+);
+const svglintIgnoredFile = path.join(__dirname, '.svglint-ignored.json');
+
+const data = JSON.parse(fs.readFileSync(dataFile, 'utf8'));
+const htmlNamedEntities = JSON.parse(
+  fs.readFileSync(htmlNamedEntitiesFile, 'utf8'),
```
LitoMore, line 21: Would you mind to change these sync function to async?

```js
import {promises as fs} from 'node:fs';

const dataFile = JSON.parse(await fs.readFile(svglintIgnoreFile));
```

sachinraja: I didn't change it because top level await requires users to be on Node v14

LitoMore: Oh, I see.

.svglintrc.mjs
```diff
@@ -1,11 +1,26 @@
...
-
-const data = require('./_data/simple-icons.json');
-const { htmlFriendlyToTitle } = require('./scripts/utils.js');
-const htmlNamedEntities = require('named-html-entities-json');
-const svgpath = require('svgpath');
-const svgPathBbox = require('svg-path-bbox');
-const parsePath = require('svg-path-segments');
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
```
LitoMore, line 3: Since we required Node.js `^12.20.0 || ^14.13.1 || >= 16.0.0` for running this script. We can use [`node:` protocol](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) for imports.

```js
import fs from 'node:fs';
import path from 'node:path';
import { fileURLToPath } from 'node:url';
```

Here is a [discussion](https://github.com/nodejs/node/issues/38343) about `node:` procotol.

scripts/utils.js
```diff
@@ -3,76 +3,96 @@
...
-   */
-  slugToVariableName: (slug) => {
-    const slugFirstLetter = slug[0].toUpperCase();
-    const slugRest = slug.slice(1);
-    return `si${slugFirstLetter}${slugRest}`;
-  },
+/**
+ * Get contents of _data/simple-icons.json.
+ */
+export const getIconDataString = () => {
```
mondeja, line 85: Shouldn't be in this function `Icon` named as `Icons` as is referring to a plural?

```suggestion
export const getIconsDataString = () => {
```

scripts/utils.js
```diff
@@ -3,76 +3,96 @@
...
-      .replace(/</g, '&lt;')
-      .replace(/>/g, '&gt;')
-      .replace(/./g, (char) => {
-        const charCode = char.charCodeAt(0);
-        return charCode > 127 ? `&#${charCode};` : char;
-      }),
+/**
+ * Get icon data as object from _data/simple-icons.json.
+ */
+export const getIconData = async () => {
```
mondeja, line 95: The same here, don't forget to change all references to them.

```suggestion
export const getIconsData = async () => {
```

.svglintrc.mjs
sachinraja: Oh, thanks for looking into that, I didn't use them because I wasn't sure if they worked on all of our supported versions. I'll change all of the std lib imports to use the `node:` prefix.

mondeja (changes requested):

mondeja (approved):
Awesome! 🎊

LitoMore (approved):

PR #6951, opened 2021-12-07
Automatically update milestone (SVG count) in README

mondeja:
Closes https://github.com/simple-icons/release-action/issues/27


jorgeamadosoria (approved):
as long as this takes the place of a Github Action, that's a win.

scripts/release/update-svgs-count.js
```diff
@@ -0,0 +1,38 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const regexMatcher = /Over\s(\d+)\s/;
+const updateRange = 500;
+
+const rootDir = path.resolve(__dirname, '..', '..');
+const dataFile = path.resolve(rootDir, '_data', 'simple-icons.json');
+const readmeFile = path.resolve(rootDir, 'README.md');
+const readmeContent = fs.readFileSync(readmeFile).toString();
```
sachinraja, line 18: Should add the `utf-8` encoding and remove `toString`

sachinraja (changes requested):

sachinraja:
I think we should update every 100 instead of every 500 to be more accurate in the count, but I'd like to hear what everyone else thinks about that too.

mondeja:
> I think we should update every 100 instead of every 500 to be more accurate in the count, but I'd like to hear what everyone else thinks about that too.

I agree. What do you think about it @jorgeamadosoria?

jorgeamadosoria:
> > I think we should update every 100 instead of every 500 to be more accurate in the count, but I'd like to hear what everyone else thinks about that too.
> 
> I agree. What do you think about it @jorgeamadosoria?

no particular opinion about it. If you and @sachinraja agree on 100, and it is not a performance hit on the build, go for it.

scripts/release/update-svgs-count.js
mondeja: Updated by https://github.com/simple-icons/simple-icons/pull/6951/commits/6d9fbfe3e5dc0617b0fd7ad717c1551bc1f55bf9 in other files as well

scripts/release/update-svgs-count.js
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env node
```
LitoMore, line 1: This is unnecessary.

```suggestion
```

scripts/release/update-svgs-count.js
```diff
@@ -0,0 +1,38 @@
...
+/**
```
LitoMore, line 2: This is unnecessary.

```suggestion
/**
```

mondeja: I think that this discussion is beyond this PR because we've the shebang in all scripts. It's true that currently are doing nothing. However, I'm inclined to give executable permissions to all scripts and maintain the shebangs, so developers can execute them directly in Linux using `./scripts/release/update-svgs-count.js`. See [this answer](https://stackoverflow.com/a/33510581/9167585) for more info about it. Do you want to open another PR for that @LitoMore?

LitoMore: > Do you want to open another PR for that or an issue to discuss it

Sure.

sachinraja (approved):

sachinraja:
Awesome, good work on this @mondeja!

PR #6964, opened 2021-12-14
Correctly add `new icon` labeled pull requests to Priority 2 column

mondeja:
This solves the problem pointed by @service-paradis on https://github.com/simple-icons/simple-icons/pull/6615#issuecomment-931322124

Currently, when an user opens a pull request using the default template, this is not linked to the correspondent issue if it would be closed. For this reason, the `Add Pull Request Labels and Assign to Project` workflow does not categorize well pull requests which have the `new icon` label to priority 2 column of reviews project.

One solution would be to replace `**Issue:**` in new icon template by `Closes` or `Close`, but this has the downside that an user could close a tracking issue if a maintainer does not notice it. So I've added the [`add_links_by_content` input](https://github.com/mondeja/pr-linked-issues-action#add_links_by_content) to [`pr-linked-issues-action`](https://github.com/mondeja/pr-linked-issues-action). This input treats issues as linked even if they are not really linked, giving a match by pull request body content.


sachinraja (dismissed):

sachinraja:
> One solution would be to replace **Issue:** in new icon template by Closes or Close, but this has the downside that an user could close a tracking issue if a maintainer does not notice it.

Doesn't this have the same issue?

mondeja:
> Doesn't this have the same issue?

Using that in the pull request template would link to issues, but for [tracking](https://github.com/simple-icons/simple-icons/labels/tracking) ones, we don't want to close them. The solution has been rejected by maintainers a couple of times mainly because of that.

sachinraja:
But I don't see how that solves this. If there's an `Issue:` linking to a tracking issue, wouldn't it still close when the PR is merged?

mondeja:
See:

image

Linked issues are issues that may be closed by a pull request, not just metioned ones. See [Linking a pull request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

sachinraja:
Sorry if I'm misunderstanding but with the changes in this PR (if this action were to run), what is the difference between `closes #1` and `Issue: #1`?

mondeja:
> Sorry if I'm misunderstanding but with the changes in this PR (if this action were to run), what is the difference between `closes #1` and `Issue: #1`?

- In terms of the closing behaviour: `closes #1` links the issue, so when merged the issue would be closed as is explained in [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), whereas `Issue: #1` is the same that just `#1`, the pull request will be shown as mentioned in the issue, but the issue will not be closed when merged.
- In terms of the action, `Issue: #1` will not be treated as a linked issue because is not added to `add_links_by_content`, but `closes #1` yes because the issue in that case is explicitly linked. If we want that the issue would treated as a linked, so taken into account for categorizing to Priority 2, we need to add `Issue: #{issue_number}` to the `add_links_by_content` input. Currently only the variants `**Issue:**` and `**Issue**:`, regarding to the word `Issue` have been added. I've only used these variants because the first is the used by the pull request template and the second could be an slight change of the first. I'm open to add more if you consider them necessary.

sachinraja:
I understood that keywords can be used to automatically link issues. I think the issue was that I was misunderstanding what the `pr-linked-issues-action`. So it doesn't actually link the issues, it only adds it to the step's output, which is used for prioritizing?

mondeja:
> So it doesn't actually link the issues, it only adds it to the step's output, which is used for prioritizing?

Yes, that's correct 👍🏼

sachinraja (approved):

PR #6965, opened 2021-12-17
remove Google Search Console

sachinraja:
There doesn't seem to be an icon anymore: https://search.google.com/search-console/about.

Part of #4205


service-paradis:
It looks like they still use their icons in the [products page](https://about.google/products/) and for the Search Console favicon:
image
image

sachinraja:
Ah you're right! I'll close this.

PR #6967, opened 2021-12-17
Add warning about v6.0.0 npm package deprecation to README

mondeja:
Added deprecation warning introduced in #6900 to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ import { siSimpleicons } from 'simple-icons/icons'
 ```
 
 You can also import the needed icons individually, where `[ICON SLUG]` is replaced by a [slug].
+
+⚠️ This way of importing icons has been deprecated in v6.0.0 and will no longer work from v7.0.0.
 ```javascript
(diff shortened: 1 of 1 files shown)


sachinraja:
We should probably just remove it from the README

mondeja:
It couldn't be strange for users that have used it, if they come back to the README to see the documentation about simple-icons, to not find any about this type of import?

sachinraja:
I don't think it'll be that confusing, but we can just add a note like this that it's been removed:
> Note: the old way of importing with `require('simple-icons/icons/[ICON SLUG])` is deprecated and will be removed in v7.0.0.

mondeja:
I like it 👍🏼 

Done in https://github.com/simple-icons/simple-icons/pull/6967/commits/c71a9b92b2978643aa58ffd17a57917bbd6f7ace but used `:warning:` (:warning:) instead of `Note:`

sachinraja (approved):

PR #6970, opened 2021-12-19
Upgrade SVGLint dependency

ericcornelissen:
### Description

Opening this PR to upgrade the [SVGLint](https://github.com/birjolaxew/svglint) dependency to `v2.0.0`, the major version bump requires a higher minimum NodeJS version for development than is currently mentioned in this project's Contributing Guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -370,7 +370,7 @@ If you have an affiliation to the brand you contributing that allows you to spea
 
 ## Testing Package Locally
 
-* Make sure you have [NodeJS](https://nodejs.org/en/download/) installed. At least version 12.20.0 is required.
+* Make sure you have [Node.js](https://nodejs.org/en/download/) installed. At least version `^12.20.0 || ^14.13.1 || >=16.0.0` is required.
 * Install the dependencies using `$ npm install`.
 * Build and test the package using `$ npm test`.
 * Run the project linting process using `$ npm run lint`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -370,7 +370,7 @@ If you have an affiliation to the brand you contributing that allows you to spea
 
 ## Testing Package Locally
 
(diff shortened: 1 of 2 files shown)


sachinraja:
I think we should do this in #6946

sachinraja:
Nevermind, I'll merge from develop in the other PR.

sachinraja (changes requested):
I think we should update `"engines"` in package.json too

ericcornelissen:
> I think we should update "engines" in package.json too

I don't think so because that also affects people with `simple-icons` as a dependency, and to them the v12.20.0 restriction does not apply.

sachinraja:
Ah ok, this looks good to me then.

sachinraja (approved):

LitoMore (approved):
> I think we should update "engines" in package.json too

@ericcornelissen is right. We don't need to update the `engines` field. We have already documented the the requirement in readme for developers.

Good job! I will move our `.svglintrc.js` to ESM in another PR.

CONTRIBUTING.md
```diff
@@ -370,7 +370,7 @@ If you have an affiliation to the brand you contributing that allows you to spea
 
 ## Testing Package Locally
 
-* Make sure you have [NodeJS](https://nodejs.org/en/download/) installed. At least version 12.4.0 is required.
+* Make sure you have [NodeJS](https://nodejs.org/en/download/) installed. At least version 12.20.0 is required.
```
LitoMore, line 373: I'm thinking about this part. The exact version range should be `^12.20.0 || ^14.13.1 || >=16.0.0`.

```suggestion
* Make sure you have [Node.js](https://nodejs.org/en/download/) installed. At least version `^12.20.0 || ^14.13.1 || >=16.0.0` is required.
```

sachinraja:
> Good job! I will move our .svglintrc.js to ESM in another PR.

I'll do that in #6946

mondeja (approved):
Great!

Issue #7035, opened 2022-01-08
Request: YOLO Icon


Si-jeong:
### Brand Name

YOLO

### Website

https://pjreddie.com/darknet/yolo/

### Popularity Metric

The Alexa rank is 191,860. See [https://www.alexa.com/siteinfo/pjreddie.com](https://www.alexa.com/siteinfo/pjreddie.com).

### Official Resources for Icon and Color

png: https://pjreddie.com/media/image/yologo_2.png
color: `#00ffff`, from website header 

### Contributing

Yes

PR #7040, opened 2022-01-09
Add YOLO icon

Si-jeong:
yolo

**Issue: #7035 **
**Alexa rank: 191,336 **
  

### Checklist

### Description
The hex value was picked based on the color of the [logo](https://pjreddie.com/media/image/yologo_2.png).
I optimized the logo with [SVGOMG Online Tool](https://jakearchibald.github.io/svgomg/).


jorgeamadosoria:
@Si-jeong thanks for your contribution!
I noticed that there is a variable width in the outer layer of the letters (specially rightmost draw of the last O and leftmost upper corner of the Y). This may be a result of the conversion from PNG to SVG.

If we could get that fixed, I don't think the icon has any other outstanding issues.

Si-jeong:
> @Si-jeong thanks for your contribution! I noticed that there is variation in width in the outer layer of the letters (specially rightmost draw of the last O and leftmost upper corner of the Y). This may be a result of the conversion from PNG to SVG.
> 
> If we could get that fixed, I don't think the icon has any other outstanding issues.

Hi, @jorgeamadosoria! Thanks for pointing it out. I updated the PR with the new SVG.

jorgeamadosoria:
excellent, looks much better. LGTM

PR #7048, opened 2022-01-12
Decrease Typescript icon size in README

mondeja:
Typescript icon is placed inside an `h4` (`####` in Markdown) in the README while other sections with icons are using `h3` but the icons has the same size (24 now). It seems reasonable to me to reduce the size of the Typescript icon so that it is more in line with the dimensions of the section title.

### Before

image

### After

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ for (const iconSlug in simpleIcons) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sachinraja (approved):
Nice job with these README updates!

Issue #7062, opened 2022-01-15
Request: NetBSD


fraggerfox:
### Brand Name

NetBSD

### Website

https://netbsd.org

### Popularity Metric

The Alex rank is: 89,285

NetBSD is a free, fast, secure, and highly portable Unix-like Open Source operating system.

It is from the BSD family of Operating Systems.

### Official Resources for Icon and Color

color: #ff6600

### Additional Comments

The image is taken from https://www.netbsd.org/gallery/logos.html

Usage guidelines can be found here https://www.netbsd.org/about/disclaimer.html

### Contributing

Yes

PR #7064, opened 2022-01-15
Add NetBSD icon.

fraggerfox:
**Issue:** #7062 
**Alexa rank:** 89,285
  

### Checklist

### Description

-  The hex value was chosen from the original logo at https://www.netbsd.org/gallery/logos.html
- ~The logo was manually vectorized.~ The Encapsulated PostScript from the website was used.

image


adamrusted:
Did you use the EPS version of the logo available on the source page? Without looking too deeply, I'd say that's probably our best bet for an accurate vector.

fraggerfox:
> Did you use the EPS version of the logo available on the source page? Without looking too deeply, I'd say that's probably our best bet for an accurate vector.

You are correct, I missed out the EPS link, I have regenerated the svg from the EPS, thank you for pointing this out.

_data/simple-icons.json
```diff
@@ -6566,6 +6566,12 @@
             "source": "http://www.netapp.com/",
             "guidelines": "https://www.netapp.com/company/legal/trademark-guidelines/"
         },
+        {
+            "title": "NetBSD",
+            "hex": "FF6600",
+            "source": "https://www.netbsd.org/gallery/logos.html",
```
dirien, line 6572: Can you write the source just with "https://www.netbsd.org" please.

dirien (changes requested):
hi @fraggerfox,

Perfect fit!

image

Left a comment on the source tag. 

After this, we can merge this PR

_data/simple-icons.json
fraggerfox: Should be done in the latest commit.

dirien:
thx @fraggerfox! Merged! :)

fraggerfox:
@dirien could you have a look at the related PR #7061 (which is the package manager for NetBSD)? (When you get time of course), Thank you.

PR #7071, opened 2022-01-17
switch from uvu to mocha

sachinraja:
uvu is small and fast, but the reporter (which cannot be configured) is not great for this project. It outputs all tests even if they pass so it is difficult to find failing tests in the 40,000 tests we run. [Mocha](https://github.com/mochajs/mocha) is also relatively small and fast and has a `min` reporter to show only the failing tests.

Benchmarks:
uvu:
```
Benchmark 1: npx uvu
  Time (mean ± σ):      2.518 s ±  0.099 s    [User: 2.878 s, System: 0.414 s]
  Range (min … max):    2.409 s …  2.702 s    10 runs
```

mocha:
```
Benchmark 1: npx mocha tests --reporter min --inline-diffs
  Time (mean ± σ):      3.312 s ±  0.110 s    [User: 4.018 s, System: 0.345 s]
  Range (min … max):    3.115 s …  3.441 s    10 runs
```

So mocha is slower, but not by much.

### DIFF
--- a/tests/min-reporter.cjs
+++ b/tests/min-reporter.cjs
@@ -1,0 +1,10 @@
+const { reporters } = require('mocha');
+
+class EvenMoreMin extends reporters.Base {
+  constructor(runner) {
+    super(runner);
+    runner.once('end', () => this.epilogue());
+  }
+}
+
+module.exports = EvenMoreMin;
--- a/tests/min-reporter.cjs
+++ b/tests/min-reporter.cjs
@@ -1,9 +1,11 @@
-const { reporters } = require('mocha');
+const { reporters, Runner } = require('mocha');
+
+const { EVENT_RUN_END } = Runner.constants;
 
 class EvenMoreMin extends reporters.Base {
   constructor(runner) {
     super(runner);
-    runner.once('end', () => this.epilogue());
+    runner.once(EVENT_RUN_END, () => this.epilogue());
   }
 }
 
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "prepare": "is-ci || husky install",
     "prepublishOnly": "npm run build",
     "postpublish": "npm run clean",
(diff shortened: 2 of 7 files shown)


mondeja (approved):
Great! I personally don't like the [screen clearing command](https://github.com/mochajs/mocha/blob/baa12fd73e59ab6139d05d5eb76222c5d7a774ba/lib/reporters/min.js#L39) executed by the min reporter, but tested with [even-more-min](https://github.com/justinmchase/even-more-min) which doesn't clear the screen and the errors are reported a bit different. Could be worth to take a look at https://github.com/mochajs/mocha/issues/2312, maybe in the future they implement `clearscreen=false` as an option for that reporter. Anyways is not a big problem I guess.

tests/min-reporter.cjs
```diff
@@ -0,0 +1,10 @@
+const { reporters } = require('mocha');
```
sachinraja, line 1: Unfortunately this file has to be cjs, mocha does not support esm reporter files.

tests/min-reporter.cjs
```diff
@@ -0,0 +1,10 @@
...
+
+class EvenMoreMin extends reporters.Base {
+  constructor(runner) {
+    super(runner);
+    runner.once('end', () => this.epilogue());
```
mondeja: I think that it would be more secure to use the constants directly:

```suggestion
const { constants } = require("mocha/lib/runner")

class EvenMoreMin extends reporters.Base {
  constructor(runner) {
    super(runner);
    runner.once(constants.EVENT_RUN_END, () => this.epilogue());
```

mondeja: We are thinking the same 😅

mondeja (approved):

LitoMore (approved):

PR #7073, opened 2022-01-19
Optimize `titleToSlug` function

mondeja:
The same change done in https://github.com/simple-icons/simple-icons-website/pull/173

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -7,6 +7,32 @@ import path from 'node:path';
 import { promises as fs } from 'node:fs';
 import { fileURLToPath } from 'node:url';
 
+const TITLE_TO_SLUG_REPLACEMENTS = {
+  '+': 'plus',
+  '.': 'dot',
+  '&': 'and',
+  đ: 'd',
+  ħ: 'h',
+  ı: 'i',
+  ĸ: 'k',
+  ŀ: 'l',
+  ł: 'l',
+  ß: 'ss',
+  ŧ: 't',
+};
+
+const TITLE_TO_SLUG_CHARS_REGEX = RegExp(
+  // Special characters such as '+' and '.' must be escaped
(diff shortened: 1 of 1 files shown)


scripts/utils.js
```diff
@@ -7,6 +7,32 @@ import path from 'node:path';
...
+  ĸ: 'k',
+  ŀ: 'l',
+  ł: 'l',
+  ß: 'ss',
+  ŧ: 't',
+};
+
+const TITLE_TO_SLUG_CHARS_REGEX = RegExp(
+  // Special characters such as '+' and '.' must be escaped
+  Object.keys(TITLE_TO_SLUG_REPLACEMENTS)
```
LitoMore, line 26: You don't have to escape them if you use Character Class `[]` instead.

scripts/utils.js
```diff
@@ -7,6 +7,32 @@ import path from 'node:path';
...
+    .map((char) => {
+      return ['+', '.'].includes(char) ? `\\${char}` : char;
+    })
+    .join('|'),
+  'g',
+);
+
+const TITLE_TO_SLUG_RANGE_REGEX = RegExp('[^a-z0-9]', 'g');
```
LitoMore, line 34:
```suggestion
const TITLE_TO_SLUG_RANGE_REGEX = /[^a-z0-9+.]/g;
```

scripts/utils.js
mondeja: But these will not work in the replacement, right? Needs to be pure chars in the replacements object and escaped in the regex string.

LitoMore: It can be:

```js
RegExp(`[${Object.keys(TITLE_TO_SLUG_REPLACEMENTS).join('')}]`, 'g')
```

mondeja: Awesome, done some performance tests and it seems to be fast also! 👍🏼

Issue #7106, opened 2022-01-28
Request: ReScript Icon


mxthevs:
### Brand Name

ReScript

### Website

https://rescript-lang.org/

### Popularity Metric

The compiler currently has 5.7k stars on Github. See https://github.com/rescript-lang/rescript-compiler

### Official Resources for Icon and Color

svg: https://raw.githubusercontent.com/jesseweed/seti-ui/master/icons/rescript.svg
color: `#E6484F`

### Additional Comments

We have the Reason icon already. It would be nice to have the ReScript icon as well.

### Contributing

Yes

PR #7108, opened 2022-01-29
Add ReScript icon

mxthevs:
image

**Issue:** #7106
**Number of stars on GitHub:** 5.7k

### Checklist

### Description

We have the Reason icon already. It would be nice to have the ReScript icon as well.


sachinraja:
I am never sure if these icons should be the inner part or outer part. In [the official icon](https://rescript-lang.org/brand#logo) the outer part is filled. In this PR, the inner part is filled. Maybe the other @simple-icons/maintainers can help with this?

jorgeamadosoria:
@sachinraja inner, since that's the way it is in the Github source. The icon as proposed is fine imo.

@mxthevs, the problem I see is that the icon as proposed is 24x24, but importing the SVG from the source yields a 23.864x24 size. It is not readily visible with the naked eye, you would have to superimpose the icons in an editor to see the difference, but it's there.

If you fix that, then this is good to merge. Let us know!

sachinraja:
@jorgeamadosoria the GitHub source is not official

jorgeamadosoria:
> @jorgeamadosoria the GitHub source is not official

but it is the one used as source. The point I'm raising is that the contribution does not conform to the source.
Do you have a source that shows that the icon is perfectly square? Because in that case we should change the source as well to that one.

jorgeamadosoria:
as it turns out, there are authoritative sources here: https://rescript-lang.org/brand#header
and it should be outer, according to that.
Given that, this contribution needs to change the source to: https://rescript-lang.org/brand#header
and the icon should be taken from: https://rescript-lang.org/static/brand/rescript-brandmark.svg

mxthevs:
I can make the requested changes @jorgeamadosoria. 

@sachinraja, I took the icon from a PR opened by one of the project maintainers, but I will change this to use the official source.

mxthevs:
@sachinraja, Can you take a look at this now, please? I have updated the icon source.

jorgeamadosoria:
it looks much better now.
I don't know what was the original PR, but this is much better, thank you for the changes.
LGTM

PR #7109, opened 2022-01-29
Add Java Third-Party Extension

silentsoft:
I just made a [Java implementation](https://github.com/silentsoft/simpleicons4j) of the simple-icons.
I hope my library will help anyone who wants to use simple-icons in Java.

Thanks,

PR #7111, opened 2022-01-30
Fix 3rd party extensions order, add test to avoid regression

mondeja:
Java 3rd party extension added in #7109 has not been added to the table in alphabetical order. This PR fixes that and adds a test to prevent this inconsistency in future PRs. Also, allows to reuse the function to get third party extensions from the README in [Webpack configuration of simple-icons-website](https://github.com/simple-icons/simple-icons-website/blob/2455e835b560859faffe5d06645eac11ae318a20/webpack.config.js#L21-L41), just the file `scripts/utils.js` needs to be downloaded as we do with SVGO configuration.


silentsoft:
Oh, I didn't even notice that was in alphabetical order.. my bad :sob: sorry about that

scripts/utils.js
```diff
@@ -117,3 +117,30 @@ export const getIconsData = async () => {
  */
 export const getDirnameFromImportMeta = (importMetaUrl) =>
   path.dirname(fileURLToPath(importMetaUrl));
+
+/**
+ * Get information about third party extensions.
+ */
+export const getThirdPartyExtensions = async () => {
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
sachinraja, line 125:
```suggestion
  const __dirname = getDirnameFromImportMeta(import.meta.url);
```
in the test file too

tests/readme.test.js
```diff
@@ -0,0 +1,62 @@
...
+  assert.ok(thirdPartyExtensions.length > 0);
+
+  const thirdPartyExtensionsNames = thirdPartyExtensions.map(
+    (ext) => ext.module.name,
+  );
+
+  const expectedOrder = thirdPartyExtensionsNames.slice().sort();
+  assert.equal(
+    thirdPartyExtensionsNames.join(', '),
+    expectedOrder.join(', '),
```
sachinraja, line 59: We could use `assert.deepEqual` here to compare the array contents instead of joining them and comparing the strings.

sachinraja (changes requested):

tests/readme.test.js
```diff
@@ -0,0 +1,62 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, test } from 'mocha';
+import { strict as assert } from 'node:assert';
+import { getThirdPartyExtensions } from '../scripts/utils.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
sachinraja, line 8: Should be changed here too

sachinraja (changes requested):

tests/readme.test.js
```diff
@@ -0,0 +1,65 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
```
sachinraja, line 3: This import isn't needed anymore

sachinraja (changes requested):

sachinraja (approved):

PR #7127, opened 2022-02-04
Update Node.js to v16 in CI workflows

mondeja:
Node.js v12 is in [maintenance state and will reach his EOL in May](https://nodejs.org/en/about/releases/).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Use Node.js 12.x
+      - name: Use Node.js 16.x
         uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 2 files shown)


sachinraja:
Shouldn't we wait for it to go EOL in May then?

mondeja:
It is to have less work on May, we need to release the next major version and update dependency lists for compute again the monetary distributions then. I'm not sure why we should freeze with v12 when we don't have compatibility problems.

sachinraja (approved):
Sounds good then. We could also update the scripts to use the latest syntax (you can do that or I may send a PR for it later).

mondeja:
> We could also update the scripts to use the latest syntax (you can do that or I might do it later).

Please, open the PR. Seems that you're more familiar with it, I don't know what the exact changes are.

PR #7155, opened 2022-02-17
fix: whatsapp source/guidelines links

maurodesouza:
Was updated the whatsapp souce/guidelines link, because the old was redirecting to an unsafe page

image


LitoMore:
> You can only use the official WhatsApp Brand Resources that are available from the WhatsApp Brand Resources website.

I think that means we can use their logo, but we must pick the logo from the [WhatsApp Brand Resources](https://www.facebook.com/brand/resources/whatsapp/logo) page.

sachinraja (approved):
Sounds good :+1: @LitoMore, thanks for this update @maurodesouza!

Issue #5001, opened 2021-02-13
PUBGm


polar-beer:
**Brand Name: PUBG mobile**
**Website: https://pubgmobile.jp/**
**Alexa rank: **
  

**Official resources for icon and color:**
https://www.pubgmobile.com/en/event/brandassets/

fbernhart:
The Alexa rank is at [326,807](https://www.alexa.com/siteinfo/pubgmobile.jp), so it's within the scope of Simple Icons.

GitHiru:
Thanks a lot!

PeterShaggyNoble:
Thanks for the request, @GitHiru.

I've yet to find an official vector version for PUBG Mobile. There are some high quality raster files available [here](https://www.pubgmobile.com/en/event/brandassets/) but it's going to be tricky to recreate manually. It we don't need to distinguish between PUBG and the mobile version (I'm not a gamer so don't know what the distinction is) then there's an SVG of just the PUBG logo available in the header of [this site](https://www.pubg.com/).

adamrusted:
@PeterShaggyNoble I would be tempted to say we just add `PUBG`. As far as I'm aware, it's just a 1:1 mobile port of the 'main' game.

PR #7199, opened 2022-03-07
Pubg

jorgeamadosoria:
pubg

issue: #5001 

This icon is for the desktop game, but they are similar enough that I think this one can cover both (Mobile version just adds "mobile" below the icon).

Color taken from the css for the header in: https://www.pubgmobile.com/en/event/brandassets/


sachinraja:
It seems like this is mixing up NetBox with PUBG. Also there is no need to request reviews on new PRs, we'll get to it when we see it.

jorgeamadosoria:
> It seems like this is mixing up NetBox with PUBG. Also there is no need to request reviews on new PRs, we'll get to it when we see it.

removed NetBox.

Also, this is not really a new PR: This is a replacement for an old abandoned PR that has been waiting review for almost a year now. I closed that and created this new one in its stead.

dirien (approved):
LGTM!

Nice @jorgeamadosoria, love it!

PR #7218, opened 2022-03-14
Add `funding` key to *package.json*

mondeja:
See https://docs.npmjs.com/cli/v6/configuring-npm/package-json#funding


sachinraja (approved):

mondeja:
This will display something like this in the npm page (example from webpack):

image

PR #7229, opened 2022-03-16
Update MobX icon

sisp:
MobX icon preview

**Issue:** n/a
**GitHub Stars:** 25k
  

### Checklist

### Description

In the previous version of the MobX icon, the text component was not vertically centered correctly. The problem was not detected by `svglint` because the text component is contained in a box, so the overall shape was centered. I've fixed the alignment with a minimal diff of the SVG path data.


sachinraja (approved):

sachinraja:
Good catch!

PR #7241, opened 2022-03-19
add "exports" for SVG files

sisp:
I was trying to import the SVG icons directly as React components using [SVGR](https://react-svgr.com/), but it failed with the following error:

> Module not found: Package path `./icons/<brand>.svg` is not exported from package `.../node_modules/simple-icons` (see exports field in `.../node_modules/simple-icons/package.json`)

So I took a look at the "exports" field in `package.json` and realized that any import from `simple-icons/icons/*` was mapped to `simple-icons/icons/*.js` which effectively prevents importing any non-JS files like SVGs in this case. When importing an SVG with SVGR in place, the import contains the `.svg` file extension, e.g.:

```ts
import SimpleIconsSVG from "simple-icons/icons/simpleicons.svg";
```

To make this case work, too, I've configured an [alternative path](https://webpack.js.org/guides/package-exports/#alternatives) without a file extension, so that

```ts
import ... from "simple-icons/icons/simpleicons"
```

maps to `simple-icons/icons/simpleicons.js` and

```ts
import ... from "simple-icons/icons/simpleicons.svg"
```

maps to `simple-icons/icons/simpleicons.svg`.


sachinraja (approved):
Makes sense, but it's odd that Webpack requires us to export a non-standard file.

sisp:
> [...] but it's odd that Webpack requires us to export a non-standard file.

Yes, I have no idea, but it works with this change. :slightly_smiling_face:

sachinraja:
I tried this:
```js
import simpleIcons from 'simple-icons/icons/simpleicons.svg'
console.log(simpleIcons)
```
and ran `node index.js`. I get the error `simple-icons/icons/simpleicons.svg.js`, so I assume this will fall back to `simple-icons/icons/simpleicons.svg` when that file cannot be resolved in a bundler like Webpack?

sisp:
Sounds reasonable, that might be it.

PR #7272, opened 2022-03-26
Add Gurobi icon

sisp:
Gurobi icon preview

[Gurobi](https://www.gurobi.com/) is an industry-standard software for mathematical optimization.

**Issue:** n/a
**Similarweb rank:** [200,482](https://www.similarweb.com/website/gurobi.com/#overview)
  

### Checklist

### Description

It was difficult to find a vector image of the logo. The only source I could eventually find was their [PDF brochure](https://cdn.gurobi.com/wp-content/uploads/2021/02/Gurobi-Optimization_Corporate-Brochure.pdf) from where I extracted the logo using Inkscape and post-processed it manually to remove redundant path points.

To make the logo monochromatic, I removed the gradients and added gaps using the following approach:
1. For each polygon, I added a stroke of 0.5px.
2. I removed path points to keep only a polyline along the sides that touched the other polygons.
3. I converted the polyline strokes to paths.
4. I subtracted the polyline paths from the polygons from which I created the polylines.

With this approach, the gaps are 0.5px because subtracting the polyline strokes of 0.5px from the original polygons removed 0.25px from each polygon. Also, the polygons were only cut at the sides that touched other polygons while the other sides remained unmodified.

The aspect ratio of the original logo was exactly 1:1. Because of the introduced gaps, the height was reduced by a tiny bit.

The hex value is the color code of the polygon in the original logo that has no gradient. I believe that's the primary color of the logo and the brand color of Gurobi.


sisp:
I've found some evidence (a promotional gift pen from Gurobi with a monochromatic icon on it) that the official monochromatic version of the Gurobi icon might look like this:

Gurobi icon v2 preview

Should we use this one instead? I have the SVG ready and could make a commit.

sachinraja:
@sisp I think we should go with that one then.

sisp:
@sachinraja Agreed. I've pushed a commit with the new icon version.

sachinraja (approved):
Well done!

PR #7287, opened 2022-03-30
Fix SVG lint error

zrfisaac:
There is some SVG file with a line ending that is wrong according to "npm run lint".


zrfisaac:
I forgot to push my modifications.

PR #7288, opened 2022-03-30
Fix SVG lint error

zrfisaac:
There is some SVG file with a line ending that is wrong according to "npm run lint"


sachinraja:
We run `npm run lint` in CI and the latest workflow runs are passing. Can you make sure you're on the latest version? I can see that there's something wrong with the commits (perhaps an incorrect merge?) as there are 56 commits in this PR that were already merged. Maybe try cloning again following the [contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md)?

PR #7357, opened 2022-04-26
Advise to use GitHub file URL also when color origin differs

### LINKED ISSUE #7356: GitHub source URL when brand color source is not the SVG file
### Kind of Issue

Mistake

### Description

I've stumbled over the [source guidelines section in the contribution guidelines](https://github.com/simple-icons/simple-icons/blob/fc042e56ff02ae3265fd7c72d8f3e245126cd181/CONTRIBUTING.md#source-guidelines) related to GitHub sources which isn't entirely clear to me:

> - GitHub: For an SVG from a GitHub (GitLab, BitBucket, etc.) repository the source URL should link to the file that was used as source material. **If the color comes from another file in the repository the URL should link to the repository itself.**
>
>   In any case, the commit hash should be part of the URL. On GitHub, you can get the correct URL by pressing y on the GitHub page you want to link to. You can get help at the [getting permanent links to files page](https://help.github.com/en/github/managing-files-in-a-repository/getting-permanent-links-to-files).

So let's say a GitHub repo contains an SVG icon which is monochromatic. Is my understanding correct that the source URL should be only the repo URL (as a permalink?) and not the precise URL (permalink) of the SVG file?

If yes, then I think the source URL should still point to the SVG file even when the color comes from elsewhere and the PR description should contain the information about the origin of the color. Losing the precise information about the source SVG file just because the original icon does not have the right color is a bit crude in my opinion and makes tracking the source unnecessarily hard.

What do you think?

### Contributing

Yes

sisp:
I've updated the contribution guidelines to use a GitHub file URL also for SVG files that do not contain the brand color. Before, in this case the GitHub _repository_ URL should be used which means losing the source information only because of a different color origin.

Closes #7356.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -293,9 +293,9 @@ If the SVG is sourced from:
 
 - **Branding page**: For an SVG from a branding page the source URL should link to the branding page and not the image, PDF, or archive (such as `.zip`) file.
 - **Company website**: If the SVG is found on the company website (but there is no branding page) the source URL should link to a common page, such as the home page or about page, that includes the source image and not the image file itself.
-- **GitHub**: For an SVG from a GitHub (GitLab, BitBucket, etc.) repository the source URL should link to the file that was used as source material. If the color comes from another file in the repository the URL should link to the repository itself.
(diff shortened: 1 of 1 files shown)


mondeja (approved):
Great, thank you!

PR #7359, opened 2022-04-27
Add hashes to all Github sources

mondeja:
Since #7357, all Github sources must have hashes, so this PR add them to all that currently haven't. I've used the hashes of the time from when the icons were added or lastly updated.


sachinraja:
Shouldn't the links also go to the icon files?

mondeja:
> Shouldn't the links also go to the icon files?

Maybe not because we can use multiple sources for an icon or the repository could contain additional information (see https://github.com/simple-icons/simple-icons/pull/4906#issuecomment-770999701), so I've preferred to respect the current sources.

PR #7360, opened 2022-04-28
Add Boxy SVG to list of vector graphics editors

sisp:
I've added [Boxy SVG](https://boxy-svg.com/) to the list of vector graphics editors in `CONTRIBUTING.md`. I've been using it for several years on Linux and like it much better than Inkscape. In my opinion, Boxy SVG has a much better UX, is cleaner and simpler, and adds significantly less editor-specific metadata to the SVG markup. It has worked very well for me also for extracting icons for Simple Icons.

It's only free on Linux and as a (regular) web app though. macOS users and users of the native browser apps need to buy in the respective app stores.


mondeja (approved):
Thanks!

Issue #7374, opened 2022-05-02
Removal of Java Coffee Cup Logo


service-paradis:
### Brand Name

Java

### Reason for Removal

Emails received at removals@simpleicons.org from an Oracle Corporation attorney.

zrfisaac:
Really ? they said why ?

sachinraja:
@zrfisaac our "use was not authorized by Oracle." No further reason was provided.

zrfisaac:
Okey, today I will do a pull request. 😕

zrfisaac:
I made a pull request #7378.

RedHeadphone:
Can we not take permission from Oracle?

mondeja:
You can always ask them.

Kawaxte:
Curse you, Oracle.

danielrdrigues:
so f*cking funny

witherslayer67:
> ### Brand Name
> 
> Java
> ### Reason for Removal
> 
> Emails received at [removals@simpleicons.org](mailto:removals@simpleicons.org) from an Oracle Corporation attorney.

how cringe of oracle

witherslayer67:
i found this
[Java Logo Guidelines](https://www.oracle.com/a/ocom/docs/java-licensing-logo-guidelines-1908204.pdf)

KryptonScript:
> ### Brand Name
> Java
> 
> ### Reason for Removal
> Emails received at [removals@simpleicons.org](mailto:removals@simpleicons.org) from an Oracle Corporation attorney.

Oh, that's bad. We will use other options, then.
God bless 🙏🏾

danielrdrigues:
Suggestion, make a cup with a 💩  in the place of the fire

PR #7378, opened 2022-05-05
Remove Java

zrfisaac:
Removal of Java Coffee Cup Logo
**Issue:** #7374

Brand Name
Java

Reason for Removal
Emails received at [removals@simpleicons.org](mailto:removals@simpleicons.org) from an Oracle Corporation attorney.

😕


sachinraja:
Thanks! You'll also need to remove our use of `assets/readme/java-white.svg` from the README. 

Marking this as draft so it doesn't accidently get merged before the next major.

zrfisaac:
Year, I saw that after, I've Already made this, that for the advice! 🙌

zrfisaac:
I should remove it from here too? because it is not going to load anymore
Capturar

mondeja:
> I should remove it from here too? because it is not going to load anymore

Yes, but I think that the best aproach would be to add an empty icon, so the `Java Library` link is not misaligned with the rest of the third party libraries. Would you mind to replace it by a 24x24 empty icon @zrfisaac?

sachinraja:
Yes, that's what I was referring to by this:
> You'll also need to remove our use of assets/readme/java-white.svg from the README.

zrfisaac:
I've removed like you said like you can see bellow :
image

But I had one idea, why not use the [OpenJDK](https://openjdk.java.net) logo !? 

_OpenJDK (Open Java Development Kit) is a free and open-source implementation of the Java Platform, Standard Edition (Java SE)._

I'd like to recomend use the OpenJDK logo. what do you think @sachinraja ?
openjdk

sachinraja:
Great idea! Let's do that.

zrfisaac:
Here is the pull request #7391 with this logo.

zrfisaac:
Alright, after the merger of the OpenJDK logo, I will do the change.

zrfisaac:
I've made the change, it's how it's going to look like.

image

sachinraja:
@zrfisaac we can make the readme changes now, so could you make a separate PR for that?

assets/readme/openjdk-white.svg
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
mondeja, line 1: Hi @zrfisaac 

You need to use the same content of `openjdk.svg` but adding `fill="white"` attribute or [the README test will not pass](https://github.com/simple-icons/simple-icons/runs/6340479990?check_suite_focus=true). See the other icons in  [assets/readme](https://github.com/simple-icons/simple-icons/tree/b2cf3885ce82351a74b3562c7a19b9e7a45ce67b/assets/readme) directory as reference. Thanks!

mondeja (changes requested):

assets/readme/openjdk-white.svg
zrfisaac: Thanks I fogot to do that, I will fix today!

mondeja: But please, separate the logo change in the third party plugin from this PR so we can update it without a major version release, see https://github.com/simple-icons/simple-icons/pull/7378#issuecomment-1120407819

mondeja (approved):
I've continued with the work here. This is ready, thanks for your work @zrfisaac 👍🏼

sachinraja (approved):

PR #7381, opened 2022-05-05
Add third-party Hugo module in README

foo-dogsquared:
Added my [Hugo module](https://github.com/foo-dogsquared/hugo-mod-simple-icons) to the list of third-party extensions. It is automatically managed by GitHub workflows to update the version weekly which is nice. :)


mondeja (approved):
Great, thank you!

Issue #7390, opened 2022-05-07
Add OpenJDK icon


zrfisaac:
### Brand Name

OpenJDK

### Website

https://openjdk.java.net/

### Popularity Metric

The Similar web rank is [- -](https://www.similarweb.com/website/openjdk.java.net). See https://www.similarweb.com/website/openjdk.java.net

### Official Resources for Icon and Color

svg: https://hg.openjdk.java.net/duke/duke/file/ca00f100dafc/vector/Agent.svg#l1
color: `#FFFFFF`, from the website header

### Contributing

Yes

PR #7391, opened 2022-05-07
Add OpenJDK icon

zrfisaac:
openjdk

**Issue:** #7390

**Similarweb rank:** [- -](https://www.similarweb.com/website/openjdk.java.net/#overview)

### Checklist

### Description
_OpenJDK (Open Java Development Kit) is a free and open-source implementation of the Java Platform, Standard Edition (Java SE)._
Even if the similarweb not show a correct rank, it's  very used by linux users instead of the JDK(Java). And After Oracle's disapproval from using the java coffey logo in this issue #7378, I think we should use this instead.


mondeja (approved):
Looks good. Thank you @zrfisaac!

PR #7427, opened 2022-05-19
Add aliases to AWS icons

### LINKED ISSUE #7242: Clarify Amazon and AWS icon names
### Brand Name

Amazon / AWS

not relevant

### Additional Comments

A couple of AWS icons use the "Amazon" brand name instead of "AWS" which I think is not optimal. Instead, I suggest to call AWS icons "AWS \<Service\>" and the files `aws<service>.svg` whereas Amazon products remain as they are. This would mean a breaking change though unless aliases are introduced.

What do you think?

### Contributing

Yes

sisp:
I've added aliases to icons of AWS services whose official names are "Amazon \<Service\>".

Resolves #7242.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -469,13 +469,23 @@
         {
             "title": "Amazon AWS",
             "hex": "232F3E",
-            "source": "https://upload.wikimedia.org/wikipedia/commons/9/93/Amazon_Web_Services_Logo.svg"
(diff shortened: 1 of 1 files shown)


mondeja (approved):
Thanks! I'll wait a couple of days to merge it just in case that someone has anything to said about.

sachinraja (approved):

PR #7441, opened 2022-05-22
deprecate index entrypoint

sachinraja:
discussion in #6789
#6788 should be merged first

Deprecate the `simple-icons` entry point as it has very little usage and can be replaced with `import * as icons from 'simple-icons'`

### DIFF
--- a/scripts/build/templates/index.js
+++ b/scripts/build/templates/index.js
@@ -1,3 +1,5 @@
+console.warn('Deprecation warning: The `simple-icons` entrypoint will be removed in the next major. Please switch to using `import * as icons from "simple-icons/icons"` if you need an object with all the icons.')
+
 var icons = {%s};
 
 Object.defineProperty(icons, "Get", {
--- a/README.md
+++ b/README.md
@@ -49,20 +49,7 @@ The icons are also available through our npm package. To install, simply run:
 npm install simple-icons
 ```
 
-The API can then be used as follows, where `[ICON SLUG]` is replaced by a [slug]:
-
-```javascript
-const simpleIcons = require('simple-icons');
-
-// Get a specific icon by its slug as:
-// simpleIcons.Get('[ICON SLUG]');
-
-// For example:
-const icon = simpleIcons.Get('simpleicons');
-
-```
-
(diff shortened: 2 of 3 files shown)


scripts/build/templates/index.js
```diff
@@ -1,3 +1,5 @@
+console.warn('Deprecation warning: The `simple-icons` entrypoint will be removed in the next major. Please switch to using `import * as icons from "simple-icons/icons"` if you need an object with all the icons.')
```
mondeja, line 1: This souldn't be inside the `Get.value` function?

sachinraja: I think we may have different perceptions of what I intended to do in #6789. I meant to remove the whole file. If we remove the `.Get` function, the index file no longer has a use as you can get an object with all the icons using `import * as icons from 'simple-icons/icons'`

mondeja: Ah OK, I think I messed up a bit here 👍🏼

mondeja: But, could be a way to import with `import * as icons from 'simple-icons'`? Perhaps an alias in exports?

mondeja (approved):

scripts/build/templates/index.js
sachinraja: Ah yes, that's a good idea. I'll do that.

mondeja: Yes, it seems simpler and more straightforward to me. That change would be in #6789 right?

sachinraja: It's not breaking so I could do it here

sachinraja: Oh also please do not release the major until my next PR. I'm still experimenting with one more change

mondeja: > It's not breaking so I could do it here

Great

> Oh also please do not release the major until my next PR. I'm still experimenting with one more change

It's scheduled in two days, so I think that you've enough time. It's done automatically like the rest of the releases. The [release-action](https://github.com/simple-icons/release-action) checks if there is at least one removal of an icon to do a major version bump.

sachinraja: > But, could be a way to import with import * as icons from 'simple-icons'? Perhaps an alias in exports?

Looked into doing this and I don't think it'll work well tree-shaking and bundle size wise so I'm reluctant to do it until we remove what is being deprecated in this PR.

sachinraja:
Merging this so I can open another PR on top

PR #7454, opened 2022-05-27
Deprecate svg prop

sachinraja:
Deprecate the `icon.svg` property in favor of a `import { getSvg } from 'simple-icons/utils'`. I set a warning until the next major. This reduces bundle size.


README.md
```diff
@@ -86,6 +85,14 @@ NOTE: the `license` entry will be `undefined` if we do not yet have license data
 */
 ```
 
+You can get the SVG string for an icon like so:
+```javascript
+import { siSimpleicons } from 'simple-icons/icons';
+import { getSvg } from 'simple-icons/utils';
```
mondeja, line 91: I'm not sure about the name, `getSvg` seems to be very general to me. Maybe `getSvgFromSimpleIcon`?

scripts/build/templates/icon-object.js
```diff
@@ -2,7 +2,8 @@
   title: '%s',
   slug: '%s',
   get svg() {
-    return '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>%s</title><path d="' + this.path + '"/></svg>';
+    d()
+    return getSvg(this)
```
mondeja, line 6: Ok, so the change is:

`'<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>%s</title><path d="' + this.path + '"/></svg>'` (125 characters) by `getSvg(this)` (12 characters), but `getSvg` seems very general as said before. Replacing with `getSvgFromSimpleIcon` would be 26 characters.

Other option is to avoid exposing functions at all and set at the top `const a = '<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>`... and adding a `htmlTitle` property, but minifying the names of the variables at the top, using for example `a` and `b`, would be something like: `a+this.htmlTitle+b+this.path+c` + `title:''` + length of the title encoded (38 characters + length of the title encoded).

Anyways, have you tested the performance decreasement of running 5 regexes for each character in the title? Maybe to reading the file is a faster process 😅 I think that the reduction in bundle size is not worth, I prefer to start with an `htmlTitle` if we want to remove the long string in the `svg` getter.

Additionally, I'm not totally against exposing an utility file, but I would prefer to expose it entirely, not only 3 functions and testing all utilities.

utils.mjs
```diff
@@ -0,0 +1,23 @@
...
+    .replace(/&#([0-9]+);/g, (_, num) => String.fromCharCode(parseInt(num)))
+    .replace(
+      /&(quot|amp|lt|gt);/g,
+      (_, ref) => ({ quot: '"', amp: '&', lt: '<', gt: '>' }[ref]),
+    );
+
+export const getSvg = (icon) =>
+  `<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>${titleToHtmlFriendly(
+    icon.title,
+  )}</title><path d="${icon.path}"/></svg>`;
```
mondeja, line 23: I've checked in the past what is faster: `string${interpolation}` or `'string' + addition` and the second is. Also, string interpolation is an ES6 feature, right? See [this thread](https://github.com/simple-icons/simple-icons/pull/6505/files#r706527151). I suppose that ESBuild converts to additions, but I'm not sure as you're targetting the build for NodeJS v14. Note that for the package [we are targetting all NodeJS versions as of 0.12.18](https://github.com/simple-icons/simple-icons/blob/986589c34c1224fcd0244e07a8fe833079fd836f/package.json#L76-L78), so the `engine` would be obsolete after this change and we are removing official support for a lot of legacy versions. I prefer not to do it.

mondeja (commented):
I'm not against the exposition of utilities, but definitely I'm against doing it in this way.

utils.mjs
sachinraja: Let's target Node 14 as Node 12 is end of life

scripts/build/templates/icon-object.js
sachinraja: Hmm ok, let's go with your change for now. I'll close this.

README.md
sachinraja: The types and function param name should help but that's fine for me if you still prefer it

mondeja:
But at the end what is the advantdge here? We would be breaking a bunch of versions and what do we get in return? Just reducing a bit the bundle size? I think that https://github.com/simple-icons/simple-icons/pull/7457 and an addition of a `htmlTitle` property would solve this problem without breaking anything.

sachinraja:
Personally I find it cleaner to have a separate function but I'm ok with closing this if it's not worth the friction.

PR #7492, opened 2022-06-10
Add guidelines link for Amazon icon

### LINKED ISSUE #7467: Amazon icon source
Currently, the Amazon icon source is [worldvectorlogo.com/logo/amazon-icon](https://worldvectorlogo.com/logo/amazon-icon) which is problematic in my opinion because the [terms of use](https://worldvectorlogo.com/terms-of-use) prohibit commercial use:

> You may only use this website and its content for personal, non-commercial purposes. By visiting our website and downloading the displayed content, you agree with these terms of use. If you do not agree with, or do not wish to abide, these terms and conditions, we strongly urge you to not visit our website and to not use its content.

This problem was also discussed in https://github.com/simple-icons/simple-icons/pull/7332#discussion_r855245765 and there we decided to not use worldvectorlogo.com as the source.

I've tried finding another source of the Amazon icon and was hoping to use Wikipedia, but the [Amazon icon from Wikipedia](https://en.wikipedia.org/wiki/File:Amazon_icon.svg) also [declares worldvectorlogo.com as its source](https://en.wikipedia.org/wiki/File:Amazon_icon.svg#filehistory). I believe using the icon from Wikipedia won't be any better than sourcing the icon directly from worldvectorlogo.com because the terms of use apply transitively (although I'm not a lawyer, that's just my common sense).

Unfortunately, I haven't found an alternative (high-quality) source of the icon. Any thoughts on how we should proceed?

sisp:
Resolves #7467.

I think this is what you mean in https://github.com/simple-icons/simple-icons/issues/7467#issuecomment-1149403511, @sachinraja, right?

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -459,7 +459,8 @@
         {
             "title": "Amazon",
             "hex": "FF9900",
-            "source": "https://worldvectorlogo.com/logo/amazon-icon"
+            "source": "https://worldvectorlogo.com/logo/amazon-icon",
+            "guidelines": "https://worldvectorlogo.com/terms-of-use"
         },
         {
             "title": "Amazon Alexa",


sachinraja (approved):
Yes, thanks! Although if someone has a source that isn't so restrictive, we should switch to that.

PR #7536, opened 2022-06-26
Add BorgBackup logo

### LINKED ISSUE #7530: Borg Backup
### Brand Name

Borg Backup

### Website

https://github.com/borgbackup/borg

### Popularity Metric

8.4k stars on github

https://github.com/borgbackup/borg/tree/master/docs/_static

### Additional Comments

_No response_

### Contributing

No

foo-dogsquared:
**Issue:** closes #7530

**Similarweb rank:** ~2 million
  
Since Similarweb ranking is not enough, [its GitHub repo](https://github.com/borgbackup/borg/) does have 8.4k stars.

### Checklist

### Description

Hex value is from the SVG.


sisp (commented):
Thanks for your contribution, @foo-dogsquared! :pray:

I wonder whether the icon should be just the text without the rectangle in the background, i.e.:


Two reasons:
1. You state that the hex value is `#00DD00` which is the fill color of the _text_, but when the icon is the rectangle minus the text, then when the hex value is applied the rectangle would be filled with this color while the text would be transparent. This differs from the look of the original logo.
2. Everywhere I see their logo in use, it seems the background color happens to be black, too. So I wonder whether the black rectangle in the logo is just an artifact that isn't actually part of the logo since it's invisible in the places where the logo is used.

foo-dogsquared:
Good looking out, @sisp. I also looked into their website and also reached the same conclusion of the text itself is the more suitable version. The closest thing I could find supporting this is the version text on the [documentation page](https://borgbackup.readthedocs.io/en/stable/) being in the same color as the logo.

sisp (approved):
Thanks for updating the icon.

LGTM now! :tada:

It's a perfect fit, there's no notable difference between the original and your icon:

PR #7551, opened 2022-06-28
Fix SVG linting errors

t-miller:
When I ran the linter in the development branch I found it difficult to look for a new contribution I was adding due to the number of other files that were throwing errors. All of them turned out to be extraneous whitespace/newline errors, so I removed the newline from each svg file that was throwing a linting error.


sisp:
That's odd. The "Lint" check also runs this command and there no errors are reported. I'll investigate it further. In any case, the errors should be fixed, but we also need to make sure the check captures them and prevents such errors in the future.

jack-mil:
I also these same errors when running `$> npm run lint`

sisp:
I just ran the same command locally and see no errors. :thinking:

```
> simple-icons@7.3.0 svglint
> svglint icons/*.svg --ci --config .svglintrc.mjs

------------------------------------------------------------------ Summary ------------------------------------------------------------------
✓ 2287 valid files.
```

But of course I believe you that you're seeing them. I just wonder why the result is not consistent.

sisp:
Without having any idea why it would make a difference, what OS are you using? I assume Windows because I'm seeing backslashes in the paths. The CI checks are run on Linux and I'm using Ubuntu locally.

sisp:
The `svglint` check is defined here:

https://github.com/simple-icons/simple-icons/blob/9b046563d4e73eca90e457d8cfed51a5620bfdea/.svglintrc.mjs#L807-L810

And the regex used for linting the SVG string is defined here:

https://github.com/simple-icons/simple-icons/blob/9b046563d4e73eca90e457d8cfed51a5620bfdea/.svglintrc.mjs#L27-L28

I imagine the problem is caused by the last bit of the regex `\n?$/` which might not match when you're on Windows and newlines are converted to `\r\n`.

It seems the icons you've listed for which the linting fails are those that have trailing newlines.

sisp:
I see two options:

1. We allow both `\n` and `\r\n` in the regex. This might lead to `\r\n` being committed, so we might end up with a mix of `\r\n` and `\n` which is not desirable IMO.
2. We enforce _no_ trailing newline (i.e. we remove the `\n` in the regex).

In any case, I think we should make sure the DX is consistent across all major platforms.

@simple-icons/maintainers What do you think?

sachinraja:
Doesn't `text=auto` in our `.gitattributes` help with line endings here? Maybe the linter will still fail before checkin though? In any case, I would agree that we should make it consistent so I'm fine with no line endings.

LitoMore:
See https://github.com/simple-icons/simple-icons/pull/6880#discussion_r754779314

I believe the `eol=lf` is necessary.

sisp:
@LitoMore I think you're right.

If I read the [Git documentation](https://git-scm.com/docs/gitattributes#Documentation/gitattributes.txt-Settostringvaluelf)

> #### Set to string value "lf"
> This setting forces Git to normalize line endings to LF on checkin **and prevents conversion to CRLF when the file is checked out**.

right, then
```diff
-*.svg              text
+*.svg              text eol=lf
```
in `.gitattributes` would force `\n` on check-out for SVGs even on Windows, so `svglint` checks should pass.

mondeja:
> in .gitattributes would force \n on check-out for SVGs even on Windows

As far as I know, that will only work for checked out files, but newly created files in Windows is possible that will still have crlf line endings.

> I imagine the problem is caused by the last bit of the regex \n?$/ which might not match when you're on Windows and newlines are converted to \r\n.

Note that we don't allow crlf in SVGs by design because is not crossplatform.

> We enforce no trailing newline (i.e. we remove the \n in the regex).

I agree with this option, is even defined [like that in *.editorconfig* file](https://github.com/simple-icons/simple-icons/blob/9b046563d4e73eca90e457d8cfed51a5620bfdea/.editorconfig#L13-L14).

sisp:
@mondeja Good point about the `.editorconfig` setting. :+1:

Then, enforcing no trailing newline seems the most consistent solution IMO. Would you also be okay with this change, @LitoMore?

LitoMore:
> Then, enforcing no trailing newline seems the most consistent solution IMO. Would you also be okay with this change? (@sisp)

Agree.

sisp:
Would you like to amend your PR accordingly, @t-miller?

t-miller:
> Would you like to amend your PR accordingly, @t-miller?

I think I made the change you were looking for

sisp:
Sorry for the possible misunderstanding. The solution we arrived at is to enforce _no_ trailing newlines in SVGs, i.e. remove `\n?` in the SVG regex in `.svglintrc.js` and remove trailing newlines in SVG files where they exist.

.gitattributes
```diff
@@ -1,7 +1,7 @@
 *                  text=auto
 
 # SVGs are treated as binary by default
-*.svg              text
+*.svg              text eol=lf
```
sisp, line 4: I think this change should be reverted again.

sisp (changes requested):

sisp (approved):
Thanks to all of you for the fruitful discussion! :tada:

PR #7567, opened 2022-07-05
Add Elementor Icon

simonfarah:
image

**Similarweb rank:**  14,871

### Checklist

### Description
Logo was retrieved from the [Elementor Logo Guidelines Page](https://elementor.com/logos/). Using Adobe Illustrator, the icon was resized to fit the required viewbox and then was optimized using SVGOMG


sisp (changes requested):
Thanks for your contribution, @simonfarah! :pray:

I see some notable differences between the original icon and yours:


Could you please check your conversion process again?

simonfarah:
Hello! Can I please know which logo source was your reference?

simonfarah:
okay thank you! Working on it...

simonfarah:
I can still see a very very tiny difference but don't actually know the reason, since the size editing is made while maintaining the original proportions.

sisp (changes requested):
Yes, I also see some differences still:


Your precision looks too low though, it seems to be just 1 decimal place whereas the contribution guidelines recommend 3. Would you mind checking the precision settings you've used for SVGO / SVGOMG?

sisp:
That's odd.

1. Can you reset the settings of SVGOMG (at e very bottom of the settings list)? The precision should be 3 then.
2. What precision does the output SVG of Adobe Illustrator have? Does it already have only a precision of 1 decimal place?

simonfarah:
I just reset the SVGOMG settings (precision to 3), then scaled the SVG in illustrator and exported it with a precision of 3, and got the same difference as [here](https://github.com/simple-icons/simple-icons/pull/7567#pullrequestreview-1031210401).

sisp:
Okay, last attempt, if that doesn't help, I can help you out. :wink:

If I understood you correctly, you exported the SVG from Adobe Illustrator with a precision of 3. If that's correct, could you scale the icon in Illustrator with the highest possible precision and let SVGOMG do the precision reduction? I've experienced it to be better to feed SVGOMG with a high-precision SVG and let it perform all optimizations.

simonfarah:
Well, this issue with scaling down the vector is still present even after trying all the ways and every single possibility of precision 😂😅. I really don't know why the proportions are changing those few millimeters.

This is the closest I could get (those few differences getting out of nowhere are driving me crazy!) :
image

I'd really would like you to try some stuff and tell me where this difference is coming from.

sisp (approved):
LGTM now! :tada:

Thanks for your patience and curiosity to figure out those differences, and of course for your contribution! :pray:

sisp:
Just an afterthought regarding those differences: If you're really curious, you could try another SVG editor. For instance, I use Boxy SVG for most conversions and sometimes Inkscape for complicated cases where Boxy lacks some features or behaves strange. I couldn't imagine why another editor should make any difference in a simple scaling conversion, but who knows ... :laughing:

simonfarah:
Hahaha yess!! I'll be definitely trying another editor. Thanks ❤️!

PR #7584, opened 2022-07-12
Add React Hook Form icon

jsun969:
reacthookform

### Checklist

### Description
React Hook Form is a popular form lib for react to improve UX.
Github: https://github.com/react-hook-form/react-hook-form (over 29k stars)
Website: https://react-hook-form.com/

I am the author of react hook form devtools browser extension


bluebill1049:
thanks for adding this icon. 👍

sisp (commented):
Thanks for your contribution, @jsun969! :pray:

There's a small but notable difference at the ring at the top of the icon:


With a precision of 3 decimal places, there were significantly more differences because of the several arcs which are often difficult to draw accurately with reduced precision. Now with a precision of 4 decimal places, only the ring at the top is a bit off though not by much. @sachinraja Do you think this is sufficiently close to the original icon?

sachinraja:
Yes I think it's fine if it's difficult to fix.

jsun969:
> Thanks for your contribution, @jsun969! 🙏
> 
> There's a small but notable difference at the ring at the top of the icon: 
> 
> With a precision of 3 decimal places, there were significantly more differences because of the several arcs which are often difficult to draw accurately with reduced precision. Now with a precision of 4 decimal places, only the ring at the top is a bit off though not by much. @sachinraja Do you think this is sufficiently close to the original icon?

I think it's quite difficult to fix. That's why I update the icon twice to pass lint🤔

jsun969:
thx a lot 🎉 @sisp

PR #7612, opened 2022-07-20
Add Medusa Icon

florian-lefebvre:
medusa

**Similarweb rank:** 275,438

### Checklist

#### Why did you pick the hex value?

The icon is made of a linear gradient going from `#592EE1` (bottom-left) `#B836D9` to (top-right). I chose the 2nd color as it seemed to me that it's better representative of the brand identity, but I'm probably influenced by the older brand identity. I'd like opinion of @srindom and @olivermrbl on this.


florian-lefebvre:
Concerning the following issue:
```bash
  x icon-size Size of <path> must be exactly 24 in one dimension; the size is c
              urrently 22.264 x 24.001
```
The original icon is not square, so I centered it vertically in Figma so that might cause the issue but I don't know what I can do about it.

florian-lefebvre:
Concerning the color, I think `#8B5CF6` would be better after all (https://github.com/medusajs/admin/blob/master/tailwind.config.js#L34).
It would look like:
medusa (1)

adamrusted:
Closing in favour of #8324

PR #7635, opened 2022-07-30
Add `Avira` icon

halacoglu:
avira

Avira Operations GmbH & Co. KG is a German multinational [computer security software](https://en.wikipedia.org/wiki/Computer_security_software) company mainly known for their [antivirus software](https://en.wikipedia.org/wiki/Antivirus_software) Avira Free Security (formerly known as Avira Free Antivirus and Avira AntiVir).

**Similarweb rank:** 12,762

### Checklist

### Description
- I retrieved the logo from the [Media Library](https://www.avira.com/en/media-library#our-logos) of the official Avira webpage. Then I edited the logo on Adobe Illustrator according to Simple Icons requirements.
- Choosing the HEX value was a bit difficult. Unfortunately, there isn't an official HEX value provided by Avira. The PNG logo and the red color used in the guideline PDF is `#DB0D15`. The CMYK.eps is `#D60325`, the RGB.eps is `#D60627` and the website logo is `#E11F27`. I chose `#DB0D15`.


_data/simple-icons.json
```diff
@@ -1181,6 +1181,12 @@
             "source": "https://www.avast.com/",
             "guidelines": "https://press.avast.com/media-materials"
         },
+        {
+            "title": "Avira",
+            "hex": "DB0D15",
+            "source": "https://www.avira.com/en",
```
sisp, line 1187: I think the source URL should be https://www.avira.com/en/media-library#our-logos as well because that's where the ZIP archive containing the logos can be downloaded.

```suggestion
            "source": "https://www.avira.com/en/media-library#our-logos",
```

sisp (changes requested):
Thanks for your contribution, @halacoglu! :pray:

I see some small differences between the original and your icon at the handle of the umbrella:



Could you please try to see whether it's possible to remove the differences?

By the way, I compared your icon with `logo_avira_horizontal_FBlack_RGB.eps`. Which file did you use for the derivation? Just in case they are not 100% identical ... (I haven't checked)

PratyushJayachandran (approved):
Thank you for the contribution @halacoglu and correction @adamrusted 🎉

PR #7706, opened 2022-08-27
Add Internet Computer icon

rvanasa:
Closes: #7721

internetcomputer

**Similarweb rank:**
[181,347](https://www.similarweb.com/website/dfinity.org/#overview)

### Checklist
  - [x] I optimized the icon with SVGO

### Description

I am a software engineer at [DFINITY](https://dfinity.org/) (a not-for-profit foundation building the [Internet Computer](https://internetcomputer.org/)). 

The primary goal of adding the Internet Computer icon is to make it possible for developers to use our logo in [Shields.io](https://shields.io/) status badges on GitHub repositories. 

The hex value is the primary color used on [internetcomputer.org](https://internetcomputer.org).


_data/simple-icons.json
```diff
@@ -5100,6 +5100,18 @@
             "hex": "1FB141",
             "source": "https://icons8.com/"
         },
+        {
+            "title": "ICP",
```
LitoMore, line 5104: There are a lot of products and organizations on the internet called ICP. I recommend using `Internet Computer Protocol` instead.

rvanasa: Renamed to `Internet Computer` (for consistency with [internetcomputer.org](https://internetcomputer.org/)). 

Thanks for taking a look at this PR!

LitoMore: @simple-icons/icon-review Any thoughts on this icon name and its aliases? Can we use `ICP` as an alias with this icon?

LitoMore:
@simple-icons/icon-review Any thoughts on this icon name and its aliases? Can we use `ICP` as an alias with this icon?

sachinraja:
I would say no because it's not a well known alias.

rvanasa:
Sounds reasonable to me. I went ahead and removed the ICP alias.

LitoMore:
CleanShot 2022-09-20 at 19 49 13@2x@rvanasa According to their [logo use guidelines](https://dfinity.frontify.com/d/XzkdhhDptijE/dfinity-brand-guide#/dfinity/logo). We should the black & white version as reference.

rvanasa:
This proposed icon is consistent with the monochromatic usage on [internetcomputer.org](https://internetcomputer.org/):

image

Since I'm an employee at Dfinity (rather than a third-party contributor), I hope this also gives a bit of extra credence to this request. At some point, we will likely update the usage guidelines to include this specific icon as a valid usage of the logo.

LitoMore:
@rvanasa OK. Please ping me once the guideline is updated.

LitoMore:
Hey @rvanasa, I think we could move forward.

Let's use the current black & white version of the brand guideline.

You could create another PR to update the icon when your proposal is accepted by your design team.

rvanasa:
The [Dfinity brand guide](https://dfinity.frontify.com/d/XzkdhhDptijE/dfinity-brand-guide#/internet-computer/logo) has been updated to include the proposed design. Cheers!

LitoMore:
@rvanasa TBH, I'd prefer the Definity one. The Internet Computer one not looks not good in a badge. Because its gap is too thin. We should take into account how icons look in various sizes.

I also have questions about this design. Why the black and white icons for Definity and Internet Computer are different, I think they should be the same.

What do you think?

rvanasa:
@LitoMore, @sachinraja, could we proceed with the current design for the time being? Many other icons in this pack use thin lines in a similar way (Google Calendar, Google Chat, Apache, Netflix, Jamboard, DVC, Cinema4D, etc.), so it doesn't seem like this would be an issue. 

Also, for clarification (since it's somewhat ambiguous in the usage guidelines), the "DFINITY" logo is the same as the "Internet Computer" logo aside from the text. The dichromatic black & white design would require two colors in the logo, whereas the current version is designed specifically for Shields.io badges (which was the original reason for opening this PR). 

Thanks, and happy holidays!

rvanasa:
Thank you so much for the reply! I've sent your suggestion to the design team for review.

Here is the generated preview for reference:

internetcomputer (1)

In my opinion, the smaller icons (which are the priority for supporting Shields.io badges) look okay but aren't as recognizable as the icon style used on our website (see below). That being said, I'll defer to the opinion of the design team, which will update the usage guidelines if needed. 

For additional context, here is the more widely adopted icon style from [internetcomputer.org](https://internetcomputer.org/):

image

Here is another embedded icon usage with the same style:

image

Thanks again for getting back to us, and I will update this PR once we make a decision.

rvanasa:
After talking with the design team, we decided to use the same icon as on [internetcomputer.org](https://internetcomputer.org/) (without the original thin lines). I've updated the PR, and here is the corresponding preview image:

internetcomputer

I've also asked the team to include this variation in the official style guide, although I hope it's clear that this is a legitimate usage of the logo given its prevalence on the official Internet Computer website. 

Cheers!

rvanasa:
@adamrusted, just quickly checking back in. Does this work for you?

adamrusted:
Should be fine, thanks @rvanasa!
Will ping other @simple-icons/icon-review folks who may be able to get this one closed off before me!

icons/internetcomputer.svg
PratyushJayachandran, line 1: I see [some differences](https://github.com/simple-icons/simple-icons/assets/35000807/162569ce-4d8e-4416-b5bf-b5fdedfdd527) when comparing with blacked out logo available on the website header, with underlying red one being the contributed file.

PratyushJayachandran (changes requested):
I have suggested changes to the links.
There is also some variation when comparing with the colored icon available on the website. I made it monochromatic before comparing.
Please provide the correct source in case I am comparing with the wrong one.

icons/internetcomputer.svg
adamrusted: With regard to this, it's not a 1:1 comparison between what is on the website, and what has been submitted (see PR history). I personally would be more than happy to accept this, given it's from the brand themselves.

PratyushJayachandran: Okay then. But can we update source and guidelines Links at least?

adamrusted: Yeah, @rvanasa - are you OK to double check the urls as suggested by @PratyushJayachandran?

rvanasa: Updated; thank you!

PratyushJayachandran (approved):
LGTM, thank you @rvanasa :tada:

PR #7755, opened 2022-09-08
Add Tamiya icon

LitoMore:
CleanShot 2022-09-11 at 18 02 08@2x

**Similarweb rank:** [52,531](https://www.similarweb.com/website/tamiya.com/#overview)
**Logo resource:** https://commons.wikimedia.org/wiki/File:TAMIYA_Logo.svg
**Color:** `#000000`
**Product reference:** https://www.tamiyausa.com/shop/signature-line/tamiya-logo-sticker-white/

### Checklist

### Description

They don't have any black and white SVG resource on internet. I processed a black and white one based on a colored version.

I also added a gap between the two stars. Because their black and white version always has a gap. BTW, they don't have any standard about the width of the gap. I bought a lot of Tamiya products, the width of each gap is not the same. Here is my Tamiya products list: https://github.com/LitoMore/tamiya-toys.

### Difference preview

No needed.


dirien (commented):
Hi @LitoMore,

love it!

I found this icon on the German site of Tamiya -> https://www.tamiya.de/build/img/svg/tamiya_logo.svg?v=2.6.0c

Maybe we could use consider this, as it is some kind of official?

LitoMore:
@dirien The resource from https://tamiya.de looks not good. The height of those letters is not the same.

CleanShot 2022-09-10 at 20 56 36@2x

Here are some of my Tamiya products:

| [item-95514](https://www.tamiya.com/english/products/95514/index.htm) | [item-94893](https://www.tamiya.com/english/products/94893_oval_circuit/index.htm) | [item-74039](https://www.tamiya.com/english/products/74039/index.htm) |
|:-:|:-:|:-:|
||||

As you can see, the letters in all those logs are in the same baseline.

So I would prefer to use the resource from Wikimedia.

LitoMore:
CleanShot 2022-09-08 at 21 24 47@2xBTW, I think we should increase the gap.

Although we have documented that the default gap is 0.5px, I think the gap between the two stars is a major gap. We should increase its width for a better view.

I also discussed this topic with another project owner https://github.com/xojs/xo/issues/689. Their logo is a typical case.

dirien:
I like it! @LitoMore! Regarding the gap with: I trust you in this!

LitoMore:
I decided to remove the text `TAMIYA` from the logo. I have some logo stickers which bundled in the Mini 4WD products, those stickers only have two stars. And my [t-shirt](https://www.tamiya.com/english/products/67235_67238/index.htm) has no text `TAMIYA` under the logo either.

Currently the gap in shields may look very thin. But I have another approach with shields.io. See https://github.com/badges/shields/issues/8372.

dirien:
You let me know, when it's ready to review!

LitoMore:
@dirien Ready to review now~

dirien (approved):
Hi @LitoMore 

for me it looks fine, the stars are all shaped the same. As there is no real official svg with the gap etc. I am fine from my side with your suggestion.

image

LitoMore:
Merging. Please let me know if you want any further changes.

PR #7774, opened 2022-09-13
Add WezTerm icon

LitoMore:
| Option 1 | Option 2 |
|:-:|:-:|
|||

**GitHub stars:** https://github.com/wez/wezterm - 5.7k
**Resource:** https://github.com/wez/wezterm/blob/fe78b5821fc106c1061f4c1cc454ff01e74bf97d/assets/icon/wezterm-icon.svg
**Color:** `#4E49EE` Picked from the resource

### Checklist


LitoMore:
@wez Which one do you prefer?

wez:
Option 1 looks more obviously like the logo as it has the outline that prevents it from looking like a random `$W`, so I'm leaning more towards that, but I'm honestly on the fence.

I think you should also be aware of https://github.com/wez/wezterm/discussions/2396 which is a more complex icon proposed by a member of the community. I haven't made any decisions about that yet.

LitoMore:
@wez I'd prefer the option 2. In most cases the icon in Simple Icons is used for shields.io.

The option 2 looks better in a badge.

For https://github.com/wez/wezterm/discussions/2396, those app icons are not the flat design and in vector format. We should use the current SVG one as reference.

adamrusted:
Marking as 'in discussion' as the discussion over on WezTerm's own repo is still ongoing.

PeterShaggyNoble:
As discussion on a potential new icon has been going on for over a year now and there's no guarantee it will be adopted, I think we should run with the current icon and update in the future if necessary.

Of the 2 options presented, my vote would would go to the first, with the background as that's the option that was chosen by @wez and it's what's used in [the website](https://wezfurlong.org/wezterm/)'s header;  I don't see the `$W` in use on its own anywhere without the background.

From the website as well it looks like `#7e56c2` could be another option for the colour - @wez; do you have a preference between the 2?

wez:
Proceeding with the `$W` here sounds good to me.  I do tend to prefer purple over blue.  The current icon has a blue-er color for contrast reasons, even though its contrast is admittedly not ideal.  I'd be happy with either or both.

PeterShaggyNoble:
Thanks for the feedback, @wez 👍🏻  Based on that, I think we can run with everything as-is, and we can easily update the icon and/or colour in the future if you develop a stronger preference for one over the other.

@LitoMore, I'm seeing a souple of _hairline_ thin differences in the 2 inner curves of the `$` (original in black below, yours in red), could you have a look at those, please?

LitoMore:
These differences are pretty slight. We can ignore it.

What do you think? @simple-icons/icon-review

image

PratyushJayachandran:
> These differences are pretty slight. We can ignore it.
> 
> What do you think? @simple-icons/icon-review

Looks pretty spot on to me. Outline view with source superimposed on contribution:
image

PratyushJayachandran (approved):

PR #7839, opened 2022-09-21
Add Electron Fiddle icon

LitoMore:
CleanShot 2022-09-21 at 16 03 50@2x

**GitHub stars:** 6.7k https://github.com/electron/fiddle
**Resource:** https://github.com/electron/fiddle/blob/19360ade76354240630e5660469b082128e1e57e/assets/icons/fiddle.svg
**Color:** `#E79537` from its circle background

I removed its circle background because it's not the main part of the logo. It's app icon has a background shape but not a circle. So we don't have to keep the circle background for Simple Icons.

### Checklist

### Difference preview

CleanShot 2022-09-21 at 16 03 13@2x


dirien:
Hi @LitoMore,

are you approving your own PRs?

FYI @simple-icons/maintainers

PR #7849, opened 2022-09-23
Add `add-icon-data` script

### LINKED ISSUE #7848: Add a script for filling the JSON file
Contributes to https://github.com/simple-icons/simple-icons/pull/7846#issuecomment-1255393569.

### Proposal

Use [`Inqurier.js`](https://github.com/SBoudrias/Inqurier.js) to implement our CLI prompt.

The prompt will look like this:

```
npm run add-icon

# Show a command line prompt
? Title:
> Micro Focus

? Source:
> https://...

? Color:
> ABCDEF

? Alias (optional)
...
...
```

Code example:

```js
import inquirer from 'inquirer';
import chalkPipe from 'chalk-pipe';
import { getIconsDataString } from './utils.js';

const prompts = [
  {
    type: 'input',
    name: 'title',
    message: 'Title',
  },
  {
    type: 'input',
    name: 'hex',
    message: 'HEX color',
    validate: (text) => /^[a-fA-F0-9]{6}$/.test(text.replace(/^#/, '')),
    transformer: (text) => {
      return chalkPipe(text.startsWith('#') ? text : `#${text}`)(text);
    },
  },
  {
    type: 'input',
    name: 'source',
    message: 'source',
  },
];

const icon = await inquirer.prompt(prompts);
const iconsData = JSON.parse(await getIconsDataString());

// ... sorting logic
```

### Questions

I have no idea how to design the UX for [aliases](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#aliases). Any thoughts?

LitoMore:
Resolves #7848 

~~I used a simple `localeCompare` on titles. So it's valid when we add `simple icons` (the lowercase). I may need to check the duplication of their slugs. Seems `our-lint` has the same issue.~~

#### Command

```sh
npm run add-icon-data
```

#### Prompts

| field | validation |
| :-- | :-- |
| `title` | `Not empty` `The title should be unique` `Its slug should be unique`  | 
| `hex` | `/^#?[a-f0-9]{3,8}$/i` |
| `source` | `/^https:\/\/[^\s]+$/` |

The confirm prompt is inspired by `npm init` command.

### Preview

CleanShot 2022-09-24 at 23 54 59@2x


LitoMore:
I don't have a good idea to implement the UX for [aliases](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md?rgh-link-date=2022-09-23T00%3A54%3A30Z#aliases). Any ideas?

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
+import fs from 'node:fs/promises';
+import inquirer from 'inquirer';
+import chalkPipe from 'chalk-pipe';
```
mondeja, line 3: Why are you using `chalk-pipe` instead of `chalk` directly? `chalk` is a direct dependence of `inquirer`, could we avoid the addition of `chalk-pipe` to our development dependencies?

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+    name: 'source',
+    message: 'Source',
+    validate: (text) =>
+      Boolean(text.startsWith('https://') || text.startsWith('http://'))
+        ? true
+        : 'This should be a URL',
+  },
+];
+
+const icon = await inquirer.prompt(prompts);
```
mondeja, line 51: After this, the icon will be added without a confirmation which is a destructive change. What about a final `confirm` question?

```suggestion
  },
  {
    type: 'confirm',
    message: (answers) => {
      return `The filename of the icon would be`
           + ` ${chalkPipe('blue')(titleToSlug(answers.title) + '.svg')}\n`
           + 'Add you sure that you want to add it to the data file?';
    },
    name: 'confirmation',
  },
];

const icon = await inquirer.prompt(prompts);
if (!icon.confirmation) {
  console.log(logSymbols.error, 'Icon not added');
  process.exit(1);
}
```

And we can remove the part `your icon filename should be: ...` of the final messsage.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+iconsData.icons.push({
+  title: icon.title,
+  hex: icon.hex.replace(/^#/, '').toUpperCase(),
+  source: icon.source,
+});
+iconsData.icons.sort((a, b) => a.title.localeCompare(b.title));
+
+await fs.writeFile(
+  getIconDataPath(),
+  JSON.stringify(iconsData, null, 4) + '\n',
```
mondeja, line 62: This is not consistent with our JSON prettification linter. Please, mimic the implementation already defined in *scripts/lint/ourlint.js*: https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/scripts/lint/ourlint.js#L50-L51

Newlines must be normalized because Windows users could read the data file using `\r\n` newlines.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+  {
+    type: 'input',
+    name: 'hex',
+    message: 'Hex',
+    validate: (text) =>
+      hexPattern.test(text) ? true : 'This should be a 6-digit hex code',
+    transformer: (text) => {
+      const color = text.startsWith('#') ? text : `#${text}`;
+      const luminance = hexPattern.test(text)
```
mondeja, line 31: Why is the `hex` validated again? The `transformer` will not be executed if the validation fails, this is redundant.

mondeja (changes requested):
Thanks for working on this 👍🏼 

I think that the current implementation onlly solves the problem of the ordering in the data file. It does not validate so many things. Additionally, `add-icon` is really confusing, should be renamed to `add-icon-data` as the script is not adding an icon, only their object to the data file?

Also, I think that we should standarize the assertions done to our data in tests and reuse them in other parts of the project like this script, but that could be in a separate PR.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+import { getIconsDataString, getIconDataPath, titleToSlug } from './utils.js';
+
+const hexPattern = /^#?[a-fA-F0-9]{6}$/;
+const iconsData = JSON.parse(await getIconsDataString());
+
+const prompts = [
+  {
+    type: 'input',
+    name: 'title',
+    validate: (text) => {
```
mondeja, line 15: What happens if a title generates an existent icon slug? That must be validated.

mondeja (changes requested):

mondeja (changes requested):
As you're adding a terminal color formatting library we can replace the unicode characters used in *[scripts/lint/ourlint.js](https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/scripts/lint/ourlint.js#L72)*:

https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/scripts/lint/ourlint.js#L72

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+        ? true
+        : 'This should be a URL',
+  },
+];
+
+const icon = await inquirer.prompt(prompts);
+
+iconsData.icons.push({
+  title: icon.title,
+  hex: icon.hex.replace(/^#/, '').toUpperCase(),
```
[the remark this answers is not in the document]
LitoMore: The transformer is only for prompt display. The transformer's return value will not affect the result of `await inquirer.prompt(prompts)`.

mondeja:
Related: #7851

scripts/add-icon.js
LitoMore: `chalk-pipe` can automatically recognize style strings. It is designed for those who need to use mutable styles. They can switch to another color scheme with a simple string: like from `blue.bold.underline` to `bg#ff99cc.yellow`.

I think we might need more features. But in the current case, the `chalk` is enough to use.

Will use `chalk` instead.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+
+await fs.writeFile(
+  getIconDataPath(),
+  JSON.stringify(iconsData, null, 4) + '\n',
+  'utf8',
+);
+
+console.log(
+  logSymbols.success,
+  `Icon added, your icon filename should be: ${chalkPipe('blue')(
```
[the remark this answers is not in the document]
LitoMore: We could validate the filename in the inquirer's validator.

But some users might process the SVG file before adding metadata. I think validate the duplication of title is enough.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+      return true;
+    },
+    message: 'Title',
+  },
+  {
+    type: 'input',
+    name: 'hex',
+    message: 'Hex',
+    validate: (text) =>
+      hexPattern.test(text) ? true : 'This should be a 6-digit hex code',
```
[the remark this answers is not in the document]
LitoMore: Sounds good.

scripts/add-icon.js
```diff
@@ -0,0 +1,71 @@
...
+        text,
+      );
+    },
+  },
+  {
+    type: 'input',
+    name: 'source',
+    message: 'Source',
+    validate: (text) =>
+      Boolean(text.startsWith('https://') || text.startsWith('http://'))
```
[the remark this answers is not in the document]
LitoMore: Yes. I just want to prevent contributors submit `the source is on their press page`-like stuff.

scripts/add-icon.js
LitoMore: Because the `getRelativeLuminance` will throw an error if the value is not valid. 🤷‍♂️

scripts/add-icon.js
LitoMore: Yep. Will use the same logic as `our-lint`.

scripts/add-icon.js
LitoMore: Sounds good. The final confirm question is a good idea.

> And we can remove the part your icon filename should be: ... of the final message.

No problem.

scripts/add-icon.js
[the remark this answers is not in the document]
mondeja: Ah, thanks for clarify it 😅

scripts/add-icon.js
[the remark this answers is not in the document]
mondeja: See https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/.jsonschema.json#L235

scripts/add-icon.js
LitoMore: `JSON.stringify(data, null, 4)` and `JSON.stringify(data, null, '    ')` are the same. And I have a `\n` at the end of the file.

`JSON.stringify`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

But I change my code to `${JSON.stringify(data, null, '    ')}\n` for making our code consistent.

scripts/add-icon.js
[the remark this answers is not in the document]
LitoMore: Thanks. Will use this pattern instead.

scripts/add-icon.js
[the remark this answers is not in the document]
mondeja: In that case, we can skip the self icon slug if doesn't fit with any other object in the data file.

scripts/add-icon.js
mondeja: Maybe would be worth to add a new utility function `writeIconsData` to normalize all calls.

scripts/add-icon.js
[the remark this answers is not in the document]
LitoMore: Good point. The filter will be a good solution here.

LitoMore:
Waiting for #7854 (the locale comparison) and #7855 (the URL regex) to merge.

scripts/add-icon-data.js
```diff
@@ -0,0 +1,87 @@
...
+  {
+    type: 'confirm',
+    name: 'confirm',
+    message: (answers) =>
+      [
+        'About to write to simple-icons.json',
+        chalk.reset(JSON.stringify(answers, null, 4)),
+        chalk.reset('Is this OK?'),
+      ].join('\n\n'),
+  },
```
mondeja, line 75: The `guidelines` field can be added using `when` and filtering the final answers:

```suggestion
  {
    type: 'confirm',
    name: 'add-guidelines',
    message: 'The icon has brand guidelines?',
  },
  {
    type: 'input',
    name: 'guidelines',
    message: 'Guidelines',
    validate: sourceValidator,
    when: (answers) => answers['add-guidelines'],
  },
  {
    type: 'confirm',
    name: 'confirm',
    message: (answers) => {
      answers = Object.assign(
        ...Object.keys(answers)
          .filter((key) => !key.includes('-'))
          .map((key) => {
            return { [key]: answers[key] };
          }),
      );
      return [
        'About to write to simple-icons.json',
        chalk.reset(JSON.stringify(answers, null, 4)),
        chalk.reset('Is this OK?'),
      ].join('\n\n');
    },
```

mondeja (changes requested):

mondeja:
Are you planning to add the rest of the fields @LitoMore?

LitoMore:
> Are you planning to add the rest of the fields? (@mondeja)

Let me try.

scripts/utils.js
```diff
@@ -116,6 +123,20 @@ export const getIconsData = async (rootDir) => {
   return JSON.parse(fileContents).icons;
 };
 
+/**
+ * Write icons data to _data/simple-icons.json.
+ * @param {Object} iconsData Icons data object.
+ * @param {String|undefined} rootDir Path to the root directory of the project.
+ *
```
mondeja, line 130: Just a stylistic minor change:

```suggestion
```

mondeja (changes requested):

scripts/add-icon-data.js
```diff
@@ -0,0 +1,137 @@
...
+        chalk.reset('Is this OK?'),
+      ].join('\n\n');
+    },
+  },
+];
+
+const answers = await inquirer.prompt(dataPrompt);
+const icon = getIconDataFromAnswers(answers);
+
+if (confirm) {
```
mondeja, line 130:
```suggestion
if (answers.confirm) {
```

mondeja (approved):
Done a couple of changes. Now looks good to me 🌠

Can be improved in the future though, feel free to merge it!

PR #7852, opened 2022-09-23
Build icons in order

LitoMore:
### Background

I noticed the build process of https://github.com/icons-pack/react-simple-icons is based on our exported module. They are using `Object.keys(simpleIcons)` stuff to map our icons. The order in which they are built depends on our order.

Every time they bump the simple-icons version, their `index.ts` file changes unpredictably. For example [this diff](https://github.com/icons-pack/react-simple-icons/commit/8a10296e3f4a889f3eba0abcb7998541698a8cfd#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80) (you may need to click `Load diff` since the diff is too large).

### Before

We are using `.push` inside the `Promise.all`. So we cannot guarantee that its order is identical to the JSON file.

https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/scripts/build/package.js#L89-L96

### After

I moved the `.push` logic outside the `Promise.all`. Now the order looks good :)

### DIFF
--- a/scripts/build/package.js
+++ b/scripts/build/package.js
@@ -81,30 +81,28 @@ const build = async () => {
   };
 
   // 'main'
-  const iconsBarrelMjs = [];
-  const iconsBarrelJs = [];
-  const iconsBarrelDts = [];
-  const buildIcons = [];
-
-  await Promise.all(
+  const buildIcons = await Promise.all(
     icons.map(async (icon) => {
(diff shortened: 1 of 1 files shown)


mondeja (approved):
This is a great catch, thank you @LitoMore :100:

PR #7854, opened 2022-09-24
Standarize locale comparisons and sort icons by title in package

### LINKED ISSUE #7851: Configure region in locales comparisons
We are using `String.prototype.localeCompare` to sort icons in *scripts/lint/ourlint.js* and *.svglintrc.mjs*:

https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/scripts/lint/ourlint.js#L21-L26

https://github.com/simple-icons/simple-icons/blob/d5444f699528a456ee2894751994b4eef7274235/.svglintrc.mjs#L49

According to the [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare), we are not specifying the locale used to compare, so the execution could be different depending on the implementation. Also, `String.prototype.localeCompare` creates a new [`Intl.Collator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) object, which can be avoided just creating the object before and using it to compare.

### Contributing

Yes

mondeja:
Resolves #7851 and contributes to an improvement over #7852

### DIFF
--- a/.svglintrc.mjs
+++ b/.svglintrc.mjs
@@ -47,7 +47,7 @@ const sortObjectByKey = (obj) => {
 
 const sortObjectByValue = (obj) => {
   return Object.keys(obj)
-    .sort((a, b) => collator.compare(`${obj[a]}`, obj[b]))
+    .sort((a, b) => collator.compare(obj[a], obj[b]))
     .reduce((r, k) => Object.assign(r, { [k]: obj[k] }), {});
 };
 
--- a/.svglintrc.mjs
+++ b/.svglintrc.mjs
@@ -3,6 +3,7 @@ import path from 'node:path';
 import {
   getDirnameFromImportMeta,
   htmlFriendlyToTitle,
+  collator,
 } from './scripts/utils.js';
 import svgpath from 'svgpath';
 import svgPathBbox from 'svg-path-bbox';
@@ -46,7 +47,7 @@ const sortObjectByKey = (obj) => {
 
 const sortObjectByValue = (obj) => {
   return Object.keys(obj)
-    .sort((a, b) => ('' + obj[a]).localeCompare(obj[b]))
+    .sort((a, b) => collator.compare(`${obj[a]}`, obj[b]))
     .reduce((r, k) => Object.assign(r, { [k]: obj[k] }), {});
 };
 
--- a/scripts/build/package.js
+++ b/scripts/build/package.js
@@ -99,9 +99,7 @@ const build = async () => {
   const iconsBarrelJs = [];
   const iconsBarrelMjs = [];
 
-  buildIcons.sort((a, b) => {
-    return collator.compare(a.icon.title, b.icon.title);
-  });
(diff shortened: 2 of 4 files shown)


.svglintrc.mjs
```diff
@@ -46,7 +47,7 @@ const sortObjectByKey = (obj) => {
 
 const sortObjectByValue = (obj) => {
   return Object.keys(obj)
-    .sort((a, b) => ('' + obj[a]).localeCompare(obj[b]))
+    .sort((a, b) => collator.compare(`${obj[a]}`, obj[b]))
```
LitoMore, line 50:
```suggestion
    .sort((a, b) => collator.compare(obj[a], obj[b]))
```

mondeja: It seems that is not needed, are always all strings.

LitoMore:
What if two brands have different titles but the same slug?

mondeja:
The next error will raise at build time:

<details>
  <summary>Error</summary>

```
$ npm run build

> simple-icons@7.12.0 build
> node scripts/build/package.js

/home/mondeja/files/code/simple-icons/node_modules/esbuild/lib/main.js:1624
  let error = new Error(`${text}${summary}`);
              ^

Error: Transform failed with 2 errors:
<stdin>:10789:13: ERROR: Multiple exports with the same name "siHandshake_protocol"
<stdin>:10789:13: ERROR: The symbol "siHandshake_protocol" has already been declared
    ... <TRACEBACK> ... {
  errors: [
    {
      detail: undefined,
      id: '',
      location: {
        column: 13,
        file: '<stdin>',
        length: 20,
        line: 10789,
        lineText: 'export const siHandshake_protocol={',
        namespace: '',
        suggestion: ''
      },
      notes: [
        {
          location: {
            column: 13,
            file: '<stdin>',
            length: 20,
            line: 10777,
            lineText: 'export const siHandshake_protocol={',
            namespace: '',
            suggestion: ''
          },
          text: 'The name "siHandshake_protocol" was originally exported here:'
        }
      ],
      pluginName: '',
      text: 'Multiple exports with the same name "siHandshake_protocol"'
    },
    {
      detail: undefined,
      id: '',
      location: {
        column: 13,
        file: '<stdin>',
        length: 20,
        line: 10789,
        lineText: 'export const siHandshake_protocol={',
        namespace: '',
        suggestion: ''
      },
      notes: [
        {
          location: {
            column: 13,
            file: '<stdin>',
            length: 20,
            line: 10777,
            lineText: 'export const siHandshake_protocol={',
            namespace: '',
            suggestion: ''
          },
          text: 'The symbol "siHandshake_protocol" was originally declared here:'
        }
      ],
      pluginName: '',
      text: 'The symbol "siHandshake_protocol" has already been declared'
    }
  ],
  warnings: []
}
```

</details>

LitoMore (approved):

PR #7855, opened 2022-09-24
Forbid the usage of non secure URLs

### LINKED ISSUE #7850: Should we allow non-secure `http://` sources?
### Description

There are currently 68 icons with HTTP sources. It seems like a risk for our users, should we enforce the usage of secured sources?

### Contributing

Yes

mondeja:
Changed all URLs to use `https://` and tests to forbid `http://` protocol in URLs. If a URL is not properly secured with a certificate, a screen like the next will appear in the browser:

image

Closes #7850

### DIFF
--- a/.jsonschema.json
+++ b/.jsonschema.json
@@ -231,6 +231,7 @@
     },
     "url": {
       "$id": "#url",
+      "description": "HTTPS-only URL for a source",
       "type": "string",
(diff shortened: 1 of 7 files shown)


LitoMore:
Do we need to document this rule in contributing doc?

LitoMore (approved):

PR #7863, opened 2022-09-25
Add `aliases` support for `add-icon-data` script

LitoMore:
Related to #7849 

The `aliases` field is more complicated. We may need a bunch of questions to collect users' inputs.

Here is the type overview:

```ts
type Aliases = {
	aka?: string[];
	dup?: SimpleIcon[];
	loc?: Record<langageCode, string>;
	old?: string[];
}
```

Maybe we could add a tip to contributing doc to let users modify the `_icons/simple-icons.json` directly.


LitoMore:
@mondeja Any thoughts about this? If we need to add aliases support, we will need to add a lot of prompts.

scripts/utils.js
```diff
@@ -6,6 +6,7 @@
 import path from 'node:path';
 import fs from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
+import { readFile } from 'node:fs';
```
mondeja, line 9: This import is not used 
```suggestion
```

scripts/add-icon-data.js
```diff
@@ -127,11 +178,12 @@ const dataPrompt = [
 const answers = await inquirer.prompt(dataPrompt);
 const icon = getIconDataFromAnswers(answers);
 
-if (answers.confirm) {
+if (answers.confirmToAdd) {
   iconsData.icons.push(icon);
   iconsData.icons.sort((a, b) => collator.compare(a.title, b.title));
   await writeIconsData(iconsData);
 } else {
+  console.log(answers);
```
mondeja, line 186: This should be removed
```suggestion
```

mondeja (changes requested):

mondeja:
It works nice 🌟  Seems like adding the `loc` field is not too much useful for the script, maybe in the future.

Feel free to merge it after resolve the suggestions 🔧

PR #7869, opened 2022-09-26
Add XO linter

### LINKED ISSUE #7870: Add XO linter
[XO code style](https://github.com/xojs/xo)

Add linting with linter at CI and pre-commit hook to prevent bugs.

### Why [XO](https://github.com/xojs/xo)?

XO has built-in support for TypeScript and Prettier. And it's designed for ESM projects.

For more features, you can find them at https://github.com/xojs/xo#highlights.

I already have a PR for this #7869

LitoMore:
[XO code style](https://github.com/xojs/xo)

Resolves #7870 

### Why [XO](https://github.com/xojs/xo)?

XO has built-in support for TypeScript and Prettier. And it's designed for ESM projects.

For more features, you can find them at https://github.com/xojs/xo#highlights.

### Linting errors

<details>
<summary>Click here to see linting errors</summary>

[xo-summary](https://github.com/LitoMore/xo-summary) is a tool for displaying output from `xo` as a list of style errors, ordered by count.

```
❯ npm i -g xo-summary
❯ npx xo | xo-summary
174    Block-scoped variables in strict mode are not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
92     Arrow functions are not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
41     Template literals are not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
21     Trailing commas in function syntax are not supported until Node.js 8.0.0. The configured version range is >=0.12.18.
16     Async functions are not supported until Node.js 7.6.0. The configured version range is >=0.12.18.
13     Unexpected use of the global variable process. Use require("process") instead.
11     Expected blank line before this statement.
9      Destructuring is not supported until Node.js 6.0.0. The configured version range is >=0.12.18.
8      Comments should not begin with a lowercase character.
6      Use for…of instead of Array#forEach(…).
4      Replace ·promises·as·fs· with promises·as·fs
4      Property shorthands are not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
3      Do not access a member directly from an await expression.
3      Prefer Number.parseInt() over parseInt().
3      Missing radix parameter.
3      Replace ·strict·as·assert· with strict·as·assert
3      Rest/spread properties are not supported until Node.js 8.3.0. The configured version range is >=0.12.18.
3      Expected method shorthand.
3      Prefer utf8 over utf-8.
2      Replace ·fileURLToPath· with fileURLToPath
2      /\[(.+)\]/ can be optimized to /\[(.+)]/.
2      Replace ·hasLicense· with hasLicense
2      Unexpected if as the only statement in a if block without else.
2      Array#reduce() is not allowed
2      Use uppercase characters for the value of the escape sequence.
2      Replace ·getDirnameFromImportMeta· with getDirnameFromImportMeta
2      Replace ·getDirnameFromImportMeta,·getIconsData· with getDirnameFromImportMeta,·getIconsData
1      Replace ·transform·as·esbuildTransform· with transform·as·esbuildTransform
1      RegExp lookbehind assertions are not supported until Node.js 8.10.0. The configured version range is >=0.12.18.
1      Replace ·code· with code
1      Replace ·icon,·iconObject,·iconExportName· with icon,·iconObject,·iconExportName
1      Replace ·iconObject,·iconExportName· with iconObject,·iconExportName
1      Replace ·icon· with icon
1      Use new RegExp() instead of RegExp().
1      /[^a-z0-9]/g can be optimized to /[^a-z\d]/g.
1      /^https:\/\/[^\s]+$/ can be optimized to /^https:\/\/\S+$/.
1      Prefer String#codePointAt() over String#charCodeAt().
1      /&#([0-9]+);/g can be optimized to /&#(\d+);/g.
1      The variable num should be named number_. A more descriptive name will do too.
1      Prefer String.fromCodePoint() over String.fromCharCode().
1      Replace ·quot:·'"',·amp:·'&',·lt:·'<',·gt:·'>'· with quot:·'"',·amp:·'&',·lt:·'<',·gt:·'>'
1      Spread elements are not supported until Node.js 5.0.0. The configured version range is >=0.12.18.
1      Replace ·describe,·test· with describe,·test
1      node:assert import should occur before import of mocha
1      URL_REGEX is defined but never used.
1      whiteIconFileName is never reassigned. Use const instead.
1      ignoreHttpLinks should be a Set, and use ignoreHttpLinks.has() to check existence or non-existence.
1      Use a regular expression literal instead of the RegExp constructor.
1      docsFile is never reassigned. Use const instead.
1      The Array.from is not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
1      fs is defined but never used.
1      getIconDataPath is defined but never used.
1      /^#?[a-f0-9]{3,8}$/i can be optimized to /^#?[a-f\d]{3,8}$/i.
1      Prefer .some(…) over .find(…).
1      Replace ·guidelines:·answers.guidelines· with guidelines:·answers.guidelines
1      Replace ·url:·answers.licenseUrl· with url:·answers.licenseUrl
1      Replace ·hasGuidelines· with hasGuidelines
1      method validate is equivalent to Boolean. Use Boolean directly.
1      Redundant Boolean call.
1      Only use process.exit() in CLI apps. Throw an error instead.
1      Replace ·getIconsDataString,·normalizeNewlines,·collator· with getIconsDataString,·normalizeNewlines,·collator
1      The variable prev should be named previous. A more descriptive name will do too.
1      Do not pass function collector directly to .reduce(…).
1      Use .length > 0 when checking length is not zero.
1      /simple-icons@v[0-9]+/g can be optimized to /simple-icons@v\d+/g.
1      Replace ·getIconSlug,·getIconsData,·titleToSlug· with getIconSlug,·getIconsData,·titleToSlug
1      mocha import should occur before import of ../index.js
1      Replace ·test· with test
1      node:assert import should occur before import of ../index.js
1      A function with a name starting with an uppercase letter should only be used as a constructor.
1      The Object.values is not supported until Node.js 7.0.0. The configured version range is >=0.12.18.
1      icon is never reassigned. Use const instead.
1      The catch parameter err should be named error.
1      The variable err should be named error. A more descriptive name will do too.
1      Replace ·describe,·it· with describe,·it
1      Replace ·URL_REGEX· with URL_REGEX
1      /^[MmZzLlHhVvCcSsQqTtAaEe0-9-,.\s]+$/g can be optimized to /^[MmZzLlHhVvCcSsQqTtAaEe\d-,.\s]+$/g.
1      Replace ·Validator· with Validator
1      Replace ·icons· with icons
1      Replace ·getIconsData,·getIconSlug· with getIconsData,·getIconSlug
1      Replace ·titleToSlug· with titleToSlug
1      Replace ·testIcon· with testIcon
1      Array.prototype.map() expects a return value from arrow function.
1      Replace ·reporters,·Runner· with reporters,·Runner
1      Replace ·EVENT_RUN_END· with EVENT_RUN_END
1      Classes in strict mode are not supported until Node.js 4.0.0. The configured version range is >=0.12.18.
1      Use a type instead of an interface.
1      Type Method name Get must match one of the following formats: strictCamelCase
1      Do not use "module".
1      Replace ·role:·'img',·xmlns:·'http://www.w3.org/2000/svg'· with role:·'img',·xmlns:·'http://www.w3.org/2000/svg'
```

After `npx xo --fix`

```
❯ npx xo | xo-summary
13     Unexpected use of the global variable process. Use require("process") instead.
3      Do not access a member directly from an await expression.
3      Missing radix parameter.
2      Array#reduce() is not allowed
1      fs is defined but never used.
1      getIconDataPath is defined but never used.
1      Prefer .some(…) over .find(…).
1      Only use process.exit() in CLI apps. Throw an error instead.
1      Do not pass function collector directly to .reduce(…).
1      Prefer String#codePointAt() over String#charCodeAt().
1      Prefer String.fromCodePoint() over String.fromCharCode().
1      URL_REGEX is defined but never used.
1      Use a regular expression literal instead of the RegExp constructor.
1      Type Method name Get must match one of the following formats: strictCamelCase
1      Do not use "module".
1      Array.prototype.map() expects a return value from arrow function.
1      A function with a name starting with an uppercase letter should only be used as a constructor.
```

</details>

#### VS Code

```json
{
  "xo.enable": true,
  "xo.format.enable": true,
  "editor.formatOnSave": true,
  "editor.defaultFormatter": "samverschueren.linter-xo"
}
```

For other editors, see https://github.com/xojs/xo#editor-plugins.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,6 +2,5 @@
 . "$(dirname $0)/_/husky.sh"
 
 git stash -q --keep-index
-npm run format
 git add .
 git stash pop -q
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -1,0 +1,5 @@
+{
+  "prettier": true,
+  "space": 2,
+  "ignores": ["scripts/build/templates/*"]
+}
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-export interface SimpleIcon {
+export type SimpleIcon = {
   title: string;
   slug: string;
   svg: string;
@@ -12,12 +12,13 @@ export interface SimpleIcon {
         url: string;
       }
     | undefined;
-}
+};
 
 /**
  * @deprecated The `simple-icons` entrypoint will be removed in the next major. Please switch to using `import * as icons from "simple-icons/icons"` if you need an object with all the icons.
  */
 declare const icons: Record<string, SimpleIcon> & {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
   Get(name: string): SimpleIcon;
 };
 
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "svg-path-segments": "1.0.0",
     "svglint": "2.1.0",
     "svgo": "2.8.0",
-    "svgpath": "2.5.0"
+    "svgpath": "2.5.0",
+    "xo": "^0.52.3"
   },
   "scripts": {
     "build": "node scripts/build/package.js",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "mocha": "10.0.0",
     "named-html-entities-json": "1.0.0",
     "npm-run-all": "4.1.5",
-    "prettier": "2.7.1",
     "rimraf": "3.0.2",
     "svg-path-bbox": "1.2.2",
     "svg-path-segments": "1.0.0",
@@ -61,10 +60,9 @@
   "scripts": {
     "build": "node scripts/build/package.js",
     "clean": "rimraf index.js icons.js icons.mjs icons.d.ts",
-    "format": "prettier --write .",
     "lint": "run-s our-lint jslint jsonlint svglint wslint",
     "our-lint": "node scripts/lint/ourlint.js",
-    "jslint": "prettier --check .",
+    "jslint": "xo",
     "jsonlint": "node scripts/lint/jsonlint.js",
     "svglint": "svglint icons/*.svg --ci --config .svglintrc.mjs",
     "wslint": "editorconfig-checker",
@@ -77,8 +75,5 @@
     "svgo": "svgo --config svgo.config.js",
     "get-filename": "node scripts/get-filename.js",
     "add-icon-data": "node scripts/add-icon-data.js"
-  },
-  "engines": {
-    "node": ">=0.12.18"
   }
 }
--- a/scripts/add-icon-data.js
+++ b/scripts/add-icon-data.js
@@ -1,4 +1,4 @@
-import fs from 'node:fs/promises';
+import process from 'node:process';
 import inquirer from 'inquirer';
 import chalk from 'chalk';
 import getRelativeLuminance from 'get-relative-luminance';
@@ -6,20 +6,19 @@ import {
   URL_REGEX,
   collator,
   getIconsDataString,
-  getIconDataPath,
   writeIconsData,
   titleToSlug,
   normalizeColor,
 } from './utils.js';
 
-const hexPattern = /^#?[a-f0-9]{3,8}$/i;
+const hexPattern = /^#?[a-f\d]{3,8}$/i;
 
 const iconsData = JSON.parse(await getIconsDataString());
 
 const titleValidator = (text) => {
   if (!text) return 'This field is required';
   if (
-    iconsData.icons.find(
+    iconsData.icons.some(
       (x) => x.title === text || titleToSlug(x.title) === titleToSlug(text),
     )
   )
@@ -46,12 +45,12 @@ const getIconDataFromAnswers = (answers) => ({
   title: answers.title,
   hex: answers.hex,
   source: answers.source,
-  ...(answers.hasGuidelines ? { guidelines: answers.guidelines } : {}),
+  ...(answers.hasGuidelines ? {guidelines: answers.guidelines} : {}),
   ...(answers.hasLicense
     ? {
         license: {
           type: answers.licenseType,
-          ...(answers.licenseUrl ? { url: answers.licenseUrl } : {}),
+          ...(answers.licenseUrl ? {url: answers.licenseUrl} : {}),
         },
       }
     : {}),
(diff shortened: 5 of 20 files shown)


mondeja:
I'm seeing that XO is built in top of ESLint, so opting by XO means to install and maintain updates for ESLint + XO, I don't see the point of using XO if there is not a very very good reason.

LitoMore:
We don't have to care or maintain ESLint in dependencies. XO has taken care of it all for you. For maintenance, we just need to update its version, no more other configurations.

If we only use ESLint directly, then we need to choose and maintain the rules ourself. But XO has already selected the most suitable rules for us, and it can even automatically filter suitable rules according to our `engines` fields. XO is made by [Sindre Sorhus](https://github.com/sindresorhus), so we can have a very good expectation of it, and its quality can be guaranteed.

You could just try it, check this commit https://github.com/simple-icons/simple-icons/pull/7869/commits/eddfe320c457cd3e4c28529a5e96a517ab9bef28, then use `npx xo --fix` to fix those fixable errors. For the remaining errors that need to be fixed manually, you can see its prompt to understand why this is an error.

mondeja:
> If we only use ESLint directly, then we need to choose and maintain the rules ourself.

This problem is still there although we would use XO, we need to maintain a set of rules between our repositories. We always used a set of rules for formatting and linting, only formalized currently in the [`simple-icons-website` repository](https://github.com/simple-icons/simple-icons-website/tree/5b741d95e05e12b398552a6bfdc83701bfa113e8), I think that is not worth to change them. You are pointing about that "we don't have to care", but that could be a nigthmare from a maintainer perspective, as no-configs set ups are less declarative, obfuscating things. Also, I don't want to have more dependencies to care of in updates, so I'm against XO as is affected by ESLint updates plus XO updates.

In fact, my opinion if we add linting is to create our own ESLint config package for all simple-icons repositories and don't need to depend on `eslint-config-prettier` nor `eslint-config-recommends`. That would be a great compromise between simplicity, control, maintenance and declarativity.

LitoMore:
Thank you for your explanation. Closing.

mondeja:
I've created the repository [simple-icons/simple-icons-eslint](https://github.com/simple-icons/simple-icons-eslint) to maintain there a declarative configuration that we can use across all projects just inheriting from the package:

```sh
$ npm install -DE eslint-config-simple-icons
```

```json5
// .eslintrc.json
...
  "extends": "simple-icons"
...
```

See the commit https://github.com/simple-icons/simple-icons-website/commit/659d040a73b04464e4cc9bf9d5e67a253ab4a83a as an example.

PR #7873, opened 2022-09-26
Check GitHub sources without hash using JSONSchema

mondeja:
This is an improvement that should have been included in #7359 which ensures that no more Github sources without hashes will be added.

### DIFF
--- a/.jsonschema.json
+++ b/.jsonschema.json
@@ -19,7 +19,7 @@
         },
         "source": {
           "description": "The website from which the icon was sourced",
-          "$ref": "#/definitions/url"
+          "$ref": "#/definitions/sourceUrl"
         },
         "guidelines": {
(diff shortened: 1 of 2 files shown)


.jsonschema.json
```diff
@@ -229,6 +229,20 @@
       "description": "The name of the brand",
       "type": "string"
     },
+    "sourceUrl": {
+      "$id": "#sourceUrl",
+      "description": "URL for icon source. If is a Github URL, is validated to contain a commit hash, to be an issue comment or to be a Github organization URL",
```
LitoMore, line 234:
```suggestion
      "description": "URL for icon source. If is a GitHub URL, is validated to contain a commit hash, to be an issue comment or to be a GitHub organization URL",
```

LitoMore (approved):

PR #7968, opened 2022-10-17
Add Jet Global icon

Jwaegebaert:
**Issue:** No related issue

**Similarweb rank:** 187,266

### Checklist

### Description

This icon is fetched from the [Insight Software](https://insightsoftware.com/) website on the [Jet](https://insightsoftware.com/jet/) product page to be more specific. 
The HEX value #0087cd is the primary color used in their logo.

jet


_data/simple-icons.json
```diff
@@ -5272,6 +5272,18 @@
             "hex": "FBBA00",
             "source": "https://de.wikipedia.org/wiki/Datei:JET.svg"
         },
+        {
+            "title": "Jet",
+            "slug": "jet_reporter",
+            "hex": "0087cd",
```
LitoMore, line 5278:
```suggestion
            "hex": "0087CD",
```

Jwaegebaert:
I'll resolve the lint error later today.

LitoMore (changes requested):

_data/simple-icons.json
Jwaegebaert: Thanks, I overlooked the hex guidelines

LitoMore:
@Jwaegebaert Could you try with precision 4? You can refer to https://github.com/simple-icons/simple-icons/pull/7908#issuecomment-1266943734.

Jwaegebaert:
Thanks @LitoMore, I'll give it a shot.

LitoMore:
@Jwaegebaert Never mind. My responsibility is to help you out.

Jwaegebaert:
@LitoMore thanks so muh for your quick assistance! Looks like the checks are valid now.

_data/simple-icons.json
```diff
@@ -5673,6 +5673,18 @@
...
         },
+        {
+            "title": "Jet",
+            "slug": "jet_reporter",
+            "hex": "0087CD",
+            "source": "https://insightsoftware.com/jet/",
+            "aliases": {
+                "aka": [
+                    "Jet Reports",
+                    "Jet Hub"
```
LitoMore, line 5684: I found these names here: https://insightsoftware.com/downloads/.

These aliases are different pricing plans, right? I'd prefer to remove these aliases because they are not common product names.

Please correct me if I am wrong.

Jwaegebaert: Well, each of them are more different products. You've `Jet Reports` and `Jet Analytics` which are two different programs. The reason I added `Jet Reports` and `Jet Hub` is because they both have the same logo. But if it becomes to complex we can just use ´Jet´ as the complete package. Which is also the case.

LitoMore: Let's keep `Jet` only. In fact, those aliases are only for search purposes. The `Jet` is enough to use.

LitoMore: BTW, I didn't find `Jet Reporter` on their homepage. Shall we use `jetglobal` as its slug?

In the meantime, maybe we could change its title to `Jet Global`. Then we don't need to specify its slug manually anymore. What do you think?

Jwaegebaert: Yeah good idea! I'll change it tomorrow.

Jwaegebaert:
I've received the SVG internally from somebody. Double checking the site I can't seem to locate the image as well. I'll contact them to verify if the SVG image is correct.

Jwaegebaert:
Seems like the logo was converted from a png image. I'm going to contact the company directly if I can get hold of their svg logo. I'll keep you up to date.

Jwaegebaert:
I haven't gotten any response so I'll close this PR as we don't have any original SVG files

PR #7969, opened 2022-10-17
Apply Elgato icon to Stream Deck icon pack

LitoMore:
I noticed this requirement at our [simple-icons/simple-icons#third-party-extensions](https://github.com/simple-icons/simple-icons#third-party-extensions) section.

We don't have the icon for Stream Deck. Since Elgato is the company of Stream Deck, we could use this icon instead of that fallback one.


sisp (approved):
LGTM! :+1:

PR #7975, opened 2022-10-18
Add Azure Synapse Analytics icon

pterdi:
azuresynapseanalytics

  

### Checklist

### Description
Recovered the SVG from Azure Synapse Analytics website's source. The SVG can be found [here](https://azurecomcdn.azureedge.net/cvt-7c05f9e517923fd679a6e186eecacb1ba4b2a7170271d8a0142c0a3e471be91c/svg/synapse-analytics.svg).
I used color from the icon for hex.


LitoMore:
Hi @pterdi, thank you for making this PR.

Seems there are two colors in the reference SVG. And the ⚡️-like shape is on the top of the hexagon. I'd prefer to add some gaps between them.

Like this, the gap width is 1px:

CleanShot 2022-11-02 at 00 51 51@2x

PR #8000, opened 2022-10-28
Add CDN usages for `cdn.simpleicons.org`

LitoMore:


README.md
```diff
@@ -41,6 +41,26 @@ Where `[ICON SLUG]` is replaced by the [slug] of the icon you want to use, for e
...
+#### CDN with colors
+
+We also provide a CDN service which allows you to use colors.
+
+```html
+<img height="32" width="32" src="https://cdn.simpleicons.org/[ICON_SLUG]" />
+<img height="32" width="32" src="https://cdn.simpleicons.org/[ICON_SLUG]/[COLOR]" />
+```
+
+Where `[COLOR]` is optional, and can be replaced by the [hex triplet](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet) or [CSS keywords](https://www.w3.org/wiki/CSS/Properties/color/keywords) of the icon you want to you use. The color is defaulted to the color of icon from [Simpleicons.org](https://simpleicons.org). For example:
```
mondeja, line 53: Is not exactly true that the color can be only formatted using an HEX triplet. I would write that also accepts HEX cuatriplets because a fourth digit can be passed for an opacity value.

mondeja (changes requested):

README.md
LitoMore: Ah yes. Let me change the name to `hex color`, and change its link to https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color.

The [logic for parsing the hex color](https://github.com/LitoMore/simple-icons-cdn/blob/main/source/utils.ts#L3) is different from the hex color. I will fix that.

LitoMore: Fixed within https://github.com/LitoMore/simple-icons-cdn/commit/9265817bdd0294e6fb9c5e8aba8db3d5c3b6c193.

README.md
```diff
@@ -41,6 +41,27 @@ Where `[ICON SLUG]` is replaced by the [slug] of the icon you want to use, for e
...
+#### CDN with colors
+
+We also provide a CDN service which allows you to use colors.
+
+```html
+<img height="32" width="32" src="https://cdn.simpleicons.org/[ICON_SLUG]" />
+<img height="32" width="32" src="https://cdn.simpleicons.org/[ICON_SLUG]/[COLOR]" />
+```
+
+Where `[COLOR]` is optional, and can be replaced by the [hex colors](https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color) or [CSS keywords](https://www.w3.org/wiki/CSS/Properties/color/keywords) of the icon you want to you use. The color is defaulted to the color of icon from [Simpleicons.org](https://simpleicons.org). For example:
```
mondeja, line 53: Sorry for this late review, just a minor suggestion to clarify the HEX colors source:

```suggestion
Where `[COLOR]` is optional, and can be replaced by the [hex colors](https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color) or [CSS keywords](https://www.w3.org/wiki/CSS/Properties/color/keywords) of the icon you want to you use. The color is defaulted to the HEX color of the icon shown in [simpleicons.org website](https://simpleicons.org). For example:
```

mondeja (approved):

PR #8058, opened 2022-11-24
replace index.js entry point

sachinraja:
follow up from #6789


index.d.ts
```diff
@@ -1,24 +0,0 @@
-export interface SimpleIcon {
```
mondeja, line 1: This file should not be renamed to `types.d.ts` instead of be removed?

sachinraja: good catch, fixed that

mondeja (approved):
Looks good, thanks @sachinraja! 🥇

mondeja (approved):
Just made some changes, now really looks good 😅

PR #8077, opened 2022-11-28
Expose SDK in NPM package

### LINKED ISSUE #8071: Expose internal utilities
### Description

With the next major v8 being released at the end of this week, we can start planning the exposure of internal utilities (*scripts/utils.js*) file. Note that this is not a breaking change. I suggest to expose the utilities in a new entrypoint `./utils`, so we could import the utilities using:

```javascript
import { getThirdPartyExtensions } from 'simple-icons/utils';
```

This would allow us to deduplicate a lot of code of other repositories like [`simple-icons-font`](https://github.com/simple-icons/simple-icons-font) and [`simple-icons-website`](https://github.com/simple-icons/simple-icons-website).

mondeja:
Closes #8071 

All functions of *scripts/utils.js* except `getJsonSchemaData` and `writeIconsData` have been moved to a new file ~~*utils.mjs*~~ *sdk.mjs*, which is exposed by the ~~`./utils`~~ `./sdk` entrypoint.


mondeja:
Hi @LitoMore,

Would you mind to review this PR? I think that if you uses `simple-icons/utils` in simple-icons-cdn a potential bug would be fixed. Currently you are using [this function](https://github.com/LitoMore/simple-icons-cdn/blob/fd6d5687237b18f45ecf6736ba6369468370503f/source/icon.ts#L10) to normalize icons slugs, but there are some replacements that are not handled like `&` characters by `and`, you can see the full implementation at [here](https://github.com/simple-icons/simple-icons/blob/3271d9e950f5b1296b23a6da776224d82368f76b/scripts/utils.js#L10) and [here](https://github.com/simple-icons/simple-icons/blob/develop/scripts/utils.js#L59).

LitoMore:
It's missing TypeScript definitions for `utils.mjs`.

You could test the behavior in this way:

1. Under `simple-icons` project root, run `npm link .`
2. Under `simple-icons-cdn` or other projects, run `npm link simple-icons`
3. Try `import {getIconSlug} from 'simple-icons/utils'` in any source file
4. Hover on the `getIconSlug`, its types will not shown

mondeja:
Now should work for you, let me know if any problem persists.

utils.d.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type IconData = {
+  title: string;
+  hex: string;
+  source: string;
+  slug: string | undefined;
+  guidelines: string | undefined;
+  license: string | any | undefined;
+  aliases: any | undefined;
+};
```
LitoMore, line 19: We should reuse the `SimpleIcon` type.

```js
export function getIconSlug(icon: SimpleIcon): string;
```

mondeja: Nor `svg` nor `path` properties are part of icon data objects.

mondeja:
I'm using Typescript in a project that uses simple-icons, so I'll try to test this file there ASAP and get this released.

mondeja:
I've changed the `./utils` entrypoint by `./sdk` for several reasons:

- `utils` is a bad module name to be exposed, too much ambiguous, `./sdk` is much better
- `utils.js` was added to the *.npmignore* file which could lead to the publishing of the file *scripts/utils.js*
- Having two modules with the same name but different extensions is a nightmare from a maintainer perspective

Also improved a lot the types and added a section to the documentation about how to use this SDK. I'm going to merge this as is. In the next major version bump I'll add a note about that we're now including a SDK for extensions authors. We can consider it as a beta meanwhile.

PR #8148, opened 2022-12-16
Remove `event=push` from status badge of README

mondeja:
See https://github.com/badges/shields/issues/8671

It seems that there are some problems with the new endpoint, I've opened an issue at https://github.com/badges/shields/issues/8733


mondeja:
Blocked until https://github.com/badges/shields/issues/8733 is closed.

mondeja:
It seems that the problem is fixed, so I'm merging this...

PR #8149, opened 2022-12-16
Modernize SVGO setup

mondeja:
- Removed `svgo` npm script encouraging the usage of NPX to run SVGO to avoid installation of dependencies.
- Use ESM in SVGO config.


LitoMore:
We can use `npx svgo file/to/optimize.svg` instead of `npm run svgo -- file/to/optimize.svg`.

LitoMore (approved):

PR #8344, opened 2023-02-13
Update LiveChat products data

jwr12135:
### Description

The [LiveChat](https://www.livechat.com/) products use separate colors for their different product designs. I don't know if that has always been the case, but we can see the different colors on their [brand guidelines](https://livechat.design/) now:


dirien (approved):
LGTM

Issue #8650, opened 2023-05-16
Proton VPN is outdated


jesusfj710:
### Brand Name

Proton VPN

### Official Resources for Icon and Color

https://proton.me/media/kit

### Contributing

Yes

Issue #8649, opened 2023-05-16
Proton Mail is outdated


jesusfj710:
### Brand Name

Proton mail

### Official Resources for Icon and Color

https://proton.me/media/kit

### Contributing

Yes

PR #8683, opened 2023-05-25
Add some stop words to the potential duplicates action

mondeja:
I've trigerred a false positive when renaming #8649 and #8650 with the word "outdated", one previously using "not updated". Added also some other words seen in other issues.


service-paradis (approved):
LGTM!

PR #8692, opened 2023-05-29
Add lint for checking redundant slashes

LitoMore:
We have tons of URLs that have a redundant trailing slash. For example:

```diff
-https://www.1001tracklists.com
+https://www.1001tracklists.com/
```

For these domain-only URLs, we can safely remove the trailing slash.

### How do I remove those trailing slashes

The VS Code has a performance bug when I try to find all incorrect URLs with regex. So I used Sublime Text to replace texts on my local.

Here is my regex pattern:

|Find|Replace|
|:--|:--|
|`(https://([0-9a-z-]\.?)*)(/)"`|`$1"`|

### Bundle size

||Before|After|
|:--|:--|:--|
|index.js|35,434.13 KB|35,428.42 KB|

### DIFF
--- a/scripts/lint/ourlint.js
+++ b/scripts/lint/ourlint.js
@@ -57,9 +57,9 @@ const TESTS = {
     }
   },
 
-  /* Check redundant slash suffix in URL */
+  /* Check redundant trailing slash in URL */
   checkUrl: (data) => {
-    const hasRedundantSlashSuffix = (url) => {
+    const hasRedundantTrailingSlash = (url) => {
       const origin = new URL(url).origin;
       return url.replace(origin, '') === '/';
     };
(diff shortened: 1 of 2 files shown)


scripts/lint/ourlint.js
```diff
@@ -55,6 +56,33 @@ const TESTS = {
       return `Data file is formatted incorrectly:\n\n${dataDiff}`;
     }
   },
+
+  /* Check redundant trailing slash in URL */
+  checkUrl: (data) => {
+    const hasRedundantTrailingSlash = (url) => {
+      const origin = new URL(url).origin;
+      return url.replace(origin, '') === '/';
```
mondeja, line 64: What about this example: `https://example.com//`? It passes but shouldn't. I think that the check could be "only has `/` characters and at least one".

mondeja (commented):
Great. The behaviour is consistent if clients implement RFC 2616, see [this quote on Stackoverflow](https://stackoverflow.com/a/2581514/9167585).

I just have one minor suggestion.

scripts/lint/ourlint.js
LitoMore: Good catch. I can change it's condition to:

```js
return /^\/+$/.test(url.repalce(origin, ''));
```

LitoMore: Updated.

mondeja (approved):
Thanks!

PR #8695, opened 2023-05-29
Proposal to autoclose Java requests

service-paradis:
Java requests is almost a weekly occurrence. This proposal would close those issues automatically with a reference to the post originating the removal.

Would maybe need to change the linter for this one though...


LitoMore:
I guess most of those guys come from our homepage. We could give a hint message when they search `java` and hide the `Icon missing? Submit a request` option.

mondeja:
For the linting issue I think that you must escape the `\` character with another. Apart from that, the action is not maintained, currently is using Node.js v12 which is not even in maintenance state. Also, if I'm not mistaken, this would close issues with "Javascript" in the title which IMHO would turn the issue tracker very anoying.

service-paradis:
> For the linting issue I think that you must escape the \ character with another.

The Regexp function would not work as expected if I escape the `\`

> Apart from that, the action is not maintained, currently is using Node.js v12 which is not even in maintenance state. 

You're right. Might need to find another one (or create one...)

> Also, if I'm not mistaken, this would close issues with "Javascript" in the title which IMHO would turn the issue tracker very annoying.

Nope. The `(?!\S)` part check that java is the end of the string or is followed by a space:
https://regexr.com/7eja4

mondeja:
> The Regexp function would not work as expected if I escape the `\`

It maybe works with `\x5C`, see [backslash in YAML spec](https://yaml.org/spec/1.2.2/#rule-ns-esc-backslash).

mondeja:
That is great! I agree with @LitoMore here to check if the issue's label is `new icon`. You can use [this step](https://github.com/simple-icons/simple-icons/blob/c7ab3a11ced7b26c366ad5adfdf08e41ffca8ee6/.github/workflows/add-labels-priority.yml#L23-L27) to get the labels and include in the `if` statement `contains(steps.get-labels.outputs.labels, 'new icon')`.

.github/workflows/autoclose-issues.yml
```diff
@@ -0,0 +1,54 @@
...
+  run:
+    runs-on: ubuntu-latest
+    steps:
+      - id: match-title
+        name: Match title
+        # if the title contains the word Java (case insensitive)
+        # we make sure that this word is the end of the string or is followed by a space character (ex. we do not want to match javascript)
+        # we make sure that this word is the beginning of the string or is preceded by a space character (ex. we do not want to match foojava)
+        run: |
+          if [[ ${{ github.event.issue.title }} =~ (^|[[:space:]])([j|J][a|A][v|V][a|A])([[:space:]]|$) ]]; then
```
LitoMore, line 16:
```suggestion
          if [[ ${{ github.event.issue.title }} =~ (^|[[:space:]])([jJ][aA][vV][aA])([[:space:]]|$) ]]; then
```

This regex will match `||||`. The `|` will be treated as a character when put to [character classes](https://www.rexegg.com/regex-quickstart.html#classes).

LitoMore:
BTW, we have some other icons kike `Matlab` and `Disney` that won't add.

We can consider this in the next PR.

service-paradis:
Hum... looks like action does not like `=~` operator...

LitoMore:
Yep, would you mind creating another PR to fix this? Thanks!

service-paradis:
Yes, I think, it is missing double quote around the issue title in fact

I will submit a PR with something like: ` if [[ "${{ github.event.issue.title }}" =~ (^|[[:space:]])([jJ][aA][vV][aA])([[:space:]]|$) ]]; then` since the action gives `if [[ I know Java icon is missing =~ (^|[[:space:]])([jJ][aA][vV][aA])([[:space:]]|$) ]]; then`

PR #9062, opened 2023-07-21
Add Fluke icon

### LINKED ISSUE #9059: Fluke
### Brand Name

Fluke

### Website

https://www.fluke.com/

### Popularity Metric

https://www.similarweb.com/website/fluke.com/#overview

None

### Additional Comments

_No response_

### Contributing

Yes

PratyushJayachandran:
fluke

**Issue:** close #9059 

**Similarweb rank:** 53,199. See https://www.similarweb.com/website/fluke.com

### Checklist

### Description
SVG: Taken from https://www.fluke.com
Color: Taken from logo `#ffc20e`


adamrusted:
@simple-icons/maintainers - I think this is an example of needing to keep the ®️ intact, what do we think? Looking at their site and socials, it's used across all brands - with no exceptions.

PratyushJayachandran:
> @simple-icons/maintainers - I think this is an example of needing to keep the ®️ intact, what do we think? Looking at their site and socials, it's used across all brands - with no exceptions.

I removed because they don't use it on their products. The multimeters at least don't have it. Didn't see any guidelines for ® either. But will let maintainers comment on this.

LitoMore:
This looks not good to me.

You can find some Fluke products on this page https://www.fluke.com/en-us.

It does not have a wide left padding in the trademark.

I noticed they're using the trademark with `®` sign for press releases.

So, let's keep the `®` sign to make sure it's consistent with their press releases in this case.

adamrusted:
Essentially, let's go with the icon [on the homepage](https://www.fluke.com/en-us) including the ®️ - as that's probably the most consistent icon to the products.

PratyushJayachandran:
Okay I'll do it as is in that case. The wide left padding is another quirk. Even some of their products have it but some of them don't.

adamrusted:
It should be mitigated by having the ®️ there... 🤞😅

PratyushJayachandran:
> It should be mitigated by having the ®️ there... 🤞😅

I have fixed it. Attaching updated preview.
fluke

adamrusted (approved):

PR #9163, opened 2023-07-30
Normalize third party extension names

mondeja:
All third party extensions normalized with the same case sensitivity. Also, added the word "package" to the _Blazor Nuget package_.


adamrusted (approved):
LGTM! Thanks @mondeja 🎉

Issue #9072, opened 2023-07-22
The great icon review of 2023


adamrusted:
HI folks, I want to make sure we get this in early, so we have plenty of time ahead of the v10.0.0 release in November.
I last did this a couple years ago, and would appreciate any support you can offer along the way. I will be reviewing every icon in our catalogue to make sure it's still up to date. This'll involve not just the icon, but making sure our source is still good - and that the terms / license hasn't changed. I'm not expecting people to tackle making changes to an entire section by themselves, but I'd appreciate any support in reviewing a section, and commenting below here with the icons for your letter that need work on them.

### How to Review a Letter
1. Go through each entry in the [`_data/simple-icons.json`](https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json) file for the appropriate letter.
2. Ensure that the source URL still links to an appropriate SVG. If not, find an [appropriate SVG source](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#1-identify-official-logos-and-colors).
3. Ensure the icon still matches what we have in the repository. We should re-work even if there are minor differences!
4. Make sure the [inclusion criteria](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon) still applies. Projects get retired and brands fall out of popularity, and we need to trim the repository if a brand no longer meets our criteria for inclusion.
5. Share a list of all icons that need work in a comment below. They'll get added to this master list, and your comment hidden.

### How to fix an icon
We'd like to get all of this done ahead of the v10 release so we know we've got a 'clean slate' at that point, and have removed any brands that we can. Please see our [contribution guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and submit a PR for any icons you see listed below. Be sure to change the PR template to say 'contributes to' instead of 'closes' - as we want to keep this issue open until everything is resolved.

> **Warning**
> Icons might be listed here before you've completed the review of a section. They are carried over from the work done by @jorgeamadosoria in #6399.

## 0-9 Review completed by @adamrusted 

- [ ] .ENV - does not link to SVG within Git repo.
- [ ] .NET - does not link to SVG within Git repo.
- [ ] /e/ - icon has been removed in later iterations of that git repo.
[...]

PR #9176, opened 2023-07-31
Update fila

PratyushJayachandran:
fila

**Issue:** closes #9072 

**Similarweb rank:** 74,201. See https://www.similarweb.com/website/fila.com

### Checklist

### Description
SVG: Taken from brand website header at https://www.fila.com
Color: Taken from the majority color in the logo `#002D62`


service-paradis:
I'm wondering if we could find a SVG version of their boxed logo as used in their social media that can be seen at https://www.filaholdings.com/en/media/media_logo.asp

It would probably be the best version for us.

PratyushJayachandran:
> I'm wondering if we could find a SVG version of their boxed logo as used in their social media that can be seen at https://www.filaholdings.com/en/media/media_logo.asp
> 
> It would probably be the best version for us.

Will recreating using the available SVG work? In the case that we can't find SVG of the box logo?

adamrusted (approved):

adamrusted:
Could we just go with the `F` shape, as they do in their favicon? Does it need the box around it?

PratyushJayachandran:
https://www.filaholdings.com/en/brand/index.asp#corporate
See images. When it's just the F it's in a box otherwise it's the full logo.

adamrusted:
Which woud we prefer then, just the F in the box - or the full wordmark?

LitoMore:
I prefer the FILA Linear Logo.

image

It appeared in more prominent places, such as pants and shoes.

PratyushJayachandran:
I agree with doing full. On my shoes it's full. I have only seen them doing just F on select clothing.

service-paradis:
Lets go for the full wordmark! I think you reviewed it already @adamrusted?

adamrusted:
I certainly had - providing this update to the branch doesn't break things, I'll merge it straight in!

PR #9180, opened 2023-07-31
Remove z-wave

service-paradis:
Part of #9072

It does not meet our popularity requirements anymore. We might need to take another look at its rank before merging in V10 just in case: https://www.similarweb.com/website/z-wave.com/#overview


cravl-dev:
I would argue that for standards alliances (e.g. Z-Wave, Zigbee, etc.), using the similarweb ranking for their website is a wildly inaccurate metric by which to measure their popularity. The **overwhelming majority** of people that use Z-Wave daily in their home automation setup probably have **never** been to the standards alliance website, because it's targeted more towards manufacturers than anything. Now, I can't really think of a _better_ metric off the top of my head, other than just _interest in the related field (i.e. home automation) as a whole_, which obviously isn't easily quantifiable. So, a simple exception to the rule may have to suffice, if allowable.

Particularly with Material Design Icons [deprecating all their brand icons soon and _encouraging users to use Simple Icons instead_](https://pictogrammers.com/docs/library/mdi/guides/brand-icons/) (including Z-Wave, Zigbee, and others), many thousands of users of Home Assistant and other home automation platforms will likely soon rely on SI for those icons (unless someone else steps in with a fork of/add-on repository to MDI or something).

adamrusted:
Thanks for the above @cravl-dev - the primary reason for opening these removal PRs early is to promote discussion!

I totally agree that for some brands, using the SimilarWeb rank alone isn't suitable - and I would always try to find a correlation between the SimilarWeb rank and the Google Trends data to inform a decision for removal.

Looking at that Google Trends comparison at a global and all time scale, Z-Wave is ranking lower than all the others mentioned, and is currently trending downward - so would also contribute toward reasoning for removal, especially if other protocols are just on the cusp for removal.

We also review GitHub stars frequently, so if any of these products are opensourced and popular, there is no question regarding removal.

That being said, we have strict guidelines for inclusion of a brand, and any that fall outside the criteria are fair game when it comes to an annual review - Material Icons reliance or not. We will likely review this in full toward the end of the month and reach a decision then.

adamrusted (approved):

adamrusted:
Still ranking very low - with a downward trend. Removing in v10, and we can re-add at a later date if ratings improve.

Vebryn:
I'm using it, any workarround to include localy ?

PR #9181, opened 2023-07-31
Remove Zettlr

service-paradis:
Part of https://github.com/simple-icons/simple-icons/issues/9072

It does not meet our popularity requirements anymore. We might need to take another look at its rank before merging in V10 just in case: https://www.similarweb.com/website/zettlr.com/


uncenter:
It does have plenty of stars though? [8.8k](https://github.com/Zettlr/Zettlr) and the threshold is only 5k:
 https://github.com/simple-icons/simple-icons/blob/e43eb0dbfd6c8ee21300c24946377ed502316cc7/CONTRIBUTING.md?plain=1#L49.

service-paradis:
Oh, you're totally right! Missed this one, thank you!

PR #9188, opened 2023-08-01
Drop bump-version script

mondeja:
npm includes the command [`npm version`](https://docs.npmjs.com/cli/v9/commands/npm-version) which accepts a version string as first parameter to update the *package.json* file, so the bump version script is not needed.


.github/workflows/create-release.yml
```diff
@@ -41,8 +41,21 @@ jobs:
...
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: Bump version
-        run: node ./scripts/release/bump-version.js "${{ needs.release-pr.outputs.new-version }}"
+        run: npm version "${{ needs.release-pr.outputs.new-version }}"
```
LitoMore, line 56: Different from the `bump-version.js`, the `npm version` will make a commit and create a new version tag.

We need to add `--no-commit-hooks` and `--no-git-tag-version` to skip these two actions.

```sh
npm version --no-commit-hooks --no-git-tag-version "${{ needs.release-pr.outputs.new-version }}"
```

LitoMore (approved):

PR #9191, opened 2023-08-01
Update Zapier

service-paradis:
Part of #9072
image

Update Zapier logo since they completely rebrand:
https://zapier.com/blog/zapier-rebrand-case-study/

Source and guidelines have been updated to https://brand.zapier.com/

Color have been slightly updated according to their guidelines.

I chose the Zapier Wordmark found in the App Icons from their brand assets (https://drive.google.com/drive/folders/1lDi7I-NC45zHfylsHU60Vloz_4zhyUfK). I picked the rounded one since I tough it looks a little bit better for us. They also have a squared one.

~The full logo cannot be used:~
image

Edit. We could **MAYBE** use the full logo according to (but it looks like they are using their earth and cream color...):
image


adamrusted:
@service-paradis - I think it might be worth asking if we can be allowed to use the monochrome treatment as part of the project. Would you mind reaching out?

adamrusted:
@service-paradis - did you hear back from Zapier at all?

adamrusted (approved):

PR #9195, opened 2023-08-01
Fix linting error and refactor `.svglintrc.mjs`

mondeja:
- Fixed `ineffective-segments` lint recognizing relative `m` commands after `z` | `Z` as invalids (Drops *Sketch* from *.svglint-ignored.json*).
- Refactor  long `if`s by `switch`s blocks, `forEach` and index loops by `for ... of` loops, made comments and error messages multiline, use `ast.source` over `$.html()`, unify SVG path regexes...


.svglintrc.mjs
```diff
@@ -298,32 +305,32 @@ export default {
...
+            const decimalCodepointCharIndex =
+              getTitleTextIndex(ast.source) + match.index + 1;
+            if (xmlNamedEntitiesCodepoints.includes(decimalNumber)) {
+              replacement = `"&${
+                xmlNamedEntities[
+                  xmlNamedEntitiesCodepoints.indexOf(decimalNumber)
+                ]
+              };"`;
+            } else {
+              replacement = String.fromCharCode(decimalNumber);
```
LitoMore, line 324:
```suggestion
              replacement = String.fromCodePoint(decimalNumber);
```

Minor: Unicode is better supported in [`String#codePointAt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt) [`String.fromCodePoint()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#compared_to_fromcharcode).

Here is [the difference between `String.fromCodePoint()` and `String.fromCharCode()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint#compared_to_fromcharcode).

.svglintrc.mjs
```diff
@@ -440,17 +433,21 @@ export default {
...
+        const commands = [].concat.apply(
+          [],
+          [
+            lowerMovementCommands,
+            lowerDirectionCommands,
+            upperMovementCommands,
+            upperDirectionCommands,
+            curveCommands,
+          ],
+        );
```
LitoMore, line 446:
```suggestion
        const curveCommands = [...lowerCurveCommands, ...upper];
        const commands = [
          ...lowerMovementCommands,
          ...lowerDirectionCommands,
          ...upperMovementCommands,
          ...upperDirectionCommands,
          ...curveCommands,
        ];
```

Enforces the use of [the spread operator (`...`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) over outdated patterns. This also helps keep consistency by using a flexible operator instead of:

- `Array.from(...)`
- `Array#concat(...)`
- `Array#slice()`
- `Array#toSpliced()`
- `String#split('')`

LitoMore (approved):

Issue #9099, opened 2023-07-24
Twitter is being rebranded as X (not a joke)


service-paradis:
### Brand Name

Twitter (X)

### Official Resources for Icon and Color

- https://www.theverge.com/2023/7/23/23804629/twitters-rebrand-to-x-may-actually-be-happening-soon
- https://x.com is being redirected to twitter.com
- twitter.com is displaying the X logo

- https://about.twitter.com/en/who-we-are/brand-toolkit is not yet updated
- https://about.twitter.com/press/brand-assets is not working

### Additional Comments

This could be made in some steps once we're sure it gonna stay:

- [x] Add X #9200
- [x] Remove Twitter as a breaking change in the next milestone #9748 
- [x] Add Twitter as an AKA for X in the next milestone (maybe?) #9318

Marked as pending since [Later Musk said he was going with the “minimalist art deco” logo but would probably make changes to refine it later.](https://twitter.com/elonmusk/status/1683324132853686272)

### Contributing

Yes

adamrusted:
Worth noting as well, that although he's going with that logo for now, it's not an original icon.
Best to let this one percolate for a while before we make any rash changes (unlike a certain Mr. Musk).

alistair3149:
Might be because of caching. Going to X.com once then going back to Twitter will force the icon to appear.

LitoMore:
We can keep the Twitter icon. They still have some products named Twitter. For example, `Twitter Blue`.

They didn't revoke the "Twitter" trademark so we can keep it.

Another point is this will be a quite big breaking change for our users. We should handle it with care.

adamrusted:
My understanding is they're rebranding everything to X, but they're taking a REAL long time doing it. As I said above, let's hold fire on any action for now - as it may or may not actually happen fully.

Regarding a breaking change, would an icon with the name X but the slug twitter not still work, without it being a breaking change?

guylepage3:
Just add X for the time being. Then deprecate Twitter when it's fully deprecated on X's end.

BUGProtectionAssociation:
It still use the bird...So add new not delete
![image](https://github.com/simple-icons/simple-icons/assets/65578116/01cf928d-0546-4a8f-89c6-7aa0fa052d6e)

adamrusted:
Just to chip in on this, although the twitter.com domains are still being used, almost everything is now branded as X. Unless there are any objections @simple-icons/maintainers, I think we can safely remove Twitter as of v10.

[...]

PR #9200, opened 2023-08-02
Add X icon

service-paradis:
image

**Issue:** Part of #9099

I think we can add this one even if the logo might not be th official one. We could always update it when needed. We'll probably need to update the source and guidelines anyway if they eventually update them.

As mentioned in #9099, we can keep Twitter til the next milestone (or even the other one). Eventually, if we really remove Twitter, we might add it as an aka for X.


LitoMore (approved):
image

Perfect!

PR #9211, opened 2023-08-04
Update Zoom icon

service-paradis:
image

**Issue:** Part of #9072

### Description

- I took `Zoom company logo (digital)/SVG (RGB)/logo_ZM_wordmark_100-RGB.svg` from `Zoom company logo (digital).zip` found at https://library.zoom.us/brand-system/public in the section "Zoom company brand kit (digital)"
- I updated metadata according to their brand guidelines


adamrusted:
Their [favicon](https://github.com/simple-icons/simple-icons/assets/1830867/95422ea3-0a06-4dc0-a538-ec012875a604) is probably a better fit for a 'simple icon'. Is there anywhere in their terms stating we can't use it?

LitoMore:
@adamrusted This is only used for their favicon. We should only use the icon if it is consistent with the main trademark on the page.

BTW, I have PR to add wide-mode support to shields.io here: https://github.com/badges/shields/pull/9191. I will pick up that PR in a short time.

So wide logos should not be a problem on the badge.

adamrusted (approved):
LGTM! Thanks @service-paradis 🎉

PR #9186, opened 2023-08-01
Autogenerate SDK TypeScript definitions on releases

mondeja:
Currently, the type definitions of the SDK function comments are slightly outdated. This PR introduces a script to update the Typescript type definitions generated from the file *sdk.mjs*, executing it on releases.

PR #9218, opened 2023-08-06
Fix missing package

LitoMore:
This bug blocked our release process.

The `typescript` is needed since we have `npx tsc` command in the release workflow.

Related to #9186


LitoMore:
Merging to unblock our release process.

PR #9237, opened 2023-08-08
Refactor tests and scripts

mondeja:
- Don't allow single `'` nor doble quotes `"` in URLs.
- Explicitly import `process` module with `node:` protocol.
- Refactor, normalize usage of APIs and remove duplicated code.
- Updated *SmugMug* source as had a single quote and the current URL is returning 404 code.


scripts/release/update-sdk-ts-defs.js
```diff
@@ -4,6 +4,7 @@
  * to match the current definitions of functions of sdk.mjs.
  */
 
+import process from 'node:process';
 import fsSync from 'node:fs';
```
LitoMore, line 8: The `fsSyns.exists` can be replaced with the Promise function `fs.access`.

```diff
-import fsSync from 'node:fs';
+import fs from 'node:fs/promises';

-const fileExists = fsSync.existsSync(sdkMts);
+const fileExists = await fs.access(sdkMts).then(() => true).catch(() => false);
```

Reference: https://nodejs.org/api/fs.html#fspromisesaccesspath-mode

LitoMore (approved):

PR #9244, opened 2023-08-09
Add documentation links checking to CI

mondeja:
This allows to remove the test of just HTTPs protocols in documentation links without being invasive in the development workflow as the check is slow.

This improvement in markdown-link-check would be great: https://github.com/tcort/markdown-link-check/issues/116


service-paradis (approved):

.github/workflows/verify.yml
```diff
@@ -44,6 +44,21 @@ jobs:
...
+        with:
+          list-files: shell
+          filters: |
+            docs:
+              - '*.md'
+              - '.github/**.md'
+      - name: Check documentation links
+        if: steps.changes.outputs.docs == 'true'
+        run: |
+          npx markdown-link-check --retry \
```
LitoMore, line 59: This will not be cached and will always install the latest version.

We can consider adding this package to `devDependencies`.

LitoMore (approved):

PR #9260, opened 2023-08-11
Update Aerospike

PratyushJayachandran:
aerospike

**Similarweb rank:** 751,760. See https://www.similarweb.com/website/aerospike.com

### Checklist

### Description
SVG: Taken from [this specific PDF](https://pages.aerospike.com/rs/aerospike/images/Acid_Whitepaper.pdf) as the logo is not available as a square elsewhere
Color: Updated to `#C22127` taken from the same logo inside PDF


adamrusted:
Looks like on their socials they just use the `A` without the square. Would it be worth us following suit? If so, we may still need to use this source, as the logo isn't available as an SVG on the website...

PratyushJayachandran:
> Looks like on their socials they just use the `A` without the square. Would it be worth us following suit? If so, we may still need to use this source, as the logo isn't available as an SVG on the website...

Their favicon is still the same is the red square. But yeah socials are different. Should we favor favicon more?

adamrusted:
> Their favicon is still the same is the red square. But yeah socials are different. Should we favor favicon more?

@simple-icons/icon-review

adamrusted:
Given we've not heard anything, let's just stick with the square version for now. Will review shortly.

adamrusted (approved):

PR #9266, opened 2023-08-12
Update Angular

PratyushJayachandran:
angular

**Similarweb rank:** 20,916. See https://www.similarweb.com/website/angular.io

### Checklist

### Description
SVG: Taken from https://angular.io/presskit one color logo version
Color: `#DD031` Taken from color version of logo


_data/simple-icons.json
```diff
@@ -799,7 +799,10 @@
         {
             "title": "Angular",
             "hex": "DD0031",
-            "source": "https://angular.io/assets/images/logos/angular/angular_solidBlack.svg"
+            "source": "https://angular.io/presskit",
```
adamrusted, line 802:
```suggestion
            "source": "https://angular.io",
```

Logo is available as SVG in the homepage. We should probably use that, with the presskit page as a 'guidelines' entry.

PratyushJayachandran: I have a general question regarding this. In some cases, I have found that the website header is slightly different than branding page. Shouldn't branding page prevail if it exists?

adamrusted: It's a difficult one. I would use the homepage icon as our source, because that's always going to be available providing the project keeps going. By adding the press kit page as a 'guidelines' entry, we're instructing our users that they should at least consult that page before using the logo.

PratyushJayachandran: I agree with that. It's just the discrepancy that bugs me. 

Recent case was clarivate #9142 
Sometimes some brands provide logos in PDF with specific use cases and versions for monotone which don't match with website header when minutely comparing paths.

I'll fix as you suggested in this PR though

adamrusted: Yeah, it's never an ideal situation to find ourselves in. 😅

PratyushJayachandran: I have fixed it.

PR #9360, opened 2023-08-21
Replace all `#FFFFFF` with `#000000`

LitoMore:
All `#FFFFFF` brand colors can be changed to `#000000`. White color is not commonly used.

### Note

- [Oh Dear](https://simpleicons.org/?q=ohdear) - The icon needs to be updated, but we can still use black on the old one
- [Zulip](https://simpleicons.org/?q=zulip) - The color `#50ADFF` was picked from its favicon


_data/simple-icons.json
```diff
@@ -7037,7 +7037,7 @@
         },
         {
             "title": "LGTM",
-            "hex": "FFFFFF",
+            "hex": "000000",
             "source": "https://lgtm.com"
         },
```
service-paradis, line 7042: Looks like we should remove this one for V10. It redirects to https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

_data/simple-icons.json
```diff
@@ -8784,7 +8784,7 @@
...
         {
             "title": "Oh Dear",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 8787: Guidelines mentions `#FF3900`

_data/simple-icons.json
```diff
@@ -12515,7 +12515,7 @@
         },
         {
             "title": "Talenthouse",
-            "hex": "FFFFFF",
+            "hex": "000000",
             "source": "https://www.talenthouse.com"
```
service-paradis, line 12519: Should plan to remove this one maybe?

service-paradis (changes requested):
I have the feeling that black or white are not really the correct colors for many of them...

_data/simple-icons.json
LitoMore:
```suggestion
            "hex": "FF3900",
```

_data/simple-icons.json
LitoMore: Sure. We can remove those icons with another pull request.

_data/simple-icons.json
LitoMore: np. Let's do this in another pull request. On this pull request, I just want to drop all `#FFFFFFF` color.

_data/simple-icons.json
```diff
@@ -1445,7 +1445,7 @@
         },
         {
             "title": "Backendless",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 1448:
```suggestion
            "hex": "5056E5",
```
They use this color a lot on their website. Might be a better choice, what do you think?

_data/simple-icons.json
```diff
@@ -4379,7 +4379,7 @@
         },
         {
             "title": "Fluxus",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 4382: Looking at https://fluxteam.net/, `#000000` does not really represents the brand. I would go for something more purpleish like `#8D73B0` or one of the dark blue value found as background.

_data/simple-icons.json
```diff
@@ -5666,7 +5666,7 @@
         },
         {
             "title": "Honor",
-            "hex": "FFFFFF",
+            "hex": "000000",
             "source": "https://www.hihonor.com"
```
service-paradis, line 5670: Vector logos are available at https://www.hihonor.com/global/brand-guideline11/basics/logo. Could also be added as guidelines. According to this document, `#000000` is the correct color to use in this case!

_data/simple-icons.json
```diff
@@ -6954,7 +6954,7 @@
         },
         {
             "title": "Leanpub",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 6957: No real use of black for their website. If we want to use a darker color instead of white, we might use the color of their header and footer (`#222222`).

_data/simple-icons.json
```diff
@@ -6976,7 +6976,7 @@
         },
         {
             "title": "Lemmy",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 6979:
```suggestion
            "hex": "14854F",
```
Might be a better choice in this case. What do you think?

_data/simple-icons.json
```diff
@@ -8918,7 +8918,7 @@
         },
         {
             "title": "OpenJDK",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 8921: Not sure black is really representative for this icon. I think we could probably use `#437291` or `#E76f00`.
What do you think?

_data/simple-icons.json
```diff
@@ -10541,7 +10541,7 @@
         },
         {
             "title": "Reason Studios",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
service-paradis, line 10544: I think a shade of yellow like `#F9CB61` might be a better choice. What do you think?

service-paradis (changes requested):
Added some more comments. The other changes looks fine to me!

_data/simple-icons.json
LitoMore: I prefer to understand that this icon has no theme color. And the `#000000` would be better. The more default scene to use our icons is a white background, if the icon is white there will be visual problems.

service-paradis: The color is there to represent the brand color. Not necessarily the icon color (even if it often correspond). In this case, the brand seems to use some color scheme, but never uses black.
We can always ask other @simple-icons/maintainers to weight in on this one!

runxel: I'm in for the purple!

_data/simple-icons.json
LitoMore: Sure.

_data/simple-icons.json
LitoMore: I think black is better. It's head and hands are in black.

adamrusted: The blue and orange are both prominently used [on their GitHub profile](https://github.com/openjdk) alongside the black/white/red mascot. Personally, I'd go with `#437291` given it's used more in the GitHub profile picture - and is the bolded text in the text logo.

LitoMore: I don't think we can tamper with the skin color of their mascot, which is black.

adamrusted: Much like @service-paradis though, I see the hex we provide as a representation of the brand, more than a representation of the mascot that accompanies it. Given the blue and orange are used prominently (albeit behind the mascot in some) I would say either of those is the 'brand colour' over black. The mascot will _display_ as white/black in our previewer anyway!

runxel: This is the pale blue for sure given how ["JDK"](https://openjdk.org/groups/build/doc/building.html) is written in it, too!

_data/simple-icons.json
```diff
@@ -8918,7 +8919,7 @@
         },
         {
             "title": "OpenJDK",
-            "hex": "FFFFFF",
+            "hex": "000000",
```
adamrusted, line 8922:
```suggestion
            "hex": "437291",
```

Just this, and then I think we're good to go on this one!

adamrusted (approved):
LGTM! Thanks @LitoMore.
Will wait for another of the requested reviewers to confirm they're happy before we merge though.

runxel (approved):
LGTM! Looks all good to me

runxel:
Only thing is, that LGTM.com is no longer active...
We should remove this in the next version!

adamrusted (approved):
Let's get this merged in then! I'll get a PR open (if there isn't one already) to remove LGTM @runxel

PR #9373, opened 2023-08-22
Add Microchip to excluded brands

adamrusted:
Have heard a pretty strong "no" back from Microchip when asking whether we can include their brands. PR serves to add them to the list of protective brands, and closed down any PRs / issues relating to them.


LitoMore:
I would call them `excluded` in the commit message.

adamrusted:
Fixed it @LitoMore - only reference now is the branch name.

LitoMore (approved):

PR #9300, opened 2023-08-16
A Review - Part 1

adamrusted:
Various changes contributing to the great icon review.
Only one icon change, to update casing in the title.

Issue #8734, opened 2023-06-06
Release action fails


LitoMore:
See https://github.com/simple-icons/simple-icons/actions/runs/5166350721/jobs/9317354356

@mondeja Could you help to check this bug?

It seems our https://github.com/simple-icons/simple-icons is not using the latest version of `release-action`. Can we bump the `release-action` to latest?

mondeja:
The action is using the previous version because the new doesn't works. I've reverted to the old commit and moved that history to https://github.com/simple-icons/release-action/pull/74

Sorry, I hadn't enough time to completely update the action.

mondeja:
Maybe is a network error calling `client.rest.pulls.listFiles()`, in that context `file.patch` must have content. Have you tried a retry?

LitoMore:
I've retried several times but still fails.

LitoMore:
I will look into the issue this weekend.

LitoMore:
This is related to the GitHub REST API:

The issue can be found with:

```sh
gh api \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/simple-icons/simple-icons/pulls/8692/files
```

Here is the response:

```json
[
  {
    "sha": "6a4bf17024d49edcb7abaae8fe323af87d508e97",
    "filename": "_data/simple-icons.json",
    "status": "modified",
    "additions": 718,
    "deletions": 718,
    "changes": 1436,
    "blob_url": "https://github.com/simple-icons/simple-icons/blob/e4443eda351184e618ac6519089ec98be18a937c/_data%2Fsimple-icons.json",
    "raw_url": "https://github.com/simple-icons/simple-icons/raw/e4443eda351184e618ac6519089ec98be18a937c/_data%2Fsimple-icons.json",
    "contents_url": "https://api.github.com/repos/simple-icons/simple-icons/contents/_data%2Fsimple-icons.json?ref=e4443eda351184e618ac6519089ec98be18a937c"
  },
  {
    "sha": "77d10bb5463bdefafa1f9b2b5b5eeeb0dd7e14b3",
    "filename": "scripts/lint/ourlint.js",
    "status": "modified",
    "additions": 28,
    "deletions": 0,
    "changes": 28,
    "blob_url": "https://github.com/simple-icons/simple-icons/blob/e4443eda351184e618ac6519089ec98be18a937c/scripts%2Flint%2Fourlint.js",
[...]

PR #9375, opened 2023-08-22
Allow skipping release note with `skip release note` label

LitoMore:
Related to https://github.com/simple-icons/simple-icons/pull/9300#issuecomment-1688461364.

Related commit: https://github.com/LitoMore/simple-icons-release-action/commit/1a0c7bb3a4a694ba3877d92e340c01a7a56cc730

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -23,7 +23,7 @@ jobs:
       did-create-pr: ${{ steps.release.outputs.did-create-pr }}
       new-version: ${{ steps.release.outputs.new-version }}
     steps:
-      - uses: LitoMore/simple-icons-release-action@9969e1266c9e470fe015dbd825a2897423e9270e
+      - uses: LitoMore/simple-icons-release-action@1a0c7bb3a4a694ba3877d92e340c01a7a56cc730
         id: release
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}


adamrusted:
This looks good to me. Any reason we're using your version of the release action instead of the one managed by the org?

LitoMore:
@adamrusted See this https://github.com/simple-icons/simple-icons/issues/8734#is.suecomment-1578410818

adamrusted (approved):
Happy to merge this in, given all we're discussing here is the version number. Would be good to get back to the org-managed release action at some point, but not today!

mondeja:
🤔 This looks strange to me. I though that all icons with data updated should be marked as updated icons after a "Icon data updated" header in release notes (see [9.8.0](https://github.com/simple-icons/simple-icons/releases/tag/9.8.0)). Why is not the case for the icons in this letters review? @LitoMore @adamrusted

LitoMore:
@mondeja They are only guidelines updates.

Here is my comment: https://github.com/simple-icons/simple-icons/pull/9300#issuecomment-1688461364

mondeja:
But are icon data updates, right? What should be considered as icon data updates instead?

adamrusted:
This is potentially where we need to get a bit more granular with the big reviews.

Just removing slashes from URLs? Great, skip the release note. Nobody wants potentially 1000+ icons in that list because you've reformatted the exact same URL.

Make fundamental changes to the source and/or data? That's gotta be in the release note.

PR #9377, opened 2023-08-22
Metadata update for the Y review

service-paradis:
Part or #9072


service-paradis:
> @service-paradis - is this ready for merging? If so, I'll review it shortly!

Yes, should be good!

PR #9430, opened 2023-08-28
Update Rocket.Chat

adamrusted:
Tinkers with the Rocket.Chat path, in theory removing the need for the `.svglint-ignored.json` file (for now, at least!)


service-paradis:
@adamrusted you can take a look at https://github.com/simple-icons/simple-icons/pull/5190#issuecomment-792842559. Unfortunately, it looks like we need to keep this one there!

PR #9480, opened 2023-09-03
Add ko-KR, ja-JP localization

sh-cho:
### Description

I wonder if it is correct to add them to the LOC if they are pronounced similarly but use different characters.

For example, "Kakao" is pronounced "카카오" in Korean, but the pronunciation is almost the same.


adamrusted:
More than happy to include some more translations! My view is if the localised display name of a brand is any different to that which is already included, we should totally include it as a localisation.

@sh-cho - would you also like to check over the Korean translation for our [website repository](https://github.com/simple-icons/simple-icons-website)?

sh-cho:
Sure :)

adamrusted:
I'll also leave this open for discussion for now, if there are any other brands that _very obviously_ have a Korean or Japanese alternative name, be good to rattle through as many as we can 😅

adamrusted (approved):
In fact, let's get these updates merged in for now. Any more we find further down the line can be added in with their own PRs! Thanks for your work on this @sh-cho! 🎉

PR #9487, opened 2023-09-03
Add `[Icon request]` to bot ignore.

adamrusted:
Adds `[Icon request]` to the duplicate detection rules, ensuring requests starting with that string don't get confused by the bot.

### DIFF
--- a/.github/workflows/potential-duplicates.yml
+++ b/.github/workflows/potential-duplicates.yml
@@ -34,6 +34,7 @@ jobs:
             assets
             for
             from
+            [Icon request]
           state: all
           threshold: 0.7
           comment: |


mondeja (changes requested):
This would raise a syntax error on the potential duplicates action. You can test [the logic](https://github.com/wow-actions/potential-duplicates/blob/34fe38ec333ab6f2af4000210ba41824d8a61edf/src/util.ts#L40-L54) with:

```js
> 'icon\nrequest\n[Icon request]'.split(/[\s\n]+/).map((keyword) => keyword.trim()).filter((keyword) => keyword.length > 0).reduce((memo, keyword) => memo.replace(new RegExp(keyword, 'igm'), ''), '[icon request] hello world')
Uncaught SyntaxError: Invalid regular expression: /[Icon/: Unterminated character class
```

I think that we should add `\[` and `\]` but not sure if we should add two escape characters `\` or one. Even better, we must escape regexes in the action to avoid errors.

See:

```js
> 'icon\nrequest\n\\[\n\\]'.split(/[\s\n]+/).map((keyword) => keyword.trim()).filter((keyword) => keyword.length > 0).reduce((memo, keyword) => memo.replace(new RegExp(keyword, 'igm'), ''), '[icon request] hello world')
'  hello world'
```

That's what you are trying to do, right?

mondeja:
Opened https://github.com/wow-actions/potential-duplicates/issues/24 to track this problem in the action.

PeterShaggyNoble:
Converting this to a draft pending resolution of the above.

PR #9498, opened 2023-09-06
Update icon count in readme to over 2700

DenverCoder1:
There are now 2700+ icons on the develop branch


service-paradis:
Thanks for your PR. But, it should be updated automatically with the new release:
https://github.com/simple-icons/simple-icons/blob/develop/scripts/release/update-svgs-count.js

PR #9503, opened 2023-09-06
Add Ubuntu MATE icon

nfoert:
ubuntumate_preview2

**Issue:** I didn't make an issue, however it's for [this](https://github.com/Ileriayo/markdown-badges/issues/550) badge request.

**Similarweb rank:** [235,790](https://www.similarweb.com/website/ubuntu-mate.org/#overview)

  

### Checklist

### Description
- Image source [here](https://github.com/ubuntu-mate/ubuntu-mate-artwork/blob/7920d3f2c810108e3032a3dc31335d145d785645/usr/share/icons/Yaru-MATE-light/256x256%402x/places/start-here.png)
- HEX color was just grabbed from this image using an online tool

### Help Required
I'm having an issue with the .svg. I'm sure you can see that the positioning for the image in the preview is offset, and I can't figure out why. Running `npm run lint` results in this error message:
```
x icon-centered <path> must be centered at (12, 12); the center is currently (154.103, 147.4)
```
I followed the directions from `CONTRIBUTING.md` and am unsure what value to change to resolve this issue.
Can I please be advised on how to fix this issue? I'm using Inkscape 1.3.


adamrusted:
Ah, looks to me like the path was using a transform, so had kept the original size, even though you'd shrunk it down. I'm unsure how to resolve this in Inkscape specifically, but in other programs I've been able to resolve by ungrouping everything, so you're working directly on the canvas with a single path, and then setting the size.

adamrusted:
Ah, looks to me like the path was using a transform, so had kept the original size, even though you'd shrunk it down. I'm unsure how to resolve this in Inkscape specifically, but in other programs I've been able to resolve by ungrouping everything, so you're working directly on the canvas with a single path, and then setting the size.

_data/simple-icons.json
```diff
@@ -13644,6 +13644,11 @@
             "source": "https://design.ubuntu.com/resources",
             "guidelines": "https://design.ubuntu.com/brand"
         },
+        {
+            "title": "Ubuntu MATE",
+            "hex": "84A454",
+            "source": "https://github.com/ubuntu-mate/ubuntu-mate-artwork/blob/7920d3f2c810108e3032a3dc31335d145d785645/usr/share/icons/Yaru-MATE-light/256x256%402x/places/start-here.png"
```
adamrusted, line 13650: There is an SVG available on the website homepage. We should probably use that as the source here.

adamrusted (changes requested):

nfoert:
ubuntumate
> Ah, looks to me like the path was using a transform, so had kept the original size, even though you'd shrunk it down. I'm unsure how to resolve this in Inkscape specifically, but in other programs I've been able to resolve by ungrouping everything, so you're working directly on the canvas with a single path, and then setting the size.

Thanks for the response. I seem to have it looking properly now. I'm not sure why I didn't try to save the image from the main page, but I switched to use that one instead.

Here's the updated preview image.

_data/simple-icons.json
nfoert: This is outdated, because I've changed it to the correct URL.

_data/simple-icons.json
```diff
@@ -13644,6 +13644,11 @@
             "source": "https://design.ubuntu.com/resources",
             "guidelines": "https://design.ubuntu.com/brand"
         },
+        {
+            "title": "Ubuntu MATE",
+            "hex": "84A454",
+            "source": "https://ubuntu-mate.org/images/ubuntu-mate/logo-white.svg"
```
adamrusted, line 13650:
```suggestion
            "source": "https://ubuntu-mate.org"
```

adamrusted (commented):
Just a quick change to the JSON, and this is grand! Thanks for your work on this @nfoert!

nfoert:
No problem! Thanks for giving me an opportunity to learn more about contributing.

PR #9508, opened 2023-09-07
Part 1 of the letter T review

service-paradis:
Part of #9072

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -13041,7 +13041,7 @@
         {
             "title": "Telegram",
             "hex": "26A5E4",
-            "source": "https://telegram.org"
+            "source": "https://telegram.org/tour/screenshots"
         },
         {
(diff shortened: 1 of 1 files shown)


adamrusted:
As discussed in #9375 - I think anything where we're altering the source / guidelines should be included in the release note. If you wanted to split out the removal of trailing slashes into another PR that we could skip the release note on, that's chill.

service-paradis:
Totally missed that! Make sense to me though!
I will split the PRs!

adamrusted:
Trailforks also appears to just be slash slashing @service-paradis  - could you bin that one from this PR, and then we can merge!

service-paradis:
Too many slashes to slash... I'm seeing slashes everywhere!

adamrusted:
> I'm seeing slashes everywhere!

SLASH 'EM!

PR #9535, opened 2023-09-12
Added azure blob icon

D4n1aLLL:
**Issue:** closes #

**Similarweb rank:**
  

### Checklist

### Description
I found the icon here: https://azure.microsoft.com/content/dam/microsoft/legacy/acom/svg/h-z/storage-blobs.svg
I used the color from the svg it self.

I am new here and a bit concerned about the white rectangle in original SVG

azureblobs


_data/simple-icons.json
```diff
@@ -1567,6 +1567,11 @@
             "hex": "CB2E6D",
             "source": "https://azure.microsoft.com/products/devops"
         },
+        {
+            "title": "Azure Blobs",
+            "hex": "0072C6",
+            "source": "https://azure.microsoft.com/content/dam/microsoft/legacy/acom/svg/h-z/storage-blobs.svg"
```
adamrusted, line 1573:
```suggestion
            "source": "https://learn.microsoft.com/azure/architecture/icons"
```

If the icon is available here, we should probably use this source. A lot of our other Azure icons come from there.

adamrusted:
Looks like there's not a particular icon for 'Azure Blob Storage' inside that package I sent across. If I remember correctly, it's just a 'type' under Azure Storage Accounts - of which there _is_ an icon we could include.

adamrusted:
Difficult to determine what to call the icon then, as I can't see it actively used in the Azure Portal.
I'm seeing `Azure Storage Account` - and there's 'Containers' or 'Blob Inventory' under that.
I guess `Azure Storage - Blob Block` or `Azure Blob Storage` would be a good name for it.

D4n1aLLL:
> Difficult to determine what to call the icon then, as I can't see it actively used in the Azure Portal. I'm seeing `Azure Storage Account` - and there's 'Containers' or 'Blob Inventory' under that. I guess `Azure Storage - Blob Block` or `Azure Blob Storage` would be a good name for it.

It is called Azure Blob(s) Storage, depending on the context Microsoft uses Azure Blob Storage/Azure Blob/Blob Storage

https://azure.microsoft.com/en-us/products/
https://azure.microsoft.com/en-us/free/#all-free-services

adamrusted:
Let's go with `Azure Blob Storage` for now. I've personally seen it referred to as that in most projects involving it. Also gives us the most keywords for people to search it with 😅

adamrusted:
So in our contribution guidelines, we suggest using cut-outs and outlines to differentiate between the colours. See the Power Platform related icons for examples!

adamrusted:
Nice work @D4n1aLLL - though at smaller sizes, the browser window gets a bit lost. If you want, I can have a look at how _I'd_ tackle this particular one, and share a preview with you? Won't be straight away, but I'll get round to it today.

adamrusted:
Perhaps something like this, @D4n1aLLL?

azureblobstorage

adamrusted:
image
Updated PR to use above icon. For comparison, download the ZIP from Microsoft's site and use:
`Azure_Public_Service_Icons\Icons\general\10780-icon-service-Blob-Block.svg`

PeterShaggyNoble:
Converted this to a draft in order to postpone any further action on it until after v11 is released and we've made contact with Microsoft. See #10016 for more details.

PR #9536, opened 2023-09-12
Add `new` to Potential Duplicates workflow

service-paradis:
The word new is recently used often in icon request. It should be part of the ignored list of the Potential Duplicates workflow.


adamrusted (approved):
LGTM! Thanks @service-paradis

PR #9564, opened 2023-09-16
Add BitTorrent

### LINKED ISSUE #6246: Request to add torrent clients
- [ ] BitComet: https://en.wikipedia.org/wiki/BitComet
Alexa rank: 62777

- [ ] BItTorrent: https://en.wikipedia.org/wiki/BitTorrent_(software)
Alexa rank: 4104

- [ ] uTorrent: https://en.wikipedia.org/wiki/%CE%9CTorrent
Alexa rank : 705

- [ ] Transmission: https://en.wikipedia.org/wiki/Transmission_(BitTorrent_client)
Alexa rank: 139777

So, I think we can all agree that like 99% of applications of the bitTorrent protocol is illegal downloads and piracy. Regardless of my personal opinions on the matter, I think that shouldn't matter, using the same criteria used to allow fileshare services and OnlyFans: Even if 99% of the content is illegal / unsavory / not preferable, there are plausible legal applications and they are not illegal nor promoted for illegal uses.

If this is approved, I will take them since I'm personally interested in having them added. Waiting on feedback...

PratyushJayachandran:
bittorrent

**Issue:** closes #6246 

**Similarweb rank:** 17,375. See https://www.similarweb.com/website/bittorrent.com

### Checklist

### Description
SVG: Taken from the website front page
image
Color: `#5d44db` Taken from one of the website elements. Their presskit logo only contains BitTorrent classic in a dull purple. I chose this one as it represents the old software icon very closely.


PeterShaggyNoble (changes requested):
While the icon in the PR looks good when compared to the source file; it looks to be like the source file is an alternative version for use in specific situations and that we should actually be using the filled version as it's used for all their products [here](https://www.bittorrent.com/products/).

I also think we should use black for the colour, the purple's only used as an accent colour on their site.

We can also omit the `guidelines` entry from the JSON as their are no guidelines available in their press kit, just logos.

PratyushJayachandran:
> While the icon in the PR looks good when compared to the source file; it looks to be like the source file is an alternative version for use in specific situations and that we should actually be using the filled version as it's used for all their products [here](https://www.bittorrent.com/products/).
> 
> I also think we should use black for the colour, the purple's only used as an accent colour on their site.
> 
> We can also omit the `guidelines` entry from the JSON as their are no guidelines available in their press kit, just logos.

I changed the source to https://www.bittorrent.com/products and re did the icon. Please review again @PeterShaggyNoble

_data/simple-icons.json
```diff
@@ -1863,6 +1863,12 @@
             "hex": "683D87",
             "source": "https://www.bitrise.io/presskit"
         },
+        {
+            "title": "BitTorrent",
+            "hex": "5D44DB",
```
PeterShaggyNoble, line 1868:
```suggestion
            "hex": "050505",
```

PeterShaggyNoble (approved):
SVG looks good now; just need to update the colour and this is ready to merge.

PR #9640, opened 2023-09-26
Remove Jenkins X

jwr1:
Issue: contributes to https://github.com/simple-icons/simple-icons/issues/9072

* Jenkins X - Remove icon: similarweb rank ([~2,823,000](https://www.similarweb.com/website/jenkins-x.io/#overview)) and GitHub ([4.4k](https://github.com/jenkins-x/jx)) is slightly below requirement (needs discussion)

PR #9639, opened 2023-09-26
Remove Jabber

jwr1:
**Issue:** contributes to #9072

* Jabber - Remove icon: similarweb rank ([~3,335,000](https://www.similarweb.com/website/jabber.org/#overview)) is below requirement

PR #9642, opened 2023-09-26
Remove JPEG

jwr1:
**Issue:** contributes to https://github.com/simple-icons/simple-icons/issues/9072

* JPEG - Remove icon: similarweb rank ([~2,079,000](https://www.similarweb.com/website/jpeg.org/#overview)) is below requirement (needs discussion)


ThatOneCalculator:
JPEG is one of the most commonly used image formats in all of computing. Just because its seldom-viewed homepage has gone down in rank doesn't mean that the entire format has gone down in popularity.

jwr1:
I understand the JPEG format is popular (I use it myself, and it's one of the most used image formats), but I wasn't sure how strongly you all followed the similarweb rank. This is why I mentioned this would **need discussion** before merging or closing.

@LitoMore, how do we remove icons, if this isn't correct? With #9639, for instance, jabber.org is completely out of the accepted rank: [2,777,993](https://www.similarweb.com/website/jabber.org/#overview), and I removed both the `_data/simple-icons.json` entry, and it's SVG file, all in a separate pull request; I'm not sure how to do it any better than that. As a reference, #9370, did the exact same thing; the website, webhint.io, has a similarweb rank of [1,737,615](https://www.similarweb.com/website/webhint.io/#overview) and not enough GitHub stars, so it's completely valid that it should be removed. For my other pull request, #9640, it's only a little out of the rank required, so I also said it'd need discussion.

PR #9651, opened 2023-09-27
Update Husqvarna

PratyushJayachandran:
husqvarna

**Similarweb rank:** 21,548. See https://www.similarweb.com/website/husqvarna.com

### Checklist

### Description
SVG: I am confused. When I compare I see a difference but when I re-do it results in the same code. The logo is now available on homepage so I updated that.


adamrusted (changes requested):
There is a _slight difference_ between the icon already in our collection and the source icon, assuming you use `husqvarna-logo-text` as your source from within the spritesheet. You can also trim the `/us` off the URL, as the SVG logo is available on the landing page.

PratyushJayachandran:
> There is a _slight difference_ between the icon already in our collection and the source icon, assuming you use `husqvarna-logo-text` as your source from within the spritesheet.
> 
> You can also trim the `/us` off the URL, as the SVG logo is available on the landing page.

I have updated the logo from https://www.husqvarna.com
Attaching the SVG here. I had to copy each path from HTLM and craft the image as my plugin was unable to fetch the image as SVG from the website.
husqvarna-original

adamrusted (approved):

PR #9696, opened 2023-10-08
Make width & height check also take iconTolerance into account

ThatOneCalculator:
**Issue:** ref: https://github.com/simple-icons/simple-icons/pull/9685#issuecomment-1751938939

### Description

`iconTolerance` is defined in the svglintrc to account for floating point tolerance, but is only used for path centering, not size. This PR makes it so that `iconTolerance` is also taken into account for sizes that report `24.001` or `23.999`


LitoMore:
Closing due to this change will make the icon overflow.

PR #9717, opened 2023-10-12
Add Alamy

zcraber:
alamy

**Similarweb rank:** [3,734](https://www.similarweb.com/website/alamy.com)

### Checklist

### Description
Logo taken from the website header. Brand colour `#00ff7b` picked from the website.


PeterShaggyNoble:
I think we can get away with using the [encircled "a"](https://s.alamy.com/logos/1.68.0/safari-pinned-tab.svg) here, as they use it themselves all over social media and it's better suited to our purposes as an icon library. What do you think, @zcraber?

zcraber:
> What do you think?

@PeterShaggyNoble Sounds great, I'll work on it! Actually, I did search for that 'a' icon before, but couldn't find the vector version.

zcraber:
@PeterShaggyNoble Done!

adamrusted (approved):

PeterShaggyNoble:
> Actually, I did search for that 'a' icon before, but couldn't find the vector version.

Spend enough time here and you'll get to the point where you're able to find vector artwork even when you're not looking for it! 😆

PR #9719, opened 2023-10-13
Add Picxy

zcraber:
picxy

**Similarweb rank:** [428,601](https://www.similarweb.com/website/picxy.com/#overview)

### Checklist

### Description
Colour `#2e3192` was picked from the website.


PeterShaggyNoble:
@zcraber, do you have an example of them using just the "P" anywhere, other than their favicon? It looks to me like they use the full logo pretty much everywhere.

zcraber:
> do you have an example of them using just the "P" anywhere, other than their favicon?

They have used it in their [X profile](https://twitter.com/picxydotcom/photo). @PeterShaggyNoble

adamrusted (changes requested):
I think I agree with @PeterShaggyNoble - they use the full wordmark on _every_ other social platform.
Could you re-work this please @zcraber?

zcraber:
@adamrusted Done!

PeterShaggyNoble (changes requested):
Everything looks perfect to me now, thanks, @zcraber 👍🏻 But, before we merge this though, would you be up to trying to clean out some of the excess points in the SVG?

PeterShaggyNoble:
Oh, wow, it's only now I zoom in fully on the original that I see just how bad it is! 😮 It actually looks like it's the result of running a raster image through a converter. We do try to walk a line between maintaining the integrity of the original icon while removing as many excess points as we can and fixing any minor imperfections that are obviously unintentional.

In this case, though, as the original is _so_ bad, we can't tell exactly what it's _supposed_ to look like. And my own attempts to delete excess points or fix imperfections meet with the same results as yours, introducing some pretty glaring differences.

So, seeing as it currently lines up perfectly with the original, I say we run with this as-is for now in the interests of getting this PR merged, and we can take another pass at it in more detail in the future.

PR #9753, opened 2023-10-18
Add BBC News

RobertHarnett:
bbcnews

**Issue:** closes #

**Similarweb rank:**
[bbc.co.uk - 88](https://www.similarweb.com/website/bbc.co.uk/#overview)
[bbcnews.com - 55,572](https://www.similarweb.com/website/bbcnews.com/#overview)
[bbc.news - 113,211](https://www.similarweb.com/website/bbc.news/#overview)

### Checklist

### Description

I have soured directly from the [BBC News page](https://www.bbc.co.uk/news).
The specific shade of red they use seems to vary; I picked the most prominent one in the header of the page, which is also the same one they use throughout their app.


adamrusted:
As per [their guidelines](https://www.bbc.co.uk/branding/logo-use) we need to seek permission to include this before we add it. This was discovered as part of the `Letter B Review` that took place as part of #9072. If we do not seek or get granted permission, then we won't be able to include this - and will also have to remove BBC and BBC iPlayer as of v10.

PR #9792, opened 2023-10-29
Replace `rimraf` with `del-cli`

LitoMore:
### Comparison

See https://github.com/sindresorhus/del-cli?tab=readme-ov-file#comparison


sisp (approved):
LGTM 👌

PR #9810, opened 2023-11-04
Fix workflows

LitoMore:


mondeja:
I've received the same error message in another project. Seems that does not happen when the action is composed. I'll send a PR in a few moments...

PR #9832, opened 2023-11-07
Remove Angular Universal

adamrusted:
Angular Universal is now part of the core CLI package. No longer operates as a separate 'product'.


adamrusted:
Bear with. I don't know why this is breaking _everything_

PeterShaggyNoble (approved):

PR #9836, opened 2023-11-08
Remove YOLO

service-paradis:
**Issue:** Part of #9072

I'm not sure it fits our requirements. While https://www.similarweb.com/website/pjreddie.com/#overview is almost correct, it does not gives us a metric for the popularity of YOLO itself. We need to take a decision about keeping this! (if we keep it, we need to remove trailing slash)


adamrusted:
Don't think much discussion will be required on this one - it was added before our inclusion criteria was refined. If there's no obvious metric we can measure 'YOLO' against - we should remove it.

adamrusted (approved):

PR #9846, opened 2023-11-12
Use `save-exact` for dependencies

LitoMore:
### Documentation

https://docs.npmjs.com/cli/v10/using-npm/config#save-exact


mondeja (approved):
Great. The error in the workflow will be fixed by #9849

PR #9848, opened 2023-11-12
Fix Autocloser Workflow

adamrusted:
The autocloser action has been failing, see https://github.com/simple-icons/simple-icons/actions/runs/6840581192.
PR serves to correct workflow.

### DIFF
--- a/.github/actions/get-labels/action.yml
+++ b/.github/actions/get-labels/action.yml
@@ -16,5 +16,5 @@ runs:
     - id: get-labels
       shell: sh
       run: |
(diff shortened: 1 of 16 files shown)


mondeja (changes requested):
That is part of the problem but, additionally, a previous checkout step is missing.

https://github.com/simple-icons/simple-icons/blob/5794f411a1c82115516e31ba8414b5ad3812c790/.github/workflows/add-labels-priority.yml#L23-L24

mondeja (approved):

Issue #9855, opened 2023-11-14
Update: Deezer Icon


sebastien46:
### Brand Name

Deezer

### Official Resources for Icon and Color

- [Announcement](https://newsroom-deezer.com/2023/11/deezer-new-brand-identity-and-logo)
- [Brand Guidelines reference](https://developers.deezer.com/guidelines/logo) [(main site currently non-functional)](https://deezerbrand.com)
- Logo with text svg: https://newsroom-deezer.com/wp-content/uploads/2023/11/logodeezer.svg
![](https://newsroom-deezer.com/wp-content/uploads/2023/11/logodeezer.svg)
- Color hex: #A238FF

### Additional Comments

- Two variants, light & dark, but only the text color changes, so it's irrelevant.
- From the [Terms of Service:](https://www.deezer.com/legal/cgu)

> Article 8 – Intellectual property
> The general structure of the Site, the Application, the Deezer Free Service and all elements composing it (such as but not limited to the logos, domain names, tracks or videos, and their associated elements such as the photos, images, texts and biography of the authors, artists and/or any legal beneficiary on the tracks or videos, as well as the visuals on the packaging) are the exclusive property of DEEZER and/or its licensors (notably the record and video producers, the record companies, authors’ rights societies, etc.).
> These elements are protected by intellectual property law and other laws, and particularly copyright. You may only use these elements for the purpose of using the Deezer Free Service, in accordance with the Terms.
> Any representation of all or part of the Site, the Application, the Deezer Free Service and/or the elements composing them (as described above), by any means, without the express consent of DEEZER, is therefore forbidden and would be treated as an infringement punishable by law.

adamrusted:
Thanks for this @sebastien46 - though looking at those terms of service we may need to remove the icon if we can't demonstrate we have explicit permission to include it. Would you be able to reach out to them seeking permission to continue including the icon? I'll generate a 'removal' PR just in case the answer is no, given our deadline for v10 removals is the last release of this month.

adamrusted:
Were you able to reach out to Deezer, @sebastien46?

sebastien46:
> Were you able to reach out to Deezer, @sebastien46?

Yes, but nothing came up unfortunately.

PeterShaggyNoble:
It can take a good while for some of these companies to respond to _anything_, so let's give it more time.

tagliala:
[...]

PR #9856, opened 2023-11-14
Remove Deezer

adamrusted:
Further to #9855 - this PR serves to remove Deezer if we are not granted permission ahead of v10 release date.


PeterShaggyNoble (approved):

service-paradis:
This one is not clear to me. This page contains at least 5 sections about Intellectual Properties about logos. Each time, for a specific service (User of the Free Service, User of the Premium Service, Developers, etc.).

There are no generic terms of service for other usage... In fact, it would be probably preferable to reach to them, but I'm not sure if the removal is that urgent...

PR #9884, opened 2023-11-20
Update Write.as

service-paradis:
image
> ℹ️ @adamrusted
> Updated PR to reflect change in icon as opposed to removal.

Part of #9072

According to their guidelines (https://write.as/brand):

> Do not alter the logos above (color, proportion, etc.).

Since they do not have any monochrome version available, it looks like we need to remove this one...

Maybe we could get away with the full wordmark?


PeterShaggyNoble:
Odd that they have a monochrome treatment of the logo but not of the icon. Might be worth contacting them before taking any action to check.

I'd be in favour of the full logo over a complete removal, though.

service-paradis:
> Odd that they have a monochrome treatment of the logo but not of the icon.

Yes, it is pretty odd. I'm not even sure it really is an official monochrome treatment or it it absolutely needs to be white on blue when using it.

As you say, we might try to contact them!

adamrusted:
@service-paradis and @PeterShaggyNoble - are you happy to remove this for now, and if we obtain permission, re-add?

PeterShaggyNoble:
My vote would be to update to the full logo while we attempt to contact them for clarification.

adamrusted:
Have updated the icon, and removed the milestone. Please review @service-paradis / @PeterShaggyNoble

PeterShaggyNoble (approved):
A lot of excess points in this one, but it will do while we wait for clarification from them as to which we should be using. If they say we should be using the wordmark, then we can take another pass at this to clean it up.

PR #9885, opened 2023-11-20
Remove W3C icon

service-paradis:
Part of #9072

Unfortunately, a lot of things at https://www.w3.org/policies/logos/ make me think we cannot have this logo right now...


PeterShaggyNoble:
The only thing I'm seeing on that page that would suggest we can't use their icon is the FAQ below, which implies the colours cannot be changed. But, as it's not _explicit_ and in the interests of trying to avoid an icon, it might be worth contacting them to confirm that.

> **Can I change a W3C logo to fit the look and feel of my site?**
> No.
> The W3C logo provides the organization with its visual identity. As a result, there are strict rules for its use. Changing the W3C logo is not permitted, as it defeats the purpose of the logo, which is to be noticed and identified as W3C.

service-paradis:
> When used on the Web, the icon must be an active link to https://www.w3.org/.

I dont know how we can achieve this on our website

> You may only post the graphic found at https://www.w3.org/Icons/w3c_home.

Since they do not have any monochrome version, I dont know how we respect that. In fact, we would at least need the ®️


Maybe we could send a request at w3t-pr@w3.org and postponed the removal to v11?

adamrusted:
> Maybe we could send a request at [w3t-pr@w3.org](mailto:w3t-pr@w3.org) and postponed the removal to v11?

I would rather remove it, and re-add if we obtain permission / a monochrome version.

adamrusted:
> Are we also attempting to contact all the other brands we're removing dues to issues with permissions or modifying logos to try to seek their approval for (re-)including them?

We're erring on the side of caution and getting removal PRs open for any that require permission for inclusion. That way, we can promote discussion on the removal PR, and if someone is willing to get in touch with the brands, they can do so there.

adamrusted (approved):

PR #9936, opened 2023-11-27
Update Webex icon color

LitoMore:
Use black instead of white as the brand color for a better default visual.


adamrusted (approved):

service-paradis:
A little late for this one, but not sure black is really representative here. Navigating through https://www.webex.com/, I think `#078855` might be a better choice.

PR #9937, opened 2023-11-27
Avoid `#FFFFFF` icon color

LitoMore:
Related to #9360

The CI will pass once #9936 is merged.


mondeja:
Why you don't just added it to the `hex` test? That would avoid another long loop.

https://github.com/simple-icons/simple-icons/blob/15cc8ea7fa21cad39987a7393715dbc756992205/tests/test-icon.js#L41

PeterShaggyNoble:
What if a brand's sole primary colour is _actually_ white, though?

I'm struggling to think of an example off the top of my head, but I think we'd need to provide the ability to override this, just in case. Perhaps by having a list of excluded slugs, in the same way we did when we first introduced SVG linting to prevent all our existing icons from throwing errors while we worked on cleaning them up?

scripts/lint/ourlint.js
```diff
@@ -82,6 +82,19 @@ const TESTS = {
...
+  /* Check icon colors */
+  checkColor: (data) => {
+    const hasWhiteColor = (color) => {
+      return color === 'FFFFFF';
+    };
+
+    const allColors = [...new Set(data.icons.flatMap((icon) => icon.hex))];
+
+    if (allColors.some((color) => hasWhiteColor(color))) {
+      return 'Some icon colors are white `FFFFFF`, please replace with black `000000`';
```
service-paradis, line 95: > please replace with black `000000`

I would not ask specifically to replace with black since it might not be the best alternative to represent the brand

mondeja:
> What if a brand's sole primary colour is _actually_ white, though?
> 
> I'm struggling to think of an example off the top of my head, but I think we'd need to provide the ability to override this, just in case. Perhaps by having a list of excluded slugs, in the same way we did when we first introduced SVG linting to prevent all our existing icons from throwing errors while we worked on cleaning them up?

Opened https://github.com/simple-icons/simple-icons/issues/9976 to discuss it. I think that we must move forward this PR with the current approach introduced by #9360 to prevent more drift about the change and use the opened issue to discuss the problem.

scripts/lint/ourlint.js
adamrusted: @LitoMore - I don't think we explicitly need to tell users to replace with black. Perhaps we just suggest they find a more appropriate colour?

```suggestion
      return 'Some icon colors are white `FFFFFF`, are you sure that is the correct color?';
```

service-paradis:
Closing this one according to #9976

PR #9955, opened 2023-11-29
Retry checking of operative links on CI

mondeja:
This step is flaky, retry 3 times before throwing an error. [Check it in action](https://github.com/simple-icons/simple-icons/actions/runs/7029602574/job/19127518959#step:8:172).


LitoMore:
I think we can ignore some URLs if they always fail.

mondeja:
> I think we can ignore some URLs if they always fail.

The problem is that is not the case. For example, `simpleicons.org` URL has failed on #9954, that is why I've opened this PR.

mondeja:
I'm closing this, seems that is not the proper solution...

PR #10000, opened 2023-12-02
Update Readme

PratyushJayachandran:
Update icon number in Readme


PeterShaggyNoble:
Is this not still done automatically by the release action?

PratyushJayachandran:
> > Is this not still done automatically by the release action?
> 
> It should be, yeah. We were just under as of last release, so will update tomorrow! 

Oof... I didn't know that. 😞

PR #10024, opened 2023-12-04
Add Microsoft and Adobe to 'excessively protective brands' list

Lordfirespeed:


PeterShaggyNoble:
Thanks, @Lordfirespeed, but we won't be doing this until we've made contact with each company to seek permission and only then if the outcome does not go in our favour. In the meantime, we will simply be pausing all open and newly created issues & PRs until we receive a response.

Lordfirespeed:
Okay! ❤️

PR #10039, opened 2023-12-06
Split issue types in README

adamrusted:
A small update to the README, to split issues out by type.
Noted that there are a lot of users who would contribute, but don't have any interest in working on SVGs per-say.
By highlighting that we have different options available, we should (in theory) be able to appeal to those who would like to contribute, but aren't up for working on SVGs.

Also filtered to remove issues with a linked PR. We can assume they're already 'in progress'


PeterShaggyNoble (approved):
Would it be worth adding these to the relevant section of our contributing guidelines, too?

Issue #10052, opened 2023-12-09
Replace FITE with TrillerTV


PeterShaggyNoble:
FITE is [rebranding as TrillerTV](https://www.fite.tv/watch/we-become-trillertv/2pe5x/) - a terrible move in my opinion - on the 13th of December, so we'll need to add a new icon for TrillerTV and queue FITE up for removal in v12.

- [x] Add TrillerTV
- [x] Remove FITE - #10060

PR #10053, opened 2023-12-09
Remove FITE

PeterShaggyNoble:
**Issue:** #10052

### Description
FITE has been rebranded as TrillerTV


PeterShaggyNoble:
And this is why I shouldn't create PRs on my phone! 🙄 Don't kow what went wrong. Will revisit on Monday.

Issue #10073, opened 2023-12-12
Update/Review Multiple icons


PeterShaggyNoble:
I fell down a rabbit hole this morning, going through [Logopedia's archives](https://logos.fandom.com/wiki/Logopedia:Recent_logos/Archive) and found a number of icons we need to update or, at least, review. Rather than try to take them all on myself and put pressure on the @simple-icons/icon-review team or open individual issues that could get lost in the mix, I've created the following list for us all to work through together.

In some cases, the icon may _already_ have been updated and I simply missed it and in others it's only the branding, logo and/or stylisation of the icon that has been changed so our icon won't need to be updated but the colour may, but a comparison should be done anyway to be sure.

If you're taking on anything on the list then please leave a comment below so we can check it off and so nobody else picks it up in the meantime.

- [x] Asus (#10090, https://github.com/simple-icons/simple-icons/labels/pending)
- [x] Bugatti (#10099)
- [x] Channel 4 (#10100)
- [x] Citroën (#10101)
- [x] Counter-Strike (#10106)
- [x] Dailymotion (#10216)
- [x] Dolby (#10107)
- [x] Google Chat (#10257)
- [x] Google Chrome (#10116)
- [ ] Google Play
- [x] ~Google Wallet (re-add)~ (no action needed; we've never had this in our library!)
- [x] Infiniti (#10118)
- [x] Instagram (#10153)
- [x] Lada (#10155)
- [x] ~LogMeIn (rename as GoTo)~ (only rebranded in America; no action needed until GoTo is requested)
- [x] Opel (#10159)
- [x] Photobucket (#10161)
- [x] Riot Games (#10105)
- [x] Roblox (#10162)
- [x] ~Roblox Studio~ (no action needed; our icon was added after their last update)
- [x] RTL (#10166)
- [x] Škoda (#10167)
- [x] Smart (#10168)
- [x] Starz (#10169)
- [x] Tele 5 (#10170)
- [x] Wish (#10171)

adamrusted:
> I fell down a rabbit hole this morning, going through [Logopedia's archives](https://logos.fandom.com/wiki/Logopedia:Recent_logos/Archive)

Don't we all have this issue?
Joking aside though - thanks for compiling this list. Frustrating that some of these icons have changed _since_ we did our big review for the year! I'll crack on with a few of these when I get a moment.

service-paradis:
[...]

PR #10090, opened 2023-12-13
Update ASUS icon

PeterShaggyNoble:
ASUS

**Issue:** making a start on #10073

**Similarweb rank:** [742](http://similarweb.com/website/asus.com)

### Checklist

### Description
Manually vectorised from [their favicon](https://www.asus.com/new_asus_ico_256x256.png). Although I can't find this icon in use anywhere else other than for the favicon, it is apparently being used on their products.


adamrusted:
Looking on their socials, although the icon is shown on _some_ phones - it would appear they're 'launching' the brand on the 9th/10th January, at their CES event. Shall we hold off until then, just in case that _is_ the big launch, and an SVG becomes available?

adamrusted:
@PeterShaggyNoble - it would appear all products revealed as part of CES are using the existing ASUS icon. Perhaps we're a bit early to the party on this one?

PeterShaggyNoble:
I'd agree, @adamrusted; we're a few weeks removed from CES now and the only place I can see this icon in use is in their main website's favicon and in the logo for [their CES event](https://www.asus.com/event/ces), which, for me, isn't enough to warrant us updating at this time.

PR #6422, opened 2021-08-28
Source update for several icons

jorgeamadosoria:
**Issue:** #6399 

### Checklist
  - [ ] I updated the JSON data in `_data/simple-icons.json`
  - [ ] I optimized the icon with SVGO or SVGOMG
  - [ ] The SVG `viewbox` is `0 0 24 24`

### Description
Source updates in simple-icons.json. More info in the linked issue.
Basically the goal is to remove specific files linked in the source field, except in the cases of github and / or favicons.

Checklist for review:
  - [ ] Aerospike
  - [x] Alipay
  - [x] Angular
  - [ ] AniList
  - [x] Apache Cassandra
  - [ ] ARAL
  - [x] Arduino
  - [x] Bose
  - [ ] Broadcom
  - [ ] CSharp
  - [ ] Chevrolet
  - [ ] Citroën
  - [x] Conda-Forge
  - [ ] DasErste
  - [ ] Data Version Control
  - [ ] DLNA
  - [ ] Facepunch
  - [ ] Fila
  - [ ] Fujifilm
  - [ ] Gmail
  - [ ] GNU Ice Cat
  - [ ] Google Cardboard
  - [X] Google Calendar
  - [ ] Google Hangouts
  - [ ] Google Keep
  - [ ] Google Maps
  - [ ] Google Meet
  - [x] Hack The Box
  - [x] Handlebars.js
  - [ ] Iata
  - [x] Ingress
  - [ ] Internet Explorer
  - [x] InVision
  - [ ] iTunes
  - [x] Jamboard
  - [x] Jamstack
  - [x] JavaScript
  - [x] Kaufland
  - [x] Kotlin
  - [x] MetaFilter
  - [ ] Noun Project
  - [ ] Nrwl
  - [ ] Nutanix
  - [ ] NVIDIA
  - [ ] OBS Studio
  - [ ] OCaml
  - [x] Open Containers Initiative
  - [x] OpenVPN
  - [x] Pro Tools
  - [ ] PyPy
  - [ ] Rakuten
  - [x] Reason
  - [ ] Redbubble
  - [x] Redux
  - [ ] Reliance Industries Limited
  - [ ] Research Gate
  - [x] Robot Framework
  - [x] Safari
  - [ ] San Francisco Municipal Railway
  - [ ] São Paulo Metro
  - [ ] Scrum Alliance
  - [ ] Sega
  - [ ] SketchUp
  - [x] ŠKODA
  - [ ] Society6
  - [ ] Swagger
[...]

PR #10092, opened 2023-12-13
Update metadata

service-paradis:
I made some changes that were initiated in #6422 and left over.


PeterShaggyNoble:
@service-paradis, before I start into reviewing this in the morning: given that #6422 was 2 years ago, have you already verified that all URLs are still valid and still contain the expected icons and/or information?

service-paradis:
Yes, I only added valid ones and edited some with new findings while verifying!

PeterShaggyNoble (approved):
All looks good to me, @service-paradis.

Although, I'm not seeing and _brand_ guidelines at the URLs provided for [OpenVPN](https://openvpn.net/legal), [Rakuten](https://global.rakuten.com/corp/news/media) or [Reliance](https://www.ril.com/legal-notice). Having said that, I don't see any harm in including them now and using them for the `guidelines.trademark` entry when we change the structure of the `guidelines` entry - we'll just need to remember to change them from `branding` to `trademark` at that time.

PR #10104, opened 2023-12-14
Add Stagetimer

adamrusted:
image

**Similarweb rank:** [416,854](https://www.similarweb.com/website/stagetimer.io/#overview)

### Checklist

### Description
Icon pulled from site header. Hex used is the primary colour used on the site.


PeterShaggyNoble (approved):
SVG looks good to me, @adamrusted. But they use a few different shades of green on their site so would we not be better to run with `#00a66c` from the logo to be safe?

PR #10131, opened 2023-12-16
Add Deutsche Post icon

lukasbestle:
deutschepost

**Similarweb rank:** 4803

### Checklist

### Description

Uses the SVG embedded into the Deutsche Post website


_data/simple-icons.json
```diff
@@ -3725,6 +3725,11 @@
             "hex": "0018A8",
             "source": "https://www.db.com"
         },
+        {
+            "title": "Deutsche Post",
+            "hex": "000000",
```
PeterShaggyNoble, line 3730:
```suggestion
            "hex": "FFCC00",
```

PeterShaggyNoble (approved):
Thanks for the contribution, @lukasbestle 👍🏻 The SVG looks perfect to me; just need to update the colour and this is ready to merge.

_data/simple-icons.json
lukasbestle: I did some research and finally found the brand guidelines that I missed on the weekend (will add the link to the JSON as well): https://www.dpdhl-brands.com/de/dp/logo

According to the guidelines, the color of the posthorn icon may not be changed from black.

I saw your comment in the other PR that the brand color is not the same as the icon color. However many uses of Simple Icons blindly color the SVG with the brand color. Which is correct for most icons, but breaks the guidelines here.

How can we solve this dilemma?

PeterShaggyNoble: @lukasbestle, we put the onus on our users to ensure they read and adhere to all branding & trademark guidelines we provide. As you can see from the [colour section](https://www.dpdhl-brands.com/de/dp/farben) of their brand guidelines, the primary _brand_ colour (which is what we aim to include in our data) here is `#ffcc00`.

PeterShaggyNoble (approved):

PR #10132, opened 2023-12-16
Add Galaxus icon

lukasbestle:
galaxus

**Similarweb rank:** 17,698

### Checklist

### Description

Uses the icon from the SVG embedded into the Galaxus website


PeterShaggyNoble (changes requested):
Thanks for the contribution, @lukasbestle. However, I'm seeing [some significant differences](https://github.com/simple-icons/simple-icons/assets/15157491/212b86f6-c705-4f36-aaf1-a4fd9a2937ba) between your version (in red) and the source (in black). I'm using the full logo for my comparison, the 3rd `svg` element within the `a` tag in the header.

Also, I don't know that black is the correct choice of colour here but I also don't know what the correct choice _should_ be! They don't seem to have very clear cut branding on that front; the logo changes to white in dark mode and the different shades of black/grey throughout the site in both light & dark mode, and then the 6 colours running across the top of the page as accent colours in each category or section. Given all that, then perhaps black _is_ the best option for us until someone can suggest something better 🤔

lukasbestle:
Thanks. Looks like they changed the source since I extracted the icon a few months ago. I'll fix this PR.

lukasbestle:
About the colors: They use the six colors from the strip at the top of the page throughout their branding, but none of them seems to be primary. So I think if we need to decide on one, black is the safest as it seems to dominate the overall branding.

PeterShaggyNoble (approved):
SVG lines up perfectly now, @lukasbestle 👍🏻 Although, there are quite a few excess points in the source file that we might be able to clean up. Would you mind having a look at that before we merge?

Fully agree with you on the case you made for using black as the colour here.

lukasbestle:
The "turtle head" uses a lot of points, but none of them can be removed without changing the quite complex shape with many convex and concave segments. However I was able to construct the central ellipse with four points instead of 15.

PeterShaggyNoble (approved):
Nicely done, @lukasbestle 👍🏻

There are a couple more points that could probably be removed, but not such a significant amount to warrant blocking this PR from being merged; we can revisit them in a future update.

Thanks once again for the contribution.

PR #6213, opened 2021-08-08
adding Hetzner logo as full name

jorgeamadosoria:
hetzner

**Issue:** https://github.com/simple-icons/simple-icons/issues/6145

Following the official logo information in https://www.hetzner.com/newsroom, I believe the logo is supposed to contain the full name of the company.
Smaller size does not read very easily however.

PR #10133, opened 2023-12-16
Add Transifex icon

lukasbestle:
transifex

**Similarweb rank:** 70,344

### Checklist

### Description

The SVG is taken directly from their web app. The color is from the `.u-fill-brand` class in their stylesheet.


PeterShaggyNoble (changes requested):
Thanks for the contribution, @lukasbestle, but I think we might need to go with [the full wordmark](https://7gdrhmqdxblg-u503.pressidiumcdn.com/wp-content/uploads/2016/08/logo-widget.svg) here. The only other place I see them using just the "tx" is in [their GH org](https://github.com/transifex) where it has a different treatment, everywhere else they use the full logo. So I don't know that, outside the context of their own website, the "tx" can be considered recognisable as a branding element.

Also, on a sidenote having reviewed a couple of your PRs now: note that the _brand_ colour isn't always necessarily the (primary) colour used in the source logo/icon. In this case, if I were taking it on myself and with a lack of any explicit guidelines on colour, I'd likely pick the grey from the logo or one of the most prevalent blue used on their website.

lukasbestle:
Thanks for your detailed reply.

I'm not sure about using the wordmark. Its aspect ratio is pretty non-square. Since this is Simple Icons and not Simple Logos, I thought the favicon would be a better fit for common use cases.

If you still think we should go with the wordmark, I'm happy to make the change. Same with the color.

PeterShaggyNoble:
> Since this is Simple Icons and not Simple Logos

Oh, trust me, I'm fully in agreement with you on that one! 😆 Unfortunately, though, I seem to be in the minority and the precedent has been _long_ since set. If we want to change it now then that's going to need to be separate discussion.

lukasbestle:
There are also other Simple Icons that use the square favicon instead of the very rectangular wordmark, for example Hetzner.

PeterShaggyNoble:
In the case of Hetzner, we received explicit permission from the company to use their favicon. See #6213 for more details.

PeterShaggyNoble:
Should have said in my last comment: if you'd like to reach out to Transifex to get their permission to use their favicon, too, that'd be great. If you do then it might be worth seeking clarification on the colour, too.

lukasbestle:
I checked their web app at https://app.transifex.com/contact/ again, which is what users of their service will get into contact with constantly (public contact page linked, the app itself requires logging in but uses the same layout and icon). I was able to pull the official SVG from there to replace the manually vectorized icon. Also I took their official brand color from the app stylesheet.

Is the prominent use in their web app enough or should I still double-check with them on whether we can use the "tx" icon?

PeterShaggyNoble (approved):
Nice find, @lukasbestle; we can definitely run with the icon based on this. Everything looks perfect here now - thanks for the contribution.

lukasbestle:
@PeterShaggyNoble Thanks for merging my PRs and for your very helpful and constructive reviews. 👍

PR #10181, opened 2023-12-23
Add Fastmail icon

### LINKED ISSUE #9826: Request: Fastmail
### Brand Name

Fastmail

### Website

https://www.fastmail.com/

### Popularity Metric

https://www.similarweb.com/website/fastmail.com/#overview

https://www.fastmail.com/media-assets/

### Additional Comments

_No response_

lukasbestle:
fastmail

**Issue:** closes #9826

**Similarweb rank:** 11,559

### Checklist

### Description

I used their official SVG but cleaned up unnecessary points in the outer ring shape. I then cut a line of 0.5 px through the color intersection to make the icon recognizable even when monochromatic. The color is the first (= primary) of the official brand colors.


PeterShaggyNoble (changes requested):
Thanks for taking this one one, @lukasbestle 👍🏻 I'm seeing some small differences throughout when comparing with [the logo](https://www.fastmail.com/wp-content/themes/fastmail/webroot/images/header/logo.svg) in the website's header (assuming that's what you used?). [Here's an example](https://github.com/simple-icons/simple-icons/assets/15157491/a9db5fc0-6f95-4657-96f3-7fc8a423bc86) (with yours in red); it may just be a precision issue or I may be comparing against a different source.

Also, I think we can probably forego the cut-out in the outer circle, at least, but will ask the other @simple-icons/maintainers for a second opinion.

lukasbestle:
I can look into the differences. Seems to be a precision issue, but could also be that it was caused by reducing the number of points.

For the outer ring I'd be fine with either the split or no split. But I think the letter icon in the center would not be recognizable anymore without the split.

lukasbestle:
Any update on the cut-out? I'd like to make both changes in one go.

service-paradis:
I think I would keep the cut, but both would be ok for me.

I'm not sure though if we can really add this logo according to https://www.fastmail.com/brand-guidelines/:

> 14. Don’t use different colours for the Logo.

lukasbestle:
> I'm not sure though if we can really add this logo according to [fastmail.com/brand-guidelines](https://www.fastmail.com/brand-guidelines/):
> 
> > 14. Don’t use different colours for the Logo.

Is it a *different* color though if we use their own brand color? I assume what they meant is that one shouldn't randomly make the logo red or pink.

adamrusted:
My interpretation of that line is that if they themselves have not made a monochrome version of the logo available, we should not be creating one ourselves. The colours of the logo available should not be altered.

lukasbestle:
I have asked the Fastmail Press Team for guidance.

lukasbestle:
Unfortunately I have not heard back from Fastmail after two emails to the press team. So it looks like they don't want to be included.

PR #10200, opened 2024-01-02
Use original release action

LitoMore:
Related to https://github.com/simple-icons/release-action/pull/77


.github/workflows/create-release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       did-create-pr: ${{ steps.release.outputs.did-create-pr }}
       new-version: ${{ steps.release.outputs.new-version }}
     steps:
-      - uses: LitoMore/simple-icons-release-action@b14cfca88575c04ea2a90ef282ca16d275cb2715
+      - uses: simple-icons/release-action@v1
```
mondeja, line 26: Would you mind to release `v1` as an alias for `v1.1.0`? v1 is currently [the old implementation](https://github.com/simple-icons/release-action/releases/tag/v1).

mondeja (changes requested):

.github/workflows/create-release.yml
LitoMore: It will fetch the latest v1.x.x. See https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management.

mondeja: Are you sure? It says:

> Move the major version tag (such as v1, v2) to point to the Git ref of the current release. For more information, see "[Git basics - tagging](https://git-scm.com/book/en/v2/Git-Basics-Tagging)."

LitoMore: Yes. See this https://github.com/actions/toolkit/blob/main/docs/action-versioning.md.

mondeja: Nope, quoting it:

> 3. Make the new release available to those binding to the major version tag: **Move the major version tag (v1, v2, etc.) to point to the ref of the current release.**

LitoMore: Thanks. I just tested it here https://github.com/LitoMore/simple-icons/actions/runs/7393719732/job/20114090327.

Let me update it to `v1.1.0`.

mondeja (approved):
Great! Thank you @LitoMore 🥇

PR #10249, opened 2024-01-09
Add Clubforce

PeterShaggyNoble:
Clubforce

**Issue:** n/a (client request)

**Similarweb rank:** [394659](https://similarweb.com/website/clubforce.com)

### Checklist

### Description
I used the icon from [the SVG](https://clubforce.com/app/uploads/2022/06/clubforce-blue-logo.svg) in the [source](https://clubforce.com/media-kit) header as a starting point to recreate the app icon seen on page 10 of [their brand guidelines](https://clubforce.com/app/uploads/2022/08/clubforce-brand-styleguide-1.pdf) and rook the colour from the same SVG.


_data/simple-icons.json
```diff
@@ -2874,6 +2874,11 @@
             "hex": "2C39BD",
             "source": "https://www.cloudways.com/en/media-kit.php"
         },
+        {
+            "title": "Clubforce",
+            "hex": "191176",
+            "source": "https://clubforce.com"
+        },
```
service-paradis, line 2881: We can add https://clubforce.com/media-centre/ for guidelines here.

icons/clubforce.svg
```diff
@@ -0,0 +1 @@
+<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>Clubforce</title><path d="M1.251 0v13.325l3.977-3.017V3.902h8.54L18.992 0ZM22.75 2.43l-3.902 2.945v5.3l-3.608-2.5-3.386 2.5 5.963 4.123-5.89 4.417-5.742-4.343-3.092 2.43L11.926 24l10.823-7.877Z"/></svg>
```
service-paradis, line 1: According to https://clubforce.com/media-centre/:

> Please do not modify the Clubforce logos

I'm wondering if we should add a background for the icon as we can see at page 10 from their [Brand Guidelines PDF](https://clubforce.com/app/uploads/2022/08/clubforce-brand-styleguide-1.pdf):
image

It is the only version we can find without wordmark.

service-paradis (approved):
Added 2 comments, what do you think @PeterShaggyNoble?

icons/clubforce.svg
PeterShaggyNoble: Unfortunately, I think you're right, @service-paradis. PR Updated.

icons/clubforce.svg
adamrusted, line 1: I'm spying a [slight difference](https://github.com/simple-icons/simple-icons/assets/1830867/0aa1dfe1-9130-4fa7-a843-e45df0394e23) when scaling the source SVG down. Also, could we not just get away with the emblem, without the background?

adamrusted (changes requested):

icons/clubforce.svg
PratyushJayachandran, line 1: Upon comparing with the app logo in guidelines PDF, the logo mark is a bit smaller compared to your contribution. See the logo in white compared to cutout. The logo in the PDF is a bit crooked but we will ignore that in our icon.
image

PratyushJayachandran (changes requested):

icons/clubforce.svg
PeterShaggyNoble: Updated to reduce the cut-out size. Note that, after talking it over with @adamrusted, we decided that, given the low quality of the icon in the PDF that we should keep the cut-out faithful to the SVG, rather than tweaking it to match the PDF, so you'll need to pull the SVG in for your comparison as well.

PratyushJayachandran: Yes, I agree @PeterShaggyNoble. I only intended the size to be similar just like the radius of the corners are already matching.

PeterShaggyNoble: Is this good to go, then, @PratyushJayachandran?

PratyushJayachandran: Good now 😄 @PeterShaggyNoble

PratyushJayachandran (approved):
Thank you for the contribution @PeterShaggyNoble 🎉

PR #10253, opened 2024-01-10
Update Aston Martin

Mvlprem:
astonmartin

**Similarweb rank:** [25,353](https://www.similarweb.com/website/astonmartin.com/#overview)

### Checklist

### Description

svg taken from [header](https://www.astonmartin.com), hex taken from stylesheet ` --black `


_data/simple-icons.json
```diff
@@ -1446,7 +1446,7 @@
         },
         {
             "title": "Aston Martin",
-            "hex": "000000",
+            "hex": "161A11",
```
PratyushJayachandran, line 1449:
```suggestion
            "hex": "F4F4F2",
```
We can stick to the color they use in the logo itself. Although their brand color is `#00665e`, I was hard pressed to find any usage. They use the green background with white logo majority of the times.
astonmartin

PratyushJayachandran (changes requested):

_data/simple-icons.json
service-paradis: @PratyushJayachandran The `hex` is there to store the brand color, not the logo color. In this case, I would really expect `#00665E`.

PeterShaggyNoble, line 1449:
```suggestion
            "hex": "00665E",
```

PeterShaggyNoble (approved):
SVG looks good to me 👍🏻 But I would agree with @service-paradis on the choice of colour.

_data/simple-icons.json
Mvlprem: According to this [page](https://www.astonmartinhouston.com/about-us/aston-martin-logo-history/) ` Aston Martin settled on its contemporary logo in the iconic silver, black, and green color scheme `

In their website the logo uses ` #fff ` on black bg and ` #161A11 ` on white bg, and as per our guidelines in the absence of brand guidelines we use the brand's primary web color which i believe this ` #161A11 ` so i think we should stick to it.

@PeterShaggyNoble What do you think?

Mvlprem: This welcome [page](https://www.astonmartin.com/en/welcome) bg img uses green too.

PeterShaggyNoble: It seems we have consensus, so 🙂 Will update the colour and get this merged.

PR #10257, opened 2024-01-10
Update Google Chat

PeterShaggyNoble:
Google Chat

**Issue:** #10073

**Similarweb rank:** n/a

### Checklist


service-paradis:
I found an [official SVG](//storage.googleapis.com/support-kms-prod/8rx5bOqxOZ8pnH49ZQR3fGIafNwtO3X9fDe1) for this icon at https://support.google.com/chat/answer/9455386
When comparing with it, it does not work. Do you think we should use this one instead?

PeterShaggyNoble:
Nice find, @service-paradis - I've updated the PR based on that.

service-paradis (approved):

PR #10265, opened 2024-01-11
Update Zapier

Mvlprem:
zapier

**Similarweb rank:** [5,032](https://www.similarweb.com/website/zapier.com/#overview)

### Checklist

### Description

svg taken from brand [page](https://brand.zapier.com/), Hex taken from brand guidelines.

ss1
ss2


PeterShaggyNoble (changes requested):
Thanks for this one, @Mvlprem 👍🏻 Your SVG lines up perfectly with the chosen source. However, page 28 of the brand guidelines mentions that they have different logos for use at different sizes. If you're using Figma, would you be able to extract the "TINY" logo on that page and use that instead as it would be better suited to our purposes.

For the colour, I _think_ we should go with the "Zapier Orange" `#ff4f00`. Even though "Earth" is the first one listed in the brand guidelines, they state that:
> This palette is centred on our Zapier Orange

And it's also the first colour listed at https://brand.zapier.com.

And, for the guidelines, you can link directly to the Figma file.

Mvlprem:
Yes orange is their brand color, but when using only single color for logo i think only the black and white are allowed from Logo guidance pic above which is why i picked black.

As for different logo if your are talking about the **Tiny** one on page 28 I've compared both **tiny and small** and i don't see any major differences they align almost perfectly.

PeterShaggyNoble (approved):
If your version lines up with the tiny one then we're good to go on that front. I would like to get a second opinion from one of the other @simple-icons/maintainers on the colour before merging, though; we generally use the brand's colour rather than the logo's.

service-paradis:
Since the `hex` is used to represent the brand color, I would expect it to be Zapier Orange (`#ff4f00`). 

The usage for the logomark looks rather strict about using only black and white when using it in monochrome while also specifying to not change colors:
image

The app icon might be what we should use. Unfortunately, it looks like they omit the brand name below 40px which is not really suitable for us.
image

adamrusted:
Good spot @service-paradis - I would agree we should therefore use the app icon with the text inside of it, perhaps the 30px one?

adamrusted (changes requested):
Good spot @service-paradis - I would agree we should therefore use the app icon with the text inside of it, perhaps the 40px one? @Mvlprem - would you mind taking another look?

PeterShaggyNoble:
@Mvlprem, what's the source file for the updated version? The only SVG I can find is in [their Google Drive folder](https://drive.google.com/drive/folders/1lDi7I-NC45zHfylsHU60Vloz_4zhyUfK), but it has square corners. If you extracted it from the Figma file, could you share the source here, please?

Mvlprem:
> @Mvlprem, what's the source file for the updated version? The only SVG I can find is in [their Google Drive folder](https://drive.google.com/drive/folders/1lDi7I-NC45zHfylsHU60Vloz_4zhyUfK), but it has square corners. If you extracted it from the Figma file, could you share the source here, please?

Sorry forgot about the source, I took it from the figma file page 32, 40px
logo-symbol
I didn't updated the source to figma link, if necessary pls update it if svg is good, sorry for the extra work.

PeterShaggyNoble (changes requested):
I'm seeing [a tiny difference](https://github.com/simple-icons/simple-icons/assets/15157491/1428297b-fe37-4a6c-a39e-83b5224ba46b) in all 4 corners, @Mvlprem, with your version in red. Most likely it can be solved by upping the precision by 1.

The imperfections in the letters in the source file are annoying me but they're present in _all_ versions of the logo and icon, so there's not much we can do about it.

PeterShaggyNoble (approved):
I'm now seeing some tiny differences in the lettering, such as [this one](https://github.com/simple-icons/simple-icons/assets/15157491/ff482dc7-f575-4333-9081-1085667e8fd3) in the "r". But, given how insignificant they are and how riddled with imperfections the source is, I think we can let them slide.

PR #10273, opened 2024-01-12
Fix license types

LitoMore:
Try this code with REPL in our project root:

```js
const icons = require('./index.js');
Object.values(icons).filter(icon => Boolean(icon.license)).filter(icon => !icon.license.url).length;
```

The result is `0`. Which means all licenses have URLs.

But the `getIconsData()` function will not return the license URL if it's a SPDX license. See this schema change:

- #9575


types.d.ts
```diff
@@ -5,12 +5,12 @@
  */
 export type License = SPDXLicense | CustomLicense;
 
-type SPDXLicense = {
+export type SPDXLicense = {
   type: string;
-  url?: string;
+  url: string;
```
mondeja, line 10: I think that we should just remove this `url` field as SPDX licenses doesn't need urls, right?

sdk.d.ts
```diff
@@ -55,7 +55,7 @@ export type IconData = {
   source: string;
   slug?: string;
   guidelines?: string;
-  license?: License;
+  license?: Omit<SPDXLicense, 'url'> | CustomLicense;
```
mondeja, line 58: And this will change accordingly.

mondeja (changes requested):
Additionally, seems to me that this object could be refactored as a `const` and the documentation changed accordingly:

https://github.com/simple-icons/simple-icons/blob/c6fb28b4876405dd4afb8482549bba363c58002b/.jsonschema.json#L601-L605

types.d.ts
LitoMore: Nope. The URL does exist when you access it in this way:

```js
import * as icons from 'simple-icons';

console.log(icons['siApache']).license;
//=> {type: 'Apache-2.0', url: 'https://spdx.org/licenses/Apache-2.0'}
```

See here: https://github.com/simple-icons/simple-icons/blob/develop/scripts/build/package.js#L49

mondeja (approved):
Ah OK, I see. Thanks @LitoMore

PR #10276, opened 2024-01-12
Refactor JSONSchema and URL checks

mondeja:
- Deduplicates URL regex checks and definitions.
- Uses `const` instead of single value enum for custom license types in JSONSchema.
- Includes *.jsonschema.json* file in JS package.
- Improves types for `getIconsData*` functions of SDK.


.jsonschema.json
```diff
@@ -617,7 +616,7 @@
     },
     "duplicate": {
       "$id": "#duplicate",
-      "description": "A \"dup\" brand",
+      "description": "A duplicated brand",
```
PeterShaggyNoble, line 619:
```suggestion
      "description": "A brand that uses the same icon",
```

mondeja:
The changes proposed on https://github.com/simple-icons/simple-icons/pull/10573#discussion_r1511863468 must be backported to this PR before the merge so we remove the URL regex entirely.

mondeja:
The previous comment (now hidden) has no sense because in this PR I'm using the same regex for JSONSchema check and in the code. The regex can't be removed.

adamrusted:
Hey @mondeja - are you okay to resolve the conflicts on this ready for next week? We can look to merge in as part of the major release then :)

LitoMore:
As I remember we are planning to move the SDK part to a module.

mondeja:
Don't worry, I'll resolve the conflicts before the Wednesday regardless of whether we move the SDK to a separate npm package. In that case, the changes proposed by this PR must be included. Feel free to open the PR to separate the SDK meanwhile @LitoMore. If not, I'll open it before the weekend if I have time for it, anyways we can always postpone that change for v13.

LitoMore:
Ok, I will try to create the new repository for SDK.

mondeja:
> Ok, I will try to create the new repository for SDK.

As you want but we can release it from this same repository.

LitoMore:
> As you want but we can release it from this same repository. Probably this approach is better because there are some files in simple-icons/simple-icons that the SDK needs to work.

Okay, we need to change this repository to [workspaces](https://docs.npmjs.com/cli/v10/using-npm/workspaces). The folder structure needs some changes.

mondeja:
This is ready @LitoMore, would you mind to review it?

sdk.d.ts
```diff
@@ -63,9 +63,9 @@ export type IconData = {
 /* The next code is autogenerated from sdk.mjs */
 /* eslint-disable */
 
-export const URL_REGEX: RegExp;
 export const SVG_PATH_REGEX: RegExp;
 export function getDirnameFromImportMeta(importMetaUrl: string): string;
+export function urlRegex(jsonschemaPath?: string): Promise<RegExp>;
```
LitoMore, line 68: We used a `fs.readFile` for the JSON schema file. This means the `.jsonschema.json` file will be missed during the tree-shaking.

This means that our `urlRegex()` cannot be used in some scenarios.

mondeja: I've added `!.jsonschema.json` to `.npmignore`, so will be included in the package. This is the same behaviour of other functions like `getIconsData`, `getThisPartyExtensions` or `getThirdPartyLibraries`, which read from the icons data file and the README.

LitoMore (approved):

PR #10277, opened 2024-01-12
Fix URL to the vue3-simple-icons version.

wyatt-herkamp:
I am the developer of [vue3-simple-icons](https://github.com/wyatt-herkamp/vue3-simple-icons). When V11 was released I redid the code that builds the components. To fix the naming convention errors.  And the resource that the version was pulled from no longer exists. 

However, I now generate a file that lists this information. So I updated the link.


PeterShaggyNoble (approved):
Thanks, @wyatt-herkamp 👍🏻

PR #10315, opened 2024-01-18
Remove Trailing Slashes from URLs

PeterShaggyNoble:
Removes the remaining trailing slashes from all `source` & `guidelines` URLs.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -455,7 +455,7 @@
         {
             "title": "AirPlay Video",
             "hex": "000000",
-            "source": "https://developer.apple.com/design/human-interface-guidelines/airplay/overview/icons/",
+            "source": "https://developer.apple.com/design/human-interface-guidelines/airplay/overview/icons",
             "guidelines": "https://www.apple.com/legal/intellectual-property/guidelinesfor3rdparties.html"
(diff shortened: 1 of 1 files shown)


_data/simple-icons.json
```diff
@@ -14790,7 +14790,7 @@
         {
             "title": "TinyLetter",
             "hex": "ED1C24",
-            "source": "https://tinyletter.com/site/press/"
+            "source": "https://tinyletter.com/site/press"
         },
```
service-paradis, line 14794: As stated in https://discord.com/channels/1142044630909726760/1151132343776591899/1151169674113978411, removing trailing slashes from path does not always work. You can try it with this one. While https://tinyletter.com/site/press/ is working, https://tinyletter.com/site/press is not.

If we want to do this, we'll need to check them all!

service-paradis (changes requested):

_data/simple-icons.json
PeterShaggyNoble: Well, feck! 😕 What sort of lunatic sets a web server up like that these days?!

Given that, I think we're better off closing this PR and tackling each one individually as part of the Great Icon Review/

PR #10321, opened 2024-01-18
Rename Amazon AWS as AWS

readboy8:
### Description
Change Amazon AWS to just AWS and with the alias Amazon Web Services
Change .svg title tag to AWS and the filename amazonaws to aws.


PeterShaggyNoble (approved):
:+1: Looks good to me. Thanks, @readboy8. But, seeing as @adamrusted originally proposed we do this the other way 'round, I'm going to ask for his feedback too.

Either way, though, this constitutes a breaking change that we're going to have to hold off on until v12.

adamrusted (changes requested):
Apologies for the delay on this. The platform is referred to as Amazon Web Services first and foremost by Amazon themselves - with AWS being more of a handy short name for the platform name when attaching to products - so we should probably match their naming prioritisation. Would you mind swapping the name and the aka please @readboy8?

readboy8:
Ok, I created a new pull request with the switched title and alias.

PR #10356, opened 2024-01-25
Add Kununu Icon

### LINKED ISSUE #9357: Add Kununu
### Brand Name

Kununu

### Website

https://www.kununu.com/

### Popularity Metric

https://www.similarweb.com/website/kununu.com/#overview

svg: kununu
color: '#020207'

### Additional Comments

_No response_

### Contributing

Yes

Mvlprem:
kununu

**Issue:** closes #9357

**Similarweb rank:** [14,149](https://www.similarweb.com/website/kununu.com/#overview)

### Checklist

### Description

svg taken from [homepage](https://www.kununu.com/), Hex taken from stylesheet ` background-color ` ` #ffc62e `


PeterShaggyNoble (dismissed):
@Mvlprem, when I first saw this request a couple of weeks ago, my initial thought was that, for an _icon_, we should use the wave/squiggle they use as a logo on all their social media pages. It's also used extensively in their printed materials - [this PDF](https://campaigns.kununu.com/wp-content/uploads/2024/01/kununu-Gehaltscheck2024.pdf), for example.

Also, for the colour, looking around their site and social media I think yellow's probably the better choice.

What do you think? If you want to go with the wave, there's an SVG of it on their homepage.

Mvlprem:
@PeterShaggyNoble Initially i wanted to go with the wave too as it serves better for the project but the only reason i took the current one is because they provided it as downloadable logo and more over other sites like glassdoor, EU startups(just for reference) are using the full logo. Except their social media the wave thing can't be found anywhere else as a proper logo, but we can see it as a bg img on their homepage and as favicon.

As for the hex well they provided an ` eps ` in blue shade so i believe blue as their primary, even the wave thing that we want to use is also in blue shade. I believe the yellow is more like a secondary color and blue is primary.

So what do you say, should we update it to wave or ask for others opinion too?

PeterShaggyNoble:
I think we're definitely OK to use the wave as it's used extensively enough to be identifiable with their brand and is how they present themselves to the world on social media. And, also, it's obviously better suited to our purposes on the icon library.

On the colour, though, as we have 2 differing opinions going here, let's ask the other @simple-icons/maintainers for theirs. If the consensus is to use the blue/purple then I agree with the shade you've chosen.

service-paradis:
Well, it is nt an easy one. Their main 3 colors are blue, yellow and orange. The primary color is never the same depending on which page we are or which document we read.

I think I have a slight preference for the yellow shade since it is often use as background for the first main section for a lot of their website pages.

_data/simple-icons.json
```diff
@@ -7890,6 +7890,11 @@
             "hex": "290B53",
             "source": "https://cncf-branding.netlify.app/projects/kuma/"
         },
+        {
+            "title": "Kununu",
+            "hex": "FFC62E",
+            "source": "https://news.kununu.com/presseinformationen"
```
adamrusted, line 7896:
```suggestion
            "source": "https://www.kununu.com"
```
Given we're using the icon version from the homepage, we should link that as our source. I would also argue the press area doesn't provide any guidelines on how the icon should be used - so shouldn't be included.

adamrusted (approved):
Icon is spot on, just a change of source to reflect us using the SVG found on the homepage. I'll merge that in now and get this merged into develop. Thanks @Mvlprem!

PR #10368, opened 2024-01-27
Use `v1` version for simple-icons/release-action

LitoMore:
Related to:

- https://github.com/simple-icons/release-action/issues/80
- #10200

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -23,7 +23,7 @@ jobs:
       did-create-pr: ${{ steps.release.outputs.did-create-pr }}
       new-version: ${{ steps.release.outputs.new-version }}
     steps:
-      - uses: simple-icons/release-action@v1.1.0
+      - uses: simple-icons/release-action@v1
         id: release
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       did-create-pr: ${{ steps.release.outputs.did-create-pr }}
       new-version: ${{ steps.release.outputs.new-version }}
     steps:
-      - uses: simple-icons/release-action@v1.1.0
+      - uses: simple-icons/release-action@v1
```
mondeja, line 26: Would you mind to create the new release in simple-icons/release-action?

LitoMore: We don't need to create one.

Like Deno does: https://github.com/denoland/setup-deno. They don't have the `v1` tag.

It will fetch the latest `v1.x.x` automatically.

mondeja: Could you provide a link of a running example? Seems to me that deno only supports [this](https://github.com/denoland/setup-deno/blob/0df5d9c641efdff149993c321fc27c11c5df8623/.github/workflows/test.yml#L16):

> [1.x, "1.33.1", canary, ~1.32, b31cf9fde6ad5398c20370c136695db77df6beeb]

LitoMore: See this job: https://github.com/LitoMore/games/actions/runs/7686593957.

My `deno.yml` is `uses: denoland/setup-deno@v1`. But they updated the Node.js from 12 to 16 since https://github.com/denoland/setup-deno/releases/tag/v1.1.1.

This means the `action/tool-name@tagName` supports the feature.

mondeja (approved):
Ah OK, should be fine.

PR #10430, opened 2024-02-10
Add Logicool aliases

### LINKED ISSUE #10426: Request: Logicool
### Brand Name

Logicool

### Website

https://www.logicool.co.jp/ja-jp

### Popularity Metric

<s>I don't know because I don't have a Similarweb account (I can't sign up for Similarweb with my email address).</s>
https://www.similarweb.com/website/logicool.co.jp/#overview

Logicool is the brand name of Logitech in Japan. Since the logos are completely identical, we would like the aliases to be set.

### Additional Comments

_No response_

adamrusted:
**Issue:** closes #10426

Adds the `loc` and `dup` aliases to Logitech for the Japanese locale.

Note for @simple-icons/maintainers - might it be worth us embedding localisations into the `dup` tag? That would've made _logi_-cal sence in this case. Either that, or we should look to allow more attributes on the `loc` tag.


mondeja:
> might it be worth us embedding localisations into the dup tag

I would vote for that.

PeterShaggyNoble:
Interesting conundrum here 🤔 "ロジクール" is obviously not a correct alias for Logitech, but we currently have no way of adding it otherwise. I think the best approach would be to starting allowing `dup` aliases to have their own alias entries but without allowing them to have their own `dup`s. I don't _think_ that would constitute a breaking change so we could use this PR to implement it.

PeterShaggyNoble (approved):
Looks good to me, @adamrusted 👍🏻 But, as I haven't touched our JSON schema in a long long time, I'll leave it to one of the others to double check before merging.

mondeja (approved):
The addition to *sdk.d.ts* is not a breaking change because the new property is optional. Looks nice, good job!

PR #10434, opened 2024-02-10
Add Hevy

patrik-csak:
simpleicons

**Similarweb rank:** [110,983](https://www.similarweb.com/website/hevyapp.com/)

### Checklist

### Description

- Source: https://www.hevyapp.com/wp-content/uploads/hevy-logo.svg
- Cofounder: @desmondmc
- GitHub organization: @hevyapp


adamrusted (changes requested):
Your SVG is a perfect match, @patrik-csak - though there are a _lot_ of excess points in this one (they're present in the original, not your doing!). Could you look to reduce those down a bit before we merge this in?

image

patrik-csak:
@adamrusted i just pushed a commit that removes a bunch of the points

CleanShot 2024-02-11 at 11 24 17@2x

PeterShaggyNoble (changes requested):
Thanks for your work on this one so far, @patrik-csak; it doesn't look like an easy file to work with. Unfortunately, though, some of the points you have removed have resulted in some differences with the original. At the same time, though, it looks like there are a few more that _could_ be removed. I also compared your first `path` against the source and noted some differences in the horizontals in that, too. Would you mind having another look at this please?

If you need any help, you can reach out to us on [Discord](https://discord.gg/vUXFa7t5xJ). My basic process for trying to remove excess points, though, is to zoom in as far as my software will allow and work my way around the path, point by point. If I notice even the _slightest_ visual shift as I delete a point then I undo it and move on to the next one. I also keep a copy of the source icon on the canvas in a contrasting colour so I can continuously compare as I go to see if I've introduced any differences - with some paths, deleting a point can result in a shift in an entirely different part of the icon.

patrik-csak:
thanks for the feedback and advice, @PeterShaggyNoble. i just gave it another shot from scratch with your suggestions

PeterShaggyNoble (approved):
Nice work, @patrik-csak, everything lines perfectly now. I would have thought that we should be able to remove even more points but, having tried it myself, I've found _very_ few that can be removed without introducing even a tiny difference. So, I'm happy to run with this as-is but will let one of the other @simple-icons/maintainers double check it before merging.

service-paradis (approved):
> Nice work, @patrik-csak, everything lines perfectly now. I would have thought that we should be able to remove even more points but, having tried it myself, I've found very few that can be removed without introducing even a tiny difference. So, I'm happy to run with this as-is but will let one of the other @simple-icons/maintainers double check it before merging.

Same here. I got something pretty clean, but it ends up with hairlines here and there. We can live with this one for now! Thank you for your work @patrik-csak, it is not an easy one to work with!

patrik-csak:
thanks guys! pleasure working with you all

PR #10440, opened 2024-02-11
Update deprecation notice in versioning policy

mondeja:
The current notice does not have much sense:

> SVGs and metadata of brands will be updated on request. This excludes brands that were removed in the new _major_ release.

These updates will be covered only by the next minor or patch release of the next major release, so this statement is not true.

Also, I think that it's great to support the API of the entire previous major version by only fixing bugs, but anyways is a really edge case scenary.


VERSIONING.md
```diff
@@ -18,9 +18,4 @@ _Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs.
 
 ## Deprecation
 
-For approximately 3 months after a _major_ release, the _major_ release preceding it will be supported. After 3 months, the previous _major_ release is deprecated and will no longer be supported.
-
-Support of a _major_ version entails the following:
-
-- SVGs and metadata of brands will be updated on request. This excludes brands that were removed in the new _major_ release.
-- Bugs in the npm library will be updated if reported.
+After a _major_ release, the next _major_ release preceding it will be supported, which includes fixing bugs bugs in the npm library or other APIs.
```
adamrusted, line 21:
```suggestion
After a _major_ release, the next _major_ release preceding it will be supported, which includes fixing bugs in the npm library or other APIs.
```

VERSIONING.md
```diff
@@ -18,9 +18,4 @@ _Major_ releases are scheduled on a half-year basis, mainly to remove old SVGs.
 
 ## Deprecation
 
-For approximately 3 months after a _major_ release, the _major_ release preceding it will be supported. After 3 months, the previous _major_ release is deprecated and will no longer be supported.
-
-Support of a _major_ version entails the following:
-
-- SVGs and metadata of brands will be updated on request. This excludes brands that were removed in the new _major_ release.
-- Bugs in the npm library will be updated if reported.
+After a _major_ release, only the _major_ release that immediately preceded it will be supported, which includes fixing bugs in the npm library or other APIs.
```
adamrusted, line 21:
```suggestion
After a _major_ release, the only other supported version is the last version of the previous _major_ release. This previous version will only receive bug fixes to either the npm library, or our internal APIs. For example, following the release of `v11.0.0` on 7 Jan 2024, the only previous version we support is `v10.4.0`, released on 24 Dec 2023. All other previous versions are considered deprecated.
```

Is it worth us giving an example, something like this?

mondeja: The next part seems strange to me, at least a bit unnecessary.

> All other previous versions are considered deprecated.

I don't think that is worth the complete example, but if others agree will be fine.

PeterShaggyNoble: I'd agree that the example is unnecessary.

adamrusted: Removed the example :)

adamrusted (approved):

PR #10497, opened 2024-02-19
Add Fediverse Metrics

### LINKED ISSUE #10172: Fediverse Popularity Metrics
### Kind of Issue

Other

### Description

I opened this is issue as suggested by https://github.com/simple-icons/simple-icons/issues/967#issuecomment-1866558369

[Fediverse](https://en.wikipedia.org/wiki/Fediverse) by its own nature is _decentralised_ so you cannot rank just one single website. Especially you cannot use the rank (Alexa, Google, whatever...) of the people visiting the platform landing page (ex. https://joinmastodon.org/) because its a number that tells you very little about all the other websites (the correct name is servers/instances) actually running the software (ex. https://joinmastodon.org/servers).

People don't have to pass through the developer's website and not even the instance he/she runs, to use the platform. In the same way that you don't have to go to wordpress.org or even wordpress.com to use WordPress. You can discover an instance in many different ways and you can register directly there (ex. https://mastodon.social/ https://climatejustice.social/ ...).

🚨 **It is moreover contrary to the very spirit of the open movement and the nature of the Fediverse to use an indicator created by a Big Tech.** 🚨

Most of the interactions happen inside each server so you need an external website that tracks and aggregates the data of most of the servers. There're different websites built to provide Fediverse stats and they are also open source:

https://fedidb.org/network
Source code (old version): https://github.com/fedidb
(by the same dev of https://github.com/pixelfed)
image

https://the-federation.info/#projects
source code: https://codeberg.org/thefederationinfo/the-federation.info
image

https://www.fediverse.to/#browse-sw
https://gitlab.com/ayush-sharma/to-the-fediverse
image

Other stats:
https://fediverse.observer/
https://gitlab.com/diasporg/poduptime

Other info/links:
https://fediverse.party/en/fediverse/

Fediverse Wiki:
https://joinfediverse.wiki/

I hope this can help to show that Fediverse is every day more relevant and new platforms are gaining popularity! 😸

adamrusted:
**Issue:** Closes #10172

Adds the briefly discussed Fediverse popularity metrics to our CONTRIBUTING file.
Open to tweak the numbers as necessary, but they felt about right.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,6 +93,9 @@ To be considered popular enough to be within our scope, a brand must be in exist
     - In the top 50, or,
     - In the top 5k, _with_ a global rank of 1m or better.
 1. In cases where a brand does not have its own website the Similarweb rank of its parent company will be accepted _if_ the brand is that company's primary product.
+1. If the brand is a Fediverse project, the following metrics apply:
+    - For **software**, Monthly Active Users (MAU) as listed on [FediDB](https://fedidb.org/software) should be 10,000 or higher.
+    - For **servers** that have a unique recognizable logo, the User Count as listed on [FediDB](https://fedidb.org/network) should be 100,000 or higher **AND** the MAU should be 10,000 or higher.
 1. Its packages meet one of the following minimum requirements:
     - [npm](https://www.npmjs.com): 100k weekly downloads,
     - [JSDelivr](https://www.jsdelivr.com): 1m daily or 35m monthly requests,


PeterShaggyNoble (changes requested):
As the 10k MAU is a base requirement, I think we could simplify this by mentioning it in the parent bullet and then only have the nested bullet for servers.

CONTRIBUTING.md
```diff
@@ -93,6 +93,9 @@ To be considered popular enough to be within our scope, a brand must be in exist
     - In the top 50, or,
     - In the top 5k, _with_ a global rank of 1m or better.
 1. In cases where a brand does not have its own website the Similarweb rank of its parent company will be accepted _if_ the brand is that company's primary product.
+1. If the brand is a Fediverse project, the following metrics apply:
+    - For **software**, Monthly Active Users (MAU) as listed on [FediDB](https://fedidb.org/software) should be 10,000 or higher.
+    - For **servers** that have a unique recognizable logo, the User Count as listed on [FediDB](https://fedidb.org/network) should be 100,000 or higher **AND** the MAU should be 10,000 or higher.
```
adamrusted, line 98:
```suggestion
1. If the brand is a Fediverse project, it needs to have a minimum of 10,000 Monthly Active Users as listed on [FediDB](https://fedidb.org/software).
    - In addition, for **servers** that have a unique recognizable logo, the User Count as listed on [FediDB](https://fedidb.org/network) should be 100,000 or higher.
```

PeterShaggyNoble (approved):
That reads better now 👍🏻 Before merging, though, should we move it down to be point 6 so that we're giving more importance to package statistics?

PR #10517, opened 2024-02-23
Add BuySellAds

### LINKED ISSUE #10515: Request: BuySellAds
### Brand Name

BuySellAds

### Website

https://www.buysellads.com/

### Popularity Metric

Not sure. I've logged into similarweb and it does not display the rank.

image

How should I get that?

Color: `#EB4714`
In their website there is a "BSA" favicon.

### Additional Comments

For maintainers: we are using this icon in the simple-icons' Github profile but the CDN is returning 404 because this icon is missing. See https://github.com/simple-icons/.github-private/blob/main/profile/README.md

adamrusted:
image

**Issue:** closes #10515

**Similarweb rank:** Unknown

### Checklist

### Description

Bit of a selfish add on our part, given it's the company behind Carbon Ads.


PeterShaggyNoble:
> Bit of a selfish add on our part, given it's the company behind Carbon Ads

Was about to post on the issue: if we add this, are we not opening ourselves up to people to use it as a point of argument in favour of adding other brands that don't have an SW rank? I _was_ going to suggest that the better option might to be to add Carbon instead, but their rank is outside our scope.

adamrusted:
Good point!
In the interim, I've hosted a local copy on the `.github-private` repo so that our icons aren't broken 😇

service-paradis:
Compared with Adroll using Google Trends is not necessary convincing unfortunately:
https://trends.google.com/trends/explore?q=BuySellAds,Adroll&hl=en

Do not try this with Google AdSense 😅

LitoMore:
Can we use this page as popularity reference?

https://opencollective.com/buysellads

It has a good activity on contributing popular open-source projects.

mondeja:
I think we could even add an exception for brands that we are using directly.

PeterShaggyNoble (approved):
> I think we could even add an exception for brands that we are using directly.

I'd agree with that, but I'd still worry about the perception of us ignoring our own metrics of popularity to our own benefit and it opening us up to people trying to make a case of why we should ignore them for other icons. Let's give it a go, though; if it becomes an issue then we can always remove it again.

Issue #10514, opened 2024-02-23
Discussion around colours chosen as part of bulk update


adamrusted:
The below table details icons that were updated in #9360 and are still part of our collection. Warrants further discussion, and I've added my suggestions to the side.

| Icon             | Previous Hex | Current Hex | Action                                                                                                                                                                                    |
| :--------------- | :----------- | :---------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Backendless      | `#FFFFFF`    | `#5056E5`   | The blue seems to be quite widely used - perhapse we should keep it for this one?                                                                                                         |
| Clarivate        | `#FFFFFF`    | `#000000`   | Logo (when not displayed on a black background) uses black, purple, and green as its' colours. The green is most prevelantly used through their website and socials. Change to `#93FF9E`. |
| Electron Builder | `#FFFFFF`    | `#000000`   | Docs site uses `#2094f3` quite extensively.                                                                                                                                               |
| Fluxus           | `#FFFFFF`    | `#8D73B0`   | Website uses current hex quite extensively, but logo is always white. Revert to white for icon colour.                                                                                    |
| Honor            | `#FFFFFF`    | `#000000`   | Almost always displayed asa black logo on a white background. Keep black as the colour.                                                                                                   |
| Jaguar           | `#FFFFFF`    | `#000000`   | Always shown as a white icon on black background. Revert to white.                                                                                                                        |
| Leanpub          | `#FFFFFF`    | `#222222`   | `#262425` or `#000000` get used on their socials. Switch to one of these.                                                                                                                 |
| Lemmy            | `#FFFFFF`    | `#14854F`   | Although green is used throughout the website, the logo is always displayed in black and white. Given we're only using the outline, I would argue `#000000` is correct.                   |
| Ludwig           | `#FFFFFF`    | `#000000`   | Website is mostly greys and white. Icon is always displayed in white, but an arguement could be made for `#404040` being the hex we use.                                                  |
| Mumble           | `#FFFFFF`    | `#000000`   | I'd agree with this one. Mumble is mainly black/white, but the mumble shape itself is displayed in black wherever it's shown.                                                             |
| Oh Dear          | `#FFFFFF`    | `#FF3900`   | Agree on this. Colour is described as the brand's 'key colour'.                                                                                                                           |
[...]

PR #10518, opened 2024-02-23
The Update PR

adamrusted:
Updates to hexes based on #10514.


PeterShaggyNoble (changes requested):
I'd suggest splitting this into 2 PRs so we can use the `skip-release-notes` label on the ones that are just being reverted to white - what do you think?

PeterShaggyNoble (approved):

PR #10519, opened 2024-02-23
The Revert PR

adamrusted:
This PR goes alongside #10518 to complete #10514. Reverts some icons back to using `#FFFFFF` as their hex.


LitoMore:
Some colors don't need to revert. Let's create a common PR instead.

adamrusted:
> Some colors don't need to revert. Let's create a common PR instead.

See the linked PR and issue. Can we keep discussion to the issue so as to not muddy the waters?

LitoMore:
np. Go ahead.

PeterShaggyNoble (approved):

service-paradis:
As mentioned in the issue (https://github.com/simple-icons/simple-icons/issues/10514#issuecomment-1961461766), I think some of those revert was not necessary (Fluxus and Reason Studios for example).

PR #10564, opened 2024-03-03
Allow to select between licenses in `add-icon-data` script

uncenter:
**Issue:** N/A

**Similarweb rank:** N/A

### Checklist

### Description

While working on a few icons I found a few places that the `add-icon-data` script could be improved. This PR adds an autocompletion prompt to make it more clear what kind of licenses are accepted; you can type/paste a SPDX identifier, or start typing to show what licenses are under that name. This PR also adds colons after the message of each prompt, improves wording in various places, and properly handles Ctrl + C (prompts being force quit). I know the script isn't really vital to the project but it helps to make it easier for new contributors!

Screenshot 2024-03-03 at 15 15 53 (Ghostty)


scripts/add-icon-data.js
```diff
@@ -88,85 +89,103 @@ const getIconDataFromAnswers = (answers) => ({
...
-  message: 'Hex',
-  validate: hexValidator,
-  transformer: hexTransformer,
-});
-
-answers.source = await input({
-  message: 'Source',
-  validate: sourceValidator,
-});
+try {
```
mondeja, line 92: Please, don't use a `try` / `catch` block. That would silently bypass all other possible exceptions. Use `process.on('SIGINT', () => {});` instead, see this thread on StackOverflow: https://stackoverflow.com/a/71543784/9167585

In this case, would you mind to add other signal interruptions? Adding SIGINT, SIGQUIT and SIGKILL should be fine.

mondeja (changes requested):
Hi. Thanks for the improvement!

scripts/add-icon-data.js
```diff
@@ -88,85 +89,103 @@ const getIconDataFromAnswers = (answers) => ({
...
+      );
+    answers.licenseType = await autocomplete({
+      message: 'License type:',
+      source: async (input) => {
+        input = (input || '').trim();
+        return Boolean(input)
+          ? licenseTypes.filter((license) =>
+              license.value.toLowerCase().includes(input.toLowerCase()),
+            )
+          : licenseTypes;
```
mondeja, line 139: You don't need a `Boolean` casting here:
```suggestion
        input = (input || '').trim();
        return input
          ? licenseTypes
          : licenseTypes.filter((license) =>
              license.value.toLowerCase().includes(input.toLowerCase()),
            );
```

mondeja (changes requested):

scripts/add-icon-data.js
```diff
@@ -88,85 +89,103 @@ const getIconDataFromAnswers = (answers) => ({
...
+      jsonSchema.definitions.brand.properties.license.oneOf[0].properties.type.enum.map(
+        (license) => {
+          return { value: license };
+        },
+      );
+    answers.licenseType = await autocomplete({
+      message: 'License type:',
+      source: async (input) => {
+        input = (input || '').trim();
+        return Boolean(input)
```
mondeja, line 135: You don't need a `Boolean` casting here.
```suggestion
        return input
```
Note that `?` implies `==`, not `===`, so `''` is evaluated to `false`.

mondeja (changes requested):

scripts/add-icon-data.js
uncenter: Ah right, thanks.

scripts/add-icon-data.js
uncenter: I think @inquirer/prompts intercepts the SIGINT and all that? How do you feel about:

```js
} catch (error) {
  if (
    error.message.includes('force closed the prompt') ||
    error.message.includes('prompt denied')
  ) {
    console.log(chalk.red('\nAborted.'));
    process.exit(1);
  } else {
    throw error;
  }
}
```

mondeja: Where is the catch of `SIGINT` in inquirer? Could you link it? I don't see that in the link that you've shared.

mondeja: I think that inquirer has a really weird behaviour and that's the reason why you're throwing an exception here. If I press `Ctrl^C` in the `title` prompt, it raises an exception:

```
file:///home/mondeja/files/code/simple-icons/node_modules/@inquirer/input/node_modules/@inquirer/core/dist/esm/lib/create-prompt.mjs:36
                    reject(new Error(`User force closed the prompt with ${code} ${signal}`));
                           ^

Error: User force closed the prompt with 0 null
    at file:///.../simple-icons/node_modules/@inquirer/input/node_modules/@inquirer/core/dist/esm/lib/create-prompt.mjs:36:28
    at Emitter.emit (file:///.../simple-icons/node_modules/signal-exit/dist/mjs/index.js:67:19)
    at #processEmit (file:///.../simple-icons/node_modules/signal-exit/dist/mjs/index.js:236:27)
    at #process.emit (file:///.../simple-icons/node_modules/signal-exit/dist/mjs/index.js:187:37)
    at process.callbackTrampoline (node:internal/async_hooks:130:17)
```

This looks a really bad idea. What is the recommended solution given by inquirer maintainers? It is documented anywhere?

uncenter: Yep. That's why that's the best solution I could come up with. It could be worth it to switch to https://github.com/terkelg/prompts, which supports better cancel handling? They have a great [autocomplete prompt built-in](https://github.com/terkelg/prompts#autocompletemessage-choices-initial-suggest-limit-style). I'll make a new PR if that sounds good?

mondeja: I think that inquirer.js seems to be better maintained and to rewrite the whole script to make a change doesn't look worth to me. Just go with the licenses selection improvement in this, we can open an issue on inquirer.js to proper handle this case.

mondeja (approved):
Looks good, thank you! 🥇 

Opened #10572 to track further improvements.

PR #8714, opened 2023-06-02
Autocloser: Only trigger on `opened` & fix issue number param

LitoMore:
Documentations:

- [GitHub Actions Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)
- [Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows)

PR #10569, opened 2024-03-04
Only check potential duplicates on `opened` event

LitoMore:
Related to #8714

We should bypass the `edited` and `reopened` event in case the issue was closed and marked as duplicated by accident.


adamrusted:
As mentioned on Discord, I don't think we need this. The potential duplicates flow should trigger on any action (create/edit/reopen) just in case someone renames their issue!

mondeja:
What is the purpose of this change?

Issue #10464, opened 2024-02-14
3D Printer Manufacturers


adamrusted:
## Printer Manufacturers

| Brand          | Popularity Metric                                                     |     In Scope?      | Website                     | Resources                                                                                                           | Resolving PR |
| :------------- | :-------------------------------------------------------------------- | :----------------: | :-------------------------- | :------------------------------------------------------------------------------------------------------------------ | :----------: |
| AnkerMake      | SW: [156,608](https://www.similarweb.com/website/ankermake.com/)      | :heavy_check_mark: | https://ankermake.com       | SVG in header, use `M`                                                                                              |    #10465    |
| Anycubic       | SW: [55,128](https://www.similarweb.com/website/anycubic.com/)        | :heavy_check_mark: | https://store.anycubic.com  | SVG text logo in some printer [user guides](https://drive.google.com/file/d/1NRMp12tRvmWXrfAqgpwHvoSWEnU5_mZG/view) |    #10466    |
| ~~Artillery~~  | SW: [707,947](https://www.similarweb.com/website/artillery3d.com/)    |        :x:         | https://www.artillery3d.com | ~~None~~                                                                                                            |              |
| Elegoo         | SW: [59,210](https://www.similarweb.com/website/elegoo.com/)          | :heavy_check_mark: | https://www.elegoo.com      | SVG in printer [user guides](https://www.elegoo.com/pages/download)                                                 |    #10467    |
| Flashforge     | SW: [226,392](https://www.similarweb.com/website/flashforge.com/)     | :heavy_check_mark: | https://www.flashforge.com  | SVG in printer [user guides](https://www.flashforge.com/download-center/48)                                         |    #10584    |
| FLSUN          | SW: [279,506](https://www.similarweb.com/website/flsun3d.com/)        | :heavy_check_mark: | https://flsun3d.com         | None                                                                                                                |              |
| ~~LDO Motors~~ | SW: [608,237](https://www.similarweb.com/website/ldomotors.com/)      |        :x:         | https://ldomotors.com       | ~~SVG in header of [docs site](https://docs.ldomotors.com/en/voron/voron2)~~                                        |              |
| Phrozen        | SW: [190,785](https://www.similarweb.com/website/phrozen3d.com/)      | :heavy_check_mark: | https://phrozen3d.com       | SVG in printer [user guides](https://phrozen3d.com/pages/manual)                                                    |              |
[...]

PR #10584, opened 2024-03-07
Add Flashforge icon

ApexDevelopment:
flashforge

**Issue:** contributes to #10464

**Similarweb rank:** https://www.similarweb.com/website/flashforge.com/

### Checklist

### Description

SVG is from the website header. https://flashforge.com/ includes the registered trademark symbol, but https://flashforge-usa.com/ does not. It is missing from some of their manuals, and their socials also do not use it. I can't find usage guidelines.

I could use some advising on the hex code. I chose black because it is how it appears on their main website, but Flashforge USA uses a shade of blue, and in their PDF manuals it's an off-black.


PeterShaggyNoble (approved):
SVG looks good, @ApexDevelopment, nice work 👍🏻 On the colour, there doesn't seem to be a clear brand colour to me, either. Black looks to be the dominant colour to me but, even then, they use many different shades, so pure black is as good a choice as any.

PR #10596, opened 2024-03-08
Fix collinear segment on United Nations

mondeja:
I've manually changed the segment, ran SVGO on it and visually checked that don't breaks the geometry.

The problem is here:

image


service-paradis (approved):
Perfect 🎉, thank you @mondeja for fixing the linter and this icon!

PR #10599, opened 2024-03-08
Fix false positives in collinear-segments rule

mondeja:
This PR handles an edge case not contemplated by `collinear-segments` lint rule. Quoting the SVG 1.1 spec:

> If a moveto is followed by multiple pairs of coordinates, the subsequent pairs are treated as implicit lineto commands.

This was raising a few false positives for BNB Chain and Deutsche Post. The solution is to only reset the start point for chained moveto commands in first segments of chains.


mondeja:
I'm merging yolo.

PR #10601, opened 2024-03-09
Use prettier's `--cache` option

mondeja:
This speeds a lot subsecuent executions of Prettier locally, specially useful when formatting on pre-commit GIT's hook.  Note that Prettier stores the cache file at `./node_modules/.cache/prettier/.prettier-cache`, so this is not used on CI.


service-paradis (approved):

mondeja:
We can always in the future include it on CI...

PR #10602, opened 2024-03-09
Only lint changed SVG icons on CI

mondeja:
Allows to run SVGLint only for a subset of icons locally and only for changed/added icons on CI. 

- `npm run lint --icons=icons/simpleicons.svg`
- `npm run svglint --icons=icons/simpleicons.svg`
- `npm run lint --icons='icons/simpleicons.svg icons/nodedotjs.svg'`

By default uses the value in *.npmrc*, see https://docs.npmjs.com/cli/v10/using-npm/config. This should be cross platform taking into account that CMD is not the default shell on Windows.


mondeja:
🤔 Seems that this wouldn't work on Windows...

LitoMore (approved):

PR #10613, opened 2024-03-12
Revert to `kceb/git-message-action@v2` on CI

mondeja:
While https://github.com/kceb/git-message-action/pull/18 is not released, we must stick with v2


LitoMore (approved):

LitoMore:
Yes. Let's use v2 til they fix the bug.

Issue #10610, opened 2024-03-11
Release Tags Now Include "v"


mackenly:
Release Tags are now prefixed by"v". I suspect it was one of the changes in #10600, however, nothing seems particularly off other than incrementing action versions. This has the effect of breaking downstream CD for 3rd party extensions. Is this something that should be fixed or just leave it?

- Old tags: 11.7.0
- Latest tag: v11.8.0

[View Releases](https://github.com/simple-icons/simple-icons/tags)

mackenly:
This issue didn't seem to fit any of the issue templates.

mondeja:
I think that this is not related to #10600. Seems that the latest release failed and @LitoMore pushed manually, I'm wrong?

LitoMore:
Ah, yes. The GitHub Release action fail so I pushed manually. Let me remove the wrong tag and release new one instead.

LitoMore:
@mackenly I've deleted the wrong tag `v11.8.0` and pushed `11.8.0` instead. The tag on https://github.com/simple-icons/simple-icons/releases/tag/11.8.0 has been updated.

mondeja:
Fixed now, thanks @LitoMore! #10614 will prevent that this will not happen again.

PR #10614, opened 2024-03-12
Add pre-push GIT hook and workflow to check release tag

mondeja:
Add a pre-push GIT hook to disallow push tags following other format than `X.Y.Z`. Prevents https://github.com/simple-icons/simple-icons/issues/10610 to happen again. EDIT: also adds a workflow to check if a release is created with the wrong tag and deletes it if so.

This is not the first time that happened. I'm not sure if I pushed two years ago a tag with a `v` prefix, but fixed the mistake 10 minutes later. The tag format of simple-icons is not really common so these things can happen and with a lot of third party extensions now a mistake like that can be a real pain.


mondeja:
I'm not sure if this is cross platform, perhaps could break pushes on Windows 🤔

mondeja:
OK, so if an user is capable of running `.husky/pre-push` before push it means that it has been installed dev dependencies. In that case, has node installed so we can write this as a node script. Rewriting...

LitoMore:
I used `create new release` to create a tag. I'm afraid this hook cannot detect that action.

LitoMore:
Here https://github.com/simple-icons/simple-icons/releases.

It will create a tag when creating new release.

mondeja:
Ahm... 🤔 I'll check if could be some workflow that can be configured at that level.

mondeja:
Hi @LitoMore. I've added a workflow to handle the case that you've faced, tested [here](https://github.com/mondeja/test-tag-release-create-ci-delete/actions/runs/8486806539/job/23253758959#step:5:12) and seems to work fine. Note that to remove releases and tags the app needs the following permission, which is currently granted:

image

Would you mind to review again?

LitoMore:
The Packagist is using the GitHub Webhook event to publish new versions. But this code still cannot intercept the webhook action.

mondeja:
Closing for now...

PR #10616, opened 2024-03-12
Fix collinear segments on Red Bull and optimize it

mondeja:
Optimized with SVGO and removed a few collinear segments manually.


service-paradis (approved):
Good job, thank you @mondeja!

PR #10617, opened 2024-03-12
Optimize OWASP Dependency-Check

mondeja:
Before:

image

After:

image

Note that there is a tiny difference in a curve. The previous bezier curve was a complete mess with a lot straigh lines of collinear segments.


service-paradis (approved):
The difference is slightly more pronounced that I would like, but considering the mess it was before, I'm approving it! We might need to revisit this one someday, but I consider this as an improvement! Thank you @mondeja!

mondeja:
> The difference is slightly more pronounced that I would like, but considering the mess it was before, I'm approving it! We might need to revisit this one someday, but I consider this as an improvement! Thank you @mondeja!

Additionally, I don't know why it has an slug, seems to be that is not needed.

service-paradis:
Hum... Good question! I suppose it was to simplify it by removing OWASP from it...Now, it would be a breaking change to remove it. Maybe we'll flag this when doing the great review for the letter O!

PR #10638, opened 2024-03-14
Use `kceb/git-message-action@v3.0.1`

LitoMore:
We cannot use `@v3` for now due to https://github.com/kceb/git-message-action/issues/19


mondeja:
Seems to me that is worth to wait for `v3` instead of having to update it two times.

kceb:
Hey I think I fixed this. LMK if it works

@mondeja @LitoMore

mondeja (approved):

PR #10643, opened 2024-03-14
Add XO linter

LitoMore:
I intentionally didn't fix those errors on the commit [`64fefbd` (#10643)](https://github.com/simple-icons/simple-icons/pull/10643/commits/64fefbd305c161479ae151b7f6f360fa45b37bfa), so you can play XO.

### Changes

- Apply most of the XO fixes. For some minor errors, I used `eslint-ignore.`
- Move the `.svglintrc.mjs` code body to `scripts/svglint-config.mjs`. It's related to https://github.com/xojs/xo/issues/673.
- Remove `prettier` dependency, XO has a built-in prettier
- Limit the `prettier --fix` command to only format markdown and JSON files
- Remove the incorrect `engines` field from `package.json`

### Features

It has a friendly error prompt, and you can simply the link on the prompt to see the reason:

image

It has tons of rules and good practices built in.

Let me know if you have any questions.

#### VS Code `.vscode/settings.json`

Some options can be omitted. It's up to your global configuration.

```
{
  "prettier.enable": true,
  "eslint.enable": false,
  "xo.enable": true,
  "xo.format.enable": true,
  "editor.formatOnSave": true,
  "[javascript]": {
    "editor.defaultFormatter": "samverschueren.linter-xo"
  },
  "[json]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[markdown]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  }
}
```

For other editors, please refer to https://github.com/xojs/xo?tab=readme-ov-file#editor-plugins.


package.json
```diff
@@ -126,8 +128,5 @@
     "posttest": "npm run postpublish",
     "get-filename": "node scripts/get-filename.js",
     "add-icon-data": "node scripts/add-icon-data.js"
-  },
-  "engines": {
-    "node": ">=0.12.18"
```
mondeja, line 131: Why this has been removed?

mondeja:
Really great, you're right about xo 🥇

package.json
LitoMore: You can add it back and see the error from `n/no-unsupported-features/es-syntax`.

I think this is unnecessary since `0.12.18` is a really old version. We can totally ignore it.

mondeja: 🤔 It seems there is something we have missed so far. There are multiple ways to use the npm package: the SDK with its TS types, the library's TS types, and the library itself, and ideally we should be able to set values of this field for each.

If is not possible, I think that we should ignore the rule to maintain proper compatibility with all supported versions, just for legacy reasons. Seems to me that do not hurts.

LitoMore: Targeting the engine to the `0.12.18` is not accurate anymore. You can see tons of errors like:

mondeja: Could you set the rule to run only for distributed files?

LitoMore: I've set the rule with `overrides`.

LitoMore: I think the engine field can be removed because the `0.12.18` is not correct anymore.

.xo-config.json
```diff
@@ -0,0 +1,47 @@
+{
```
mondeja, line 1: > Move the .svglintrc.mjs code body to scripts/svglint-config.mjs. It's related to https://github.com/xojs/xo/issues/673

Can you use `"!.,svglintrc.mjs"` in the `ignores` field of `.xo-config.json`?

mondeja (changes requested):

package.json
```diff
@@ -98,21 +99,21 @@
...
     "svgpath": "2.6.0",
-    "typescript": "5.2.2"
+    "typescript": "5.2.2",
+    "xo": "0.58.0"
   },
   "scripts": {
     "build": "node scripts/build/package.js",
     "clean": "del-cli index.js index.mjs index.d.ts sdk.js",
-    "format": "prettier --cache --write .",
+    "format": "prettier --cache --write '**/*.md' '**/*.json'",
```
mondeja, line 113: We can't remove the prettier dependency if is defined here.

mondeja: Also, why these globs? Why there have been so many removes of the *.prettierignore* file?

package.json
mondeja: Most of the people is using the library, not the SDK. What happens if you readd `engines.node`? Seems to me that doesn't hurts.

mondeja (changes requested):

.xo-config.json
```diff
@@ -0,0 +1,47 @@
...
+        },
+        "warnOnUnassignedImports": true,
+        "newlines-between": "never"
+      }
+    ]
+  },
+  "overrides": [
+    {
+      "files": ["sdk.mjs"],
+      "nodeVersion": ">=14"
```
mondeja, line 37: What happens if you decrease this version. Could you explain why is Node.js v14 the minimum version supported by the SDK?

scripts/build/package.js
```diff
@@ -93,33 +101,33 @@ const build = async () => {
...
   const rawIndexJs = `${constantsString}module.exports={${iconsBarrelJs.join(
     '',
   )}};`;
   await writeJs(indexJsFile, rawIndexJs);
-  // and ESM
+  // And ESM
   const rawIndexMjs = constantsString + iconsBarrelMjs.join('');
   await writeJs(indexMjsFile, rawIndexMjs);
-  // and create a type declaration file
+  // And create a type declaration file
```
mondeja, line 121: What about this?

```js
// ... and ESM
// ... and create a type declaration file

File: scripts/get-filename.js [hunk 6]
```diff
@@ -5,15 +5,19 @@
...
   process.exit(1);
 } else {
   const brandName = process.argv
     .slice(3)
-    .reduce((acc, arg) => `${acc} ${arg}`, process.argv[2]);
+    // eslint-disable-next-line unicorn/no-array-reduce
+    .reduce(
+      (accumulator, argument) => `${accumulator} ${argument}`,
+      process.argv[2],
+    );
```
mondeja, line 20: This code is a complete madness, please change it by just:

```js
const brandName = process.argv[2];
```

mondeja (changes requested):

sdk.mjs
```diff
@@ -201,11 +207,11 @@ export const getThirdPartyExtensions = async (
       module = module.split('<img src="')[0];
       return {
         module: {
-          name: /\[(.+)\]/.exec(module)[1],
+          name: /\[(.+)]/.exec(module)[1],
           url: /\((.+)\)/.exec(module)[1],
         },
         author: {
-          name: /\[(.+)\]/.exec(author)[1],
+          name: /\[(.+)]/.exec(author)[1],
```
mondeja, line 214: Is this a bug in xo? Note that the first `[` is escaped.

types.d.ts
```diff
@@ -5,6 +5,7 @@
  */
 export type License = SPDXLicense | CustomLicense;
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
mondeja, line 8: Why this has been disabled?

.xo-config.json
```diff
@@ -0,0 +1,47 @@
...
+    },
+    {
+      "files": ["scripts/**/*", "tests/**/*"],
+      "nodeVersion": ">=18",
+      "rules": {
+        "unicorn/no-process-exit": "off"
```
mondeja, line 43: Why do you've ignored this rule. It seems to me that is fine to add, just we need to add `#!/usr/bin/env node` to the scripts, see https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-process-exit.md

sdk.mjs
```diff
@@ -116,43 +116,45 @@ export const titleToHtmlFriendly = (brandTitle) =>
...
+      String.fromCodePoint(Number.parseInt(number_, 10)),
+    )
+    .replaceAll(
       /&(quot|amp|lt|gt);/g,
-      (_, ref) => ({ quot: '"', amp: '&', lt: '<', gt: '>' })[ref],
+      (_, reference) => ({quot: '"', amp: '&', lt: '<', gt: '>'})[reference],
     );
 
 /**
  * Get path of *_data/simpe-icons.json*.
```
uncenter, line 128:
```suggestion
 * Get path of *_data/simple-icons.json*.
```
Should we fix this little typo in this PR lol?

scripts/add-icon-data.js
uncenter, line 1: Thoughts on these new "compact" imports? I must say I much prefer having a space after the opening curly brace and before the closing one, as it was before. Seems more readable to me that way.

.xo-config.json
LitoMore: This won't work. You can try this on your local. This will affect the original XO configuration.

package.json
LitoMore: I would remove the incorrect field. If you really want this, we should use `>=12.16.0` instead.

We're using conditional exports in `packcage.json`. See https://nodejs.org/api/packages.html#conditional-exports.

.xo-config.json
LitoMore: image

Because we're using `fs/promises`.

scripts/build/package.js
LitoMore: Good idea.

scripts/get-filename.js
LitoMore: Haha, sure.

sdk.mjs
LitoMore: CleanShot 2024-03-19 at 19 39 09@2x

See.

types.d.ts
LitoMore: I strongly recommend you please with XO on your local so that you can know why this was added.

We're using some `UPPER_CASE` types/variables, and it's exported to our users. We can change its case right now.

.xo-config.json
LitoMore: Then you should change those files to executable files. I don't think it's reasonable to have a shebang for a file that can't be executed.

mondeja: What do you mean with "for a file that can't be executed". Yes, can be executed using `./scripts/file.js` directly if we add shebangs for them.

scripts/add-icon-data.js
LitoMore: Please no. This stuff makes lines longer and wrapped. I don't want this.

In the meantime, I'm using XO's default style; we should avoid adding too much personal coding style to it.

LitoMore: If you have questions about the default style of XO, you can go to XO to search for related issues, and Sindre Sorhus will always give you a good reason.

.xo-config.json
uncenter: > and make them executables

I think this is what @LitoMore was getting at, we would have to make them executable with `chmod +x`. Simple fix though and I think it makes sense!

sdk.mjs
mondeja: Yes, seems to be that if there is a `\[` at the start of a `[]` group, the entire group is treated as literal, but IMHO anyways looks like unreadable and counterintuitive. Don't worry, I'll replace this regular expressions on #10637

.xo-config.json
LitoMore: If you agree to update those `node scripts/file.js` to `./scripts/files.js` with `chmod +x`, then it's OK.

.xo-config.json
mondeja: Great. So can we add a check of the major Node.js version used to the start of the SDK file? By that approach can have something like `node.engine` checks for each export.

scripts/add-icon-data.js
uncenter: I found some related issues but no explanation for this choice: https://github.com/xojs/xo/issues/614, https://github.com/xojs/xo/issues/440. I wouldn't say having spaces in objects like that is a "personal coding style" as it has been the general guideline/practice in the ecosystem for years. Prettier itself has it as a default: https://prettier.io/docs/en/options#bracket-spacing.

LitoMore: Whatever, you can always change the code style if you prefer something. Getting involved in coding-style wars is really boring. So both are OK for me. I just want a easily configurable, auto-formatting style.

.xo-config.json
mondeja: I've tested locally and works. See the [`ignores` documentation](https://github.com/xojs/xo?tab=readme-ov-file#ignores), it says:

> Additional ignores can be added here.

package.json
mondeja: See this comment: https://github.com/simple-icons/simple-icons/pull/10643#discussion_r1530223931 We can check the Node.js version in the SDK itself because almost nobody is using it.

.xo-config.json
LitoMore: > I've tested locally and works.

What about other files? XO on other files is not working anymore.

package.json
mondeja: Can we add a negative glob here to format all files except `.m?js` ones with Prettier?

LitoMore: > Can we add a negative glob here to format all files except .m?js ones with Prettier?

Not sure. I will give a try.

types.d.ts
mondeja: I've tested and I can't see errors running xo.

.xo-config.json
LitoMore: How about publishing the SDK as a new package?

.xo-config.json
mondeja: Oh I see, what a crazy feature. Let me see if this can be fixed using a *xo-config.mjs* config file and trying to import the rest of the ignore.

scripts/add-icon-data.js
mondeja: Altough strange, I'm fine with the change.

types.d.ts
LitoMore: CleanShot 2024-03-19 at 20 15 16@2x

It's recommended case is `SpdxLicense`.

.xo-config.json
LitoMore: Sure. Thanks a lot.

.xo-config.json
mondeja: Yep, that can be another option. Something like `@simple-icons/sdk`. But that would be a breaking change that we can postpone for the v12.

I'm fine if you don't want add the check here but still thinking that the `node.engine` is worth for the main package.

types.d.ts
mondeja: Ah OK, that would be a breaking. Fine enough :+1: Sorry it was not being executed locally.

.xo-config.json
LitoMore: Sure. I will add the `node.engine` back. And it should be `>=12.16.0` due to https://nodejs.org/api/packages.html#conditional-exports.

.xo-config.json
mondeja: Yes, please do it 👍🏼 just `+x` is enough!

mondeja (changes requested):
The file *.pretierrc.json* must be removed. The config option `"singleQuote": true` is not doing nothing now.

sdk.mjs
mondeja: Do you know about a mispell fixer like [mispell-fixer](https://github.com/vlajos/misspell-fixer) or [codespell](https://pypi.org/project/codespell/) for JS?

LitoMore: I would avoid cspell for our project. You will need to maintain a super-long dictionary like this https://github.com/netless-io/flat/blob/main/cspell.config.js.

You can install it in your local environment.

Typos are OK for me. Just fix it when you find it.

.xo-config.json
mondeja: Okay, seems a bug in the globby dependency of XO.

I think that the most suitable alternative is to add ` --config svglintrc.mjs` to `svglint` script (at *package.json*) move *scripts/svglint-config.mjs* to *svglintrc.mjs* and drop *.svglintrc.mjs*. That will keep the blame of the SVGLint config file, which I'm most concerned about.

sdk.mjs
LitoMore: You could fix those typos in another PR. Those typos are not related to this topic.

package.json
LitoMore: Prettier does not support `!**/*` pattern.

mondeja: Yes, Prettier uses fast-glob, which uses micromatch that [accepts negation patterns](https://github.com/micromatch/micromatch?tab=readme-ov-file#extglobs). Use `'**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)'`

mondeja (changes requested):

package.json
mondeja: Note that you don't have added `.json5`, which is the extension of the file *.github/renovate.json5*. I don't want to remember to add more extensions here, a false positive reformat would catch it in the future if we add more files with other extensions.

LitoMore: Tried `'**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)'` but this will include all other files like `.npmrc`, `.gitignore`, etc.

Should I add them to the `.prettierignore`?

mondeja: Yes, please. Anyways XO doesn't formats them.

package.json
```diff
@@ -97,34 +98,34 @@
...
     "test": "mocha tests --reporter tests/min-reporter.cjs --inline-diffs",
     "pretest": "npm run prepublishOnly",
     "posttest": "npm run postpublish",
-    "get-filename": "node scripts/get-filename.js",
-    "add-icon-data": "node scripts/add-icon-data.js"
+    "get-filename": "./scripts/get-filename.js",
+    "add-icon-data": "./scripts/add-icon-data.js"
   },
   "engines": {
-    "node": ">=0.12.18"
```
mondeja, line 128: Sorry I think thay we are not in the same boat. This would make simple-icons to be uninstallable on Node.js versions prior to 12. Could you revert this change?

My suggested solution is to either:

a) Add a check to *sdk.mjs* to verify that imports of this file are being doing on Node.js >= 12.16.
b) Separate SDK on a different library.

Any of these can be done in another PR.

mondeja (changes requested):

package.json
LitoMore: Oh no. I can't add those patterns to `.prettierignore`, and the XO would ignore those files as well.

I just added an `--ignore-unknown` to bypass those unknown files.

package.json
LitoMore: np.

package.json
```diff
@@ -97,32 +98,32 @@
...
-    "clean": "node scripts/build/clean.js",
-    "format": "prettier --cache --write .",
+    "build": "./scripts/build/package.js",
+    "clean": "./scripts/build/clean.js",
+    "format": "prettier --cache --write --ignore-unknown '**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)' && xo --fix",
     "lint": "npm run ourlint && npm run jslint && npm run jsonlint && npm run svglint && npm run wslint",
-    "ourlint": "node scripts/lint/ourlint.js",
-    "jslint": "prettier --cache --check .",
-    "jsonlint": "node scripts/lint/jsonlint.js",
-    "svglint": "svglint --ci $npm_config_icons",
```
mondeja, line 116: We need to add a  `prettierlint` script to run Prettier with `--check` in the `lint` script with the rest of the parameters of the `format` script.

mondeja (changes requested):
It's almost there, just one more small change...

mondeja (approved):
Awesome. Thanks for this great work @LitoMore! 🥇

PR #10651, opened 2024-03-15
Drop npm-run-all dependency

mondeja:
npm-run-all currently adds 87 dependencies (total are 456) and a lot are repeated of other versions, which adds a lot of time to dependency tree resolution.

Just for sintactic sugar of something that we never touch.We are only using it to run a lot of `npm run` synchonusly in the `lint` script. After this change, fresh installs are much more faster (~2 seconds less on CI).


LitoMore (approved):
How about rename `our-lint` to `ourlint`? Only this script is in a different case.

mondeja:
Makes sense, done!

PR #10652, opened 2024-03-15
Drop del-cli dependency

mondeja:
23 packages less to install, just to remove 4 files. Better with a simple script.


LitoMore:
I'd prefer to keep the `del-cli`. See here https://github.com/sindresorhus/del-cli#comparison

The `del` has safe checks. It's more stable in various scenes. Here is the source code: https://github.com/sindresorhus/del/blob/main/index.js

mondeja:
But we are just removing 4 files that are created by our own script. What safety provides to us in **this** context?

mondeja:
What do you think about this @uncenter?

LitoMore:
> I don't think that the comparison with rimraf is even relevant.

Users may execute the script under another directory.

mondeja:
> > I don't think that the comparison with rimraf is even relevant.
> 
> Users may execute the script under another directory. You're using a relative path for deleting files.

They can always install `del-cli` globally, should be the recommended way. I don't know why we should provide them a CLI to delete files "safely" locally for the repository when we are a icons library, that point doesn't makes sense at all.

adamrusted:
> Users may execute the script under another directory. You're using a relative path for deleting files.

Most of our users, myself included, would just run the script from the VS Code 'NPM Scripts' box at the bottom of their IDE. This would surely run it at the root of the project you're currently in, mitigating the issue we're having. If people are executing the script via the CLI then they should have enough knowledge to parse the error code generated.

mondeja:
Now the paths are absolute, no matter from what directory is executed.

LitoMore:
I still think it did not optimize our project, but it added a `clean.js` that needs to be maintained. I reserve my opinion, but it's up to you to do whatever you want.


BTW, the `npm run clean` fails if there is no file to delete. This needs to be fixed.

mondeja:
> BTW, the npm run clean fails if there is no file to delete. This needs to be fixed.

What do you mean with fix it? The error is handled:

```
$ npm run clean

> simple-icons@11.8.0 clean
> node scripts/build/clean.js

Error cleaning files: ENOENT: no such file or directory, unlink '/.../simple-icons/index.js'
```

scripts/build/clean.js
```diff
@@ -0,0 +1,21 @@
...
+
+Promise.all([
+  fs.unlink(path.join(rootDirectory, 'index.js')),
+  fs.unlink(path.join(rootDirectory, 'index.mjs')),
+  fs.unlink(path.join(rootDirectory, 'index.d.ts')),
+  fs.unlink(path.join(rootDirectory, 'sdk.js')),
+]).catch((error) => {
+  console.error(`Error cleaning files: ${error.message}`);
+  process.exit(1);
+});
```
LitoMore, line 21:
```suggestion
const __dirname = getDirnameFromImportMeta(import.meta.url);
const rootDirectory = path.resolve(__dirname, '..', '..');
const files = ['index.js', 'index.mjs', 'index.d.ts', 'sdk.js'];

Promise.all(
  files.map(file => fs.unlink(path.join(rootDirectory, file)))
).catch((error) => {
  console.error(`Error cleaning files: ${error.message}`);
  process.exit(1);
});
```

I coded on GitHub manually, you may need a reformat with Prettier.

LitoMore:
@mondeja Just let the script always return exit code 0. It doesn’t matter if there are no files to delete.

This script should work fine even if there is only one file. Even there is only one `index.js` file.

uncenter:
Hmm well ideally it would exit 1 on an error outside of the files not existing.

mondeja:
Let me improve it...

mondeja:
Now looks better.

LitoMore:
We should always return exit code 0. Otherwise, it will block subsequent command execution.

As I mentioned: *The script should work fine even if there is only one file. Even there is only one index.js file*.

mondeja:
> We should always return exit code 0. Otherwise, it will block subsequent command execution.
> 
> As I mentioned: _The script should work fine even if there is only one file. Even there is only one index.js file_.

Note that is the current behaviour. You can test it with:

```sh
$ touch index.js
$ node scripts/build/clean.js 
File index.mjs does not exist, skipping...
File index.d.ts does not exist, skipping...
File sdk.js does not exist, skipping...
$ echo $?
0
$ node scripts/build/clean.js 
File index.js does not exist, skipping...
File index.d.ts does not exist, skipping...
File index.mjs does not exist, skipping...
File sdk.js does not exist, skipping...
$ echo $?
0
```

It only returns 1 exit code if there is an error unlinking a file.

uncenter (approved):

LitoMore:
Great.

LitoMore (approved):

PR #10686, opened 2024-03-21
Update OSGeo hex and guidelines

adamjstewart:
Very slight change in color, also added links to document the source.

### Before

before

### After

after


_data/simple-icons.json
```diff
@@ -10929,8 +10929,9 @@
         },
         {
             "title": "OSGeo",
-            "hex": "5CAE58",
-            "source": "https://www.osgeo.org"
+            "hex": "4CB05B",
+            "source": "https://github.com/OSGeo/osgeo/blob/5ee0625fbef339d6badc14cdd121d0174161a192/marketing/branding/logo/osgeo-logo-bw.svg",
+            "guidelines": "https://github.com/OSGeo/osgeo/blob/5e3df1231cfd1f3a1cb087dd541a4f9b8ed85a89/marketing/branding/styleguide-osgeo.pdf"
```
PeterShaggyNoble, line 10934: For `guidelines` on GitHub, we use the default branch rather than the commit hash to ensure that we're always providing the latest version.
```suggestion
            "guidelines": "https://github.com/OSGeo/osgeo/blob/master/marketing/branding/styleguide-osgeo.pdf"
```

PeterShaggyNoble (changes requested):
Thanks, @adamjstewart 👍🏻 Did you do a comparison between our current icon and the new source to make sure they line up perfectly? If so then just the one small change below and this is ready to merge.

_data/simple-icons.json
adamjstewart: I noticed a lot of other guidelines use commit hashes for stability/reproducibility. Is this a new policy? I'm fine with whatever you decide.

_data/simple-icons.json
```diff
@@ -10929,8 +10929,9 @@
         },
         {
             "title": "OSGeo",
-            "hex": "5CAE58",
-            "source": "https://www.osgeo.org"
+            "hex": "4CB05B",
+            "source": "https://github.com/OSGeo/osgeo/blob/5ee0625fbef339d6badc14cdd121d0174161a192/marketing/branding/logo/osgeo-logo-bw.svg",
```
adamjstewart, line 10933:
```suggestion
            "source": "https://github.com/OSGeo/osgeo/blob/master/marketing/branding/logo/osgeo-logo-bw.svg",
```

adamjstewart:
I did not compare the new and old source to make sure they're the same, I only updated the URL. I'm happy to undo that change if you want me to.

_data/simple-icons.json
adamjstewart: Ah, it looks like source is required to use the hash but guidelines isn't. Would be good to make these consistent and add tests to enforce it.

PeterShaggyNoble:
> I noticed a lot of other guidelines use commit hashes for stability/reproducibility. Is this a new policy? I'm fine with whatever you decide.

Yeah, after looking at a few more entries, it seems we've become super-inconsistent about whether we use the default branch or the commit hash for the `guidelines`. We'll need to sort that out and address it. For this one, though, I've since found [this page](https://www.osgeo.org/about/branding-material) on their website, which I think would work better for the `guidelines` as it's more likely to be kept up to date with any future changes.

> I did not compare the new and old source to make sure they're the same, I only updated the URL. I'm happy to undo that change if you want me to.

If we're not at least doing a comparison against the new `source` then, yes, we should keep our existing one.

PeterShaggyNoble (approved):
Thanks, @adamjstewart :+1:

PR #10726, opened 2024-03-28
Use `simple-icons[bot]` for actions

LitoMore:
I've set up the `BOT_PRIVATE_KEY` and `BOT_APP_ID` in organization settings for the action https://github.com/actions/create-github-app-token. I've changed all role-related visible actions to the token of our GitHub App.

### References

This is the main idea for the GitHub App in GitHub Actions:
- https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/making-authenticated-api-requests-with-a-github-app-in-a-github-actions-workflow

And this is how I generate the private key:
- https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps

### DIFF
--- a/.github/workflows/add-labels-priority.yml
+++ b/.github/workflows/add-labels-priority.yml
@@ -8,18 +8,27 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.action == 'opened'
     steps:
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.BOT_APP_ID }}
+          private-key: ${{ secrets.BOT_PRIVATE_KEY }}
       - uses: simple-icons/labeler@v1
         with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          repo-token: ${{ steps.app-token.outputs.token }}
 
   assign-to-project:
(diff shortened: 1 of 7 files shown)


.github/workflows/add-labels-priority.yml
```diff
@@ -8,18 +8,27 @@ jobs:
...
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          repo-token: ${{ steps.app-token.outputs.token }}
 
   assign-to-project:
     runs-on: ubuntu-latest
     name: Assign to Project
     if: github.event.action == 'opened'
     needs: triage
-    env:
-      MY_GITHUB_TOKEN: ${{ secrets.AUTO_ASSIGN_WORKFLOW_TOKEN }}
```
mondeja, line 21: I think that the next permission must be configured on the app too at organization level rather than at repository level.

image

LitoMore: Agree.

mondeja (approved):
Should be fine now. I'm going to keep the secrets configured just in case that something will not work and we would must revert this. If all works as expected will remove them ASAP.

Awesome work! 🤖

PR #10727, opened 2024-03-28
Switch single-quotes for double-quotes to allow scripts to run on Powershell.

adamrusted:
The husky hooks were causing issue on my machine, which boils down to the use of single-quotes in our package.json commands. Amending those make the commands run universally.


package.json
```diff
@@ -109,10 +109,10 @@
   "scripts": {
     "build": "./scripts/build/package.js",
     "clean": "./scripts/build/clean.js",
-    "format": "prettier --cache --write --ignore-unknown '**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)' && xo --fix",
+    "format": "prettier --cache --write --ignore-unknown \"**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)\" && xo --fix",
```
mondeja, line 112: Would you mind to apply the next suggestion instead to deduplicate the pattern?

1. Set the pattern at *.npmrc* without using quotes like `prettier_files=**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)` and add a comment indicating that is used at `format` and `prettierlint` scripts
2. Set them in these two script with `$` syntax as environment variables:

```suggestion
    "format": "prettier --cache --write --ignore-unknown $prettier_files && xo --fix",
```

And check if works on PorwerShell. It should work because seems to be no problems with the `svglint` script, that uses this approach.

adamrusted: Unfortunately, Windows doesn't want to play nice with that.
image

I've also noticed the SVGLint command isn't running against icons for me at the moment (I have to specify them individually after `npm run svglint` to get it to parse correctly...)

mondeja (approved):
Okey, thanks! I'll check another solution for the svglint script.

package.json
mondeja: > I've also noticed the SVGLint command isn't running against icons for me at the moment (I have to specify them individually after `npm run svglint` to get it to parse correctly...)

Fixed by #10733

PR #10733, opened 2024-03-29
Fix PowerShell support for `svglint` script

mondeja:
Fixes https://github.com/simple-icons/simple-icons/pull/10727#discussion_r1544624796


svglint.config.mjs
```diff
@@ -180,7 +180,7 @@ const ignoreIcon = (linterName, path, $) => {
   iconIgnored[linterName][path] = iconName;
 };
 
-export const config = {
```
mondeja, line 183: Prevented in future by https://github.com/birjj/svglint/pull/96

PR #10804, opened 2024-04-11
Fix issue templates

adamrusted:
There are various labels mentioned in our issue templates that are no longer used. PR fixes them for their current equivelants.

### DIFF
--- a/.github/ISSUE_TEMPLATE/package.yml
+++ b/.github/ISSUE_TEMPLATE/package.yml
@@ -1,16 +1,16 @@
 name: Packages
-description: Report problems and suggest ideas for the packages
-labels: [meta]
+description: Report problems and suggest ideas for the NPM and Packagist packages
+labels: [package]
 
 body:
   - type: markdown
     attributes:
       value: |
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/package.yml
```diff
@@ -1,16 +1,16 @@
 name: Packages
 description: Report problems and suggest ideas for the packages
-labels: [package]
+labels: [meta]
```
mondeja, line 3: The label `package` is still in use for npm/Packagist packages issues or PRs and it shouldn't be confused with `meta`, which is for the internal repo mangement, the build process, the lint process... etc. Why you've changed it?

adamrusted: We can revert it - I just went by the descriptions :)

mondeja: We can add another template for meta describing what it is. Sometimes users open issues that fall into this category.

mondeja (approved):

PR #10863, opened 2024-04-26
Use `import.meta.dirname` & `import.meta.filename`

LitoMore:
This API requires Node.js `^20.11 || >=22` . See https://nodejs.org/api/esm.html#importmetadirname.

In the meantime, we can consider publishing the SDK with a new package, for example `@simple-icons/sdk`.


mondeja:
> In the meantime, we can consider publishing the SDK with a new package, for example @simple-icons/sdk.

But this would still be a breaking change, right? It would be nice to do it at v12.

LitoMore:
>.But this would still be a breaking change, right?

Yes. It's just a draft PR.

mondeja:
> > .But this would still be a breaking change, right?
> 
> Yes. It's just a draft PR.

I refer to publishing the SDK with a new package.

mondeja (approved):

sdk.mjs
```diff
@@ -190,10 +174,7 @@ export const normalizeColor = (text) => {
  * @returns {Promise<ThirdPartyExtension[]>} Information about third party extensions
  */
 export const getThirdPartyExtensions = async (
-  readmePath = path.join(
-    getDirnameFromImportMeta(import.meta.url),
-    'README.md',
-  ),
+  readmePath = path.join(import.meta.dirname, 'README.md'),
```
mondeja, line 177: This change needs to be applied to the new function `getThirdPartyLibraries` too.

mondeja (changes requested):

adamrusted:
@LitoMore - is this likely to be ready for v12, or should we push it out to v13?

adamrusted:
@LitoMore - reminder that this is still outstanding. Are you likely to get this completed for v15?

sdk.mjs
```diff
@@ -40,29 +39,14 @@ const TITLE_TO_SLUG_RANGE_REGEX = /[^a-z\d]/g;
...
- * @param {string} [jsonschemaPath] Path to the *.jsonschema.json* file.
  * @returns {Promise<RegExp>} Regex to validate HTTPs URLs.
  */
-export const urlRegex = async (
-	jsonschemaPath = path.join(
-		getDirnameFromImportMeta(import.meta.url),
-		'.jsonschema.json',
-	),
-) =>
+export const urlRegex = async () =>
```
mondeja, line 46: I think that this function `urlRegex` can be moved to *scripts/add-icon-data.js*, which is the unique script that is using it. Could you do that on this PR and run again `node scripts/release/update-sdk-ts-defs.js`?

.xo-config.json
```diff
@@ -47,7 +47,7 @@
 	"overrides": [
 		{
 			"files": ["sdk.mjs", "sdk.d.ts"],
-			"nodeVersion": ">=14.13.1",
+			"nodeVersion": "^20.11 || >=22",
```
mondeja, line 50: Just letting a comment here for further reviews...

The version `20.11` added `import.meta.dirname` and `import.meta.filename` (see https://nodejs.org/es/blog/release/v20.11.0).

LitoMore: I'm unable to add a comment since it's a JSON file. But we can add the comment after the #13015.

sdk.mjs
LitoMore: Done!

PR #10870, opened 2024-04-28
Increase similarity threshold on potential duplicates workflow

uncenter:
See title. I've been seeing many many false positives at exactly 70%, bumping up to 80% should still catch some actual duplicates while ignoring the obvious false positives.

### DIFF
--- a/.github/workflows/potential-duplicates.yml
+++ b/.github/workflows/potential-duplicates.yml
@@ -41,7 +41,7 @@ jobs:
             for
             from
           state: all
-          threshold: 0.7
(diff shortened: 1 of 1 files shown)


mondeja (commented):
What about 75?

uncenter:
> What about 75?

We could try that I guess. I looked back through a bunch of issues that had been marked as duplicates at some point or another and 80% looked to be the right threshold (if not higher!).

mondeja (approved):
Okay, let's try with 80 👍🏼

PR #10884, opened 2024-05-03
Remove license field from third-party extensions

LitoMore:
The field has misleading information. It shows the license of the source code. But some extensions are published to some specific platforms such as Stream Deck, Miro, and Webflow. Users should follow the platform terms rather than source code license.

What do you think?


adamrusted:
I would suggest we only remove the license fields for those closed-source platforms. Where code owners make a license available I think there's still merit to us highlighting it.

LitoMore:
Or split the extensions to two sections?

One for `useful ports`, another one for `third-party extensions`. Like this below:

```markdown
### Useful Ports

- Angular module
- Laravel package
- Java library
- ...

### Third-Party Extensions

- Drawio library
- Figma plugin
- Framer component
- Miro app
- Stream Deck icon pack
- Webflow app
- ...
```

mondeja:
I think this column should not be deleted. I vote for creating badges that point to platform terms or leaving the cell empty if this is not appropriate for the extension.

LitoMore:
Closing.

Issue #7993, opened 2022-10-25
n8n


LudovicSterlin:
### Brand Name

n8n

### Website

https://n8n.io/

### Popularity Metric

The Similarweb rank is 109,467 

https://www.similarweb.com/fr/website/n8n.io/#overview

### Official Resources for Icon and Color

svg https://runacap.com/wp-content/uploads/2020/03/n8n-logo.svg
png: https://github.com/n8n-io/n8n/blob/master/assets/n8n-logo.png
color: 
- logo #EA4B71 from header website
- text #101330 from header website

### Additional Comments

The logo and text are not the same color.

### Contributing

Yes

PR #10909, opened 2024-05-06
Add n8n icon

CodeShakingSheep:
image

**Issue:** closes https://github.com/simple-icons/simple-icons/issues/7993

**Popularity metric:**

The Similarwebrank is 60,987. See https://www.similarweb.com/fr/website/n8n.io

### Checklist

### Description

I forgot where I downloaded the original svg icon (had it on my computer for quite a while). I just combined the paths and applied the transformations to them. Then optimized it.


icons/n8n.svg
uncenter, line 1: image

I'm seeing a bunch of differences as illustrated by the screenshot above. Are you using the logo from their website's header? Also I'd like to note that you said " I just combined the paths and applied the transformations to them." - there are no paths to combine for this logo!

uncenter (changes requested):

icons/n8n.svg
uncenter: They also have a press kit from which you can source the icon: https://n8n.io/press/. `n8n-symbol-only.svg` would be the file to grab.

CodeShakingSheep: Hm, right. As I wrote I had this other n8n icon svg on my computer for a while (and don't know anymore from where I downloaded it). With the official svg icon you linked, I have a problem with removing `fill-rule="evenodd" clip-rule="evenodd"`, see screenshot. The problem is that 2 of the small circles get filled. I don't know how to solve that. Do you have any idea?

image

uncenter: Hmm I'm not sure. @PeterShaggyNoble I recall you helped me before with an icon that had `fill-rule="evenodd"`, how can we get rid of it here?

_data/simple-icons.json
```diff
@@ -10074,6 +10074,11 @@
             "hex": "48AC98",
             "source": "https://n26.com"
         },
+        {
+            "title": "n8n",
+            "hex": "EA4B71",
+            "source": "https://n8n.io"
```
adamrusted, line 10080:
```suggestion
            "source": "https://n8n.io/press"
```

icons/n8n.svg
adamrusted: I don't know the exact science with this - but I've always been able to 'split' the path in my editor, move the circle shapes up/down (changes per program) and then re-do the merging.

CodeShakingSheep: Ok, I managed to solve this by breaking apart the path (in Inkscape), then creating 4 little circles in the size of the 4 small paths and then took the difference between the big path and the 4 little circles respectively. Is it good now?

uncenter (approved):

adamrusted (approved):

PR #10932, opened 2024-05-10
Update Tencent QQ icon

LitoMore:
**Popularity metric:** Similarweb global rank - `qq.com` - `#63`

### Checklist

### Description

According to their guideline, we should use the `QQ Blue` one instead.

The color was picked from https://qq.design/brand/BrandDesign/Color.


uncenter:
(FYI you don't need to provide a popularity metric when updating an icon.)

_data/simple-icons.json
```diff
@@ -15298,8 +15298,9 @@
         },
         {
             "title": "Tencent QQ",
-            "hex": "EB1923",
-            "source": "https://en.wikipedia.org/wiki/File:Tencent_QQ.svg"
+            "hex": "1EBAFC",
+            "source": "https://qq.design/brand/BrandDesign/Logo",
```
uncenter, line 15302: Where are you getting a downloadable logo from? I'm only seeing PNG images.

uncenter (changes requested):

_data/simple-icons.json
LitoMore: I just modified the previous one.

uncenter: Ah. Well my point stands that the link listed isn't really a source then.

LitoMore: OK. Then let's use the `https://en.wikipedia.org/wiki/File:Tencent_QQ.svg` instead.

Updated.

adamrusted (approved):

PR #11051, opened 2024-05-28
Add `permissions` to Remove Labels workflow

mondeja:
Changes the event from `pull_request_target` to `pull_request` also. Currently the workflow is using an app token, so it's not needed to run in the base side.


LitoMore (approved):

mondeja:
Seems that `pull_request` will not work on forks because `app-id` environment variable is not available there ([see this run](https://github.com/simple-icons/simple-icons/actions/runs/9269460441/job/25500303405)), reverted at https://github.com/simple-icons/simple-icons/commit/01dd04f14aadd1323a94e9f4080e45e7f2d61bbb.

@LitoMore why the app-token action recommends storing it in a environment variable? It wouldn't be better in a secret?

PR #11053, opened 2024-05-28
Add SVGLint rule to check invalid characters after final closepath

mondeja:
In v11.15.0 we introduce the FreeCAD icon (#8879), which has one error after the final command. There is a `55` number after the final closepath (`z`), that is not a valid SVG path command. The problem is reproduced by some SVG path parsers, [like the Blender one](https://github.com/mondeja/simple-icons-blender/actions/runs/9261590293/job/25477207504#step:11:60).

This error is not being caught by our SVG linter, so added another rule to catch it. The reporter name "final-closepath" is enough open to specify in the same rule another logic to check that a path doesn't need to end with an explicit closepath, which is an optimization rule that I've in mind from a long time ago but it would only reduce ~200 characters in the whole library, so it's not worth for now.


adamrusted:
@mondeja - does the path not need the `z` at the end?

service-paradis:
Thanks @mondeja! It fixed [the issue I had with this icon using Inkscape](https://github.com/simple-icons/simple-icons/pull/8879#issuecomment-2069233521)!

Issue #6094, opened 2021-07-10
Organic Maps


victorbnl:
**Brand Name:** Organic Maps
**Website:** https://organicmaps.app
**Alexa rank:** 1,765,904
  

**Official resources for icon and color:** https://github.com/organicmaps/organicmaps.github.io/tree/master/static (see logo-on-*.svg)

service-paradis:
Unfortunately, this is out of scope currently as we required a rank below 500k (or at least 5k stars on Github). Feel free to open a new request once it reach the requirements!

CodeShakingSheep:
**Website**
https://organicmaps.app/

**Official resources for icon and color**
svg: https://organicmaps.app/logos/green-on-transparent.svg
color: `#006C35`

**Popularity Metric**
Organic Maps now has ~8.7k stars on GitHub which puts it into scope. See https://github.com/organicmaps/organicmaps

Can this issue therefore be reopened please @service-paradis  ? I would like to work on this icon.

service-paradis:
LGTM now @CodeShakingSheep! Reopening this one!

PR #11057, opened 2024-05-28
Add Organic Maps icon

CodeShakingSheep:
image

**Issue:** closes https://github.com/simple-icons/simple-icons/issues/6094

**Popularity metric:**

Organic Maps now has ~8.7k stars on GitHub which puts it into scope. See https://github.com/organicmaps/organicmaps

### Checklist

### Description

Just resized and optimized it.


adamrusted (changes requested):
Couple of issues on this one - would you mind having a look please @CodeShakingSheep?

CodeShakingSheep:
> Couple of issues on this one - would you mind having a look please @CodeShakingSheep?

Ah yes, just updated the source. Any other issues @adamrusted ?

icons/organicmaps.svg
adamrusted, line 1: Sorry, yes @CodeShakingSheep - there's a difference between your PR (black) and the source icon (red) below. Looks like it's just been broken in optimization - might be worth another try?
image

CodeShakingSheep: Strange, I don't know how the difference was created. The source svg only has 1 path. So, the only things I did was removing the <g> element and resizing it. 

Anyway, I just edited the source svg from scratch in Inkscape but when using an optimization precision of 4, I end up with the exact same path. So, I increased the precision to 5 although not sure if that solves the difference. Is it better now?

adamrusted (changes requested):
Still seeing exactly the same difference when using [this image](https://organicmaps.app/logos/green-on-transparent.svg) from the site header as source. Could you try re-generating please @CodeShakingSheep? 🙏
image

CodeShakingSheep:
> Still seeing exactly the same difference when using [this image](https://organicmaps.app/logos/green-on-transparent.svg) from the site header as source. Could you try re-generating please @CodeShakingSheep? 🙏

Thanks for reviewing @adamrusted . I used exactly that image from the website's header. I think there's something odd with the your downloaded svg from this PR. You wrote that black is my PR, right? However, when I zoom into my svg in Inkscape the edge is sharp (not rounded as in your screenshot). See here:
image

Also, when you zoom into the svg on GitHub you can see that it's sharp and not rounded. So, I assume that somehow the svg from this PR was altered while you downloaded it? It's really weird.

adamrusted (approved):
You were right @CodeShakingSheep - viewing the source file in another software fixed the issue, and it's a perfect match! Thanks for you patience on this one!

Issue #11104, opened 2024-06-03
Request: Astra


CodeShakingSheep:
### Brand Name

Astra

### Website

https://wpastra.com/

### Popularity Metric

The Similarweb rank is 70,654. See https://www.similarweb.com/website/wpastra.com

### Official Resources for Icon and Color

- SVG: https://cdn.shortpixel.ai/spai/ret_img/wpastra.com/wp-content/uploads/2019/05/astra-logo.svg
- Color: `#FFC300`

### Additional Comments

Edit: Instead of using the website header's SVG, I think the monochrome icon from this page should be used as base: https://wpastra.com/brand-assets/ .

Edit 2: The Brand assets only provide the icon as png. So, the website header's svg needs be used instead.

CodeShakingSheep:
> This issue is potentially a duplicate of one of the following issues:
> 
> * [Request: Astral #11096](https://github.com/simple-icons/simple-icons/issues/11096) (91.67%)

No, this is not a duplicate :smile:

uncenter:
We are well aware, hence the *potentially*! I'll remove that label.

PR #11118, opened 2024-06-06
Add Astra icon

CodeShakingSheep:
image

**Issue:** closes https://github.com/simple-icons/simple-icons/issues/11104

**Popularity metric:**

The Similarweb rank is 70,654. See https://www.similarweb.com/website/wpastra.com

### Checklist

### Description

Downloaded icon from website's header, removed wordmark, resized, unified paths, optimized.


_data/simple-icons.json
```diff
@@ -1587,6 +1587,11 @@
             "hex": "00665E",
             "source": "https://www.astonmartin.com"
         },
+        {
+            "title": "Astra",
+            "hex": "FFC300",
+            "source": "https://cdn.shortpixel.ai/spai/ret_img/wpastra.com/wp-content/uploads/2019/05/astra-logo.svg"
```
uncenter, line 1593: See https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#source-guidelines, this is incorrect.

uncenter (changes requested):

_data/simple-icons.json
CodeShakingSheep: Yeah right. Just updated the source.

_data/simple-icons.json
```diff
@@ -1587,6 +1587,11 @@
             "hex": "00665E",
             "source": "https://www.astonmartin.com"
         },
+        {
+            "title": "Astra",
+            "hex": "FFC300",
+            "source": "https://wpastra.com"
+        },
```
uncenter, line 1594: This should actually probably be sourced from https://wpastra.com/brand-assets/. They have logos there, and the first listed brand color seems more accurate than the yellow: `#5c2ede`.

icons/astra.svg
uncenter, line 1: Logo looks like a perfect match compared to the header though I'd like to make sure it matches once we update the source to their branding page.

uncenter (changes requested):

_data/simple-icons.json
uncenter: Screenshot 2024-06-07 at 10 03 56 (Finder)

You sure?

CodeShakingSheep: Ah yes, I just downloaded the icons zip and there were only svgs. 
image


You're right about the logo zip though. There are svg files present. I just updated color and source.

icons/astra.svg
CodeShakingSheep: Ok, if I need to redo the icon based on the svg from the branding page, let me know.

_data/simple-icons.json
uncenter: Hmm well actually none of the logos here match the outline treatment you've done here. Maybe change the source back if we want to stick with it this way... or we could change to the filled in treatment? Up to you.

CodeShakingSheep: `astra_logo03.svg` from the logo zip file matches the outline version of the logo.

image

So, I think we can leave the source and icon as they are? What do you think @uncenter?

uncenter: From what I saw the outline was significantly thinner then your outline? I could have been wrong though I didn't fully check.

CodeShakingSheep: Really? I didn't fully check it either. Anyway, I did a version without the outline. This is how it looks like now.
image

adamrusted (approved):
Spot on, thanks @CodeShakingSheep!

PR #11157, opened 2024-06-09
Fix SVGO `sortAttrs` configuration

LitoMore:
The `xmlns` in `order` field won't work. We should use the `xmlnsOrder: "alphabetical"` option instead.

Docmentation: https://svgo.dev/docs/plugins/sortAttrs/


mondeja (commented):
Is this documented? I can't see the `'alphabetical'` value handling in [sortAttrs plugin source code](https://github.com/svg/svgo/blob/main/plugins/sortAttrs.js)... 🤔

mondeja:
I think that it is a bug in SVGO. If `"xmlns"` is included in the `order` property, it should take preference over `xmlnsOrder`.

LitoMore:
> I can't see the 'alphabetical' value handling in [sortAttrs plugin source code](https://github.com/svg/svgo/blob/main/plugins/sortAttrs.js?rgh-link-date=2024-06-09T15%3A38%3A01Z)... 🤔

It's in their type definition.

> I think that it is a bug in SVGO. If "xmlns" is included in the order property, it should take preference over xmlnsOrder.

I'm not sure if it's a bug or a feature, but the `xmlns` is unnecessary so far. Also, their code hasn't changed for years.

mondeja (approved):

Issue #8399, opened 2023-03-03
Request: phpBB


GwynethLlewelyn:
### Brand Name

phpBB forum software

### Website

https://www.phpbb.com/

### Popularity Metric

Similarweb global rank: 130,708 - https://www.similarweb.com/website/phpbb.com/#overview
Netcraft site rank: 3,599 - https://sitereport.netcraft.com/?url=https%3A%2F%2Fwww.phpbb.com
GitHub stars: 1,600+ - https://github.com/phpbb/phpbb

phpBB is a free open-source bulletin board written in PHP.

### Official Resources for Icon and Color

Everything you need for the logo/icon, including official colours, and the fonts used in the logo:

https://www.phpbb.com/about/logos/

### Additional Comments

phpBB is a well-known, popular forum software. It's free and open source, licensed under the GNU General Public License v2
The icon/logo and the name itself, however, are registered trademarks, to be used — without modification — to identify the forum's brand.
As for the rest of your questions related to brand guidelines, I believe that the link above will explain pretty much everything, and allow you to download the relevant SVG files

Thank you so much for your work 😀 

### Contributing

Yes

PR #11163, opened 2024-06-09
Add phpBB icon

CodeShakingSheep:
image

**Issue:** closes https://github.com/simple-icons/simple-icons/issues/8399

**Popularity metric:**

Similarweb rank is 111,115. See https://www.similarweb.com/website/phpbb.com

### Checklist

### Description

Resized and optimized.


icons/phpbb.svg
PratyushJayachandran, line 1: The R symbol is not aligning:
image

PratyushJayachandran (dismissed):

icons/phpbb.svg
CodeShakingSheep: Thanks for reviewing @PratyushJayachandran :pray: 
According to https://www.phpbb.com/about/logos/ (scroll down until the end) I used the SVG file for the small version of the logo (it has exactly 24px height originally) which has the R symbol placed differently.

LitoMore: According to our contributing guideline. The `®` symbol can be removed.

CodeShakingSheep: Can you elaborate why @LitoMore?

When I read through the guidelines, I think these 2 would apply here:

> - If the brand itself includes the symbol with all uses of the logo, even at small sizes, it must be included.
> - If there is ambiguity about the conditions under which the symbol is required, it must be included if it is a registered trademark symbol (®) but not if is a trademark symbol (™).

service-paradis: @CodeShakingSheep You're right. Since they are using the symbol themselves everywhere, we need to keep it.

PR #11237, opened 2024-06-16
Add GSMA icon

PratyushJayachandran:
gsma

**Popularity metric:** Similarweb rank 120,779. See https://www.similarweb.com/website/gsma.com

### Checklist

### Description
Icon: Taken from website header https://www.gsma.com
Color: Taken from logo `#000000`


_data/simple-icons.json
```diff
@@ -6724,6 +6724,11 @@
                 ]
             }
         },
+        {
+            "title": "GSMA",
+            "hex": "000000",
```
adamrusted, line 6729: The company tends to use a red as their brand colour (see all the app store icons!). We might be best placed to use one of those instead of black.

adamrusted (changes requested):
Agree on the inclusion of the ™️symbol - they seem to include it pretty universally. Only query is around the colour chosen.

PeterShaggyNoble (approved):
SVG looks good to me 👍🏻 

I agree with @adamrusted on the choice of colour, though, so I'll get that changed so this can be merged.

Issue #11289, opened 2024-06-20
Request: Spotube


CodeShakingSheep:
### Brand Name

Spotube

### Website

https://spotube.krtirtho.dev/

### Popularity Metric

It has 26.1k stars on GitHub. See https://github.com/KRTirtho/spotube

### Official Resources for Icon and Color

- SVG: https://github.com/KRTirtho/spotube/blob/master/assets/spotube-logo.svg
- Color: `#3B82F6`

uncenter:
No need to create an issue if you know it to be within our popularity requirements and you are already planning to submit a PR :)

CodeShakingSheep:
> No need to create an issue if you know it to be within our popularity requirements and you are already planning to submit a PR :)

Thanks for the hint. Makes sense.

PR #11290, opened 2024-06-20
Add Spotube icon

CodeShakingSheep:
image

**Issue:** closes https://github.com/simple-icons/simple-icons/issues/11289

**Popularity metric:**

It has 26.1k stars on GitHub. See https://github.com/KRTirtho/spotube

### Checklist

### Description

Took the color from the primary button on the website.


uncenter:
Before we look at the icon itself we need to make sure adding Spotube doesn't violate our guidelines. I fear that this may fall under the piracy clause of "Illegal services (e.g. piracy, malware, threatening material, spam, etc.)", a [forbidden category under our guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon), if it "bypasses" Spotify Premium (I'm not totally clear on how the "audio sources" works, and if the YouTube audio source is allowed either). Is there any element of the project that could be called piracy or break Spotify/YouTube's TOS?

uncenter:
I think we've reached the consensus that Spotube is fine for adding to our collection! You may ignore my previous comment.

adamrusted (changes requested):
I'm seeing quite a few differences between your path (black) and the source file, edited down (red). Could you please take another look @CodeShakingSheep?

image

PR #11364, opened 2024-06-30
Add more criteria for popularity to CONTRIBUTING.md

MichaelHinrichs:
Everything in the template is N/A.


CONTRIBUTING.md
```diff
@@ -100,7 +100,11 @@ To be considered popular enough to be within our scope, a brand must be in exist
     - [npm](https://www.npmjs.com): 100k weekly downloads,
     - [JSDelivr](https://www.jsdelivr.com): 1m daily or 35m monthly requests,
     - [crates.io](https://crates.io): 100k weekly downloads,
-    - [PyPi Stats](https://pypistats.org): 100k weekly downloads, or,
+    - [PyPi Stats](https://pypistats.org): 100k weekly downloads,
+    - [pkg.go.dev](https://pkg.go.dev): 100k importers,
+    - [Homebrew Formulae](https://formulae.brew.sh): 100k weekly downloads,
+    - [Arch linux User Repository](https://aur.archlinux.org) popularity of 7.00,
```
uncenter:
```suggestion
    - [Arch User Repository](https://aur.archlinux.org) popularity of 7.00,
```

uncenter (changes requested):

CONTRIBUTING.md
```diff
@@ -100,7 +100,11 @@ To be considered popular enough to be within our scope, a brand must be in exist
     - [npm](https://www.npmjs.com): 100k weekly downloads,
     - [JSDelivr](https://www.jsdelivr.com): 1m daily or 35m monthly requests,
     - [crates.io](https://crates.io): 100k weekly downloads,
-    - [PyPi Stats](https://pypistats.org): 100k weekly downloads, or,
+    - [PyPi Stats](https://pypistats.org): 100k weekly downloads,
+    - [pkg.go.dev](https://pkg.go.dev): 100k importers,
+    - [Homebrew Formulae](https://formulae.brew.sh): 100k weekly downloads,
```
uncenter, line 105: These need to be updated to be a) realistic for the package manager and b) actually tracked by the package manager! Homebrew doesn't have a weekly downloads count, and 100k is very high for a weekly count anyway. We can and should reference the statistics from Homebrew for packages/apps/products already in our collection to determine the right number.

| Package | Last 30 days | Last 90 days | Last 365 days |
| --- | --- | --- | --- |
| `deno` | 13k | 40k | 145k |
| `go` | 87k | 261k | 1m |
| `1password` (cask) | 3k | 9k | 39k |
| `firefox` (cask) | 11k | 35k | 140k |
| `inkscape` (cask) | 2k | 6k | 26k |
| `obsidian` (cask) | 5k | 15k | 58k |
| `raycast` (cask) | 7k | 21k | 70k |
| `spotify` (cask) | 6k | 21k | 83k |

uncenter (changes requested):

CONTRIBUTING.md
MichaelHinrichs: Oh, dang it. I just defaulted to the same minimum as everything else.

adamrusted (commented):
I've mentioned this elsewhere, but the keyword for me is that we're a repository of  _popular_ brands. While some of these numbers seem quite high, a lot are across the lifetime of a product and do not reflect current popularity. I would argue if we can't see weekly / monthly stats as @uncenter mentioned - then we can't use that metric.

adamrusted:
> Ah okay so maybe we take out the 365 days metric from Homebrew and just keep the 30 days one?

Happy for us to have an 'annual' metric, but I was more making the point of the metrics needing to be time-sensitive. No point us looking at a download total if we can't validate its trajectory.

LitoMore:
These popular metrics are unnecessary in the contributing document. We can consider the popularity of those items based on actual situations. We just need to explain the reason in the relevant issue/PR.

For example this one: https://npmjs.com/is-yarn-global. This package has millions of download counts per week. But I can't entirely agree it is popular stuff.

adamrusted:
Hard disagree @LitoMore .
We've had stats-led metrics since before either of us joined the project, and have only refined them since. We should always be open to alternative official metrics being added if they're relevant.

In your example, looks as though you've not given your project a logo - so we wouldn't be adding it anyway 🤷‍♂️

LitoMore:
@adamrusted I was talking to you about a library that can have a lot of downloads but is not really popular. You steered the conversation to whether or not it has a logo. That's different from the point and not what I want to discuss.

Going back to what I said, if we only used downloads as a metric, it would greatly lower our threshold, and many simple, basic libraries would qualify for a logo.

We may need to add other dimensions to ensure our quality.

adamrusted:
@LitoMore - I think you misunderstood my point. The lack of a logo was just to highlight another reason why that particular package wouldn't be added. I kind of think we _should_ allow any popular package to be included, whether that's because of general use of the package, influencer marketing, etc **PROVIDING** it meets the metrics we've already set out. Again, they're quite robust and have been worked on for the duration of time I've known of the package.

The point of this PR is to add alternative metrics that can be used in specific instances we don't already account for, which I welcome as a way for us to include more packages, and provide alternative metrics for packages we already have in the collection. It'll only make our lives as maintainers easier to have more metrics to compare against.

If we see an influx of what we deem 'not popular enough' packages - then yes we can refine the numbers, but we have to start with something.

LitoMore:
@adamrusted Thank you for your explanation. It makes sense to me. 👍

adamrusted (approved):

CONTRIBUTING.md
```diff
@@ -100,7 +100,9 @@ To be considered popular enough to be within our scope, a brand must be in exist
     - [npm](https://www.npmjs.com): 100k weekly downloads,
     - [jsDelivr](https://www.jsdelivr.com): 1m daily or 35m monthly requests,
     - [crates.io](https://crates.io): 100k weekly downloads,
-    - [PyPi Stats](https://pypistats.org): 100k weekly downloads, or,
+    - [PyPi Stats](https://pypistats.org): 100k weekly downloads,
+    - [Homebrew Formulae](https://formulae.brew.sh): 5k installs in the last 30 days or 50k installs in the last year,
+    - [Arch User Repository](https://aur.archlinux.org): popularity of 7.00 or,
```
uncenter, line 105: Can you give some more context to this? What does a popularity of 7.00 mean? What icons do we already have for projects/brands that meet this level of popularity?

mondeja: Quoting the [ArchLinux forum](https://bbs.archlinux.org/viewtopic.php?id=283934):

> Popularity is calculated as the sum of all votes with each vote being weighted with a factor of 0.98 per day since its creation.
>
> Popularity_today = Popularity_yesterday * 0.98 + New_votes_from_today

mondeja (approved):
LGTM

uncenter (approved):

PR #11542, opened 2024-07-27
Add LATAM Airlines logo

CS-DaviMagalhaes:
preview

**Issue:** none

**Popularity metric:**
Global rank 10,427

### Checklist

### Description
Picked the color from the source.


LitoMore (changes requested):
Thank you for submitting this PR!

I can see some differences compared with the resource on Wikipedia:



Can you check this?

BTW, I noticed the homepage https://latamairlines.com has an SVG resource in the header. We could use that one intead.

adamrusted:
As per [this page](https://www.latamairlines.com/gb/en/legal/terms-of-use) we would need permission to include the brand. If you're still wanting to contribute this (totally appreciate it's been _a while_!) then would you mind reaching out to them @CS-DaviMagalhaes?

PeterShaggyNoble:
Closing due to inactivity and lack of permission. If someone can make contact with them and get permission then we can revisit this one.

PR #11546, opened 2024-07-28
Remove part about going to https://similarweb.com/website/google.com to bypass login

MichaelHinrichs:
Not anymore!

Template is N/A.


uncenter:
Still works for me, only if you visit it multiple times in a short time span you then have to use incognito to make it work.

MichaelHinrichs:
Nevermind then.

PR #11716, opened 2024-08-31
Use `in` keyword instead of Object.hasOwn in ourlint script

uncenter:
Using `'x' in y` is supported by TypeScript, which allows us to get rid of the `TODO` comment & `@ts-ignore` for the previous usage of `Object.hasOwn`.


LitoMore:
They are different. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#direct_vs._inherited_properties.

This is fine for some comments in our code. We can remove them once TypeScript implements this.

uncenter:
> They are different. See [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#direct_vs._inherited_properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#direct_vs._inherited_properties).
> 
> This is fine for some comments in our code. We can remove them once TypeScript implements this.

Sure but `license` isn't a property on anything in native JavaScript... there are no conflicts and no apparent downsides to this option.

mondeja:
I'm more inclined to stick to complete accuracy in these types of situations. Although is very unlikely that Node.js implement a `url` in `Object`'s prototypes, this change will possibly introduce a silent bug if we edit the `url` property name in the future. Additionally, can set a precedent in the source code to how we check properties in objects, probably being spread.

As @LitoMore said, is fine to have some comments in our code.

uncenter:
> I'm more inclined to stick to complete accuracy in these types of situations. Although is very unlikely that Node.js implement a `url` in `Object`'s prototypes, this change will possibly introduce a silent bug if we edit the `url` property name in the future. Additionally, can set a precedent in the source code to how we check properties in objects, probably being spread.

Sure.

> As @LitoMore said, is fine to have some comments in our code.

It is confusing when users of the lint script get a warning every time. Seemed like a simple fix.

mondeja:
> It is confusing when users of the lint script get a warning every time. 

I'm not totally sure but we can probably disable the TODO comments warnings in the XO configuration. Open a PR if you find them too annoying.

PR #11720, opened 2024-09-01
Update Create React App icon

LitoMore:
There is no visual difference.

### Checklist

### Description

We don't need that much precision. I optimized this icon with `npx svgo icons/createreactapp.svg` to reduce its file size.


uncenter (approved):
Looks fine.

PR #11785, opened 2024-09-17
Avoid raw asset URLs

LitoMore:
The raw image path file does not make it easy for maintainers to view the relevant history of the source file.


uncenter (approved):

uncenter:
Should wait for the second review but this is a nice change!

mondeja (commented):
I think that we can use `if then else` JSONSchema constructs for this, just like the github one, see: https://github.com/simple-icons/simple-icons/blob/55b5a43a032ada6e2f0b78dc65c5bb16561517f0/.jsonschema.json#L682-L689

mondeja:
I'm not totally against this approach but I think that the JSONSchema spec would be more reusable and would save us code to maintain.

LitoMore:
The error message from JSONSchema is not human-friendly. It shows a super long regex pattern to our contributors. We've had a few cases where they didn't understand the error message.

mondeja (approved):
Okay, let's go with this 👍🏼 

> The error message from JSONSchema is not human-friendly. It shows a super long regex pattern to our contributors. We've had a few cases where they didn't understand the error message.

Probably it would be worth to migrate the Github URLs checking of JSONSchema to our linting script. Other solution would be to use [this trick](https://www.npmjs.com/package/jsonschema#localizing-error-messages) to make error messages more human friendly, thus maintaining all the possible validation at JSONSchema side.

PR #11876, opened 2024-09-27
Re-enable pull requests labeling

mondeja:
Just add the labels but don't categorize pull requests in projects. After this we can reenable the `add-labels-priority` workflow.


LitoMore (approved):
Let's go.

PR #11879, opened 2024-09-27
Add SPDX license check to ourlint

### LINKED ISSUE #11746: Manage SPDX license list with package manager
Our current SPDX license list is hardcoded in the `.jsonschema.json`. However, some records have been left behind. To manage the list, we can use the npm package [spdx-license-ids](https://github.com/jslicense/spdx-license-ids).

We might need to add a new script `bump-spdx-license-list` to programmatically update the SPDX list in `.jsonschema.json`. This might be placed in `scripts/build`.

We also need a lint to check that the SPDX is updated.

Could we consider moving SPDX away from `.jsonschema.json` and just using Node.js scripts to manage and lint the list directly? For example:

```js
// scripts/lint/ourlint.js
import spdxLicenseIds from 'spdx-license-ids/index.json' with {type: 'json'};

const allGood = iconsData.icons.every(({license}) =>
  !license || license.url ? true : spdxLicenseIds.includes(license.type),
);
console.log(allGood);
```

Let's not discuss the experimental feature `with {type: 'json'}`, we can change it to `fsPromises.readFile()` at will.

### References

- https://spdx.org/licenses/
- https://github.com/spdx/license-list-data
- https://github.com/jslicense/spdx-license-ids
- https://spdx.dev/learn/handling-license-info/#where

LitoMore:
Resolves #11746


scripts/lint/ourlint.js
```diff
@@ -201,6 +206,37 @@ const TESTS = {
       return `Invalid URLs:\n\n${invalidUrls.join('\n\n')}`;
     }
   },
+
+  /* Check if all licenses are valid SPDX identifiers */
+  async checkLicense(data) {
+    const spdxLicenseIds = JSON.parse(
+      await fs.readFile(
+        path.join(
+          import.meta.dirname,
```
mondeja, line 215: Scripts are currently supporting Node.js v18 onwards

https://github.com/simple-icons/simple-icons/blob/11c864ea28d0a4cfe7f5b7d89eea1ce909ffca54/.xo-config.json#L44-L50

and `import.meta.dirname` is only available since v20 (see https://nodejs.org/api/esm.html#importmetadirname).

That's the reason we're using the `path.dirname(fileURLToPath(import.meta.url))` trick in *sdk.mjs*, provided to the whole project with `getDirnameFromImportMeta()` SDK function. Please, use that instead.

scripts/lint/ourlint.js
```diff
@@ -201,6 +206,37 @@ const TESTS = {
       return `Invalid URLs:\n\n${invalidUrls.join('\n\n')}`;
     }
   },
+
+  /* Check if all licenses are valid SPDX identifiers */
+  async checkLicense(data) {
```
mondeja, line 211: As this function is the first async function in the ourlint script, you should also change the type of `TESTS` variable. The return of the functions must be `Promise<string | undefined> | string | undefined` instead of `string | undefined` now.

Note that the consumer `Promise.all` accepts either promises or not.

mondeja (changes requested):

scripts/lint/ourlint.js
```diff
@@ -199,6 +205,37 @@ const TESTS = {
...
+    const spdxLicenseIds = JSON.parse(
```
mondeja, line 211: Shouldn't this be a `Set`?

scripts/lint/ourlint.js
```diff
@@ -199,6 +205,37 @@ const TESTS = {
...
+    );
+    const badLicenses = [];
+    for (const {title, slug, license} of data.icons) {
+      if (
+        license &&
+        license.type !== 'custom' &&
+        !spdxLicenseIds.includes(license.type)
+      ) {
+        badLicenses.push(
+          `${title} (${slug ?? titleToSlug(title)}) has an bad license: ${license.type}`,
```
mondeja, line 230: I think that the message is not very descriptive. Would expect something like `{title} does not have a valid SPDX license. See the valid licenses at {spdxLicensesUrl}.`

scripts/lint/ourlint.js
LitoMore: Sorry, I didn't get you. What do you mean Set? It reads the list from the JSON file.

scripts/lint/ourlint.js
LitoMore: Good point.

scripts/lint/ourlint.js
mondeja: To convert the list to a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) because SPDX license identifiers are unique values. Mainly because the intent is clear and its faster, especially for large sets:

> The [has](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has) method checks if a value is in the set, using an approach that is, on average, quicker than testing most of the elements that have previously been added to the set. In particular, it is, on average, faster than the [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method when an array has a length equal to a set's size.

LitoMore: Sounds good. Done.

mondeja (approved):
Great. Thank you!

PR #11885, opened 2024-09-28
Update New Japan Pro-Wrestling icon

LitoMore:
Optimized the file size from 31 KB to 23 KB.


LitoMore:
They are different shapes. The icon you provide is not even a monochrome version.

PratyushJayachandran (approved):

PR #11890, opened 2024-09-29
Add ElevenLabs icon

z-alsamarai:
elevenlabs

**Issue:** closes NA

**Popularity metric:**
* Similarweb Global Rank: [3,665](https://www.similarweb.com/website/elevenlabs.io/)

### Checklist

### Description
  - Picked Hex value 000000, because ElevenLabs uses a classic black and white color scheme, and it's the main color of their logo.
  - Used boxy-svg to resize to the proper viewbox, and used svgomg for optimization.


uncenter:
Please see the linting errors, your SVG isn't properly formatted. Additionally, we should use https://elevenlabs.io/icon.svg as the source as it is already an SVG and we don't need to vectorize.

z-alsamarai:
@uncenter My apologies, I pushed my commit before I saw your comment.
The SVG should be properly formatted now, and I tried using the SVG from the [link](https://elevenlabs.io/icon.svg) provided, it seems to yield the same result after removing all extraneous elements such as the rect in their version and scaling the viewbox down to 24x24.

uncenter:
>  it seems to yield the same result after removing all extraneous elements such as the rect in their version and scaling the viewbox down to 24x24.

What was the source you used before then? If you traced the PNG I find it hard to believe the SVG version lines up exactly 😅

LitoMore (changes requested):
I can see some differences.



Can you check this?

z-alsamarai:
@LitoMore You are absolutely right, now I see it! This is my first time working with icons, and SVGs this way. 

Here's what I changed:

- Downloaded https://elevenlabs.io/icon.svg
- Modified the following to satisfy the guidelines: 

1. Removed the rectangle
2. Merged the two paths into one
3. Remove the grouping so the svg is just a single path
4. Made the viewbox 24x24
5. Centered the new path in the viewbox so that the top and bottom sides touch the border.
6. Optimized the new icon with svgomg

For reference, here's the original Icon elements:
image

I hope this solves what you're looking for! If not I'm happy to make the necessary changes to make it work.

_data/simple-icons.json
```diff
@@ -4842,6 +4842,11 @@
             "source": "https://elementor.com/logos/",
             "guidelines": "https://elementor.com/logos/"
         },
+        {
+            "title": "ElevenLabs",
+            "hex": "000000",
+            "source": "https://elevenlabs.io/icon.svg"
```
uncenter, line 4848:
```suggestion
            "source": "https://elevenlabs.io/brand",
            "guidelines": "https://elevenlabs.io/brand"
```

icons/elevenlabs.svg
```diff
@@ -0,0 +1 @@
+<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>ElevenLabs</title><path d="M19.314 24H14.4V0h4.914v24ZM9.6 24H4.686V0H9.6v24Z"/></svg>
```
uncenter, line 1:
```suggestion
<svg role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><title>ElevenLabs</title><path d="M4.6035 0v24h4.9317V0zm9.8613 0v24h4.9317V0z"/></svg>
```

The icon they provide on the brand page is apparently slightly different then the one from the previous source.

uncenter (changes requested):

uncenter (approved):

PR #11891, opened 2024-09-29
Update Unilever icon

LitoMore:
This optimized the file size from 25.652 KiB to 20.747 KiB.


PratyushJayachandran:
The logo is now hosted in a different space now: [https://www.unilever.com/news/press-and-media/media-assets/](https://www.unilever.com/news/press-and-media/media-assets/)
Can we update that too?

PratyushJayachandran (changes requested):
When comparing with the logo on their media assets page, I see some imperfections. Attaching the comparison below.
unilever-comparison
image
I used the white version from here: [Media Assets](https://www.unilever.com/news/press-and-media/media-assets/)

LitoMore:
Update. It only has a slight difference. You may have some misconfiguration on your image editor.

PratyushJayachandran:
> Updated. It only has a slight difference. You may have some misconfiguration on your image editor.

unilever-comparison
It is better than before but some of the circles in the artwork aren't matching. Screenshot of the outline view attached:
image

LitoMore:
> You may have some misconfiguration on your image editor.

Please use Inkscape or Figma to review this.

mondeja:
I'm using Inkscape and still seeing some significative differences.

icons/unilever.svg
adamrusted, line 1: I'm also seeing issues with the circles in the right-hand side of the logo:
image

The eye of the fish at the bottom:
image

And some slight differences to the shapes in the top-left:
image

This is in Affinity Designer, with no snap-to-pixel or precision rules applied.

LitoMore: Thanks. I just reprocessed the icon. I'm not sure what happened to the previous file.

Here is the difference preview:

adamrusted: I would prefer to have a _slightly_ larger file for accuracy sake (I would always prefer we go that route, for the record). Comparing your version (black) with the source file (white) I'm seeing _more_ differences now than I was with the last version:

image
image

LitoMore:
Thanks for reviewing. The precision 3 looks not good on this icon. So let's keep the precision 4.

PR #11902, opened 2024-09-30
Use `Number.parseInt()` for hexadecimal/decimal conversions

uncenter:
Is equivalent to our current function. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt. We use it as `Number.parseInt(hexString, 16)` where 16 is the radix/base of the string (hexadecimal).


LitoMore:
It has a slight difference from the previous implementation. But this should be fine.

```js
hexadecimalToDecimal('p')
//=> -1

Number.parseInt('hexadecimalToDecimal', 16)
//=> NaN
```

mondeja (approved):
Good catch! Thanks 👍🏼 

Pretty sure that I was reinveting the wheel in #6522, probably due to the confusion about `Number()`, `parseInt()` and `Number.parseInt()` support then.

LitoMore (approved):

Issue #11741, opened 2024-09-09
Change indentation of `simple-icons.json` to tab


LitoMore:
As the number of icons increases, our files become larger and larger. We are currently using 4-space indentation. This can be optimized with tabs.

Here are file sizes in different formats:

|4-space|tab|minified|
|:-|:-|:-|
|563 KB|422 KB|347 KB|

We can also change the distribution file to minified JSON.

### TODO

- [x] Optimize distribution file to minified JSON (#11903)
- [ ] Change all space-indent to tab-indent (except YAML)

uncenter:
We might as well switch to tabs everywhere right? It's pretty much all benefits - especially in this case of file size being important - and it would be nice to have the codebase in sync. That being said I think you are right that we should definitely have the distributed data file minified.

LitoMore:
Agree. The tab is also the default indent style of XO.

The advantage of tab indent in code is that developers can modify its width at will.

mondeja:
I agree, but note that the change that would bring greater benefits in terms of file size by a long way would be to remove the `icons` property, exporting just the array.

uncenter:
That too! Would this be something we do somewhat soon, or wait until we can reduce the number of open PRs somewhat as to reduce conflicts?

LitoMore:
No rush. Removing indents from the `simple-icons.json` will break its readability.

We also need to consider how to design this part. We can use the minified file for the `npm pack` only. But this means we need to use another path for the distribution file.

The file size of the pack will only affect the speed when they install it, and there will be no performance change when accessing it through import.

Of course, we also have a few users who access `simple-icons.json` directly through CDN, but there are not many such users.

uncenter:
> Removing indents from the `simple-icons.json` will break its readability.

Right but we wouldn't be removing it from https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json, it would be a/the compiled/minified file.

LitoMore:
Here https://github.com/LitoMore/simple-icons-figma/blob/a5c2fae7ddcb68985c54d7ac7a0f3a9c75cca34f/source/utils.ts#L43-L49 

We are unsure if other users are accessing our `simple-icons.json` file this way.

So we'd better not change its location. My current thinking is to add a process to minify the file during the `prepublishOnly` stage and then change it back at the `postpublish`

PR #11903, opened 2024-09-30
Minify `_data/simple-icons.json` on publish

LitoMore:
Related to #11741

This added two functions, `minifyIconsData` and `reformatIconsData`, to the distributing process.

I kept the file location the same due to https://github.com/simple-icons/simple-icons/issues/11741#issuecomment-2384129213.

I also added a new environment variable `SI_ENV=distribution` to the `publish.yml` action as an actual distribution. This is useful for our developer running `npm run build` locally. It can avoid generating diffs in our local.

I didn't use `NODE_ENV=distribution` because other packages may use this variable to determine whether it's CI or production. We'd better not to overwrite it.

The `SI_ENV=distribution` is not in `verify.yml` because there is a `npm run lint` after the `npm run build`. If we add it, a linter error will occur.

### Update

I change its logic to the `scripts/release/minify-icons-data.js`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -46,6 +46,8 @@ jobs:
         run: node scripts/release/reformat-markdown.js "${{ steps.get-version.outputs.version }}"
       - name: Update SDK Typescript definitions
         run: node scripts/release/update-sdk-ts-defs.js
+      - name: Minify icons data file
+        run: node scripts/release/minify-icons-data.js
       - name: Build NodeJS package
         run: npm run build
       - name: Deploy to NPM
--- a/scripts/release/minify-icons-data.js
+++ b/scripts/release/minify-icons-data.js
@@ -1,0 +1,9 @@
+/**
+ * @file
+ * Minify _data/simple-icons.json file.
+ */
+import {getIconsData} from '../../sdk.mjs';
+import {writeIconsData} from '../utils.js';
+
+const icons = await getIconsData();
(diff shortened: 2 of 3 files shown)


mondeja:
> I also added a new environment variable `SI_ENV=distribution` to the `publish.yml` action as an actual distribution. This is useful for our developer running `npm run build` locally. It can avoid generating diffs in our local.

I don't see the point. If it would generate diffs seems like an error with the build process, right?

LitoMore:
> I don't see the point. If it would generate diffs seems like an error with the build process, right?

The only difference is that it is formatted with indentation. In `publish.yml`, we already have the `npm run lint` before the `npm run build`. It already checks all the stuff before building.

mondeja:
I'm not a fan of conditional build process nor adding build environments if they are not strictly necessary. Consider that we run a couple of scripts to prepare things before publishing:

https://github.com/simple-icons/simple-icons/blob/14ccea31a8a2d3c3937277a38d577b6ecfa00332/.github/workflows/publish.yml#L45-L48

Why don't add the minification there avoiding the reformat at all?

LitoMore:
> Why don't add the minification there avoiding the reformat?

On my... I almost forgot this part. Yes, we can do the process here.

LitoMore:
Done with new script `scripts/release/minify-icons-data.js`.

mondeja (approved):
Clean and simple!

Hope no one complains 🤞🏼

PR #11979, opened 2024-10-10
Update Deno icon

LitoMore:
### Checklist


_data/simple-icons.json
```diff
@@ -4203,7 +4203,7 @@
         {
             "title": "Deno",
             "hex": "000000",
```
uncenter, line 4205: Have we ever considered using `#70FFAF` as the hex here? They seem to use it as their accent/brand color in some places, though not necessarily in the logo itself.

CleanShot 2024-10-10 at 21 23 33

image

image

icons/deno.svg
uncenter, line 1: Looks good!

_data/simple-icons.json
LitoMore: We should not use that color on the icon. They didn't use this color on the logo.

uncenter: What do you mean the homepage theme color is subject to change... everything is subject to change? We are changing the "blank-and-white Deno" logo literally right now 😅

uncenter: And for the record [our guidelines say](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#color-guidelines) that the brand color isn't necessarily the color of the icon, just the accent color the brand uses in general.

LitoMore: They didn't use color on the logo. This is a color that has nothing to do with the icon, so we shouldn’t use it on the icon.

uncenter: Like I said, our guidelines don't say that the brand color must be from the icon. It looks pretty clear to me that Deno uses this green color as their "accent", so I don't see why not. We can always ask someone working at the company for confirmation too...

LitoMore: It's official is just black-and-white. They even added a white background to the icon to ensure that the icon is always black-and-white.

LitoMore:
Hi @lucacasonato, could you help review this icon change? I believe the icon color should be always black-and-white. What do you think?

We had a similar discussion before at https://github.com/denoland/dotland/pull/1946.

PratyushJayachandran (approved):

lucacasonato:
Either is acceptable as per our [logo guidelines](https://github.com/user-attachments/files/17339441/Logo.guidelines.pdf) (attached).

The preferred logo to use is the transparent background one. The one with an explicit background should only be used in exceptional cases. You can find all versions of the logo here: https://drive.google.com/drive/folders/14oL13lfBUZ6kAFkZy4GrE-IDVpr6HbGP

LitoMore:
@lucacasonato Thank you. Then I would choose to use a colored one in case the user needs this color for other purposes.

PR #11989, opened 2024-10-11
Add Bagisto

zcraber:
bagisto

**Popularity metric:** [14.8k stars](https://github.com/bagisto/bagisto)

### Checklist

### Description
Colour picked from the icon


mondeja (changes requested):
Would you mind to create a wider separation here to distingish it at smaller sizes?

image

We've used 0'5 cm in the past as a standard and worked very well.

PeterShaggyNoble (changes requested):
Thanks for making that change, @zcraber. However, I think the cut-out at the bottom should overlap equally with the two shapes in the original. Also, the cut-out around the handle should be updated to the same width.

PeterShaggyNoble:
Closing due to inactivity

PR #11999, opened 2024-10-12
Add openSUSE Leap icon

DerBraky:
image

**Issue:** n/a

**Popularity metric:**
- Similarweb Global Rank: [72,281](https://www.similarweb.com/website/opensuse.org) (Leap is one of openSUSE's core products/brands)

### Checklist

### Description

- Hex value `#FBC75D` from [website background](https://get.opensuse.org/leap)


PeterShaggyNoble (changes requested):
Could you try increasing the precision on this one, please, @DerBraky? We're seeing some hairline thin differences throughout, most noticeably in the outer top right of the diamond shape.

Also, there is an excess point in both top corners of the bottom shape that could be removed.

Issue #11901, opened 2024-09-30
Update various broken source/guidelines links


uncenter:
I ran a link checker (https://github.com/lycheeverse/lychee) on our data file, combed through to eliminate the false positives, and did a little investigation into each to see if there was an easy new link to use. I ended up with 161 links, many of which have little notes. I'll plan to fix the ones with notes in a large batch PR later, the rest will still need have sources/new links tracked down. The following list was generated from a little script I made, so my apologies if a link is incorrectly associated with the wrong icon!

- [ ] Géant: https://www.geantcasino.fr/
- [ ] CentOS: https://wiki.centos.org/ArtWork/Brand/Logo
- [ ] TUI: https://www.design.tui/brand/logos
- [ ] Groupon: https://about.groupon.com/press/
- [ ] SlideShare: https://www.slideshare.net/ss/creators/
- [ ] Ello: https://ello.co/
- [ ] Scribd: https://brand.scribd.com/d/oFZcsq7FVpSh/about-scribd#/media-press-kit/downloads
- [ ] SmugMug: https://www.smugmughelp.com/articles/409-smugmug-s-logo-and-usage
- [ ] Qi: https://www.wirelesspowerconsortium.com/knowledge-base/retail/qi-logo-guidelines-and-artwork.html
  - Should be https://www.wirelesspowerconsortium.com/knowledge-base/for-retail-and-brands/qi-logo-guidelines-and-artwork/
- [ ] Equinix Metal: https://metal.equinix.com/
- [ ] SNCF: https://www.sncf.com/
- [ ] Fujitsu: https://www.fujitsu.com/global/about/brandmanagement/logo/
- [ ] containerd: https://cncf-branding.netlify.app/projects/vitess/
  - Now https://vitess.io/
- [ ] MEGA: https://mega.io/corporate
  - Now https://mega.io/media
- [ ] Ripple: https://brand.ripple.com/article/brand-policy
  - Logo (maybe) and color (definitely) need updating as well
- [ ] DeviantArt: https://help.deviantart.com/21
  - Can't find any version of a brand/media/logo page anymore
- [ ] Ford: https://secure.ford.com/brochures/
  - Can't find any version of a brand/media/logo page anymore
- [ ] Wolfram: https://company.wolfram.com/press-center/mathematica
  - There is https://company.wolfram.com/press-center/ but I'm not sure it has the same content. The link needs to be updated for all 3 Wolfram-related icons.
[...]

PR #12001, opened 2024-10-13
Use lychee instead of markdown-link-check on CI

mondeja:
Just use lychee instead of markdown-link-check on CI. See a comparsion between both tools: https://github.com/lycheeverse/lychee?tab=readme-ov-file#features

### DIFF
--- a/.github/markdown-link-check.json
+++ b/.github/markdown-link-check.json
@@ -1,7 +1,0 @@
-{
-  "ignorePatterns": [
-    {
-      "pattern": "^https://www.similarweb.com"
-    }
-  ]
-}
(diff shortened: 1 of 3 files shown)


uncenter:
Sweet! Love lychee, used it for #11901. Does this just check our docs for broken links?

mondeja:
Yes, only for changed documentation files. In `develop` branch exits always with code 0 but still creates a report. Feel free to open other PRs to create reports of possible positive broken links and run on CI at releases to create reports, that would make the work that you've done in #11901 much more easy.

The configuration could eventually bypass some links that became broken if no editions have been made on files, but I think that's ok for now to not bother us so much.

mondeja:
Yes, only for changed documentation files. In `develop` branch exits always with code 0 but still creates a report. The configuration could eventually bypass some links that became broken if no editions have been made on files, but I think that's ok for now to not bother us so much.

Feel free to open other PRs to create reports of possible positive broken links and run on CI at releases to create reports, that would make the work that you've done in #11901 much more easy.

LitoMore (approved):
Looks good.

uncenter:
I considered that, but also note that only half of those reports were real - many if not most were false positives. Just seems too unreliable to fully automate :/

PR #12006, opened 2024-10-13
Use lychee config file

LitoMore:
This moved lychee args to the `lychee.toml`. This also allows users to use the configuration while runnig `lychee` on local.

I also added the `^https://www.figma.com/community/plugin` pattern to fix the error in #12003.

The `curl -I https://www.figma.com/community/plugin/1149614463603005908` will get a 403 response. This may be because Figma rejects requests with unknown user-agents.

### DIFF
--- a/sdk.d.ts
+++ b/sdk.d.ts
@@ -2,7 +2,7 @@
  * @file
  * Types for Simple Icons SDK.
  */
-
+// eslint-disable-next-line n/file-extension-in-import
 import type {CustomLicense, SPDXLicense} from './types';
 
 /**
--- a/sdk.d.ts
+++ b/sdk.d.ts
@@ -2,7 +2,7 @@
  * @file
  * Types for Simple Icons SDK.
  */
-// eslint-disable-next-line n/file-extension-in-import
+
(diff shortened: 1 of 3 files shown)


sdk.d.ts
```diff
@@ -2,7 +2,7 @@
  * @file
  * Types for Simple Icons SDK.
  */
-// eslint-disable-next-line n/file-extension-in-import
```
mondeja, line 5: This removal I think that happens because you've a different version of xo on your PATH, perhaps because you're using another version of Node.js from the one that installed deps locally. I've experienced it one time. Revert it and the lint will pass.

LitoMore: Yes. Fixed now.

mondeja:
See https://github.com/lycheeverse/lychee-action/issues/254

LitoMore:
> See https://github.com/lycheeverse/lychee-action/issues/254

See the latest commit on the v2 branch https://github.com/lycheeverse/lychee-action/tree/v2.

It should be fixed now.

mondeja (approved):
Ah, OK!

PR #12092, opened 2024-10-31
Update Lerna icon

LitoMore:
### Checklist

### Description

Update the icon according to https://github.com/lerna/lerna/issues/4088. The color was picked from the their primary color on the project homepage.


_data/simple-icons.json
```diff
@@ -8827,8 +8827,8 @@
         },
         {
             "title": "Lerna",
-            "hex": "334155",
-            "source": "https://github.com/lerna/website/blob/dcef9c248b2b67bdd2214058beca437732690226/images/lerna-logo-dark.svg"
+            "hex": "9333EA",
+            "source": "https://user-images.githubusercontent.com/900523/173044458-fd0b57f6-6374-4265-98b5-eb8f55fe1fb3.svg"
```
uncenter, line 8831: I don't think we should use user uploaded images as sources like this. Where is this taken from?

uncenter (changes requested):

_data/simple-icons.json
LitoMore: From their GitHub readme, uploaded by their owner. There are no more links for it. It's the most accurate one.

GitHub asset links are always permanent. So it's OK to use the `user-images.githubusercontent` link.

LitoMore:
@uncenter Can I get an approval? Thanks~

_data/simple-icons.json
```diff
@@ -8827,8 +8827,8 @@
         },
         {
             "title": "Lerna",
-            "hex": "334155",
-            "source": "https://github.com/lerna/website/blob/dcef9c248b2b67bdd2214058beca437732690226/images/lerna-logo-dark.svg"
+            "hex": "9333EA",
```
uncenter, line 8830:
```suggestion
            "hex": "C084FC",
```

The color of the favicon is a darker purple which I'm guessing is what you used here, but on the website itself the primary purple used is `#c084fc`. Shall we use that instead?

icons/lerna.svg
uncenter, line 1: Looks good!

uncenter (changes requested):

_data/simple-icons.json
LitoMore: Sounds good. Let's use the more colorful one.

uncenter (approved):

Issue #12107, opened 2024-11-04
Standard icons?


LitoMore:
Are those standard icons like FCC, CE, RoHS, etc icons in our scope?

These are all standards issued by local governments, but this is more a standard for the world. You can usually see them on various electrical appliances.

PratyushJayachandran:
I'm in agreement of this.

mondeja:
I agree too!

PR #12109, opened 2024-11-04
Add CE icon

LitoMore:
**Popularity metric:** It is a symbol that can be seen everywhere in life.

### Checklist

### Description

The scope discussion:

- #12107


uncenter (approved):

uncenter:
This might violate the "No symbols, including flags and banners" stipulation of https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon. Maybe we should add some sort of exception if we want to do #12107, though I'm not sure exactly what would be exempted.

LitoMore:
> No symbols, including flags and banners.

These don't have a global standard. You draw the sign whatever you want.

But electronic standard symbols are different. They have official usage guidelines, like those links I referred to in the JSON.

uncenter:
> But electronic standard symbols are different.

Right, so I'm saying we should specify that those are allowed 😅

PR #12121, opened 2024-11-10
Update Signal icon

uncenter:
image

### Checklist

### Description

The source we previously used was also updated in their brand refresh (https://github.com/signalapp/Signal-Desktop/commits/main/images/signal-logo.svg), but I thought it would be best to use their direct brand page.


LitoMore (approved):
Perfect.

PR #12149, opened 2024-11-17
Make single `aka`s and `old`s render in one line on data-file

mondeja:
The idea has been taken from [this commit](https://github.com/simple-icons/simple-icons/pull/12146/commits/72da9a6281f56daf2b0f1116fb6134574588818f) at #12146. 

Prettify the data file making `aka` and `old` aliases render in a single line if they contain one item. This format is accepted by Prettier, the input election is respected.

Now is more comfortable and readable in my opinion.

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -5,9 +5,7 @@
             "hex": "ECD53F",
             "source": "https://github.com/motdotla/dotenv/blob/40e75440337d1de2345dc8326d6108331f583fd8/dotenv.svg",
             "aliases": {
-                "aka": [
-                    "Dotenv"
-                ]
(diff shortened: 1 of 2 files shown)


LitoMore:
This is not a huge benefit, it's more of a personal preference, I prefer to simplify our linting process and use the default prettier settings.

Also this doesn't help with auto-saving in the editor, which makes it more cumbersome.

Finding icons from a JSON file is a cumbersome task in itself. If you want to increase readability, you can consider using commands to do it, for example:

```shell
npm run show-icon-data
```

```shell
# Inquirer auto-complete search prompt
? Search an icon with title/slug/alias? simpleicons
```

PR #12191, opened 2024-11-25
Add Markdown linting and formatting

mondeja:
To follow a consistent style across all docs. Fixes two errors in third party integrations tables.


LitoMore:
How about using the Prettier instead?

It's much easier for the developers to configure the formatter in the editor.

uncenter:
Sorry, I'm looking for `"proseWrap": "never"` 😅

mondeja:
Now is using `"proseWrap": "never"` 👍🏼

uncenter:
You'll need to remove the changes from when you had it enabled earlier - prettier won't undo that for you, it'll preserve existing formatting like that.

mondeja:
> You'll need to remove the changes from when you had it enabled earlier - prettier won't undo that for you, it'll preserve existing formatting like that.

`"proseWrap": "never"` wraps all paragraphs into one line (see https://prettier.io/docs/en/options.html#prose-wrap).  The rest of the changes are from Prettier or are intended for the Markdownlint checking.

LitoMore (approved):

.github/workflows/publish.yml
uncenter, line 1: Is there a specific reason to make these multiline? Looks messy to me.

LitoMore: It looks more like @mondeja's preference. Both formats are acceptable to me.

In my experience, if you use an editor on the command line and do not turn on line wrapping, the part that exceeds the screen will be difficult to read.

But see here: https://github.com/simple-icons/simple-icons/actions/runs/11991639440/job/33430520601. We have to expand the line to see the rest of the content.

For long commands with some not-that-important parameters, it is okay to use the line wrap. But here is only one version string. I think it's fine to write in one line.

mondeja: Sorry, this was formatted by Prettier. I've reverted these changes.

PR #12203, opened 2024-11-26
Remove Google Fit

mondeja:
Contributes to #11901

Google Fit is deprecated and will not be available at 2025-06-30.


service-paradis:
The API is deprecated to migrate to Health Connect, but I'm not sure if Google Fit App will also be deprecated?

PeterShaggyNoble:
Should we hold off on this until v16 in case the app isn't also being deprecated? [Wikipedia](https://en.wikipedia.org/wiki/Google_Fit#History) would suggest that it _is_ being deprecated, in favour of Fitbit, although it doesn't state as much explicitly.

PeterShaggyNoble:
Deprecation has now been pushed back until next year.

PR #12228, opened 2024-11-27
Update multiple broken sources

mondeja:
Contributes to #11901


uncenter:
Awesome, thanks for doing this. Just curious - did you have a chance to check the logos at the new URLs to see if they vaguely matched what we have in stock? Might be worth it to just triage these for changes in logo especially if the branding URL has changed. I don't have time for this now but I might be able to get to it this weekend of you haven't already by then :)

mondeja:
Reviewed all now and I can confirm that are the same sources. I checked it for most of it while I was editing the sources.

uncenter (approved):

PR #12243, opened 2024-11-28
Ensure that icons data object fields are sorted

mondeja:
This PR implements another rule to `ourlint` script to check that fields in icon data objects of *_data/simple-icons.json* file are sorted given a convenient ordering of fields:

- `title`
- `slug`
- `hex`
- `source`
- `guidelines`
- `license`
- `aliases`

Only 24 icons affected.


scripts/lint/ourlint.js
```diff
@@ -225,6 +225,60 @@ const TESTS = {
       return `Bad licenses:\n\n${badLicenses.join('\n')}\n\nSee the valid license indentifiers at https://spdx.org/licenses`;
     }
   },
+
+  /* Ensure that all fields are sorted in the same way for all icons */
+  fieldsSorted(_, iconsDataString) {
```
LitoMore: You can load the JSON as an array of icons and use the `Object.keys(icon)` to get the fields.

```js
const icons = json.icons;
for (const icon of icons) {
	const keys = Object.keys(icon);
	// ... sort and stringify comparison
}
```

mondeja: Are guaranteed that the fields are sorted in the same order than the file?

mondeja: Okay, I've refactored with your suggestion.

LitoMore: Yes. This should not be a problem when using `JSON.parse`. We have been parsing and stringifying data in `add-icon-data` for a long time and have not seen any problems.

The order only changes during the `JSON.stringify` with a number-keys object. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify.

LitoMore (approved):

PR #12303, opened 2024-12-07
Bump xo to 0.60.0

LitoMore:
See https://github.com/xojs/xo/releases/tag/v0.60.0


adamrusted (approved):
LGTM, and lints still seem to be running fine. Will wait for @mondeja or @uncenter to review though.

scripts/lint/ourlint.js
```diff
@@ -170,7 +170,7 @@ const TESTS = {
 
     const invalidUrls = [];
     for (const [isSourceUrl, url] of allUrlFields) {
-      const $url = new global.URL(url);
+      const $url = new globalThis.URL(url);
```
uncenter, line 173: Silly question but do we need to prefix the URL constructor with global-anything in the first place?

LitoMore: Yes. It's a good practice and good for other runtime/compiler like Bun, Deno, ncc, SWC, etc.

uncenter (approved):

PR #12374, opened 2024-12-17
Remove 1&1 Icon

adamrusted:
Removes 1&1 from our list. While it is still shown at source, the primary domain (1and1.co.uk) now redirects to IONOS.


mondeja:
Why don't you include it in v14?

adamrusted:
Can do, just figured it'd give a 'grace' period before removal

mondeja:
I think that we can include it. The renaming from "1&1 IONOS" to "IONOS" was at October 2023, but both companies were merged long time ago. This icon should have been an aka, but at the moment of the inclusion AKAs didn't exist.

adamrusted:
Apologies, @DerBraky - when visiting the media center that was shown as source, the 'brands' list only included 3 brands - 1&1 was not one of them.
image

Every other international instance of 1&1 now redirects the user to IONOS as a primary domain, hence the deletion.

mondeja:
🤔 Okay, I'll revert the deletion. Sorry, I've been confused with another icon...

PR #12381, opened 2024-12-17
Add script for formatting icon data

LitoMore:
This script can help users automatically fix the `simple-icons.json` order and final newline.

I also fixed the compare function for two icons that have the same title, like `Hive (hive)` and `Hive (hive_blockchain)`.


scripts/format-icon-data.js
uncenter, line 1: "format" seems misleading, I would use "sort". This would be `sort-icon-data.js`.

package.json
```diff
@@ -113,7 +113,8 @@
   "scripts": {
     "build": "node scripts/build/package.js",
     "clean": "node scripts/build/clean.js",
-    "format": "npm run prettier -- --write && npm run xo:fix",
+    "format:icondata": "node scripts/format-icon-data.js",
+    "format": "npm run format:icondata && npm run prettier -- --write && npm run xo:fix",
```
uncenter, line 117: What's the benefit in splitting the icon data into its own script here? I can't see myself ever using it separately from format.

LitoMore: To make our scripts consistent. Like `xo:fix` does. The change was requested by @mondeja.

scripts/format-icon-data.js
LitoMore: This also reformats the file. Even you use other indentations.

mondeja (approved):

package.json
mondeja: When you separate npm scripts and run them from another one, npm shows in their output each `npm run` command executed. If a failure happens or the command takes some apreciable time, it's easy to know what script is causing it.

scripts/format-icon-data.js
mondeja: I agree with @LitoMore here, sorting is something included in the formatting process.

PR #12405, opened 2024-12-21
Generate provenance statements when publishing

LitoMore:
Documentation: https://docs.npmjs.com/generating-provenance-statements


.github/workflows/publish.yml
```diff
@@ -29,6 +29,8 @@ jobs:
     name: NPM Package
     needs: sanity-check
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
     steps:
```
mondeja, line 34: What does this?

mondeja: Ah OK, I see it in the other PRs.

.github/workflows/publish.yml
```diff
@@ -54,6 +56,8 @@ jobs:
         uses: JS-DevTools/npm-publish@v3
         with:
           token: ${{ secrets.NPM_TOKEN }}
+          provenance: true
+          access: public
```
mondeja, line 60: I think that `access: public` is not needed as is the default. Or is related by some way?

.github/workflows/publish.yml
```diff
@@ -29,6 +29,8 @@ jobs:
     name: NPM Package
     needs: sanity-check
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
mondeja, line 33: Seems to me that this permission should be documented, as is not clear what it does. Is missing in the documentation of the `JS-DevTools/npm-publish` action.

LitoMore: It's in npm's documentation: https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions

.github/workflows/publish.yml
LitoMore: Ah, yes. It's not necessary since it's not our first time to publish the package.

.github/workflows/publish.yml
mondeja: Ah, OK then...

mondeja (approved):

PR #12380, opened 2024-12-17
Use tabs

LitoMore:
Resolves #11741

Markdown keeps spaces due to https://github.com/prettier/prettier/issues/8782

### How do I resolve conflicts on this PR

1. Copy the code from develop to the branch
2. Run `npm run format`
3. Done

### How to resolve conflicts after this PR

1. Press the dot key <kbd>.</kbd> on the PR page to open the VS Code web
2. Open `simple-icons.json` from file explorer
4. Click the indentation button `Spaces: 4` on the bottom status bar
5. Select `Convert Indentation to Tabs` from the popup command palette
6. Save and commit changes (There is a Git icon on the left sidebar)
7. Done

PR #12424, opened 2024-12-23
Fix version output from `get-version` action

LitoMore:
Fixes https://github.com/simple-icons/simple-icons/actions/runs/12458105634/job/34773364133

I used `xargs` to trim spaces and tabs.

### DIFF
--- a/.github/actions/get-version/action.yml
+++ b/.github/actions/get-version/action.yml
@@ -12,5 +12,5 @@ runs:
     - id: get-version
       shell: sh
       run: |
-        version="$(grep version -m 1 -i package.json | sed 's/[ \",:version]//g' | xargs)"
+        version="$(grep version -m 1 -i package.json | cut -d '"' -f4)"
         echo "version=$version" >> $GITHUB_OUTPUT
--- a/.github/actions/get-version/action.yml
+++ b/.github/actions/get-version/action.yml
@@ -12,5 +12,5 @@ runs:
     - id: get-version
       shell: sh
       run: |
-        version="$(grep version -m 1 -i package.json | sed 's/[ \",:version]//g')"
+        version="$(grep version -m 1 -i package.json | sed 's/[ \",:version]//g' | xargs)"
         echo "version=$version" >> $GITHUB_OUTPUT


.github/actions/get-version/action.yml
```diff
@@ -12,5 +12,5 @@ runs:
     - id: get-version
       shell: sh
       run: |
-        version="$(grep version -m 1 -i package.json | sed 's/[ \",:version]//g')"
+        version="$(grep version -m 1 -i package.json | sed 's/[ \",:version]//g' | xargs)"
```
mondeja, line 15: This has been broken by #12380. Please, use this instead:

```sh
grep version -m 1 -i package.json | cut -d '"' -f4
```

Your fix looks like a patch over an outdated command.

mondeja (changes requested):

mondeja (approved):

uncenter:
Why aren't we using jq for extracting the version field? Would be `jq .version package.json -r` I think.

mondeja:
> Why aren't we using jq for extracting the version field? Would be just `jq .version package.json -r`, and works regardless of formatting or location of the key in the package.json. jq is included by default on GitHub Actions runners iirc.

I always write no-dependency UNIX-Posix Shell syntax by default, which is totally portable to all shells. This `sed` shit was legacy code added on v4, with `cut` would have worked regardless the tabs change.

It seems to me that if we depend on jq, we depend on another thing which is always a bad decission. I don't really care so much because probably GH will depend on jq forever, but even depending on GH itself seems wrong to me.

mondeja:
Tested performance and jq is 10 times slower:

```bash
$ time for i in $(seq 1 100); do grep version -m 1 -i package.json | cut -d '"' -f4; done
real	0m0,447s
user	0m0,215s
sys	0m0,522s

$ time for i in $(seq 1 100); do jq .version package.json -r; done
real	0m3,958s
user	0m3,586s
sys	0m0,388s
```

uncenter:
```
$ hyperfine 'jq .version package.json -r'
Benchmark 1: jq .version package.json -r
  Time (mean ± σ):       1.7 ms ±   2.4 ms    [User: 0.9 ms, System: 1.1 ms]
  Range (min … max):     0.5 ms …  19.0 ms    129 runs
```

We are talking about 2 milliseconds. There is zero point in benchmarking this.

uncenter:
```
$ hyperfine 'grep version -m 1 -i package.json | cut -d \'"\' -f4' 'jq .version package.json -r'
Benchmark 1: grep version -m 1 -i package.json | cut -d '"' -f4
  Time (mean ± σ):       2.6 ms ±   0.4 ms    [User: 3.3 ms, System: 0.8 ms]
  Range (min … max):     2.0 ms …   5.1 ms    523 runs

Benchmark 2: jq .version package.json -r
  Time (mean ± σ):       2.3 ms ±   0.5 ms    [User: 1.1 ms, System: 0.5 ms]
  Range (min … max):     1.5 ms …   4.8 ms    490 runs

Summary
  jq .version package.json -r ran
    1.17 ± 0.31 times faster than grep version -m 1 -i package.json | cut -d '"' -f4
```

mondeja:
> We are talking about 2 milliseconds.

What does hyperfine? Why the result is so different against a Bash for loop?

mondeja:
Ah I see what it does... just another "efficient box"...

> There is zero point in benchmarking this.

Just to aggregate something anecdotic, consider the other arguments.

LitoMore:
`cut` is clear enough for trimming text.

If we need to operate JSON frequently, I would recommend `jq` because it is more intuitive.

But here we only have one line to operate. The `jq` is too much.

uncenter:
Fwiw there is also `npm pkg get version` that we could use. It is marginally slower but makes it readable and consistent.

mondeja:
> Fwiw there is also `npm pkg get version` that we could use. It is marginally slower but makes it readable and consistent.

`npm pkg get version` prints the version wrapped by `"` characters that need to be stripped anyways. The full command would be something like `npm pkg get version | tr -d '"'` which needs npm as dependency and is not very much readable anyways.

PR #12426, opened 2024-12-23
Remove CSS3 icon

LitoMore:
Related to:

- #12134

The CSS3 icon can be removed since we have a new CSS icon.


uncenter:
This removal seems rather quick. Seeing as the CSS3 logo has been used for many years and is more recognizable than the *brand new*, months-old logo we just added - I would refrain from making this change for a few years at least.

adamrusted:
Years seems a bit overkill. We can add an alias to CSS for people to find it via CSS3 if necessary, but we should look to remove as of v15. If React changed their logo tomorrow, we wouldn't keep both the new and old for YEARS.

PR #12562, opened 2025-01-12
Add VRR icon

DerBraky:
Untitled

**Issue:** n/a

**Popularity metric:**
- Similarweb Global Rank: [51,370](https://www.similarweb.com/website/vrr.de) 

### Checklist


PeterShaggyNoble (changes requested):
I'm noticing a [small difference](https://github.com/user-attachments/assets/83b73d4b-f6ea-492e-aed8-6fff17fd005c) in both Rs (yours in red), @DerBraky. That may either be a precision issue or because I'm comparing to a different source (I'm using [this SVG](https://www.vrr.de/assets/img/logo-vrr.svg) from the website's header. There are also a few excess points in the top corners and in the holes in the Rs that you should be able to move.

And, as there are no vector files available in their styleguide, I would switch the `source` & `guidelines` URL, simplifying the Frontify URL to just https://company-141992.frontify.com/d/bJ2KdVH3D6Kx/vrr-styleguide

PR #12587, opened 2025-01-17
Add Dev Container configurations

LitoMore:
This allows users to use our pre-configured runtime environment with Dev Containers.

### How to try it out?

1. Press the <kdb>.</kbd> on the current pull request page to enter online VS Code editor.
2. Then open the VS Code terminal, you might be able to see the info that lead to the GitHub Codespaces.
3. Open with GitHub Codespaces, then you will see a popup window that tells you can build the Dev Container for current project. If you can't see any popup, you can still manually build the Dev Container from VS Code's command palette.
4. The extensions should be set up for you after the build.

Hint: Press <kbd>Shift + Command + P</kbd> or <kbd>Shift + Ctrl + P</kbd> to open the command palette.


mondeja (commented):
I think that you need to use the `image` property to configure another version of Node.js with a docker image. Currently on the Dev Container is v17.6.0

mondeja (approved):

PR #12640, opened 2025-01-25
Add PHPStan logo

### LINKED ISSUE #11435: Add PHPStan icon
### Brand Name

PHPStan

### Website

https://phpstan.org/

### Popularity Metric

860,419

None

### Additional Comments

_No response_

elcreator:
**Issue:** closes #11435

**Popularity metric:**
https://www.similarweb.com/website/phpstan.org/ in 500k
GitHub 13.2k stars, actively used

### Checklist


LitoMore:
@elcreator Could you share where you get the vector resource? Or did you just vectorize the icon with other tools?

elcreator:
I vectorized it https://github.com/phpstan/phpstan using Inkscape and optimized it using SVGO
68747470733a2f2f692e696d6775722e636f6d2f5761524b506c432e706e67

LitoMore:
Did you notice this?



Can you fix it?

Also, the precision 4 is unnecessary. Can you process the icon with precision 3?

LitoMore (changes requested):

elcreator:
https://github.com/simple-icons/simple-icons/pull/12651

PeterShaggyNoble (changes requested):
Thanks for your work on this so far, @elcreator 👍🏻 

However, the handle of the magnifying glass doesn't line up with the source. Also, I'd keep the frame solid, with a cut-out around it, cut out the lens, and make the eye solid, too.

elcreator:
Yes, I moved it to the right to compensate for the lack of color difference.
Which width of the cutout you prefer? should it be outer or inner?
Outer cutout will cause too tiny piece of solid ink near the ear so the elephants right ear will look detached (i.e. one on the left side of the picture).

PeterShaggyNoble:
> What width of cutout do you prefer? Should it be outside or inside?
We usually go with a width of 0.5dp. And, in this case, it would be an outer cut-out.

> I moved it to the right to compensate for the lack of color difference
I should have spotted that 🤦🏻‍♂️ It will need to be realigned to the original, so, and then a cut-out around the whole shape of the magnifying glass.

elcreator:
If you launch potrace on the original then it will look like this:
or
Outer cut-out around the lens (i.e. adding circle of black background around the black frame or white background around the white frame if we reverse colors) will make the frame even thicker and will make ear look detached.
While handle will be still not outlined properly as part of it surrounded with black and another part - with white.
Can you please draw a sketch how it should be drawn?

PR #12651, opened 2025-01-26
Fixed build failing if some non-required parameters of simple-icons.json are missing

elcreator:
Improved security by replacing non-string values with empty string

**Issue:** closes #

**Popularity metric:**

### Checklist


LitoMore:
That's because your file format is wrong. See https://github.com/simple-icons/simple-icons/actions/runs/12976088588/job/36187961864?pr=12640.

elcreator:
I still think it should not fail in value.replaceAll in such cases (neither validation function with replaceAll should bypass anything except strings).

LitoMore:
I would enhance the error message instead.

Issue #4878, opened 2021-01-27
Add Plausible Analytics


metmarkosaric:
**Brand Name:** Plausible Analytics
**Website:** https://plausible.io
**Alexa rank:** 64,891 
  
**Official resources for icon and color:**
https://github.com/plausible/docs/blob/master/static/img/logo.svg

I'm the co-founder of Plausible Analytics. Thanks!

service-paradis:
Thanks for the request @metmarkosaric!

Would you like to take a look at [our contributing guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md) and gave it a try?

metmarkosaric:
Thanks @service-paradis! Will try again.

service-paradis:
@metmarkosaric 
Sorry for the confusion, your issue was correct. I was just asking you if you wanted to make a PR by yourself :)

metmarkosaric:
Ahh I see. I'm not a developer unfortunately so I'm not sure how to do it by reading the instructions there.

service-paradis:
No problem! Another contributor or a maintainer will do it eventually :)
Out of curiosity, do you have monochrome version available?

PeterShaggyNoble:
@metmarkosaric, as one of the founders, do you have an existing monochrome version of your logo or, if not, do you have a preferred treatment for converting it to monochrome? We could either just flatten it completely or use a cut-out.

![](https://user-images.githubusercontent.com/15157491/109178057-5cb7ca00-7780-11eb-8d0d-1f83e6c3b13c.png)
![](https://user-images.githubusercontent.com/15157491/109178225-8a9d0e80-7780-11eb-8239-fb32cf02fc4a.png)

metmarkosaric:
Thanks @PeterShaggyNoble! Let's go with the cutout version!

PR #12661, opened 2025-01-27
Update Plausible Analytics

romain-dartigues:
Plausible Analytics logo has been updated in early 2024.

Refs: 7a9e4001

*Popularity metric:**
* https://github.com/plausible over 21 k☆ on GitHub
* refer to #4878

### Checklist

- [x] I optimized the icon with ~~SVGO or SVGOMG~~ [SVG Path Editor](https://github.com/Yqnn/svg-path-editor) ♥

### Description

Plausible Analytics logo has been updated in early 2024, they have not removed the previous version of the logo.
I did the job for https://github.com/edent/SuperTinyIcons/pull/886 and thought it might also fit here.

> Why did you pick the hex value?

It's the color of the text in the new version of the logo; might be a tad dark though.

@metmarkosaric, do you have an opinion? Should the previous version of the icon be kept as well?


_data/simple-icons.json
```diff
@@ -12255,8 +12255,8 @@
 	},
 	{
 		"title": "Plausible Analytics",
-		"hex": "5850EC",
-		"source": "https://github.com/plausible/docs/blob/be5c935484e075f1e0caf3c9b3351ddd62348139/static/img/logo.svg"
+		"hex": "1F2961",
+		"source": "https://github.com/plausible/docs/pull/471"
```
mondeja, line 12259: Please, use the full URL to the file:

```suggestion
		"source": "https://github.com/plausible/docs/blob/cf53502f7e7a8673f75e3f05517f8a61d5cb7953/static/img/plausible_logo.svg"
```

metmarkosaric:
thanks for the update @romain-dartigues! 🙏 looks good to me and there's no need to keep the old icon as we don't use it anymore

_data/simple-icons.json
romain-dartigues: Thank you for your feedback, it should be good now.

LitoMore (changes requested):
I can see some differences. Can you check it?



It's better to use a perfessional SVG editor app like Inkscape for editing. And the precision should at least 3.

romain-dartigues:
@LitoMore < Thank you for your comment, however I kindly refuse; my goal was to update the icon with file size in mind.
The difference you are mentioning is less than one millimeter for a print of forty-two centimeters.
You can SVG Optimizer the original file ♥.

PR #12695, opened 2025-02-01
Add .nvmrc and .node-version files

mondeja:
Most Node.js package managers support reading the Node.js version from a *.nvmrc* file. For example:

- https://github.com/Schniz/fnm
- https://github.com/nvm-sh/nvm

With fnm, the `node` version is changed when cding in the repository:

```sh
$ node --version
v18.18.2
$ cd simple-icons
Using Node v22.9.0
simple-icons$ node --version
v22.9.0
```

Some other tools support *.node-version*, so adding it too for maximum compatibility.


LitoMore:
Be careful. This modification allows our stranger contributors to modify the version of Node.js that our GitHub Actions runs through `.nvmrc`.

.github/workflows/verify.yml
```diff
@@ -106,10 +110,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - id: get-node-version
+        uses: ./.github/actions/get-node-version
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: ${{ steps.get-node-version.outputs.version }}
```
uncenter, line 118:
```suggestion
          node-version-file: .node-version
```

.github/workflows/verify.yml
```diff
@@ -106,10 +110,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - id: get-node-version
+        uses: ./.github/actions/get-node-version
```
uncenter, line 114:
```suggestion
```

uncenter (changes requested):
`actions/setup-node` supports the `node-version-file` configuration option to read from that file. No need to add this extra stuff to extract it on our own. I think we can also drop the .nvmrc and just use the .node-version.

uncenter:
Nevermind, nvm still doesn't support .node-version. See https://github.com/shadowspawn/node-version-usage/issues/4. (Yet another reason to not use that outdated tool, fnm is much better.)

.github/workflows/create-release.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version-file: .node-version
           cache: npm
           cache-dependency-path: '**/package.json'
```
uncenter, line 59:
```suggestion
```
I know this isn't directly related to this pull request but after reviewing the actions/setup-node documentation I don't believe we need to set this key every time.

> The action defaults to search for the dependency file (package-lock.json, npm-shrinkwrap.json or yarn.lock) in the repository root, and uses its hash as a part of the cache key. Use cache-dependency-path for cases when multiple dependency files are used, or they are located in different subdirectories.

Since the only relevant package.json is in the repository root, we don't need it to glob through all subdirectories. (There are package.jsons in the scripts and test subdirectories, but only to set the `type` to module, not containing dependencies.)

LitoMore (approved):
LGTM. I've also posted a question to https://github.com/actions/setup-node/issues/1224 about the security concern.

.github/workflows/create-release.yml
mondeja: Fair. But note that we _should_, eventually, separate the SDK in another npm package, residing in a subdirectory. In that case, we'll need to add the line again.

Issue #5859, opened 2021-06-05
Implement functionality based on the `old` alias


ericcornelissen:
Relates to #4797 and #5627

---

### Description

In Pull Request #5369 we added the possibility for four types of aliases to the JSON data, one of which is `old`. This issue is intended to decide on how we want to use the `old` alias. Feedback and suggestions from anyone are welcome :smile: 

To me the intention of `old` is to establish a clear and easy way to provide backwards compatibility (*1) for brands that have been renamed or slugs that need to be changed for another reason.

### Proposal

With that in mind I propose we implement the following based on the `old` alias: For every brand in [`simple-icons.json`](https://github.com/simple-icons/simple-icons/blob/562aaac0488c17b5288e3018278896dafb2b1735/_data/simple-icons.json) that has an `old` alias, [the build script](https://github.com/simple-icons/simple-icons/blob/562aaac0488c17b5288e3018278896dafb2b1735/scripts/build/package.js) should generate an `.svg` file based on the "original" `.svg` as well as a `.js` file for [the NPM package](https://www.npmjs.com/package/simple-icons) (*2) using the old brand name/slug as file name. The build script should also include the brand using the old brand name/slug in [the NPM package's index.js file](https://github.com/simple-icons/simple-icons/blob/562aaac0488c17b5288e3018278896dafb2b1735/scripts/build/templates/index.js).

### Considerations

Unfortunately this somewhat complicates keeping track of unique slugs. However, as the `old`'s slug was already occupied before the brand renamed, it will never happen that the slug "introduced" by the `old` alias is already in use. That said, we should try to make sure that a new addition that clashes with an `old` alias is detected automatically by tests or linting.

### Open Questions

- (*1) How long should we keep the backwards compatibility? This can be discussed in #5627.
- (*2) Should the SVG `<title>` and JavaScript `title` and `svg` properties have the old brand name or the new brand name. I think it should be the new brand name as the SVG should be the up-to-date version that relates to the new brand name.

jorgeamadosoria:
Can't we just treat the old alias as the aka?

and as for the package, isn't there a way to get a particular svg from different slugs, so you don't have to generate duplicated svgs?

ericcornelissen:
> Can't we just treat the old alias as the aka?

[...]

PR #12795, opened 2025-02-19
Sort icon data object keys

LitoMore:
Sort all icon data object keys.

Also fixed a missing type.

### DIFF
--- a/scripts/format-icon-data.js
+++ b/scripts/format-icon-data.js
@@ -3,8 +3,7 @@
  * Format _data/simple-icons.json.
  */
 import {getIconsData} from '../sdk.mjs';
-import {sortIconsCompare, writeIconsData} from './utils.js';
+import {formatIconData, writeIconsData} from './utils.js';
 
-const iconsData = await getIconsData();
-iconsData.sort(sortIconsCompare);
-writeIconsData(iconsData);
+const icons = await getIconsData();
+writeIconsData(formatIconData(icons));
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -81,3 +81,88 @@ export const sortIconsCompare = (a, b) => {
 		? collator.compare(getIconSlug(a), getIconSlug(b))
 		: collator.compare(a.title, b.title);
 };
+
+/**
+ * Sort icon data obeject.
+ * @param {IconData} icon The icon data as it appears in *_data/simple-icons.json*.
+ * @returns {IconData} The sorted icon data.
+ */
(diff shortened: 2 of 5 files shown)


LitoMore:
> This seems duplicated, see https://github.com/simple-icons/simple-icons/pull/12243

This script can fix all errors. This also sorted all other keys and members.

It's used for Husky to auto-format before committing.

scripts/format-icon-data.js
```diff
@@ -2,9 +2,88 @@
...
+	if (!aliases) return undefined;
+	const sortAliases = Object.fromEntries(
+		Object.entries(aliases).sort(([key1], [key2]) => (key1 > key2 ? 1 : -1)),
+	);
+	return sortAliases;
+};
+
+const icons = await getIconsData();
+const mappedIcons = icons.map((icon) => {
+	return sortIcon({
```
mondeja, line 68: Would you mind to share the same sorting implementation in the `fieldsSorted` function of *ourlint.js*?

sdk.d.ts
```diff
@@ -32,6 +32,7 @@ export type Aliases = {
 	aka?: string[];
 	dup?: DuplicateAlias[];
 	loc?: Record<string, string>;
+	old?: string[];
```
mondeja, line 35: The `old` alias is not implemented yet (see #5859). Can you remove it? I think that only adds confusion for type consumers.

scripts/utils.js
```diff
@@ -81,3 +81,88 @@ export const sortIconsCompare = (a, b) => {
...
+						dup: icon.aliases.dup
+							? icon.aliases.dup.sort(sortIconsCompare).map((d) =>
+									sortIcon({
+										...d,
+										loc: sortAlphabetically(d.loc),
+									}),
+								)
+							: undefined,
+						loc: sortAlphabetically(icon.aliases.loc),
+						old: icon.aliases.old?.sort(collator.compare),
```
mondeja, line 161: And remove from here too. Currently we don't have icons with `old` aliases.

mondeja (changes requested):

scripts/utils.js
LitoMore: There are five icons that have the `old` aliases.

mondeja: Ok. So we are defining a version of the `old` implementation now. I'll open a PR to document it.

mondeja (approved):
Looks good. Thanks!

PR #12841, opened 2025-03-04
Use square brackets for JSDoc optional arguments

mondeja:
See https://jsdoc.app/tags-type

### DIFF
--- a/scripts/build/package.js
+++ b/scripts/build/package.js
@@ -100,10 +100,10 @@ const iconToJsObject = (icon) => {
  * Write JavaScript content to a file.
  * @param {string} filepath The path to the file to write.
  * @param {string} rawJavaScript The raw JavaScript content to write to the file.
- * @param {EsBuildTransformOptions | null} options The options to pass to esbuild.
+ * @param {EsBuildTransformOptions} [options] The options to pass to esbuild.
  */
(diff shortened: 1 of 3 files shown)


scripts/build/package.js
```diff
@@ -100,10 +100,10 @@ const iconToJsObject = (icon) => {
  * Write JavaScript content to a file.
  * @param {string} filepath The path to the file to write.
  * @param {string} rawJavaScript The raw JavaScript content to write to the file.
- * @param {EsBuildTransformOptions | null} options The options to pass to esbuild.
+ * @param {EsBuildTransformOptions} [options] The options to pass to esbuild.
  */
-const writeJs = async (filepath, rawJavaScript, options = null) => {
-	options = options === null ? {minify: true} : options;
+const writeJs = async (filepath, rawJavaScript, options) => {
+	options = options === undefined ? {minify: true} : options;
```
LitoMore, line 106: The `undefined` would be better, IMO. See https://github.com/sindresorhus/meta/discussions/7.

mondeja: Agree. This should be the way https://boxed.cool/option but `undefined` is enough using TypeScript.

LitoMore (approved):

PR #12852, opened 2025-03-06
Catch Ctrl+C keys to abort execution in scripts

mondeja:
Currently, strange errors appear in the screen when you try to stop these scripts using <kbd>Ctrl</kbd>+<kbd>C</kbd>.

### DIFF
--- a/scripts/add-icon-data.js
+++ b/scripts/add-icon-data.js
@@ -28,6 +28,14 @@ import {
 	writeIconsData,
 } from './utils.js';
 
+// Ctrl+C to abort
+process.stdin.on('data', (key) => {
+	if (key.toString() === '\u0003') {
+		process.stdout.write('Aborted\n');
+		process.exit();
+	}
+});
+
 /** @type {import('../sdk.js').IconData[]} */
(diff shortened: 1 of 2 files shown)


LitoMore:
See https://github.com/SBoudrias/Inquirer.js/tree/main/packages/prompts#handling-ctrlc-gracefully. We can handle more cases in this way.

mondeja:
I tried that but unfortunately generates an ugly error on my local environment:

```
$ npm run remove-icon
(Move up and down to reveal more choices)
> simple-icons@14.9.0 remove-icon
> node scripts/remove-icon.js

? Select an icon to remove: Warning: Detected unsettled top-level await at file:///home/mondeja/files/code/simple-icons/scripts/remove-icon.js:38nv)
const found = await autocomplete({
              ^
  1.1.1.1 (1dot1dot1dot1)
  1&1 (1and1)
  1001Tracklists (1001tracklists)
```

LitoMore (approved):
Ah, okay. The documentation might be outdated. I will look into the Inquirer.js to see if we can have a better solution.

PR #12905, opened 2025-03-15
Add voelkner icon

lukasbestle:
voelkner

**Popularity metric:** [Similarweb global rank 36k](https://www.similarweb.com/website/voelkner.de/)

### Checklist

### Description

I extracted the SVG from the SVG favicon from their website. The hex value was verified in the icon sprite, the main header CSS and the favicon.


PeterShaggyNoble (changes requested):
Thanks for the contribution, @lukasbestle :+1: However your version (in red below) got vertically squashed somewhere along the way, when compared with the `shop` symbol in [this SVG](https://www.voelkner.de/themes/voelkner/images/16b349edec1516cfaadf3097b36fe52e/sprite_logo.svg).

lukasbestle:
The `shop` symbol in the sprite is the full logo (`voelkner`). For this PR I used the `shop-small` symbol of the same sprite, which is just the icon (`v`). I think it makes sense to use the icon directly and not extract the first letter from the full logo.

PeterShaggyNoble:
Well spotted, @lukasbestle. However, I don't see that `symbol` in use anywhere on their homepage. Did you come across it in use anywhere on the site? If not I don't think we're going to be able to use it. In fact, they use the more vertically stretched version from the full wordmark for their [favicon](https://www.voelkner.de/themes/voelkner/images/16b349edec1516cfaadf3097b36fe52e/favicon/favicon.svg), too.

lukasbestle:
Good point. To be honest I didn't even realize they were different and assumed that the icon from their sprite must be the correct one. I've now used the favicon SVG instead.

PeterShaggyNoble (approved):
Perfect. Thanks, @lukasbestle.

PR #13015, opened 2025-04-04
Bump to xo@1.0.0

LitoMore:
Release note: https://github.com/xojs/xo/releases

~~Currently I'm using `xo.config.js` for configurations, we can change it to `xo.config.ts` after this PR:~~

- https://github.com/xojs/xo/pull/780

It looks like the declaration of the `xo.config.mts` still has some issues. We can change it to `.mts` in the future PR.


mondeja (changes requested):
Remove *.xo-config.json*.

PR #13022, opened 2025-04-05
Add script for auto-closing issues

### LINKED ISSUE #13010: Enhance Auto-Close Script to include all forbidden brands
It's becoming a near daily thing now that somebody requests LinkedIn, or a Microsoft icon. With that slew of icons being banned until Microsoft give us a clear steer one way or the other, and other large brands such as Adobe likely to be next on the chopping block - we need to look at having a way to automate the closure of these issues and PRs.

We already have Java issues being auto-closed with a suitable comment, I would like to propose that we store a custom message and reference for each forbidden brand, and enable the automation to identify any permutation of any of these brands, closing them and using the relevant message and 'duplicate of' reference.

LitoMore:
Resolves #13010


adamrusted:
@LitoMore we had discussed previously keeping a separate JSON of forbidden brands that populates the Markdown file, as well as serving as an array for mapping in a function like this. Might be best to take that as a base, and ensure we're normalising and lower-casing any issue titles so we match regardless of what shenanigans a user tries to bypass it.

adamrusted (commented):
LGTM @LitoMore - do we want to add in the auto-management of forbidden brands in the markdown  here, or handle that elsewhere? Also worth including all our forbidden brands - just for completeness. I'm sure there are issue comments / PRs explaining why we can't add them.

LitoMore:
I think we can change the `reasons` field to a string so that we can customize the reason for each rule.

LitoMore:
I've added some of the most frequently requested icons to the rules list.

autoclose.rules.mjs
```diff
@@ -0,0 +1,34 @@
...
+	{
+		patterns: [/matlab/i],
+		reason: autocloseTerm + '#1233.',
+	},
+	{
+		patterns: [/disney/i],
+		reason: autocloseTerm + '#2309.',
+	},
+	{
+		patterns: [/oracle/i, /java\s/i, /java$/i],
```
mondeja, line 17: These two regexes for Java can be composed:

```suggestion
		patterns: [/oracle/i, /java(?:\s|$)/i],
```

mondeja (changes requested):

autoclose.rules.mjs
```diff
@@ -0,0 +1,34 @@
+/**
```
mondeja, line 1: I would put this file rules file and the script inside a new folder at the path *.github/actions/autoclose-issues/* because is only part of our CI. Adding it to the root and the script to the root of the *scripts/* folder seems to me a bit confusing.

scripts/autoclose.js
```diff
@@ -0,0 +1,159 @@
...
+			);
+			return pattern.test(title) || brandNamePattern.test(body);
+		}),
+	);
+	if (!matched) return undefined;
+	return matched.reason;
+};
+
+/**
+ * Check if the issue is a won't add icon issue.
```
mondeja, line 101: The documentation of this function looks outdated.

mondeja (changes requested):

autoclose.rules.mjs
mondeja: Or probably put the rules file into *.github/* folder.

scripts/autoclose.js
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * Fetch data from GitHub API.
+ * @param {string} url The URL to fetch.
+ * @param {globalThis.RequestInit} options The options to pass to the fetch function.
+ * @returns {Promise<any>} - The response data.
+ */
+const githubFetch = async (url, options) => {
+	const response = await globalThis.fetch(new URL(url), {
```
uncenter, line 47:
```suggestion
	const response = await globalThis.fetch(url, {
```
Is there a reason we need to create a new `URL` instance from the raw string url? `fetch` accepts a plain string just fine afaik.

scripts/autoclose.js
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/** @type {Issue} */
+	const json = await response.json();
+	const {labels, state, title, body} = json;
+	const labelNames = new Set(labels.map((label) => label.name));
+	if (state === 'closed') return undefined;
+	if (labelNames.has('meta')) return undefined;
+	if (labelNames.has('in discussion')) return undefined;
+	if (!labelNames.has('new icon')) return undefined;
```
uncenter, line 86:
```suggestion
	if (state === 'closed' || labelNames.has('meta') || labelNames.has('in discussion') || !labelNames.has('new icon')) return;
```

scripts/autoclose.js
```diff
@@ -0,0 +1,159 @@
...
+	const matched = rules.find((rule) =>
+		rule.patterns.some((pattern) => {
+			const brandNamePattern = new RegExp(
+				`### Brand Name\n*${pattern.source.replaceAll('$', '')}\n*###`,
+				'i',
+			);
+			return pattern.test(title) || brandNamePattern.test(body);
+		}),
+	);
+	if (!matched) return undefined;
```
uncenter, line 96:
```suggestion
	if (!matched) return;
```
Is there a reason (lint rule?) we are specifying an explicit undefined return value here and above?

uncenter:
This looks great overall. Nice work @LitoMore, I think we all can't wait to stop having to close all those spammy issues!

scripts/autoclose.js
LitoMore: URL is also available. The benefit is that it can validate the URL string before use.

scripts/autoclose.js
LitoMore: We should return `undefined` for more clear intention. And `return` only is more for void rather than `undefined`.

scripts/autoclose.js
uncenter: What's the significance of the order?

adamrusted: I agree on state being it's own item, if we were being pedantic we could probably merge the meta and in discussion lines into one, but I won't lose any sleep over that. Separating the 'new icon' bit makes sense, given it's a different query than the others too.

scripts/autoclose.js
adamrusted: +1 for validating the URL where possible. The amount of times assuming a string will be a URL (in my own stuff, not here) has tripped me up is unbelievable.

autoclose.rules.mjs
mondeja: I would just remove `.github export-ignore` from *.gitattributes*.

LitoMore: > dot-files (except for .eslintrc.*) as well as dot-folders and their contents are ignored.
> https://eslint.org/docs/latest/use/configure/ignore-deprecated

It looks like we can't unignore the `.github` folder. I also tried some methods in #13015, but unfortunately, none worked.

mondeja: 🤔 In that case, I would put both files in a new folder at *scripts/autoclose-issues/*. I'll try to debug the problem in the future. Reading the documentation of ESLint v9, it seems to me that it could be possible or should reported as a bug.

LitoMore: Ok, I've moved those files to `scripts/autoclose-issues/`.

scripts/autoclose-issues/autoclose.rules.js
```diff
@@ -0,0 +1,34 @@
...
+	{
+		patterns: [/matlab/i],
+		reason: autocloseTerm + '#1233.',
+	},
+	{
+		patterns: [/disney/i],
+		reason: autocloseTerm + '#2309.',
+	},
+	{
+		patterns: [/oracle/i, /java\s/i, /java$/i],
```
adamrusted, line 17:
```suggestion
	{
	        patterns: [/bp/i, /british\s?petroleum/i],
	        reason: autocloseTerm + '#5849.',
	},
	{
	        patterns: [/mattel/i, /barbie/i, /hot\s?wheels/i, /fisher(?:\s|-)price/i],
	        reason: autocloseTerm + '#6656.',
	},
	{
	        patterns: [/microchip/i],
	        reason: autocloseTerm + '#9373.',
	},
	{
		patterns: [/disney/i, /disney\s?(?:plus|+)/i],
		reason: autocloseTerm + '#2309.',
	},
	{
		patterns: [/oracle/i, /java(?:\s|$)/i],
```

autoclose.rules.mjs
```diff
@@ -0,0 +1,34 @@
...
+ * @file Auto-close rules for closing won't add icons.
+ */
+
+const autocloseTerm = 'This issue was automatically closed. Please refer to ';
+
+const rules = [
```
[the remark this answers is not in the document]
adamrusted: Added to https://github.com/simple-icons/simple-icons/pull/13022#discussion_r2042226439

scripts/autoclose-issues/autoclose.rules.js
LitoMore: The `/bp/` pattern may accidentally close other brand requests. We can use `/british\s?petroleum/i` only.

adamrusted (approved):
LGTM @LitoMore!

scripts/autoclose-issues/autoclose.app.js
```diff
@@ -0,0 +1,159 @@
...
+ * @returns {Promise<string | undefined>} - Returns reason if the issue is a won't add icon issue, undefined otherwise.
+ */
+const checkIfCanBeClosed = async () => {
+	const url = `https://api.github.com/repos/${GITHUB_REPOSITORY}/issues/${ISSUE_NUMBER}`;
+	const response = await githubFetch(url, {method: 'GET'});
+	if (!response.ok) {
+		throw new Error(
+			`Failed to fetch: ${response.status} (${response.statusText}).`,
+		);
+	}
```
mondeja, line 77: This is already checked at `githubFetch`, so it will never be reached.

mondeja:
Nevermind, I'll open another PR...

PR #13027, opened 2025-04-06
Add `update icon/data` label when only data file is changed

mondeja:
With this change, the labeler would have been added the `update icon/data` label to #13020.

I'm not totally sure if we're adding this label to data-only changed PRs; looks to me like that's the case. Probably, we need to add the label `skip release note` too. Thoughs about it @LitoMore?

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,6 +4,8 @@ new icon:
 update icon/data:
   - any: [icons/*.svg]
     status: modified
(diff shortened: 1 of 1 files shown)


adamrusted:
I don't know we need to skip release note if there is a genuine change to the JSON. I think we're better placed adding that manually if it's a format change or time-sensitive URL change.

adamrusted (approved):
LGTM @mondeja! 🎉

Issue #11676, opened 2024-08-23
Request: MacPorts


jcs090218:
### Brand Name

MacPorts

### Website

https://www.macports.org/

### Popularity Metric

The similarweb rank is 304,263.

### Official Resources for Icon and Color

- SVG: https://raw.githubusercontent.com/macports/macports-www/3e81994e41d62f6879ef4307610581823e6f75be/img/macports-avatar.svg
- Color: #1E79E9

afzaadkhan:
is anyone doing this?

Jukelyn:
I also found a logo [here](https://github.com/macports/macports-webapp/blob/2ee818a4be1d1fe73c368dd86f281735c5e2a423/app/static/images/macports-flat-logo.svg), I personally think this one is better, thoughts?

JonathanRosh:
I'm working on it.

PR #13043, opened 2025-04-13
Add MacPorts icon

JonathanRosh:
Hex value from the opened issue (checked and correct), Used Inkscape to vectorize the logo.
This is my first contribution to a real life project.


JonathanRosh:
- SVG source: https://github.com/macports/macports-www/blob/8de17ad57e97d5ff87b16c084cef36059fa6c2f4/img/macports-avatar.svg
- Hex color: `#1E79E9` from logo text

adamrusted (changes requested):
Thanks for this, @JonathanRosh - and congrats on your first contribution!
I can see our linter is failing because you've left the icon 'too precise'. I'm also seeing some slight differences between your icon (black) and the source icon (blue) below.
image

Any chance you could take another look? You'll need to pull changes down from GitHub before continuing, as I had to reformat the JSON file for you :)

JonathanRosh:
Hi, thank you very much for the appreciation and motivation! :) 
I have remade the icon from scratch in Inkscape, and made sure it is not too persice (optimized it in SVGOMG). I couldnt make it exactly like the original shape so It is not perfect but i hope it is good enough.

icons/macports.svg
adamrusted, line 1: Thanks for your continued effort @JonathanRosh - though I'm now seeing the below issues between the source SVG (blue) and your PR (black).

image


What editing software are you using? In something like Inkscape you can just merge / exclude the paths down to one, set the viewport, and then optimise without much issue.

JonathanRosh: Yes, I have used Inkscape, 
I have imported the original icon,  resized it to 24x24 px and using the trace-to-bitmaps i have created the new icon with  vector in the path element.  I couldnt get any better results this way (I have played a little bit with the properties trace-to-bitmap such as the threshold).

adamrusted: You shouldn't need to use trace-to-bitmap, as the icon is just comprised of 3 paths that can be joined/excluded from each other. The tracing functionality will always try and 'optimise' by screwing with the paths in the way we're seeing.

JonathanRosh: Thanks @adamrusted!  I see — I misunderstood and thought bitmap tracing was necessary.
I’ll go back to the original SVG and try to cleanly combine the existing vector paths using Inkscape’s path tools instead of tracing. That should preserve the original shape and avoid the artifacts you pointed out.
Thanks again for the guidance — I really appreciate your help on this, learning a lot!

JonathanRosh:
I've made the final adjustments based on the feedback. Let me know if there's anything else I should update — thanks again!

PeterShaggyNoble (changes requested):
Thanks for your continued work on this one, @JonathanRosh.

However, our linter is now reporting [multiple issues](https://github.com/simple-icons/simple-icons/actions/runs/14452089101/job/41625899325?pr=13043) with the SVG. Could you see if you can fix them, please. Let us know if you need any assistance.

prem-k-r:
Closes #11676

PeterShaggyNoble:
Closing due to inactivity.

PR #13054, opened 2025-04-14
Update OpenAI Data

prem-k-r:
Added ChatGPT as alias

### DIFF
--- a/_data/simple-icons.json
+++ b/_data/simple-icons.json
@@ -11412,7 +11412,14 @@
 		"title": "OpenAI",
 		"hex": "412991",
 		"source": "https://openai.com/brand",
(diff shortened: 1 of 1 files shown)


adamrusted:
Thanks for this @prem-k-r - though looking at the logo terms for OpenAI - we would need to change our logo to be the text representation, or remove. Their guidelines pretty specifically state we cannot use the 'bloom' as a standalone icon. Would you be willing to repurpose this as an 'update' PR to change to the 'OpenAI' text? We therefore wouldn't be able to have the AKA as the text logo is not the same as the 'ChatGPT' icon.

prem-k-r:
Ah ok, but they changed their icons early Feb, do we have permission to modify the icons?

PR #13182, opened 2025-05-15
Add Polygon Icon

prem-k-r:
image

**Issue:** closes -

**Popularity metric:**

image

### Checklist

### Description
Why did you pick the hex value?
- Original logo have gradient, used the dominating color which is also used throughout the website


PeterShaggyNoble:
🤔 Their parent company's [terms of use](https://www.valnetinc.com/en/terms-of-use) are a little contradictory, first stating:
> Valnet allows its original content to be made available for non-commercial reproduction with a mandatory credit to the Sites from which the content originates, including both a link to such Sites and **the brand logo displayed**.

And later:
> The name and the logo of each of the Sites constitute trademarks, trade names, service marks or logos (collectively, the “Marks”) belonging exclusively to Valnet. You are not authorized to use any such Marks.

But immediately after that:
> Nothing contained on the Sites shall be construed as granting, by implication or otherwise, any license or right to use the Marks without the explicit permission of Valnet in writing.

So, to err on the side of caution, I think we should try to contact them to seek clarification & permission, if you wouldn't mind doing that, @prem-k-r?

If we do get permission to proceed then, unless you found something that explicitly stated we couldn't, we should remove the wordmark and use just the icon. And I think we should use `#e90c59` from the website's stylesheet for the `hex`.

prem-k-r:
I’ve sent the email and will update here when I receive their response.

adamrusted:
Assuming you've not heard anything from them @prem-k-r - I'm going to close this PR for now. Should you hear from the brand, let us know here and we can re-open here.

PR #13192, opened 2025-05-16
Add Google Password Manager

prem-k-r:
googlepasswordmanager

**Issue:** closes -

**Popularity metric:**

image

### Checklist


MDJAVEED7:
can you check mines pr #13193

prem-k-r:
> can you check mines pr #13193

A maintainer can review, especially difference between original and modified svg and give their feedback

PeterShaggyNoble:
🤔 What do you think of a treatment along these lines, @prem-k-r? I'm not too certain it works well myself.

prem-k-r:
umm, how about something like:
image

PeterShaggyNoble:
Yep, that works better than my version 👍🏻 How would it look with 0.5dp cut-outs?

prem-k-r:
> Yep, that works better than my version 👍🏻 How would it look with 0.5dp cut-outs?

I used offset/outset for this one but I think it may not be the conventional way. and stroke doesn't seems good idea for this one.
Can you please guide me briefly how to do cut-outs in Inkscape if you don't mind?

prem-k-r:
I think we should use official monotone version instead of using cut outs

Earlier they used fill:
IMG_20250611_103714

IMG_20250611_103838

---

Now their current version is just using strokes:

IMG_20250611_103910

IMG_20250611_104113

prem-k-r:
image

image

`--google-grey-700` on css

prem-k-r:
Updated to 

image

prem-k-r:
lint failing for source link: `chrome://password-manager` as no `https`

[Original SVG](https://github.com/user-attachments/assets/d9fd3336-dfd8-475b-83d5-3c7bab0296c2)

PeterShaggyNoble:
🤔 I'm torn on this monochrome treatment. It's used in a menu item to represent passwords in general, so isn't really reflective of the Google Password Manager _branding_. But, then again, in the absence of  an official monochrome version... 🤷🏻‍♂️

Personally, my vote would go to your previous treatment. But let's see what some of the other @simple-icons/maintainers think.

If it's decided to run with the outlined version then we could use [Google Fonts Material Icons](https://fonts.google.com/icons) for the source as it looks like they're using the [VPN key icon](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:vpn_key:FILL@0;wght@400;GRAD@0;opsz@24&icon.size=24).

We should keep the colour as Google blue, though; the grey is just the styling of that menu, and it changes between light & dark modes.

prem-k-r:
That's true they haven't explicitly mentioned anywhere that's their brand icon, partially other than following, rest it indicated as passwords icon

image

image

they have used brand icons for lens, translate, and cast(?), but cant say for certain if Google password manager icon is their brand icon or just representation for passwords.

the vpn key icon is slightly different, it have sharp corners.

previous version with cut-outs, corners at ring were curve so have to manually straighten that's ok but I was seeing difference when performing difference operation on path in Inkscape

prem-k-r:
Hi @PeterShaggyNoble ,
I retried previous version, this time I used Affinity for difference, and rest using Inkscape, but I found out that when exporting as svg through Affinity it only keep 3 digits after decimals, so there are very small differences when zoomed in, let me know if it works.
Thanks

image

adamrusted:
@prem-k-r - you can edit the decimals that Affinity uses via `Edit -> Settings -> User Interface` - I quite often leave my 'pixel' decimals to 5 - so that I'm using the _most_ accurate version of the icons that we can.

prem-k-r:
> @prem-k-r - you can edit the decimals that Affinity uses via `Edit -> Settings -> User Interface` - I quite often leave my 'pixel' decimals to 5 - so that I'm using the _most_ accurate version of the icons that we can.

I have but that settings doesn't seems to propagating to export

PR #13203, opened 2025-05-16
Add Jotform

prem-k-r:
image

**Issue:** closes -

**Popularity metric:**

image

### Checklist


PeterShaggyNoble (changes requested):
SVG looks perfect, @prem-k-r 👍🏻 

However, the first colour listed in their [brand guidelines](https://www.jotform.com/resources/) is `#00A3E1` so we should probably use that, even though they don't seem to use it much. Or perhaps the orange? What do you think?

And, can you update the `source` to those brand guidelines and also add a `guidelines` entry with the same URL.

prem-k-r:
> SVG looks perfect, @prem-k-r 👍🏻
> 
> However, the first colour listed in their [brand guidelines](https://www.jotform.com/resources/) is `#00A3E1` so we should probably use that, even though they don't seem to use it much. Or perhaps the orange? What do you think?


They have mentioned
> The Jotform logo represents our brand and identifies Jotform’s product offerings — please do not distort, recolor, or reconfigure it in any way.

and used either white with orange bg or black for monotone as in:
image
image
image

So maybe black?

> And, can you update the `source` to those brand guidelines and also add a `guidelines` entry with the same URL.

Done

PeterShaggyNoble:
Oh, yeah, that's a good point: that line would actually prevent us from adding this one at all as we would be recolouring it (or, more accurately, removing the colour).

Would you like to try contacting them before we close this out, though, to see if they'd allow us to do that? And to seek clarification on the brand colour? We select the brand colour independently of any colours used in logos & icons (although it is often used in them, of course) but we ask our users to adhere to the brand guidelines we provide in our metadata.

prem-k-r:
Ah, ok sure

prem-k-r:
I’ve sent the email and will update here when I receive their response.

adamrusted:
Assuming you've not heard anything from them @prem-k-r - I'm going to close this PR for now. Should you hear from the brand, let us know here and we can re-open here.

PR #13225, opened 2025-05-17
Add ISRO

prem-k-r:
image

**Issue:** closes -

**Popularity metric:**

https://www.similarweb.com/website/isro.gov.in/#overview
image

### Checklist


prem-k-r:
I've updated the icon from the pdf, SVG: [ISRO](https://github.com/user-attachments/assets/fe4dc8f1-9af9-4c6f-b706-8e79d255c171)

For color I highly suggest to keep either `#f58220` or `#0089cf`

PeterShaggyNoble (changes requested):
👍🏻 Point taken on the colour.

There's been an ever so slight visual shift in some of the curves of the "isro" wordmark on the right, most likely a precision issue. And there are also a few excess points in  both wordmarks that you might be able to remove.

Having said that, though, I now wonder if we could get away with removing the wordmarks and just using the symbol. That they include the wordmarks in the site's favicon would suggest perhaps not. What do you think?

prem-k-r:
I hope it's correct now(?)

PR #13236, opened 2025-05-17
Add Wuxiaworld

prem-k-r:
image

**Issue:** closes -

**Popularity metric:**

image

### Checklist


PeterShaggyNoble (changes requested):
All looks good as-is. But the source SVG has a lot of (potentially) excess points; could you see if any of them could be removed, please, @prem-k-r?

prem-k-r:
A dummy question- in the best-case scenario, if they provide confirmation via email to allow usage, will this override ToS?

PeterShaggyNoble:
Yes, @prem-k-r, provided the confirmation comes from someone with the authority to make such an override?

@adamrusted, if you're referring to the line below, that reads to me as being specifically about the software they make available, and doesn't include the logo & trademarks.
> Permission is granted to temporarily download one copy of the materials (information or software) on Wuxiaworld™ for personal, non-commercial transitory viewing only.

PR #13254, opened 2025-05-19
Add JSTOR

prem-k-r:
image

**Issue:** closes -

**Popularity metric:**

image

### Checklist


prem-k-r:
Updated, hopefully fixed now
Also removed 30 nodes

adamrusted (changes requested):
I'm still seeing a difference between the PR version (black) and the source version (red). Knowing that Affinity sometimes presents weird issues, could another @simple-icons/maintainers take a look please?

image

PeterShaggyNoble:
Well, this seems to be a tricky little divil! While I'm not seeing the difference Adam pointed out above, but I am seeing significant differences throughout. Which leads me to suspect I may have gone wrong somewhere along the way in my attempts to recreate it for comparison. Let's get a few more eyeballs on it to see if we can get a consensus on if there are differences and, if so, what those differences are.

icons/jstor.svg
PratyushJayachandran, line 1: I am not able to see the differences pointed out by @adamrusted but as @PeterShaggyNoble said, there are differences throughout the icon when being compared to the original. Try increasing the precision to 4 decimal places.

Comparison attached
Comparison attached

LitoMore:
This should be a precision-related issue.

prem-k-r:
Thanks everyone for the review. This icon is quite time-consuming, and I don’t have the time or patience to continue working on it right now, so I’m abandoning it.

PR #13270, opened 2025-05-21
Fix fields-sort lint

LitoMore:
The `formatIconData()` is changing the same array. I've used [`structuredClone ()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone) to create a deep clone for resolving this issue.

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -170,7 +170,8 @@ const sortAlphabetically = (object) => {
  * @returns {IconData[]} The sorted icons data.
  */
 export const formatIconData = (iconsData) => {
-	const icons = iconsData.map((icon) => {
+	const iconsDataCopy = structuredClone(iconsData);
+	const icons = iconsDataCopy.map((icon) => {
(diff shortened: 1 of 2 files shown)


scripts/utils.js
```diff
@@ -170,7 +170,8 @@ const sortAlphabetically = (object) => {
  * @returns {IconData[]} The sorted icons data.
  */
 export const formatIconData = (iconsData) => {
-	const icons = iconsData.map((icon) => {
+	const iconsDataCopy = structuredClone(iconsData);
```
mondeja, line 173: Shouldn't this be?

```suggestion
	const iconsDataCopy = globalThis.structuredClone(iconsData);
```

LitoMore: XO didn't report this issue, so it should be fine. And we only used `node` runtime.

But both are OK for me. Feel free to apply your suggestion if you prefer the `globalThis.structuredClone`.

PR #13301, opened 2025-05-26
Add Java third-party library

silentsoft:
### Description

The library has been modified to support version v14.15.0 and has been released with the same version.
It was also configured today to automatically update every morning from Monday to Wednesday.

### Related Issue
- https://github.com/silentsoft/simpleicons4j/issues/1


PeterShaggyNoble:
Note: as our releases are on a Sunday, you should be safe updating on Monday's only.

PR #13352, opened 2025-05-30
Update Netflix icon

dhruvkb:
**Issue:** Not linked to any issue

**Popularity metric:**

Similarweb global rank 28

### Checklist

### Description

This PR only updates the branding source and guidelines for Netflix. The previous URL returns 404.


PeterShaggyNoble:
@dhruvkb, did you double check that our current icon still matches the assets available at the new source?

dhruvkb:
No I did not, I only updated the URL. However I did note a few things:

- We use a non-standard monochrome version of the icon that's not present on the brand guidelines page. That makes it harder to check if the icon is up-to-date.
- The guidelines mention that the icon "always appears in its signature color", which isn't true for Simple Icons.
- The guidelines also say "Don’t flatten to one color" which Simple Icons does by nature.

If we do want to follow the guidelines to the letter, we might have to drop support for the brand.

PeterShaggyNoble:
> The guidelines mention that the icon "always appears in its signature color", which isn't true for Simple Icons.
By including the brand guidelines in our JSON, we put it upon the user to ensure they use the icon with the colour.

> The guidelines also say "Don’t flatten to one color" which Simple Icons does by nature.
There are no cut-outs in the example they provide, while there are in our version, so I think we're OK there. What do you think @adamrusted?

dhruvkb:
I would not be able to update the icon unfortunately as I don't have the necessary skills for working with SVGs. I'm really sorry about that.

PeterShaggyNoble:
I've pushed an updated icon to your branch, but we'll need one of the other @simple-icons/maintainers to review this now.

LitoMore (approved):
@PeterShaggyNoble LGTM and feel free to merge it.

Issue #12314, opened 2024-12-10
Request: Mozilla VPN


DerBraky:
### Brand Name

Mozilla VPN

### Website

https://vpn.mozilla.org/

### Popularity Metric

- Similarweb Global Rank: [1,233](https://www.similarweb.com/website/mozilla.org) (Mozilla VPN is one of the seven products/brands featured on Mozilla's homepage)

### Official Resources for Icon and Color

- SVG: https://www.mozilla.org/media/protocol/img/logos/mozilla/vpn/logo.c648f487bfb8.svg
- Color: `#000000`, from icon

### Additional Comments

cannot be found at https://mozilla.design/

adamrusted:
Do we have a popularity metric for Mozilla VPN alone? I'm guessing the Similarweb is for the TLD of mozilla.org...

adamrusted:
Unfortunately then, this would be out of scope based on the GitHub and Similarweb metrics. Given one of the git repos is a mobile client, is there some way we can view download statistics, and compare against other mobile VPN apps we may already have in the collection?

LitoMore:
`vpn.mozilla.org` is just a domain for redirecting, so it will never have a rank.

It has 500K+ downloads on Google Play. I think this one can be considered.

See https://play.google.com/store/apps/details?id=org.mozilla.firefox.vpn&hl=en

PeterShaggyNoble:
Did we not decide a while back that total downloads on Google Play wasn't an accurate metric of real-time popularity? And, 5 months later, it's still only at 500k+. I'd say this is out of scope.

PeterShaggyNoble:
We have now added metrics of popularity for apps to our contributing guidelines and set an initial cut-off of being ranked in the top 1000 in any one country. Unfortunately, that currently puts Mozilla VPN outside our scope.

PR #13379, opened 2025-06-04
Add Metrics of Popularity for Apps

PeterShaggyNoble:
**Issue:** https://github.com/simple-icons/simple-icons/issues/12314#issuecomment-2912480134

### Description
This PR serves to, in order:
1. Further downplay the importance of GitHub stars as a measure of popularity by trying to stop people from using them as a first resort.
1. Draw the attention of our regular contributors to [Similarweb's browser extensions](https://www.similarweb.com/corp/extension/), which doesn't have any rate limiting - I installed it last Monday and it made a _huge_ difference to my workflow.
1. Most importantly, formalise metrics of popularity for Android & iOS apps using their Similarweb country ranks.
1. Reenforce that GitHub stars should only ever be considered as a last resort.
1. Require that the number of stars a project has should grow over the lifetime of the project.

### Points of Discussion
- Does the 1k cut-off for apps seem reasonable? If we take the UN's count of there being 195 countries then that gives us, potentially, 390k new brands that we'll be able to add. Although, obviously, there will be a lot of crossover between the 2 platforms as well as with websites.
- I know I was overruled on this the last time I suggested it, but I do feel strongly that we _should_ expect a project's GH stars to grow over time. Judging a 2-year-old project and a 10-year-old one by the same static 5k cut-off doesn't make sense to me.


CONTRIBUTING.md
```diff
@@ -83,8 +83,17 @@ If you are in doubt, feel free to submit it and we'll have a look.
...
+
 > [!TIP]\
-> Similarweb is now trying to force users to log in to view statistics. You can bypass this by going directly to `https://similarweb.com/website/google.com`, replacing `google.com` with the TLD you would like to get the stats on. You may need to use incognito mode if you visit these ranking pages often.
+> Similarweb is now trying to force users to log in to view statistics. You can bypass this by going directly to:
+>
+> - `https://similarweb.com/website/google.com` for websites, replacing `google.com` with the TLD you would like to get the stats on,
+> - `https://www.similarweb.com/app/google/com.android.chrome` for Android apps, replacing `com.android.chrome` with the Google Play package ID of the app you would like to get the stats on, or,
+> - `https://www.similarweb.com/app/apple/535886823` for iOS apps, replacing `535886823` with the numerical Apple Store ID of the app you would like to get stats on.
+>
```
PeterShaggyNoble, line 97: The linter is currently failing because of blank lines within blockquotes. But, if I remove this particular line then that breaks the intended formatting.

CONTRIBUTING.md
```diff
@@ -83,8 +83,17 @@ If you are in doubt, feel free to submit it and we'll have a look.
 
 ### Assessing Popularity
 
+> [!NOTE]\
+> Please make your way through the following metrics _in order_, only moving on to the next metric if the current metric is not applicable to the brand being assessed or if it would put the brand outside our scope. In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.
+
```
LitoMore: We can use an empty comment here as a workaround.

Or we can ignore this rule `MD028/no-blanks-blockquote`. @mondeja What do you think?

```suggestion
> [!NOTE]\
> Please make your way through the following metrics _in order_, only moving on to the next metric if the current metric is not applicable to the brand being assessed or if it would put the brand outside our scope. In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.

<!---->

```

CONTRIBUTING.md
mondeja: Add this to the beginning of the file:

```markdown
<!-- markdownlint-disable no-blanks-blockquote -->
```

The rule [`no-blanks-blockquote`](https://github.com/DavidAnson/markdownlint/blob/main/doc/md028.md) is intended to prevent between differences in Markdown parsers.

> Rationale: Some Markdown parsers will treat two blockquotes separated by one or more blank lines as the same blockquote, while others will treat them as separate blockquotes.

We are not distributing this file, which means that is primarily designed to be consumed by the GitHub UI.

CONTRIBUTING.md
PeterShaggyNoble: Tried that, Prettier still doesn't like it. Of course, Prettier's error reporting is useless, so maybe this isn't the problem at all 🤷🏻‍♂️

mondeja:
See https://github.com/simple-icons/simple-icons/pull/13379#discussion_r2128628265

CONTRIBUTING.md
PeterShaggyNoble: It's still not passing after adding that line, @mondeja.

CONTRIBUTING.md
```diff
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable no-blanks-blockquote -->
```
mondeja, line 1:
```suggestion
<!-- markdownlint-disable no-blanks-blockquote -->

```

CONTRIBUTING.md
mondeja: Run `npm run format`.

PeterShaggyNoble: I don't have npm

CONTRIBUTING.md
```diff
@@ -83,8 +85,18 @@ If you are in doubt, feel free to submit it and we'll have a look.
 
 ### Assessing Popularity
 
+> [!NOTE]\
+> Please make your way through the following metrics _in order_, only moving on to the next metric if the current metric is not applicable to the brand being assessed or if it would put the brand outside our scope. In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.
+
+<!---->
```
mondeja:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -83,8 +85,18 @@ If you are in doubt, feel free to submit it and we'll have a look.
 
 ### Assessing Popularity
 
+> [!NOTE]\
+> Please make your way through the following metrics _in order_, only moving on to the next metric if the current metric is not applicable to the brand being assessed or if it would put the brand outside our scope. In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.
+
```
mondeja:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -83,8 +85,18 @@ If you are in doubt, feel free to submit it and we'll have a look.
...
+<!---->
+
 > [!TIP]\
-> Similarweb is now trying to force users to log in to view statistics. You can bypass this by going directly to `https://similarweb.com/website/google.com`, replacing `google.com` with the TLD you would like to get the stats on. You may need to use incognito mode if you visit these ranking pages often.
+> Similarweb is now trying to force users to log in to view statistics. You can bypass this by going directly to:
```
mondeja, line 92:
```suggestion
> Similarweb is now trying to force users to log in to view statistics. You can bypass this by going directly to:
>
```

mondeja (changes requested):

CONTRIBUTING.md
mondeja: It's fixed.

PeterShaggyNoble: Thanks, @mondeja

CONTRIBUTING.md
```diff
@@ -83,8 +85,17 @@ If you are in doubt, feel free to submit it and we'll have a look.
 
 ### Assessing Popularity
 
+> [!NOTE]\
+> Please make your way through the following metrics _in order_, only moving on to the next metric if the current metric is not applicable to the brand being assessed or if it would put the brand outside our scope. In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.
```
mondeja: I feel this part too much verbose:

> In particular, please do not jump straight to GitHub stars as a measure of popularity as that is a last resort metric.

It seems to me that the emphasis _in order_ is enough. I would remove that sentence.

PeterShaggyNoble: Should we remove that note completely, then, do you think, @mondeja? My main reason for adding was specifically to emphasise that people shouldn't immediately resort to GitHub stars. Outside of that, I would imagine that most people would work their way through the list in order.

CONTRIBUTING.md
```diff
@@ -112,10 +126,13 @@ To be considered popular enough to be within our scope, a brand must be in exist
...
    - Trending equal to or higher than the compared brand will be considered in scope.
    - Trending lower than the compared brand but with an upward trajectory will require the consensus of the person providing the comparison and at least 2 project maintainers.
-1. Where applicable, the primary repository for the brand's GitHub project meets the following requirements (see note below):
-   - A minimum of 5k GitHub stars will be required for consideration, providing,
-   - The repository is still active, and,
-   - Its [star history](https://star-history.com) is on a consistently upward trajectory.
+
+As a last resort, after exhausting all other options above, and where applicable, the primary repository for the brand's GitHub project will be considered provided it meets the following requirements (see note below):
+
+- It is active and receiving regular commits,
```
uncenter:
```suggestion
- It is active and receiving regular commits,
```

Might make sense to link something like `https://github.com/<owner>/<repository>/graphs/contributors` (e.g. https://github.com/louislam/dockge/graphs/contributors) and/or mention that "commits" as referenced refers to _significant_ commits, not the monthly dep upgrade merges I sometimes see in dormant projects 😅

CONTRIBUTING.md
```diff
@@ -95,6 +106,9 @@ To be considered popular enough to be within our scope, a brand must be in exist
 1. The website's Similarweb rank in any one country is either:
    - In the top 100, or,
    - In the top 10k, _with_ a global rank of 1m or better.
+1. Its app's Similarweb country rank, in any one country, is in the top 1k for either Android or iOS.
```
uncenter, line 109: This feels very wide ranging, I'd favor a smaller number like top 100 since who knows what tiny apps are in use in smaller countries ahah. Top 1k makes sense for the United States or Canada or whatever but maybe less so in others?

CONTRIBUTING.md
PeterShaggyNoble: In hindsight, I shouldn't have added the bit about commits as it doesn't allow for projects like [MDI](https://github.com/Templarian/MaterialDesign) where our primary repo is for handling requests only, or ones where the primary repo is used for releases only with all development taking place in secondary repos, or - and this is one I came across recently - software that is considered stable and only receiving irregular maintenance updates. Also, as GitHub is a last resort metric, we can do a bit of additional assessment on a case-by-case basis.

CONTRIBUTING.md
PeterShaggyNoble: I wanted to mirror the point above about a website's country rank, which is there intentionally to allow for brands that are popular in smaller countries without being globally popular, but still with a cut-off for global rank.

As we don't have a global rank for apps, I thought 1k was more reasonable than 10k. If that turns out to be _too_ inclusive then we can easily reduce that further. And, of course, if we get requests for lower ranking apps in countries with extremely low populations, I'd say we can reject those.

100, I would say, is certainly too low, anyway.

uncenter (approved):

PR #13414, opened 2025-06-07
Add yt-dlp

prem-k-r:
image

**Issue:** closes #-

**Popularity metric:**

Github stars: [114k](https://github.com/yt-dlp/yt-dlp)

### Checklist

### Description

-


PeterShaggyNoble:
I think we might have to pull a [Popcorn Time](https://github.com/simple-icons/simple-icons/pull/1145) on this one and say that it falls into the category of an illegal service as it's intended to be used to download copyrighted material without going through the proper distribution and payment channels. And even if that's _not_ the intent and they explicitly state somewhere _not_ to use it for those purposes, it's definitely violating the terms of service of numerous sites & services. That its predecessor, YouTube-dl, was shut down after being issued a takedown notice the RIAA does not bode well for it.

Side note: does anyone else find the 2 year gap and then sudden 75k jump in its [star history](https://star-history.com/#yt-dlp/yt-dlp) suspicious?

PR #13426, opened 2025-06-08
Add Padlet

prem-k-r:
**Issue:** closes #

**Popularity metric:**

### Checklist

- [ ] I have reviewed the [forbidden brands](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#forbidden-brands) list and confirm the brand I am submitting a PR for is not one of them, nor is it a subsidiary of one of those brands
- [ ] I have reviewed the brand's terms of service, and am confident we can add this icon
- [ ] I updated the JSON data in `data/simple-icons.json`


PeterShaggyNoble (approved):
Nice work, @prem-k-r 👍🏻

I wasn't entirely sure of the colour choice when I visited their site and started looking around but then I found their [documentation](https://docs.padlet.dev/), where they use the same pink in the header and call it the primary colour in the stylesheet, so I think we're good.

PR #13429, opened 2025-06-08
Add license to OpenJDK Duke icon

Marcono1234:
### Description
Updates the data for the OpenJDK Duke icon based on https://wiki.openjdk.org/display/duke/Main:
- > Duke is the Java mascot, which was open sourced by Sun on November 13, 2006 under the New BSD license.
- ~~That page also mentions some "guidelines", therefore I have added is as `guidelines`~~
  > All we ask is that you treat Duke with the same respect that Sun (and now Oracle) has.
  > Note that you do not need to include the BSD license text to include Duke images on your website.

I hope I understood the wiki page correctly and it really applies to the image files, and not just the mascot on its own (?).


PeterShaggyNoble (changes requested):
Thanks for this, @Marcono1234 👍🏻 I don't think we need the `guidelines`, though, as they're not _specifically_ related to the OpenJDK brand or any one version of the icon, rather they relate to granting people permission to create their own versions of the mascot.

Could you also remove the `#l1` from the end of the source URL while you're at it, please; that shouldn't have slipped through when this was added.

Marcono1234:
Have removed the `guidelines` entry again, and adjusted the `source` URL.

PR #13455, opened 2025-06-12
Add lint in Husky

cscnk52:
### Description

This Pull Request aim to add lint script in `pre-commit` hook, it checks icon before commit, reduce reviewer pressure, ensure common mistake founded before commit.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,3 @@
 npm run format
+npm run lint
 git add -u


mondeja:
We didn't added that because SVGLint script is still slow, taking ~5s in my local environment. But at the moment we decided to not add it was so much slower, taking ~8s even with less icons in the library. 

I've added an optimization to the CI that only runs SVGLint for changed icons, which is blazingly fast. See:

https://github.com/simple-icons/simple-icons/blob/31244ac00f22d47738a00752e2f4d2d41fb27316/.github/workflows/verify.yml#L75-L76
https://github.com/simple-icons/simple-icons/blob/31244ac00f22d47738a00752e2f4d2d41fb27316/.github/workflows/verify.yml#L85-L88
image


The underlying trick is that `$npm_config_icons` will take the value of `--icons` when `npm run lint --icons=...` is executed. We can probably add the same trick to the Husky script.

Thoughs @LitoMore? Are you willing to add it to the Husky script @cscnk52?

cscnk52:
I think checking only the changed files is a better solution, but it will take more time to implement.

LitoMore:
> This Pull Request aim to add lint script in pre-commit hook, it checks icon before commit, reduce reviewer pressure, ensure common mistake founded before commit. (@cscnk52)

I would keep our dev experience fast and straightforward. It's also annoying every time you commit the code to run the lint.

We already have `pretter --check` and `xo --fix` in the `npm run format` script, but there are still lots of contributors sending broken PRs to us.

> I've added an optimization to the CI that only runs SVGLint for changed icons, which is blazingly fast. (@mondeja)

The solution looks better.

mondeja:
Yes, I agree with @LitoMore that the pre-commit hook should be fast. Mainly because is a pain to run the whole linting process every time each person wants to submit an icon.

Thanks for trying and bringing this up for discussion again @cscnk52. If you're not willing to implement the fast solution, please, close this PR and I'll open an issue to track the improvement.

cscnk52:
Thanks, if I find how to only check added file, I will open this PR again, now just temporary close it.

PR #13466, opened 2025-06-13
Require `// @ts-check` at the beginning of JS files

mondeja:
Continuation of #12840. This adds linting to ensure that all JS files include a `// @ts-check` comment at the beginning.


tests/min-reporter.cjs
```diff
@@ -16,7 +17,10 @@ class EvenMoreMin extends reporters.Base {
 	 */
 	constructor(runner) {
 		super(runner);
-		runner.once(EVENT_RUN_END, () => this.epilogue());
+		runner.once(EVENT_RUN_END, () =>
+			// @ts-expect-error
```
LitoMore: Let's add a description about this line to introduce what error was ignored and why.

```
// @ts-expected-error: Xxx
```

mondeja: I've realized that we weren't using `@types/mocha`. Added and the problem is solved now.

LitoMore (approved):
LGTM. The PR conflict needs to be resolved before merging.

PR #13481, opened 2025-06-16
Update Material Design icon

prem-k-r:
image

**Issue:** closes #-

**Popularity metric:**

Similarweb [Global Rank 35,727](https://www.similarweb.com/website/material.io/#overview)

### Checklist

- [ ] I have reviewed the brand's terms of service, and am confident we can add this icon

### Description

Current icon is mainly used for Material Design 2, updated that of MD3

Their ToS same as Google's, so no need to seek permission?

Color: From website's stylesheet `--mio-theme-color-primary: #6442d6;`
But the icon itself is `#1c1b1d` from `--mio-theme-color-neutral-10: #1c1b1d;`, so a bit of doubt on color choice

SVG used: favicon- https://m3.material.io/static/assets/m3-favicon.svg
On the website they used Google Symbols font's M glyph with some styles to generate the icon


prem-k-r:
Other color option:

image

PeterShaggyNoble (changes requested):
SVG looks good, @prem-k-r 👍🏻 Although, there are 2 overlapping points at the bottom of the inside of the circle; we can remove one of them. I tried removing the other seemingly excess points on the inside of the circle, but it introduced some _tiny_ differences.

I agree with you on the choice of purple for the colour. But, seeing as the primary colour on the site changes between light & dark mode, I went looking to see if I could find confirmation of which one was _actually_ the primary, and I found [this](https://m3.material.io/styles/color/static/baseline), where "baseline color scheme" is defined [elsewhere](https://m3.material.io/styles/color/system/overview) as:
> ... the default static color scheme for Material products.

Running with the default settings for the colour definitions on that page, I think we should go with `#6750A4`.

Finally, outside of specific exceptions (e.g., our PlayStation icons), we try to avoid including versioning wherever possible, so I'd say we should run without the alias here.

prem-k-r:
@PeterShaggyNoble done

PeterShaggyNoble (approved):

PR #13510, opened 2025-06-20
Add type declarations for data JSON entry

LitoMore:
This PR adds a declaration file for `simple-icons/data.json`.

I've copied the `IconData` from `sdk.d.ts` to `types.ts`. We can consider removing the `IconData` from the `sdk.d.ts` in the next major release.

The `icons.json.d.ts` has two types of default export `export default Icons` and `export = Icons;` because we need to support both ESM and CommonJS.

I also polished our export entries, those shared declaration files can be defined at the same level as the export conditions.


scripts/build/templates/icons-json.d.ts.template
```diff
@@ -0,0 +1,6 @@
+import type {IconData} from '../types.js';
```
mondeja: I think that this doesn't require a `.template` file because is not a true template as the file is rendered as is. Several options here:

1. Remove `.template` suffix from the file name so we can have proper syntax highlighting, lint, etc.
2. Put it into the *data/* folder as a regular TS file. The problem with that approach is that it could confuse contributors, as is not a data file.
3. Put it into the root of the project as a regular TS file.
4. Render it from a string directly from the build script.

My preference is, in this order: 3, 2, 1 and 4.

mondeja: You probably want to import it from `../types.d.ts` directly.

```suggestion
import type {IconData} from '../types.d.ts';
```

mondeja (changes requested):
> We can consider removing the `IconData` from the `sdk.d.ts` in the next major release.

Meanwhile, please, re-export the type from *sdk.d.ts*. Only change the imports line to:

```typescript
import type {CustomLicense, IconData, SPDXLicense} from './types.d.ts';
```

And later

```typescript
// TODO: remove re-export in v16
export type {IconData} from './types.d.ts';
```

mondeja (changes requested):
Change the typedef in *sdk.mjs* to:

```diff
-  * @typedef {import("./sdk.d.ts").IconData} IconData
+ * @typedef {import("./types.d.ts").IconData} IconData
```

scripts/build/templates/icons-json.d.ts.template
LitoMore: See our `sdk.d.ts` file, it's using `types.js`. I just used the same format as our existing file.

LitoMore:
I would prefer not to import anything from the SDK part since our `simple-icons/index` and `simple-icons/sdk` are targeting in the different Node.js versions.

mondeja:
> I would prefer not to import anything from the SDK part since our `simple-icons/index` and `simple-icons/sdk` are targeting in the different Node.js versions.

But TypeScript types are not part of Node.js anyways. Note that typedef are just comments, nothing relevant in the context of package.json `engines`.

scripts/build/templates/icons-json.d.ts.template
mondeja: Rewrite the import in `sdk.d.ts` too. I think that we were importing from `types.js` due to a bug in TypeScript, now solved.

mondeja: Any reason why this was resolved, @LitoMore?

LitoMore: I've updated the code according to your suggestion on my local.

Sorry for confusing you. I will pick this PR up when I have time.

data/simlpe-icons.d.ts
```diff
@@ -0,0 +1,10 @@
+// @ts-check
```
mondeja: There is a typo in the file name.

scripts/lint/ourlint.js
```diff
@@ -6,7 +6,7 @@
  */
 
 /**
- * @typedef {import("../../sdk.mjs").IconData} IconData
+ * @typedef {import("../../types.js").IconData} IconData
```
mondeja, line 9: Import from a `.d.ts` file here, please.

scripts/utils.js
```diff
@@ -11,8 +11,8 @@ import path from 'node:path';
 import {collator, getIconSlug, getIconsDataPath, titleToSlug} from '../sdk.mjs';
 
 /**
- * @typedef {import("../sdk.js").IconData} IconData
- * @typedef {import("../sdk.js").DuplicateAlias} DuplicateAlias
+ * @typedef {import("../types.js").IconData} IconData
+ * @typedef {import("../types.js").DuplicateAlias} DuplicateAlias
```
mondeja, line 15: Here too.

sdk.mjs
```diff
@@ -8,7 +8,7 @@ import path from 'node:path';
 
 /**
  * @typedef {import("./sdk.d.ts").ThirdPartyExtension} ThirdPartyExtension
- * @typedef {import("./sdk.d.ts").IconData} IconData
+ * @typedef {import("./types.js").IconData} IconData
```
mondeja, line 11: And here. I think that this PR should add these rewrites to importing all corresponding types from `.d.ts` files.

mondeja (commented):
As the source of the type `IconData` is _data/simple-icons.json_, why not moving the type to _data/simple-icons.d.ts_?

scripts/lint/ourlint.js
LitoMore: I can't. The tslint will throw an error. You can try it on your local.


```
scripts/lint/ourlint.js:9:21 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file '../../types.js' instead?

9  * @typedef {import("../../types.d.ts").IconData} IconData
                      ~~~~~~~~~~~~~~~~~~

scripts/lint/ourlint.js:10:21 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file '../../types.js' instead?

10  * @typedef {import("../../types.d.ts").CustomLicense} CustomLicense
                       ~~~~~~~~~~~~~~~~~~

Found 2 errors in the same file, starting at: scripts/lint/ourlint.js:9
```

LitoMore:
> As the source of the type IconData is data/simple-icons.json, why not moving the type to data/simple-icons.d.ts?

Good point. How about moving all other related types (Aliases, SpdxLicense, etc) to the file?

mondeja:
> > As the source of the type IconData is data/simple-icons.json, why not moving the type to data/simple-icons.d.ts?
> 
> Good point. How about moving all other related types (Aliases, SpdxLicense, etc) to the file?

Yes, please.

scripts/lint/ourlint.js
mondeja: Ah OK!

mondeja (approved):
Thanks!

PR #13533, opened 2025-06-23
Enforce consistency of titles and slugs

mondeja:
Add linting for slugs and titles. The enforcements are:

- Titles must be unique from other titles of icons that don't have a slug.
- Custom slugs must be necessary (this was a test but now moved to a lint with a proper error message).
- Custom slugs must be normalized, but underscore characters (`_`) are allowed.
- Custom slugs must have their corresponding icon.
- Custom slugs must be unique between custom slugs and those inferred from titles.

This ensures that consumers of simple-icons can rely in a safe prefix to differentiate between simple-icons and other icon sets. For example, after this change, slugs can't have the character `+`, so consumers can safely add a suffix like `si+` to slugs. See the next concern at https://github.com/badges/shields/issues/11183 as an use case:

> It could also present potential new issues like name collision.


scripts/lint/ourlint.js
```diff
@@ -333,6 +341,106 @@ ${invalids.map((icon) => `${format(icon)} ${findPositon(expectedOrder, icon)}`).
...
+			if (slug === undefined) {
+				continue;
+			}
+
+			// Custom slugs must be necessary. If a title is converted to a slug
+			// and it is the same as the slug, then it is not necessary.
+			if (titleToSlug(title) === slug) {
+				errors.push(
+					`Icon "${title}" has a slug "${slug}" that is the same as the slug inferred from its title.` +
+						' Please, remove the slug or change the title.',
```
PeterShaggyNoble: We probably shouldn't be inviting people to change the `title` of an icon as it could be implied that altering a brand's name is permissible.

LitoMore (approved):

PR #13544, opened 2025-06-24
Add Odido icon

### LINKED ISSUE #12442: Request for adding Odido
### Brand Name

Odido

### Website

https://www.odido.nl/

### Popularity Metric

The similar web rank is 7,865 https://www.similarweb.com/website/odido.nl

Odido is the new T-Mobile Rebrand for the Netherlands

Here are the official Brand colors
https://merk.odido.nl/en/colour

The Logo is divided into 3 types https://merk.odido.nl/en/logo

The Color logo Using (Glow One, Glow Two and Glow Three and Glow Four from the color palette)
odido-logo 1
Original Source https://www.odido.nl/Assets/Lfwd/Static/odido-logo.svg

The Solid Black logo
odido-logo-black 1

Original Source https://www.odido.nl/Assets/Lfwd/Static/odido-logo-black.svg

The Solid White logo
odido-logo-white 1

Original Source https://www.odido.nl/Assets/Lfwd/Static/odido-logo-white.svg

### Additional Comments

The Color Logo is the official logo used everywhere but the other two alternatives are accepted by Odidio in general

xxsuperdreamxx:
odido

**Issue:** closes #12442

**Popularity metric:** Source https://www.similarweb.com/website/odido.nl/
| Global Rank | Country Rank |
|-------------|--------------|
| 8,220       | NL: 117      |

### Checklist


data/simple-icons.json
```diff
@@ -11382,6 +11382,20 @@
 		"source": "https://en.facebookbrand.com/oculus/assets/oculus?audience=oculus-landing",
 		"guidelines": "https://en.facebookbrand.com/oculus/"
 	},
+	{
+		"title": "Odido",
+		"hex": "000000",
+		"source": "https://www.odido.nl/Assets/Lfwd/Static/odido-logo-black.svg",
```
PeterShaggyNoble: For better future proofing, we use the URL of the page the source file was found but, in this case, the logo page of the brand guidelines works better.
```suggestion
		"source": "https://merk.odido.nl/en/logo",
```

data/simple-icons.json
```diff
@@ -11382,6 +11382,20 @@
...
+		"guidelines": "https://merk.odido.nl/en/logo",
```
PeterShaggyNoble: We should link to the full brand guidelines here.
```suggestion
		"guidelines": "https://merk.odido.nl/en",
```

data/simple-icons.json
```diff
@@ -11382,6 +11382,20 @@
...
+		"aliases": {
+			"aka": [
+				"ODIDO"
```
PeterShaggyNoble: This alternative casing alias is unnecessary. And I also don't see them referring to themselves this way anywhere.

data/simple-icons.json
```diff
@@ -11382,6 +11382,20 @@
...
+			],
+			"old": [
+				"T-Mobile"
```
PeterShaggyNoble: We only use the `old` alias when updating existing icons in our library.

PeterShaggyNoble (changes requested):
I can't get you version to line up with any of the ones they make available to download through their brand centre, @xxsuperdreamxx. We should use one of the EPSes available in the `Millen_URL_SolidLogo_SmallUseALL` zip found on [this page](https://dam.odido.nl/web/52967ad9a20e29eb/b2c-logos---downloads-brand-portal/?viewType=grid). See also notes below on metadata.

data/simple-icons.json
xxsuperdreamxx: Got it, since people still Refer to Odido as T-Mobile that's why

data/simple-icons.json
xxsuperdreamxx: Mostly because people think Odido logo is capitalised, even though it is just an abstract design

data/simple-icons.json
xxsuperdreamxx: I did the logo, specifically because they explained all of their guidelines and preferences for the logo, specifically on how to handle. since the main merk's is also for their fonts and their advertisement blocks

data/simple-icons.json
xxsuperdreamxx: Noted, I will use the logo from the ESP file, but this is important, the ESP is using CMYK, should I make it RGB/8 @PeterShaggyNoble

xxsuperdreamxx: Both of them are 
image

PeterShaggyNoble: No need, no; the final SVG will be colourless. And the colour for our `hex` can be found at https://coder.together.agency/#6-primary-color-palette

xxsuperdreamxx: Cool, then I will update the pull request with their version., Additionally, my version differed because I converted their layer into a single curve. ~~I will keep their curves separated~~

PeterShaggyNoble: Sorry, that link above should, of course, be https://merk.odido.nl/en/colour

adamrusted:
Just spotted on the source page:

> Minimum size of the small use logo is 60px or 16mm.

We cannot guarantee the minimum size will be used, and in fact we store images at 24px - so we are going against their terms by storing the icon in that style.

CC @PeterShaggyNoble

xxsuperdreamxx:
> Just spotted on the source page:
> 
> > Minimum size of the small use logo is 60px or 16mm.
> 
> We cannot guarantee the minimum size will be used, and in fact we store images at 24px - so we are going against their terms by storing the icon in that style.
> 
> CC @PeterShaggyNoble

If you mean in the preview in the icon list in simpleicons? I can try ask them, but the guidelines are for all the partners and users who will advertise themselves as Odido specialists/partners/solutionists.  Or, I can also email them and CC you guys too

adamrusted:
@xxsuperdreamxx - in order for us to include the icon, it needs to be transformed to a 24x24 shape, which in itself goes against their guidelines. The primary use-case of our project is also badges like shields.io - which display at less than 24x24. I would argue therefore that we couldn't add this into our project knowing that we'd be breaching their guidelines. I've tagged another maintainer for review on that line of thought though.

PeterShaggyNoble:
@adamrusted, have we not worked around that in the past through our inclusion of brand guidelines and requirement that users read & adhere to them? If they don't then that's on them not us. And I wouldn't consider Shields.io to be exempt from that requirement.

PR #13530, opened 2025-06-22
Deprecate `simple-icons/icons` entrypoint

mondeja:
Contributes to #13520

PR #13545, opened 2025-06-24
Remove `simple-icons/icons` entrypoint

### LINKED ISSUE #13520: Remove `simple-icons/icons` entrypoint from npm package
We have 2 entrypoints for the same files in the npm package, `simple-icons` and `simple-icons/icons`. Which means that we've two ways to import all the icons:

```javascript
import * as simpleIcons from 'simple-icons';
import * as simpleIcons from 'simple-icons/icons';
```

This is not ideal because it adds additional mental workload for developers. I myself had to check again if they were two different entrypoints a while ago.

The `simple-icons/icons` entrypoint should be deprecated as is not currently documented. It's just legacy code from 4 years ago, when the package had a `.Get` method in the API:

```javascript

// Get a specific icon by its slug as:
// simpleIcons.Get('[ICON SLUG]');

// For example:
const icon = simpleIcons.Get('simpleicons');
````

That was the reason to add the `simple-icons/icons` entrypoint in #6767. Discussion at #6789 and replacing at #8058.

Now we should add the deprecation of the `simple-icons/icons` entrypoint. And remove it completely, probably, in two major versions at v17.

BTW, some projects are using the export in OSS, but not so much, see https://github.com/search?q=simple-icons%2Ficons+language%3AJavaScript+OR+simple-icons%2Ficons+language%3ATypeScript+OR+simple-icons%2Ficons+language%3ATSX&type=code&p=2

mondeja:
Deprecated at #13530. Closes #13520

Marked as draft. This should be merged just before v17.0.0 release.


LitoMore (approved):

LitoMore:
Ah, we targeted this PR to v17. Let's keep it draft until the v17 release.

Issue #13508, opened 2025-06-20
Request: Longhorn


atoy3731:
### Brand Name

Longhorn

### Website

https://longhorn.io

### Popularity Metric

The similarweb rank for RKE2 (via Rancher, the company who maintains it) is: 190,681

https://www.similarweb.com/website/rancher.com

### Forbidden Brands

- [x] I have reviewed the list of [forbidden brands](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#forbidden-brands) and can confirm the brand I am requesting is not one of them, nor is it a subsidiary of one of them.

### Official Resources for Icon and Color

Branding Guidelines: https://rancher.com/brand-guidelines/

### Additional Comments

NOTE: RKE2 is maintained by Rancher, which is already an approved icon on the list.

PR #13553, opened 2025-06-26
Remove Potential Duplicates workflow

mondeja:
[`wow-actions/potential-duplicates`](https://github.com/wow-actions/potential-duplicates) is not maintained anymore. I think that, at some point, the GitHub API started paginating issues and the Octokit definition that the action is using is now outdated. This means that the action is only checking the last 30 PRs and issues to check for duplicates. 

That's the reason why, e.g., [this run](https://github.com/simple-icons/simple-icons/actions/runs/15785363349 ) didn't marked the related issue as duplicated and  it was marked manually later by @PeterShaggyNoble at https://github.com/simple-icons/simple-icons/issues/13508

We need to move to a new action, but meanwhile this must be removed. It only brings noise to our pipeline.


PeterShaggyNoble (approved):
In addition to the problems outlined above: Having been through _hundreds_ of issues in the past few weeks, it's apparent that the automated testing for duplicates that are anything other than an exact match produces far too many false positives. So, whatever we replace this with, I'd suggest that we don't have it apply the https://github.com/simple-icons/simple-icons/labels/duplicate label - it's less work to apply it ourselves then it is to remove it from the false positives.

PR #13567, opened 2025-06-30
Use node to extract SVGLint command

PratyushJayachandran:
The pre-commit hook is failing on Windows system as some commands within the script are not available on Windows.
image

1. This change is to use node instead, so that it works on all platforms
2. Removal of staging of un-staged files. The user must have already staged the files they want to commit before the hook is run so this is an unnecessary step in my opinion.


mondeja:
> 2\. Removal of staging of un-staged files. The user must have already staged the files they want to commit before the hook is run so this is an unnecessary step in my opinion.

`git add -u` only adds changes to existing files on the staging area that have been edited by `npm run format`, it does not add new files.

I've reverted the removal.

mondeja (approved):

PR #13779, opened 2025-09-07
Update AssemblyScript color

cscnk52:
### Description

From https://www.assemblyscript.org/, the true color should be `#007ACC`


mondeja (approved):
Thanks!

Issue #80, opened 2013-03-11
jSON and XML


martijn94:
I don't really know if there are "official" icons for these two languages. But I would love to have icons for those 2!

Regards,

Martijn

Issue #79, opened 2013-03-11
Xbox 360


samuelfine:
The original: http://deadeyemodz.com/product_images/uploaded_images/xbox360logo1.jpg

A flat white version of the sphere with the "X" as negative space would be perfect. Really fantastic icon set!

PR #13828, opened 2025-09-17
Update color for Lua

Kraggle09:
### Description
The current color for the Lua logo is #00007C, but the color listed on Simple Icons is #2C2D72. This PR fixes that.

The color is sourced from this page:
[Lua: Logos](https://www.lua.org/images/)

### Preview:
**Old Color:**  
lua (2)
**New Color:**  
lua

### DIFF
--- a/data/simple-icons.json
+++ b/data/simple-icons.json
@@ -9627,7 +9627,7 @@
 	},
 	{
 		"title": "Lua",
-		"hex": "2C2D72",
+		"hex": "00007C",
 		"source": "https://www.lua.org/images/",
 		"guidelines": "https://www.lua.org/images/"
 	},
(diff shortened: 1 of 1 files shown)


data/simple-icons.json
```diff
@@ -9627,7 +9627,7 @@
 	},
 	{
 		"title": "Lua",
-		"hex": "2C2D72",
+		"hex": "00007C",
```
LitoMore:
```suggestion
		"hex": "000080",
```

The color picker is inaccurate because the value depends on your monitor display profile.

I checked the color with my color picker, and it returns `#000079`. The `#000080` is from `Logos` text on the page https://www.lua.org/images. It should be the accurate one.

LitoMore (approved):

PR #13846, opened 2025-09-23
Remove `permission in review` when closing PRs/issues

LitoMore:
Here is an example: https://github.com/simple-icons/simple-icons/issues/13832.

~~We don't need a https://github.com/simple-icons/simple-icons/labels/permissions%20in%20review label for autoclose issues. I've changed the labeling method to set.~~

~~The https://github.com/simple-icons/simple-icons/labels/new%20icon label is not needed either since we have the https://github.com/simple-icons/simple-icons/labels/won%27t%20add label.~~

### DIFF
--- a/.github/workflows/remove-labels.yml
+++ b/.github/workflows/remove-labels.yml
@@ -34,6 +34,7 @@ jobs:
             invalid
             out of scope
             pending
+            permission in review
             permission required
             won't add
 
@@ -56,6 +57,7 @@ jobs:
             in discussion
             pending
             assessing
+            permission in review
 
   remove-closed-issue-labels:
     name: Remove closed issue labels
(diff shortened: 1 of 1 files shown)


LitoMore:
Both are OK for me.

mondeja:
I think that we're probably missing something here... Why about adding the https://github.com/simple-icons/simple-icons/labels/permissions%20in%20review label to _.github/workflows/remove-labels.yml_?

adamrusted:
Agree with @mondeja we should be removing the https://github.com/simple-icons/simple-icons/labels/permissions%20in%20review label as part of the Remove Labels script. The other permission ones can be helpful for us to keep as reminder of where people got to with the requests.

LitoMore:
Updated

adamrusted (approved):

PR #13854, opened 2025-09-25
Add CSS3 icon

LitoMore:
image

### Checklist

### Description

We can still see some use cases for using both HTML5 and CSS3. This PR brings the CSS3 icon back to our collection.


data/simple-icons.json
```diff
@@ -3757,6 +3757,11 @@
 		"hex": "F43059",
 		"source": "https://csswizardry.com"
 	},
+	{
+		"title": "CSS3",
+		"hex": "1572B6",
+		"source": "https://www.w3.org/html/logo/"
```
adamrusted, line 3763: Where on this page is the CSS3 badge shared? I can see a number of mentions of just the `3` - but not as it's own badge...

LitoMore: I just used the old link. I can't find any CSS3 badges either.

adamrusted: If we're re-adding, we could do with a legitimate source for this 'addition'.

PratyushJayachandran: https://www.w3.org/policies/logos/#more-logos
This is where it should have been 😞

adamrusted: We'll need another official source then that can provide an SVG for us. Unfortunately without that, we may not be able to re-add.

LitoMore:
I cannot find any official resources about the CSS3 logo.

I'm closing this PR. If anyone can find an official resource for CSS3, feel free to create a new PR for it.

PR #14053, opened 2025-11-25
Remove Stadia

cscnk52:
Stadia was shut down on January 18, 2023. See: https://stadia.google.com/gg/


LitoMore:
See https://github.com/simple-icons/simple-icons/pull/11360.

PR #14231, opened 2026-01-11
Fix trailingSlash not work

cscnk52:
This PR fix trailingslash not work as expect and fix trailingslash.


mondeja:
I commited the same mistake implementing it 😅

PR #14245, opened 2026-01-16
Replace husky by prek

mondeja:
Instead of trying to add checks in pre-commit hook with Shell scripts with Husky, which is not scalable, use a *.pre-commit-config.yaml* file to configure some hooks to be executed when certain files are edited.

- This is possible with the addition of `--git-changed` option to SVGLint, see https://github.com/simple-icons/svglint/pull/142
- Prek is available as npm package since https://github.com/j178/prek/pull/819

In order to avoid dependency duplication nor slow down the first execution, we should not use third party repos in the config file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
 			"license": "CC0-1.0",
 			"devDependencies": {
 				"@inquirer/prompts": "7.10.1",
-				"@j178/prek": "0.2.29",
+				"@j178/prek": "0.2.30",
 				"@types/mocha": "10.0.10",
 				"@types/node": "24.10.1",
 				"chalk": "5.6.2",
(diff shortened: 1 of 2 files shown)


cscnk52:
The Windows compatibility issue has been fixed in the development version; we’re just waiting for the next prek release.

I have two minor nitpicks:

1. The commands `xo`, `svglint`, and `markdownlint-cli2` require global installation. I think it would be better to run them via the corresponding npm scripts instead.
2. prek v0.2.29 supports file glob patterns for filtering files. I think using glob patterns would be better than regex expressions. see https://github.com/j178/prek/releases/tag/v0.2.29

mondeja:
> The Windows compatibility issue has been fixed in the development version; we’re just waiting for the next prek release.

Now prek v0.2.30 is released with the change. Thank you for taking your time to catch it!

> I have two minor nitpicks:
> 
> 1. The commands `xo`, `svglint`, and `markdownlint-cli2` require global installation. I think it would be better to run them via the corresponding npm scripts instead.
> 2. prek v0.2.29 supports file glob patterns for filtering files. I think using glob patterns would be better than regex expressions. see https://github.com/j178/prek/releases/tag/v0.2.29

Resolved in https://github.com/simple-icons/simple-icons/pull/14245/commits/b0d6ae63024d81735c01bbf82aab557c8665a6f2, but I used `npx` instead because `npm run` is slow.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - repo: local
+    hooks:
+      - id: format-icon-data
+        name: Format icon data
+        entry: npm run format:icondata
+        language: system
+        pass_filenames: false
+        files:
+          glob:
+            - data/simple-icons.json
```
j178, line 11: nit:

```suggestion
          glob: data/simple-icons.json
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,51 @@
...
+        types: [json]
+      - id: svglint
+        name: SVG Lint
+        entry: npx svglint --ci --config svglint.config.mjs --git-changed
+        pass_filenames: false
+        language: system
+        types: [svg]
+        files:
+          glob:
+            - icons/*.svg
```
j178, line 43:
```suggestion
          glob: icons/*.svg
```

LitoMore:
@j178 Nice to meet you here. I remember we met on Fanfou.

PR #14248, opened 2026-01-17
Fix rendering of README table icons in released packages

mondeja:
Add a `&nbsp;` space after icons that are inside a HTML link and placed before the text that the link contains. In theory, this fixes the bad rendering in npmjs and Packagist:

image

Should render like:

image


mondeja:
Well, that worked on npmjs but not on Packagist 🤔

PR #14299, opened 2026-01-29
Added About You Icon

mariavitkou:
I added About You , which is a modern European fashion and technology company focused on personalized online shopping. It combines e-commerce, data, and fashion brands to deliver a tailored fashion experience through its app and website.


cscnk52:
Low quality and not follow Pull Request template, close.

PR #14336, opened 2026-02-09
Migrate to prek.toml

cscnk52:
Migrate to `prek.toml` in https://github.com/simple-icons/simple-icons/pull/14245, `file.glob` was used, it's prek-only feature, so migrate to new config file.

Related to:
- https://github.com/j178/prek/issues/593
- https://github.com/simple-icons/simple-icons/pull/14245


prek.toml
```diff
@@ -0,0 +1,60 @@
...
+name = "Format icon data"
+entry = "npm run format:icondata"
+language = "system"
+pass_filenames = false
+files.glob = "data/simple-icons.json"
+
+[[repos.hooks]]
+id = "prettier"
+name = "Prettier"
+entry = "npm run prettier -- --write"
```
LitoMore: How about using `node --run` instead?

FYI: https://nodejs.org/en/learn/command-line/run-nodejs-scripts-from-the-command-line

LitoMore (approved):

PR #14338, opened 2026-02-10
Use `node --run` for scripts

LitoMore:
I replaced all `npm run` with `node --run` for better performance in CI/CD.

Note that those scripts in the documentation are still `npm run` for better compatibility with older Node.js users.


.github/workflows/verify.yml
```diff
@@ -121,4 +123,4 @@ jobs:
       - name: Install dependencies
         run: npm ci --no-audit --no-fund
       - name: Run tests
-        run: npm run test
+        run: node --run test
```
mondeja, line 118: `node --run test` is not executing the `pretest` script, hence the index.mjs is not built

Please, add a step to build before this.

PR #14358, opened 2026-02-14
Replace Prettier by Oxfmt

cscnk52:
This PR replaces Prettier with Oxfmt. Oxfmt is about 2× faster than Prettier on warm runs. On cold start, Prettier takes 4.366s, making it nearly 9× slower.

Oxfmt also sort package.json by default, i think it's good.

Tested using [hyperfine](https://github.com/sharkdp/hyperfine):

```
❯ hyperfine "node --run oxfmt"
Benchmark 1: node --run oxfmt
  Time (mean ± σ):     528.1 ms ±  34.1 ms    [User: 3321.3 ms, System: 1526.6 ms]
  Range (min … max):   488.0 ms … 591.6 ms    10 runs

❯ hyperfine "node --run prettier"
Benchmark 1: node --run prettier
  Time (mean ± σ):      1.168 s ±  1.124 s    [User: 0.973 s, System: 0.405 s]
  Range (min … max):    0.797 s …  4.366 s    10 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (4.366 s). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,15 @@
 	"version": "16.8.0",
 	"description": "SVG icons for popular brands https://simpleicons.org",
 	"keywords": [
-		"svg",
-		"icons"
+		"icons",
+		"svg"
 	],
 	"homepage": "https://simpleicons.org",
 	"bugs": {
 		"url": "https://github.com/simple-icons/simple-icons/issues"
 	},
+	"license": "CC0-1.0",
+	"author": "Simple Icons Collaborators",
 	"repository": {
 		"type": "git",
 		"url": "git+ssh://git@github.com/simple-icons/simple-icons.git"
@@ -24,9 +26,9 @@
 			"url": "https://github.com/sponsors/simple-icons"
 		}
 	],
-	"license": "CC0-1.0",
-	"author": "Simple Icons Collaborators",
 	"sideEffects": false,
+	"main": "index.js",
+	"types": "index.d.ts",
 	"exports": {
 		".": {
 			"types": "./index.d.ts",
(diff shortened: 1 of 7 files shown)


mondeja (changes requested):
The problem with this is that XO is still running Prettier under the hood and disabling it makes XO fail with the formatting applied by Oxfmt. If you change `prettier: true` to `prettier: false` in _xo.config.js_ and execute formatting and jslint scripts you'll reproduce it.

package.json
```diff
@@ -60,24 +62,22 @@
...
 		"get-filename": "node scripts/get-filename.js",
 		"jslint": "xo",
 		"jsonlint": "node scripts/lint/jsonlint.js",
-		"lint": "node --run ourlint && node --run prettierlint && node --run jslint && node --run jsonlint && node --run svglint && node --run wslint && node --run tslint && node --run markdownlint",
+		"lint": "node --run ourlint && node --run oxfmtlint && node --run jslint && node --run jsonlint && node --run svglint && node --run wslint && node --run tslint && node --run markdownlint",
 		"markdownlint": "rumdl check .",
 		"ourlint": "node scripts/lint/ourlint.js",
 		"prepare": "node -e \"process.exit(process.env.CI ? 0 : 1)\" || prek install",
 		"prepublishOnly": "node --run build",
-		"prettier": "prettier --ignore-unknown \"**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)\"",
```
mondeja, line 79: In the `prettier` script, we're currently executing Prettier for non JS files only, that's why the script has `\"**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts|svg)\"` hardcoded. With your changes, we're executing `oxfmt` for all files and then prettier for JS files.

PR #14383, opened 2026-02-22
Refactor redundant Object.assign wrappers and fix JSDoc types in  scripts/utils.js

Nandann018-ux:
This PR removes three redundant `Object.assign()` wrappers around `Object.fromEntries()` inside `scripts/utils.js`. Because `Object.assign` invoked with a single argument simply returns the argument untouched, this wrapper served no functional purpose at runtime.

Previously, `Object.assign()` implicitly swallowed TypeScript check errors caused by sorting by letting the result infer as `any`. With its removal, the implicit TypeScript bypass fails under `npm run tslint`. To correct this, standard and strict JSDoc inline casts (`/** @type {T} */ ( ... )`) are introduced. This provides clean syntactic JavaScript, maintains strong static type analysis, and avoids an unnecessary runtime function call.

- Passes all tests
- Passes all standard linting checks and pre-commit hooks

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -100,13 +100,12 @@ const sortIconOrDuplicate = (icon) => {
 		'loc',
 	];
 
-	/** @type {T} */
-	const sortedIcon = Object.assign(
+	const sortedIcon = /** @type {T} */ (
 		Object.fromEntries(
 			Object.entries(icon).sort(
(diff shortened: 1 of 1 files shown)


LitoMore:
We already have a plan to refactor our scripts to TypeScript. Closing.

PR #14484, opened 2026-03-17
Upgrade `xo` to 2.0.2

LitoMore:


scripts/build/package.js
```diff
@@ -133,9 +133,9 @@ const buildIcons = async () =>
...
 			const svg = await fs.readFile(svgFilepath, UTF8);
 			/** @type {IconDataAndObject} */
-			const icon = {};
-			Object.assign(icon, iconData);
-			icon.svg = svg;
+			// @ts-expect-error: Some properties does not exist. We can polish it in TypeScript migration.
+			const icon = {...iconData, svg};
+			// eslint-disable-next-line unicorn/no-immediate-mutation
 			icon.path = svgToPath(svg);
 			icon.slug = slug;
```
cscnk52: Nit: This could be simplified by constructing the object in one step.

Right now it's created and then mutated (`icon.path`, `icon.slug`),
which also requires disabling ESLint rules.

Maybe something like:

```ts
const icon = {
  ...iconData,
  svg,
  path: svgToPath(svg),
  slug,
};

PR #14515, opened 2026-03-22
Warning for unoptimized icons

cscnk52:
This pull request add warning for icon was not optimized by SVGO, just running in CI, because lots of our exist icon not optimized.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -80,7 +80,10 @@ jobs:
         if: steps.pr.outputs.is_normal == 'true' && steps.changes.outputs.icons == 'true'
         run: sed -i 's/\\"icons\/\*\.svg\\"/$npm_config_icons/' package.json
       - name: Run linter
-        run: node --run lint
+        run: |
+          node --run lint
+          node --run svgo $npm_config_icons
+          git diff --exit-code
         env:
           npm_config_icons: ${{ steps.changes.outputs.icons_files }}
           # Authorise GitHub API requests for editorconfig-checker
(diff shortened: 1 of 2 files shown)


LitoMore:
I don't think we need this. Please note that not all of our icons are optimized using svgo's default parameters.

cscnk52:
This PR intend for warning for unoptimized new icon, not affect to our exist icon.

LitoMore:
We do not mandate that all icons must use SVGO.

mondeja:
I'm not sure about this.

> We do not mandate that all icons must use SVGO.

I think this statement is not true. Quoting CONTRIBUTING.md:

> 3. Optimize the Icon
>
> All icons in Simple Icons have been optimized with the [SVGO tool](https://github.com/svg/svgo).

We mandate all icons must use SVGO.

---

> Please note that not all of our icons are optimized using svgo's default parameters.

That's right, but AFAIK, the unique setting that we change sometimes is the amount of floating point decimals. Can you provide more info about it @LitoMore? Do you've faced icons that needed other tweaks?

LitoMore:
I mean, "We do not mandate that all icons must use SVGO. Users can use other compression tools, such as SVGOMG."

They are different. Even if a user uses SVGOMG, the results may still differ from those of SVGO.

And some icons are provided directly by the copyright holders, and we should not process these icons again.

mondeja:
Fair enough. We can add the step as a warning to help reviewers know if the icon matches the output of SVGO. What do you think about it, @LitoMore?

PR #14730, opened 2026-05-16
Add Await widget third-party extension

LitoMore:
This is an iOS app extension that lets you add a Simple Icons wall to your desktop as a widget.

See https://github.com/LitoMore/await-widgets/tree/main/widgets/simple-icons-wall.


mondeja (commented):
We should add an icon for Await to Simple Icons, right? We've required this for the other extensions in the past. See #14731

LitoMore:
The Await app is very new, so it may not reach our popularity requirement. That's why I self-hosted the icon.

But if this icon is good to go, then I will create a PR to add it.

cscnk52 (approved):
I prefer self-host icon first, wait for meet our standard then add it.

LitoMore:
Or I can only add the Await icon to the Simple Icons CDN as an extra icon.

mondeja:
I think that we can make an exception and add the icon to Simple Icons. Having an extension is a good case of usage for ourselves and it encourages brand developers to create extensions for Simple Icons.

LitoMore:
I've created a PR for it:

- https://github.com/simple-icons/simple-icons/pull/14743

PR #14731, opened 2026-05-16
Add test for image URLs in README 3rd party extension tables

mondeja:
Assert that images in 3rd parth extension tables in README.md are consumed from the Simple Icons CDN. This test would have prevented the pipeline in #14730 to pass.

~~This is a breaking change in the SDK, so marking as draft for the next major version~~


cscnk52:
Nits: Since `node:test` have been stabilized in [node@v18.0.0](https://nodejs.org/api/test.html), i prefer move it to native modules, to reduce third party dependencies.

cscnk52:
Maybe migrated to `node:test` in next PR, I will merge this first.

cscnk52 (approved):

